PR #59, opened 2017-03-22
Create strings.xml - Hungarian translation file

aligitor:


r10s:
Thank you very much, I've just added your file.

Hungarian will be available in the upcoming release, 0.1.22, then.

PR #74, opened 2017-04-16
Russian language strings.xml

guland2000:
Hi! Please add Russian language to apk, it will be great to test it by me, before real use.


r10s:
Thank you very much, Delta Chat will be available in Russian starting with 0.1.29 then.

guland2000:
Thank YOU for great messenger !

r10s:
You're welcome!

toddy15:
Hi @guland2000,

thanks a lot for the Russian translation. However, it seems that in the Russian language, there is one more plural item needed, named "few".

The plural rules for Russian specify that the language has a "one" form (for 1, 21, 31, 41, 51, …), a "few" form (for 2–4, 22–24, 32–34, …), and an "other" form (for everything else: 0, 5–20, 25–30, 35–40, …, decimals).

Could you please add it to each plural string? I've tried to create an example for you with an automatic translation program, so the translation might be completely wrong, but the complete entry should look similar to this:

`<plurals name="Members">`
`        <item quantity="one">%d член</item>`
`        <item quantity="few">%d участников</item>`
`        <item quantity="other">%d членов</item>`
`</plurals>`

guland2000:
_It must be look like this:_
<plurals name="Members">
<item quantity="one">%d **участник**</item>
**<item quantity="few">%d участников</item>**
<item quantity="other">%d **участники**</item>
</plurals>
Not sure how make but i add <item quantity="few">%d участников</item> in branch

toddy15:
Hi @guland2000,

thanks a lot! However, I think that you need to add this line to *every* plural item, not only for "Members", but also for the following plurals like "Contacts", "MeAndMembers", "NewMessages" and so on.

Could you please do that as well? Thanks in advance!

PR #92, opened 2017-06-02
Update strings.xml

guland2000:
1. Fix some error, linguistic and semantic (contextual) : strings 86,118,129,223,343,345,351 
2. and added New strings 377, 378, 379,380,381,382,383,384,385.


r10s:
thank you very much, the updated translation will be available in v0.9.1 then

PR #128, opened 2017-07-25
gradle: use default debug keystore path

vitalyster:


r10s:
I'm no Android expert. What is this change for?

vitalyster:
Android build tools provide debug keystore automatically, with this change no need to create one

r10s:
Okay, thank you very much. 

@toddy15 I do not think, this will affect F-Droid.

r10s:
I've reverted your commit, I tried to explain the reasons here https://github.com/deltachat/deltachat-android#build

I'm not sure, if there is a better way to achieve using the same keyfile on different machines. But for now, I do not want to lose all my testing device installations.

However, anyway thank you for your pull request that at least started this discussion. Any further pull request are welcome and I'm looking forward to them :-)

Best regards,
Björn

vitalyster:
You can just put your keystore in '~/.android/debug.keystore', isn't it?

r10s:
I swear, it did not worked for me as I tried it first. Maybe I did something wrong.

However, yes, finally this seems to work. However, won't this get complicated for developers developing different apps?

vitalyster:
That was the main reason to make things easier for those who already develop different apps :) I have contributed to multiple projects and never saw debug keystore path redefinition.

r10s:
Okay.

But as there are already existing development setup out there and such a change may force the requirement of reinstalling Delta and losing all chats, I think it is a good idea to use config/debug.keystore if it exist. If not, we use the default.  Sth. like this https://gist.github.com/lifuzu/8009968

This would not require extra steps on new setups and leaves old setups intact.

What do you think about this?

vitalyster:
like this? https://github.com/vitalyster/deltachat-android/commit/e5507d77629224943ad58f581c17c711f1949fda

r10s:
This looks perfect to me :)
Do you like to create a pull request?

vitalyster:
done. We can also call ndk-build automatically and get rid of ndk-build line in readme too :) Are there any issues with auto build?

r10s:
Not as far as I know. I simply have no gradle android experiences; the gradle files is a copy of Telegram-FOSS. 
We only should keep in mind that it has to work on F-Droid.

PR #131, opened 2017-07-26
gradle: use default debug keystore if there is no configured one

vitalyster:


r10s:
Merged - thank you very much.

PR #171, opened 2017-09-30
Intro5 Message updates in strings.xml (de)

Ampli-fier:
To align with (coming) web updates


r10s:
I've also added the change to Transifex at https://www.transifex.com/delta-chat/delta-chat-android/translate/#de/stringsxml/114886472 , so that it won't be overwritten on the next language update ...

PR #191, opened 2017-10-25
String rework, come say hi or despair

comradekingu:
"**Filetype**" as per https://en.wiktionary.org/wiki/filetype
This avoided compound-error, along with "**e-mail**" and "**word…**" make for fewer errors in translation.
"**Successful**" is always redundant.
"**Pass.**" saves space, while being easier to comprehend than "passcode/passphrase". One could argue it falls into ambiguity with "skip" if used at the end of a sentence, but i see no problem there.

### DIFF
--- a/MessengerProj/src/main/res/values/strings.xml
+++ b/MessengerProj/src/main/res/values/strings.xml
@@ -10,21 +10,21 @@
     <string name="NoResult">No results.</string>
     <string name="NoChats">No chats yet.</string>
     <string name="DeleteChat">Delete chat</string>
-    <string name="SelectChat">Select chat …</string>
+    <string name="SelectChat">Select chat…</string>
     <string name="Search">Search</string>
     <string name="MuteNotifications">Mute notifications</string>
     <string name="MuteFor">Mute for %1$s</string>
     <string name="UnmuteNotifications">Unmute</string>
     <string name="Draft">Draft</string>
     <!--audio view-->
-    <string name="NoAudio">Please add files to the music library on your device to see them here.</string>
+    <string name="NoAudio">Files added to the music library of your device will appear here.</string>
     <string name="AttachMusic">Music</string>
     <!--documents view-->
     <string name="SelectFile">Select File</string>
-    <string name="FreeOfTotal">Free %1$s of %2$s</string>
+    <string name="FreeOfTotal">%1$s of %2$s free</string>
     <string name="ErrorHint">Unknown error</string>
     <string name="AccessError">Access error</string>
-    <string name="NoFiles">No files yet …</string>
+    <string name="NoFiles">No files yet…</string>
     <string name="NotMounted">Storage not mounted</string>
     <string name="UsbActive">USB transfer active</string>
     <string name="InternalStorage">Internal Storage</string>
@@ -41,9 +41,9 @@
     <string name="From">From</string>
     <string name="NoRecent">No recent</string>
     <string name="TypeMessage">Message</string>
-    <string name="SlideToCancel">SLIDE TO CANCEL</string>
-    <string name="SaveToDownloads">Save to downloads</string>
-    <string name="SaveToMusic">Save to music</string>
+    <string name="SlideToCancel">Slide to cancel</string>
+    <string name="SaveToDownloads">Save to \"Downloads\"</string>
+    <string name="SaveToMusic">Save to "\Music\"</string>
     <string name="Share">Share</string>
     <string name="SendItems">Send %1$s</string>
     <string name="ClearRecentEmoji">Clear recent emoji?</string>
@@ -56,23 +56,23 @@
     <!--contacts view-->
     <string name="NoContacts">No contacts yet.</string>
     <!--group create view-->
-    <string name="SendMessageTo">Send message to …</string>
+    <string name="SendMessageTo">Send message to:</string>
     <string name="EnterGroupNamePlaceholder">Enter group name</string>
     <!--group info view-->
-    <string name="AddMember">Add member</string>
+    <string name="AddMember">Grant membership</string>
     <string name="Notifications">Notifications</string>
-    <string name="RemoveMember">Remove member</string>
+    <string name="RemoveMember">Revoke membership</string>
     <!--contact info view-->
     <string name="NewContactTitle">New contact</string>
     <string name="BlockContact">Block contact</string>
     <string name="DeleteContact">Delete contact</string>
     <string name="Info">Info</string>
     <!--settings view-->
-    <string name="TextSize">Messages font size</string>
+    <string name="TextSize">Font size of messages</string>
     <string name="UnblockContact">Unblock contact</string>
     <string name="NoBlocked">No blocked contacts yet</string>
     <string name="DefaultForNormalMessages">Normal messages</string>
-    <string name="MessagePreview">Message preview</string>
+    <string name="MessagePreview">Preview of message </string>
     <string name="DefaultForGroupMessages">Group messages</string>
     <string name="Sound">Sound</string>
     <string name="InAppNotifications">In-app notifications</string>
@@ -81,27 +81,27 @@
     <string name="NotificationsAndSounds">Notifications and sounds</string>
     <string name="BlockedContacts">Blocked contacts</string>
     <string name="Default">Default</string>
-    <string name="OnlyIfSilent">Only if silent</string>
+    <string name="OnlyIfSilent">Only when silent</string>
     <string name="ChatBackground">Chat background</string>
-    <string name="SendByEnter">Send by "enter"</string>
+    <string name="SendByEnter">Send with \"⏎"\"</string>
     <string name="Help">Help</string>
     <string name="Enabled">On</string>
     <string name="Disabled">Off</string>
     <string name="LedColor">LED color</string>
-    <string name="BadgeNumber">Show count on icon if possible</string>
+    <string name="BadgeNumber">Use icon to keep count if possible</string>
     <string name="Short">Short</string>
     <string name="Long">Long</string>
     <string name="RaiseToSpeak">Raise to speak</string>
     <string name="EditName">Edit name</string>
     <string name="NotificationsPriority">View</string>
-    <string name="NotificationsPriorityDefault">Normal priority</string>
-    <string name="NotificationsPriorityHigh">High priority</string>
-    <string name="NotificationsPriorityMax">Highest priority</string>
+    <string name="NotificationsPriorityDefault">Normal</string>
+    <string name="NotificationsPriorityHigh">Prioritized</string>
+    <string name="NotificationsPriorityMax">Paramount</string>
     <string name="RepeatNotifications">Repeat notifications</string>
     <string name="NotificationsOther">Other</string>
     <string name="InChatSound">In-chat sounds</string>
     <string name="SmartNotifications">Limit notifications</string>
-    <string name="SmartNotificationsSoundAtMost">Sound at most</string>
+    <string name="SmartNotificationsSoundAtMost">Max. volume</string>
     <string name="SmartNotificationsTimes">times</string>
     <string name="SmartNotificationsWithin">within</string>
     <string name="SmartNotificationsMinutes">minutes</string>
@@ -110,27 +110,27 @@
     <!--cache view-->
     <string name="CacheSettings">Storage</string>
     <string name="KeepMedia">Keep Media</string>
-    <string name="KeepMediaInfo">Photos, videos and other files from cloud chats that you have <![CDATA[<b>not accessed</b>]]> during this period will be removed from this device to save disk space.</string>
+    <string name="KeepMediaInfo">Photos, videos and other files from cloud chats you have <![CDATA[<b>yet to access</b>]]> during this period will be removed from this device to save disk space.</string>
     <string name="KeepMediaForever">Forever</string>
     <!--passcode view-->
-    <string name="Passcode">Passcode Lock</string>
-    <string name="ChangePasscode">Change Passcode</string>
-    <string name="ChangePasscodeInfo">When you set up an additional passcode, a lock icon will appear on the chats page. Tap it to lock the app.\n\nNote: if you forget the passcode, you\'ll need to delete and reinstall the app.</string>
+    <string name="Passcode">App Lock</string>
+    <string name="ChangePasscode">Change app lock</string>
+    <string name="ChangePasscodeInfo">When you set up an app lock, a lock icon will appear atop the chat page. Tap it to lock the app.\n\nNote: Should you forget it, you\'ll need to delete and reinstall the app.</string>
     <string name="PasscodePIN">PIN</string>
-    <string name="PasscodePassword">Password</string>
-    <string name="EnterCurrentPasscode">Enter your current passcode</string>
-    <string name="EnterNewFirstPasscode">Enter a passcode</string>
-    <string name="EnterNewPasscode">Enter your new passcode</string>
-    <string name="EnterYourPasscode">Enter your passcode</string>
-    <string name="ReEnterYourPasscode">Re-enter your new passcode</string>
-    <string name="PasscodeDoNotMatch">Passcodes do not match</string>
+    <string name="PasscodePassword">Pass.</string>
+    <string name="EnterCurrentPasscode">Enter your current pass.</string>
+    <string name="EnterNewFirstPasscode">Enter pass.</string>
+    <string name="EnterNewPasscode">Enter your new pass.</string>
+    <string name="EnterYourPasscode">Enter your pass.</string>
+    <string name="ReEnterYourPasscode">Re-enter your new pass.</string>
+    <string name="PasscodeDoNotMatch">Mismatching pass.</string>
     <string name="AutoLock">Autolock</string>
-    <string name="AutoLockInfo">Require passcode if away for a time.</string>
-    <string name="UnlockFingerprint">Unlock with Fingerprint</string>
-    <string name="FingerprintInfo">Confirm fingerprint to continue</string>
-    <string name="FingerprintNotRecognized">Fingerprint not recognized. Try again</string>
+    <string name="AutoLockInfo">Require pass. following inactivity.</string>
+    <string name="UnlockFingerprint">Unlock with fingerprint</string>
+    <string name="FingerprintInfo">Confirm fingerprint</string>
+    <string name="FingerprintNotRecognized">Not recognized. Try again</string>
     <!--photo gallery view-->
-    <string name="SaveToGallery">Save to gallery</string>
+    <string name="SaveToGallery">Save to \"Gallery"\</string>
     <string name="Of">%1$d of %2$d</string>
     <string name="Gallery">Gallery</string>
     <string name="AllPhotos">All Photos</string>
@@ -177,7 +177,7 @@
     <string name="Cancel">Cancel</string>
     <string name="Edit">Edit</string>
     <string name="Send">Send</string>
-    <string name="CopyToClipboard">Copy to clipboard</string>
+    <string name="CopyToClipboard">Copy</string>
     <string name="Delete">Delete</string>
     <string name="Forward">Forward</string>
     <string name="FromCamera">From camera</string>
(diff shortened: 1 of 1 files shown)


r10s:
Thank you very much. Before I can merge this, I have to figure out how to avoid Transifex from throwing away all translations affecting these changes (Transifex seems to use the english string as the ID, not the name-attribute)

r10s:
But again, I think it's email, not e-mail.

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -10,21 +10,21 @@
     <string name="NoResult">No results.</string>
     <string name="NoChats">No chats yet.</string>
     <string name="DeleteChat">Delete chat</string>
-    <string name="SelectChat">Select chat …</string>
+    <string name="SelectChat">Select chat…</string>
```
r10s: Well, it's not clear if the ellipsis is surrounded by space or not, see https://en.wikipedia.org/wiki/Ellipsis , eg. in menus, Ubuntu adds a space and Firefox does not.
I personally prefer the space, as without a space, to me it always suggests, the the _word_ is not complete. The space suggests, the _sentence_ is not complete. In our scenario, it is the action that is not complete, but I think this is closes to the latter one.

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -110,27 +110,27 @@
...
     <string name="CacheSettings">Storage</string>
     <string name="KeepMedia">Keep Media</string>
-    <string name="KeepMediaInfo">Photos, videos and other files from cloud chats that you have <![CDATA[<b>not accessed</b>]]> during this period will be removed from this device to save disk space.</string>
+    <string name="KeepMediaInfo">Photos, videos and other files from cloud chats you have <![CDATA[<b>yet to access</b>]]> during this period will be removed from this device to save disk space.</string>
     <string name="KeepMediaForever">Forever</string>
     <!--passcode view-->
-    <string name="Passcode">Passcode Lock</string>
-    <string name="ChangePasscode">Change Passcode</string>
-    <string name="ChangePasscodeInfo">When you set up an additional passcode, a lock icon will appear on the chats page. Tap it to lock the app.\n\nNote: if you forget the passcode, you\'ll need to delete and reinstall the app.</string>
+    <string name="Passcode">App Lock</string>
```
r10s: Yes - it is the app that is locked, not the passcode. Sounds reasonable to me.

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -304,107 +304,107 @@
...
     <!-- Translators: please use a very short string here, it should not much longer than
     the string needed for video time+size (0:00, 12,3 Mib) -->
-    <string name="OneMoment">One moment …</string>
-    <string name="NoChatsHelp">Start messaging by pressing the new chat button in the bottom right corner or tap the menu button for more options.</string>
+    <string name="OneMoment">One moment…</string>
+    <string name="NoChatsHelp">Start messaging by pressing the \"New chat"\ button in the bottom right, or tap the \"Menu"\ button for more options.</string>
     <string name="Intro7Message"><![CDATA[<b>Delta Chat</b>]]> is safe for business use, compatible and standards-based.</string>
     <string name="InviteMenuEntry">Send invitations</string>
-    <string name="InviteText">I\'m using Delta Chat messenger now - %1$s - you can text me at %2$s</string>
+    <string name="InviteText">Shhh, Delta Chat messenger now - %1$s - you can text me at %2$s</string>
```
r10s: Don't know - what does "Shhh" mean in this case?

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -304,107 +304,107 @@
...
     <string name="MsgGroupImageChanged">Group image changed.</string>
     <string name="MsgMemberAddedToGroup">Member %1$s added.</string>
     <string name="MsgMemberRemovedFromToGroup">Member %1$s removed.</string>
     <string name="AskAddMemberToGroup">Add <![CDATA[<b>]]>%1$s<![CDATA[</b>]]> to group?</string>
     <string name="AskRemoveMemberFromGroup">Remove <![CDATA[<b>]]>%1$s<![CDATA[</b>]]> from group?</string>
-    <string name="ErrSelfNotInGroup">You must be a member of the group to perform this action.</string>
-    <string name="MsgGroupLeft">Group left.</string>
-    <string name="NoMessagesHint">Send message to <![CDATA[<b>]]>%1$s<![CDATA[</b>]]>:\n\n• It is okay if  <![CDATA[<b>]]>%2$s<![CDATA[</b>]]> does not use Delta Chat.\n\n• Delivering the first message may take a moment.</string>
+    <string name="ErrSelfNotInGroup">You must be a group member to perform this action.</string>
+    <string name="MsgGroupLeft">Group disbanded.</string>
```
r10s: well, it's only a single member that has left the group, the group itself stays intact and unchanged.

MessengerProj/src/main/res/values/strings.xml
Ampli-fier: I prefer to have the space as well.

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -56,23 +56,23 @@
...
     <!--group create view-->
-    <string name="SendMessageTo">Send message to …</string>
+    <string name="SendMessageTo">Send message to:</string>
     <string name="EnterGroupNamePlaceholder">Enter group name</string>
     <!--group info view-->
-    <string name="AddMember">Add member</string>
+    <string name="AddMember">Grant membership</string>
     <string name="Notifications">Notifications</string>
-    <string name="RemoveMember">Remove member</string>
+    <string name="RemoveMember">Revoke membership</string>
```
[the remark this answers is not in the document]
Ampli-fier: This gets displayed when the group is opened.
It is like a toast message on top of the entry "Add member", so "Remove member" looks better.

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -56,23 +56,23 @@
...
     <string name="DeleteContact">Delete contact</string>
     <string name="Info">Info</string>
     <!--settings view-->
-    <string name="TextSize">Messages font size</string>
+    <string name="TextSize">Font size of messages</string>
     <string name="UnblockContact">Unblock contact</string>
     <string name="NoBlocked">No blocked contacts yet</string>
     <string name="DefaultForNormalMessages">Normal messages</string>
-    <string name="MessagePreview">Message preview</string>
+    <string name="MessagePreview">Preview of message </string>
```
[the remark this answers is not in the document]
Ampli-fier: I vote for: "Display Message Preview".

BTW: The german translation only says "Vorschau", could be changed to "Nachrichtenvorschau anzeigen".

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -81,27 +81,27 @@
     <string name="NotificationsAndSounds">Notifications and sounds</string>
     <string name="BlockedContacts">Blocked contacts</string>
     <string name="Default">Default</string>
-    <string name="OnlyIfSilent">Only if silent</string>
+    <string name="OnlyIfSilent">Only when silent</string>
     <string name="ChatBackground">Chat background</string>
-    <string name="SendByEnter">Send by "enter"</string>
+    <string name="SendByEnter">Send with \"⏎"\"</string>
```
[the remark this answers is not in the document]
Ampli-fier: Even better: "Send message with button ..."

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -81,27 +81,27 @@
...
+    <string name="BadgeNumber">Use icon to keep count if possible</string>
     <string name="Short">Short</string>
     <string name="Long">Long</string>
     <string name="RaiseToSpeak">Raise to speak</string>
     <string name="EditName">Edit name</string>
     <string name="NotificationsPriority">View</string>
-    <string name="NotificationsPriorityDefault">Normal priority</string>
-    <string name="NotificationsPriorityHigh">High priority</string>
-    <string name="NotificationsPriorityMax">Highest priority</string>
+    <string name="NotificationsPriorityDefault">Normal</string>
```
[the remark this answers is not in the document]
Ampli-fier: I would label it "View priority" to be more precise what the setting means.
Then only have the value defined as: "Normal", "High", etc. _priority_ is redundant here.

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -81,27 +81,27 @@
...
+    <string name="NotificationsPriorityHigh">Prioritized</string>
+    <string name="NotificationsPriorityMax">Paramount</string>
     <string name="RepeatNotifications">Repeat notifications</string>
     <string name="NotificationsOther">Other</string>
     <string name="InChatSound">In-chat sounds</string>
     <string name="SmartNotifications">Limit notifications</string>
-    <string name="SmartNotificationsSoundAtMost">Sound at most</string>
```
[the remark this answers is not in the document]
Ampli-fier: "Max." is sufficient here.

MessengerProj/src/main/res/values/strings.xml
Ampli-fier: To me this is the "Lock Passcode"

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -177,7 +177,7 @@
     <string name="Cancel">Cancel</string>
     <string name="Edit">Edit</string>
     <string name="Send">Send</string>
-    <string name="CopyToClipboard">Copy to clipboard</string>
+    <string name="CopyToClipboard">Copy</string>
```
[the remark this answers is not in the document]
Ampli-fier: Please make it clear. As you said, it can be interpreted to copy sth. to a folder, eg.

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -304,107 +304,107 @@
...
     <string name="NotSet">Not set</string>
     <string name="NewChat">New chat</string>
     <string name="Deaddrop">Contact requests</string>
     <string name="DeaddropInChatlist">Show contact requests in chatlist</string>
     <string name="MuteAlways">Always muted</string>
-    <string name="AskStartChatWith">Start a chat with <![CDATA[<b>]]>%1$s<![CDATA[</b>]]>?</string>
-    <string name="DeaddropHint">To start a chat, click on the reply-arrows.</string>
-    <string name="NotYetImplemented">This function is not available or incomplete.</string>
-    <string name="DefaultStatusText">Sent with my Delta Chat Messenger: https://delta.chat</string>
+    <string name="AskStartChatWith">Chat with <![CDATA[<b>]]>%1$s<![CDATA[</b>]]>?</string>
```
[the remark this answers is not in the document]
Ampli-fier: +1

MessengerProj/src/main/res/values/strings.xml
Ampli-fier: What is the context here?

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -110,27 +110,27 @@
...
-    <string name="KeepMediaInfo">Photos, videos and other files from cloud chats that you have <![CDATA[<b>not accessed</b>]]> during this period will be removed from this device to save disk space.</string>
+    <string name="KeepMediaInfo">Photos, videos and other files from cloud chats you have <![CDATA[<b>yet to access</b>]]> during this period will be removed from this device to save disk space.</string>
     <string name="KeepMediaForever">Forever</string>
     <!--passcode view-->
-    <string name="Passcode">Passcode Lock</string>
-    <string name="ChangePasscode">Change Passcode</string>
-    <string name="ChangePasscodeInfo">When you set up an additional passcode, a lock icon will appear on the chats page. Tap it to lock the app.\n\nNote: if you forget the passcode, you\'ll need to delete and reinstall the app.</string>
+    <string name="Passcode">App Lock</string>
+    <string name="ChangePasscode">Change app lock</string>
+    <string name="ChangePasscodeInfo">When you set up an app lock, a lock icon will appear atop the chat page. Tap it to lock the app.\n\nNote: Should you forget it, you\'ll need to delete and reinstall the app.</string>
```
Ampli-fier: My preference is to keep this:

\n\nNote: Should you forget **the passcode**, ...

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -110,27 +110,27 @@
...
     <string name="PasscodePIN">PIN</string>
-    <string name="PasscodePassword">Password</string>
-    <string name="EnterCurrentPasscode">Enter your current passcode</string>
-    <string name="EnterNewFirstPasscode">Enter a passcode</string>
-    <string name="EnterNewPasscode">Enter your new passcode</string>
-    <string name="EnterYourPasscode">Enter your passcode</string>
-    <string name="ReEnterYourPasscode">Re-enter your new passcode</string>
-    <string name="PasscodeDoNotMatch">Passcodes do not match</string>
+    <string name="PasscodePassword">Pass.</string>
```
Ampli-fier: Is there really a need to "save space"?
I have a pretty small phone and this does not need to get chopped to fit.

We could think about an alternative version of DC without vowels ;-)
"ntr yr crrnt psscd"
"Lmt ntfctns"

As an advanced setting - it could bring (a lot) attention to DC ... ... (just dreaming)

Ampli-fier:
Regarding the word _email_:
http://www.future-perfect.co.uk/grammar-tip/is-it-e-mail-email-e-mail-or-email/

My vote goes for "e-mail". And "e-mail server" instead of "e-mail-server" (as an example).

jeanm:
*E-mail-server* made me smile, it sounds very German :) I would definitely avoid the compound form.

As for e-mail vs email, a few more data points:
* [Associated Press stylebook recommends *email* over *e-mail*](https://twitter.com/APStylebook/status/48798366980780033)
* [So does the Guardian](https://www.theguardian.com/guardian-observer-style-guide-e)
* On the other hand, the Economist seems to prefer *e-mail*
* [Google Trends shows *e-mail* declining over time](https://trends.google.co.uk/trends/explore?date=all&q=%22email%22,%22e-mail%22)
* [Lengthy SE post](https://english.stackexchange.com/questions/1925/email-or-e-mail)
* [Don Knuth's opinion](http://www-cs-faculty.stanford.edu/~knuth/email.html)

Ampli-fier:
Is it because people are (just too) lazy?

- [Google Trend _tshirt_ vs _t-shirt_](https://trends.google.co.uk/trends/explore?date=all&q=%22tshirt%22,%22t-shirt%22)

:)

MessengerProj/src/main/res/values/strings.xml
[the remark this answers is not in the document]
r10s: "Display message preview" also sounds good to me, but I am no native speaker, unfortunately.

MessengerProj/src/main/res/values/strings.xml
r10s: MsgGroupLeft ist shown as in place of the message (in the bubble) text a user has left a group by himself.
It's a sort of a system message, however, as the action comes from the user, it is shown in its name. (if someone finds some time, however, it might be a good idea show these messages eg. in italics)

MessengerProj/src/main/res/values/strings.xml
r10s: nc id

r10s: sorry, it must read "nc d" then :)

MessengerProj/src/main/res/values/strings.xml
Ampli-fier: My opinion is to show a full sentence: 

- "I have left the group!" or
- "I have left this group!"

r10s: this "I have" implies the message was written by the user in this form. I do not think, this is a good idea. I would prefer a more neutral message (which could be formatted as a clear "system message" then, if in doubt, just using eg. italics)

Ampli-fier: Hmmm, bubble? In my test scenario this text was shown as a real message.
The user did not write the text (in this form), but he triggered it by pressing "the button" and telling the others "I quit".

It was even shown in the users (the user who left) chat as outgoing text.
So, I tought it makes more sense to use the "I-form" (strange word).

r10s: we're talking about the same stuff. Currently, the message is shown as if it was typed in this form by the user - however, this was only to keep things simple, this is not my desired form. I would like the "status messages" coming from a user - but not typed by him - to be clearly distinguishable from normal text messages - eg. by using italics and a neutral speech.

MessengerProj/src/main/res/values/strings.xml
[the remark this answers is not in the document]
comradekingu: How about just "Preview message"?

MessengerProj/src/main/res/values/strings.xml
comradekingu: The sound you make to emphasise quietness.

comradekingu:
E-mail makes for fewer errors in translation, seeing as the equivalent of "email" is actively wrong in other languages, whereas local variants of "e-mail" are correct.

MessengerProj/src/main/res/values/strings.xml
r10s: Ahhh, okay.

--
Shhh, I am not yet really awake.

r10s:
@comradekingu The supprt of Transifex directed me to the following way for updating the source strings without removing the translations:

_- - - - - - - - - s n i p - - - - - - - - -_
In order to easily export and import files at once, I would suggest you use our command line client.

Once it is installed, it's a matter of 4 commands so as to achieve what you want:

    Initialize a Transifex project:
    $ tx init
    Synchronize your project on transifex.com with your local instance:
    $ tx set --auto-remote https://www.transifex.com/organization_slug/project_slug/
    Pull all the files at once
    $ tx pull -s -a
    Push the files back to Transifex
    $ tx push -t 

More information about how this works can be found on our documentation guide [here](https://docs.transifex.com/client/introduction).
_- - - - - - - - - s n i p - - - - - - - - -_

comradekingu:
@r10s I don't see how that was any different from how their documentation details it:

> Here's how Transifex handles differences between the old and new source files:
> 
>     New strings are added.
>     Modified strings are considered new ones and added as well.
>     Strings that aren't in the new source file – including ones which have been modified –  are removed from Transifex. However, your old translations are kept in the Translation Memory of your project.

The warning for overwriting string was also present, and is very real. You can easily wipe all strings, and upon uploading a manual backup, all strings are attributed to the uploader, which makes translation extra painful.

I would advise using either Weblate or Pootle.

MessengerProj/src/main/res/values/strings.xml
comradekingu: Well, what if it isn't an active decision on part of the user (?), stemming instead from something going wrong, closing the chat as a result.

MessengerProj/src/main/res/values/strings.xml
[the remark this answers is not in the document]
comradekingu: I thought active actions were followed by "…"
In any event, "Copy text" should suffice.(?)

MessengerProj/src/main/res/values/strings.xml
comradekingu: @Ampli-fier How I read that is "to lock the passcode" rather than "a passcode for the lock".

MessengerProj/src/main/res/values/strings.xml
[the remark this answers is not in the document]
comradekingu: If you just remove a member, other than it being an amputation, is not as clear as what actually happens. Is the member removed temporarily?
Membership is granted or revoked.

MessengerProj/src/main/res/values/strings.xml
Ampli-fier: It might be my german brain :)

What can I do as a user, when I press this button?

"Set the passcode" 
"Set the passcode to lock the app"

in short:

"Set the lock passcode"

shorter:

"Lock Passcode"

MessengerProj/src/main/res/values/strings.xml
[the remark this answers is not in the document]
comradekingu: It would have to be "Send with … button", but it just means an extra word for no reward. "… to send" is even shorter, but maybe a little lackluster?

MessengerProj/src/main/res/values/strings.xml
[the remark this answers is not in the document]
comradekingu: "Viewing priority" ? To differentiate it from infinite action "view".

MessengerProj/src/main/res/values/strings.xml
comradekingu: @Ampli-fier The problem arises from trying to go back the other way, especially in a language that does not use compound-words correctly.

MessengerProj/src/main/res/values/strings.xml
comradekingu: My main motivation was to avoid the different pass -word -phrase -code triopoly.
I suppose oblig. xkcd. But still, this maintains parity with all, without invoking either.

MessengerProj/src/main/res/values/strings.xml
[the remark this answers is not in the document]
comradekingu: I went with that rather than "Max. reps. within."

MessengerProj/src/main/res/values/strings.xml
Ampli-fier: "Set the lock passcode" probably avoids this for most languages (which I cannot _really_ tell only knowing two).

r10s:
okay, now i have my toolchain ready (see new "tools" folder), tomorrow I will continue (for the moment I am too tiered and i fear I will break things ... )  Good night!

r10s:
Okay, I have commited most of the changes by https://github.com/deltachat/deltachat-android/commit/06a3a6b2858fcdcf06db5fdf8208e239f738f8ed .

I close this PR now, but I, of course, further discussions are very welcome. As I have a way to easily modify the source files without changing the translations now (see "tools" directory), upcoming PR on values/strings.xml can be handled much easier and faster.

PR #216, opened 2017-12-14
Update Gradle Wrapper to latest Gradle version (4.4)

jlelse:
I updated the Gradle Wrapper version to the latest one, which is currently Gradle 4.4


r10s:
where do you got the 4.4 from?
I just updated android studio and git 4.1.

anyway, I'm also updateing currently (please forgive me if i do not merge this PR), and found a bug regarding F-Droid (we were just chatting about this): https://github.com/deltachat/deltachat-android/issues/217

jlelse:
4.4 is the newest version available on the official [download site](https://services.gradle.org/distributions/).

r10s:
ahh okay. I just use the one offered by Android Studio. I think this is the better approach as this is the version that is there out-of-the-box - F-Droid, however, seems to use its own - all in all I cannot say I have understood the gradle stuff completely...

jlelse:
You can either use your locally installed gradle with `gradle` or the gradle wrapper with `gradlew`.

r10s:
Thank you very much.

PR #238, opened 2018-01-05
Update the Notification after import

Ampli-fier:
"Account not configured" was shown instead of the configured mail address.


r10s:
works like a charm :) thank you very much :)

Ampli-fier:
that was quick ... welcome :)

PR #241, opened 2018-01-06
Notification clickable

Ampli-fier:
Click on Notification opens Delta Chat

### DIFF
--- a/MessengerProj/src/main/java/com/b44t/messenger/KeepAliveService.java
+++ b/MessengerProj/src/main/java/com/b44t/messenger/KeepAliveService.java
@@ -24,6 +24,7 @@ package com.b44t.messenger;
 
 import android.app.Notification;
 import android.app.NotificationManager;
+import android.app.PendingIntent;
 import android.app.Service;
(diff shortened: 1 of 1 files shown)


r10s:
Cool :) 
This give the notification more sense.
Have you tested that not multiple Delta activities laying over each other when clicking on the notification while Delta is already in foreground? (I had such stuff here and there, always with unexpected results when hitting the back-button) (not saying that this also happens for this PR :)

r10s:
btw: we have basic Python bindings now :) see https://github.com/deltachat/deltachat-core/pull/83 :)

Ampli-fier:
I do not see more than one instance when I test this. Going back works as expected and also the "app history" only shows one instance. I think the LaunchActivity is also started when one clicks on the DC icon. At least this was my intention.

I do not know what _Python bindings_ are. But I hope I will get to know it soon :)

r10s:
Okay, then I will merge now :)

r10s:
Just tested it on my device: Works like a charm!

PR #242, opened 2018-01-11
Spelling: A similie ate ion

comradekingu:
E-mail makes for fewer errors in translation.
: Bigversal likewise.
The single entendere of making "member" "membership", has an no unfortunate duality in English.
"Determinated" is not a word
"Sent with" gets mentally filtered.

### DIFF
--- a/MessengerProj/src/main/res/values/strings.xml
+++ b/MessengerProj/src/main/res/values/strings.xml
@@ -83,7 +83,7 @@
     <string name="Default">Default</string>
     <string name="OnlyIfSilent">Only when silent</string>
     <string name="ChatBackground">Chat background</string>
-    <string name="SendByEnter">Send with \"⏎\"</string>
+    <string name="SendByEnter">Send with \"↵\"</string>
     <string name="Help">Help</string>
     <string name="Enabled">On</string>
     <string name="Disabled">Off</string>
@@ -421,7 +421,7 @@
     <string name="AutocryptKeyTransferMsgAfterSend">Your key has been sent to yourself. Open the setup message on the other device and type the following digits:\n\n%1$s\n\nOnce done, your other device will be ready to use Autocrypt.</string>
     <string name="AutocryptSetupMessageSubject">Autocrypt Setup Message</string>
     <string name="AutocryptSetupMessageGeneralBody">The setup code must be entered on the other device to use your key there.</string>
-    <string name="AutocryptSetupMessageTapBody">Use this Autocrypt setup message to transfer your key between clients.\n\nTo decrypt and use your key, tap this message.</string>
+    <string name="AutocryptSetupMessageTapBody">Tap this Autocrypt setup message to decrypt your key, transferring it between programs.</string>
     <string name="AutocryptKeyTransferPleaseEnterCode">Please enter the setup code displayed on the other device.</string>
     <string name="AutocryptKeyTransferBadCode">Bad setup code.\n\nUnless you remember the setup code, you must initiate the key transfer on the other device again.</string>
     <string name="AutocryptKeyTransferSucceeded">This device can now Autocrypt with the transferred key.</string>
--- a/MessengerProj/src/main/res/values/strings.xml
+++ b/MessengerProj/src/main/res/values/strings.xml
@@ -17,7 +17,7 @@
     <string name="UnmuteNotifications">Unmute</string>
     <string name="Draft">Draft</string>
     <!--audio view-->
-    <string name="NoAudio">Files added to the music library of your device will appear here.</string>
+    <string name="NoAudio">Files from the music library.</string>
     <string name="AttachMusic">Music</string>
     <!--documents view-->
     <string name="SelectFile">Select File</string>
@@ -59,9 +59,9 @@
     <string name="SendMessageTo">Send message to …</string>
     <string name="EnterGroupNamePlaceholder">Enter group name</string>
     <!--group info view-->
-    <string name="AddMember">Add member</string>
+    <string name="AddMember">Grant membership</string>
     <string name="Notifications">Notifications</string>
-    <string name="RemoveMember">Remove member</string>
+    <string name="RemoveMember">Revoke membership</string>
     <!--contact info view-->
     <string name="NewContactTitle">New contact</string>
     <string name="BlockContact">Block contact</string>
@@ -83,7 +83,7 @@
     <string name="Default">Default</string>
     <string name="OnlyIfSilent">Only when silent</string>
     <string name="ChatBackground">Chat background</string>
-    <string name="SendByEnter">Send with Enter key</string>
+    <string name="SendByEnter">Send with \"⏎\"</string>
     <string name="Help">Help</string>
     <string name="Enabled">On</string>
     <string name="Disabled">Off</string>
@@ -115,7 +115,7 @@
     <!--passcode view-->
     <string name="Passcode">Passcode Lock</string>
     <string name="ChangePasscode">Change Passcode</string>
-    <string name="ChangePasscodeInfo">When you set up an additional passcode, a lock icon will appear on the chats page. Tap it to lock the app.\n\nNote: if you forget the passcode, you\'ll need to delete and reinstall the app.</string>
+    <string name="ChangePasscodeInfo">When you set up an additional passcode, a lock icon will appear on the chats page. Tap it to lock the app.\n\nNote: If you forget the passcode, you\'ll need to delete and reinstall the app.</string>
     <string name="PasscodePIN">PIN</string>
     <string name="PasscodePassword">Password</string>
     <string name="EnterCurrentPasscode">Enter your current passcode</string>
@@ -148,7 +148,7 @@
     <string name="Vignette">Vignette</string>
     <string name="Shadows">Shadows</string>
     <string name="Grain">Grain</string>
-    <string name="Sharpen">Sharpen</string>
+    <string name="Sharpen">Sharpness</string>
     <string name="Fade">Fade</string>
     <string name="Tint">Tint</string>
     <string name="TintShadows">Shadows</string>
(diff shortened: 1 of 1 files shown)


MessengerProj/src/main/res/values/strings.xml
```diff
@@ -204,23 +204,23 @@
     <!--permissions-->
     <string name="PermissionContacts">Access to contacts needed to connect with your friends across all your devices.</string>
     <string name="PermissionStorage">Storage access needed to send and save photos, videos, music, and other media.</string>
-    <string name="PermissionNoAudio">Microphone access needed to send voice messages.</string>
+    <string name="PermissionNoAudio">Mic access needed to send voice messages.</string>
```
r10s: Is this abbreviation so common? I simply don't know :)

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -204,23 +204,23 @@
...
     <string name="Intro3Headline">Fast</string>
-    <string name="Intro3Message"><![CDATA[<b>Push messages</b>]]> in seconds.<![CDATA[<br/>]]>Rapid interface.</string>
+    <string name="Intro3Message"><![CDATA[<b>Delivery</b>]]> in seconds.<![CDATA[<br/>]]>Rapid interface.</string>
 
     <string name="Intro4Headline">Powerful</string>
     <string name="Intro4Message"><![CDATA[<b>Unlimited</b>]]> chats, images, videos, audio messages and more. Multi-client capable.</string>
 
     <string name="Intro5Headline">Free</string>
-    <string name="Intro5Message"><![CDATA[<b>Delta Chat</b>]]> is free forever.<![CDATA[<br/>]]>Open-source. No ads. No subscription. No vendor lock-in.</string>
+    <string name="Intro5Message"><![CDATA[<b>Delta Chat</b>]]> is forever gratis.<![CDATA[<br/>]]>Libre copyleft. No ads. No subscription. No vendor lock-in.</string>
```
r10s: This is not the same. I think the "free forever" mainly means "freedom" here, see https://en.wikipedia.org/wiki/Gratis_versus_libre

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -329,11 +329,11 @@
     <string name="AskStartChatWith">Chat with <![CDATA[<b>]]>%1$s<![CDATA[</b>]]>?</string>
     <string name="DeaddropHint">Tap the reply-arrows to start chatting.</string>
     <string name="NotYetImplemented">This function is not available or incomplete.</string>
-    <string name="DefaultStatusText">Sent with my Delta Chat Messenger: https://delta.chat</string>
+    <string name="DefaultStatusText">Corresponded under the aegis of Delta Chat Messenger: https://delta.chat</string>
```
r10s: ?

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -417,14 +417,14 @@
...
     <string name="AutocryptSetupMessageSubject">Autocrypt Setup Message</string>
-    <string name="AutocryptSetupMessageGeneralBody">This is the Autocrypt Setup Message used to transfer your key between clients.\n\nTo decrypt and use your key, open the message in an Autocrypt-compliant client and enter the setup code presented on the generating device.</string>
-    <string name="AutocryptSetupMessageTapBody">This is the Autocrypt Setup Message used to transfer your key between clients.\n\nTo decrypt and use your key, tap on this message.</string>
-    <string name="AutocryptKeyTransferPleaseEnterCode">Please enter the setup code that is displayed on the other device.</string>
-    <string name="AutocryptKeyTransferBadCode">Bad setup code. Please try again.\n\nIf you do not remember the setup code, just initiate the key transfer on the other device again.</string>
-    <string name="AutocryptKeyTransferSucceeded">Key transferred successfully. This device is now ready to use Autocrypt with the same key as the other device.</string>
+    <string name="AutocryptSetupMessageGeneralBody">The setup code must be entered on the other device to use your key there.</string>
```
r10s: I think, this is a little too short. The string will be the body of a message that may lack all context  of the app.
The hint with the Autocrypt-compliant client is important. Eg. the user may see the message in Thunderbird without Enigmail where taping in the message does not work.

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -17,7 +17,7 @@
     <string name="UnmuteNotifications">Unmute</string>
     <string name="Draft">Draft</string>
     <!--audio view-->
-    <string name="NoAudio">Files added to the music library of your device will appear here.</string>
+    <string name="NoAudio">Files from the music library.</string>
```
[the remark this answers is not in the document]
comradekingu: The old version states that by alluding to it more-so vividly, the new version implies it.
I think the gained time in reading it, and visual disturbance avoided, is worth more than what was removed amounts to for any tangible user.

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -59,9 +59,9 @@
     <string name="SendMessageTo">Send message to …</string>
     <string name="EnterGroupNamePlaceholder">Enter group name</string>
     <!--group info view-->
-    <string name="AddMember">Add member</string>
+    <string name="AddMember">Grant membership</string>
```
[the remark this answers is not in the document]
comradekingu: Add member has more dualities in English, resulting in more errors in translation, resulting in less precision for anglos, and a wider share of potential errors for anyone else.

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -59,9 +59,9 @@
...
     <string name="Notifications">Notifications</string>
-    <string name="RemoveMember">Remove member</string>
+    <string name="RemoveMember">Revoke membership</string>
```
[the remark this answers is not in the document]
comradekingu: Membership is more-so tied to also revoking further insight, and the person, in their ability to enjoy it.

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -83,7 +83,7 @@
     <string name="Default">Default</string>
     <string name="OnlyIfSilent">Only when silent</string>
     <string name="ChatBackground">Chat background</string>
-    <string name="SendByEnter">Send with Enter key</string>
+    <string name="SendByEnter">Send with \"⏎\"</string>
```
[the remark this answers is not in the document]
comradekingu: ⏎ is the ISO-standard denomination of return, which is not enter. Why is it left out?

MessengerProj/src/main/res/values/strings.xml
comradekingu: I would argue yes, for the current userbase, which in turn is a sub-ideal argument.
The pictogram for microphone is universally recognized.

MessengerProj/src/main/res/values/strings.xml
comradekingu: Me thinking you think you are right is the only concept i am certain about there.
That idea was also conveyed that way to me, which tells me something about me. What that same reasoning tells me it tells someone else, is any variation thereof.
The precision with which the intended concept of "free" and "open-source" is stated, (especially in that order,) allows for meaningless interpretation.

So, promising it to always to be gratis is then actually a new claim, and the connotation between that and the rest is troublesome.
How about: 
> "Delta chat **is copyleft**, forever libre. No ads …

MessengerProj/src/main/res/values/strings.xml
comradekingu: Meaning: Under the protection, sponsorship, or patronage of; under the supervision, control, or management of.
"Sent with" is the mark of uncanny comparison.

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -417,14 +417,14 @@
...
-    <string name="AutocryptSetupMessageTapBody">This is the Autocrypt Setup Message used to transfer your key between clients.\n\nTo decrypt and use your key, tap on this message.</string>
-    <string name="AutocryptKeyTransferPleaseEnterCode">Please enter the setup code that is displayed on the other device.</string>
-    <string name="AutocryptKeyTransferBadCode">Bad setup code. Please try again.\n\nIf you do not remember the setup code, just initiate the key transfer on the other device again.</string>
-    <string name="AutocryptKeyTransferSucceeded">Key transferred successfully. This device is now ready to use Autocrypt with the same key as the other device.</string>
+    <string name="AutocryptSetupMessageGeneralBody">The setup code must be entered on the other device to use your key there.</string>
+    <string name="AutocryptSetupMessageTapBody">Use this Autocrypt setup message to transfer your key between clients.\n\nTo decrypt and use your key, tap this message.</string>
```
comradekingu: This is the part that says "tap", which still states "Autocrypt setup message".
Does Autocrypt compliance somehow require a touchscreen?

MessengerProj/src/main/res/values/strings.xml
comradekingu: AutocryptSetupMessage_Subject_ states it is in fact the Autocrypt setup message
With this being AutocryptSetupMessageGeneral_Body_
See other comment. I am unsure how or what to reinstate, because why does the tapping appear out of context?

r10s: The tapping may just not be possible if the client is not autocrypt-capable. But the client will display the text nevertheless in this case. So the text should give the user a hint into the right direction. In fact, when displaying the message with an autocrypt-capable client, this client may replace the text by its own - eg. Thunderbird without Enigmail displays our text, Thunderbird with Enigmail its own text.

MessengerProj/src/main/res/values/strings.xml
r10s: *GeneralBody is the text that is written to the mail and may be displayed by any client.
*TapBody is the text that is displayed by Delta Chat on receiving a setup message (instead of the original text which is *GeneralBody if the message comes from another Delta client or a completely different text from other clients)

So, inside Delta Chat, the *GeneralBody messages are never displayed (exception: older Delta Chat clients that do not process the Setup message).

See https://delta.chat/assets/blog/20171215-asm-delta-enigmail.jpg to get an idea: the text in thunderbird is not the text provided by Delta Chat ...

Issue #236, opened 2018-01-02
remove requirement for contacts permission.


nutpantz:
The contacts database on android is less secure than a postcard.
While I can deny access to the permission delta chat complains it needs it to access your friends across all devices insistently.
A internal secure contract database would be far more secure than goggle and any unscrupulous developers information mine.
Or at least let me deny access just once understanding I have a reason to deny the permission 

I think this has been causing delta chat to freeze up requiring me to force closer it often.

r10s:
> Or at least let me deny access just once understanding I have a reason to deny the permission

Thank you for your post. At least since Mashmallow (Android 6/SDK 23), if you have denied the access to the contacts database once, Delta Chat should no longer ask for permissions. So, I am not sure, if this is still an issue. Which Android version are you using?

Regarding the access to the contacts database at all: We use it because it is used by many Android users and we want to allow the users to access their contacts in a convenient way. If users regard the contacts database as insecure and do not add their contacts to it, I think there is no additional security risk in reading an empty database.

nutpantz:
Yes using Samsung marsh mellow.
It asks me for contacts permission regularly
And since I started blocking it  I have had errors that delta chat is not responding and had to force close it almost as regularly. So there is definitely a issue there.

I understand most people use the built in contacts but then I know most people don't realize just what a security risk they are.
Promoting good security practices is essential

(Btw is deltacrypt compatible with k9mails new autocrypt support(not that I can figure out how to turn it on in k9mail))

r10s:
> Yes using Samsung marsh mellow.

Okay, don't know about the Samsung Android, but at least in stock Android, the permission should requested only once (however, the whole progress can be improved some time ...) 

> It asks me for contacts permission regularly

How often? Every time you open the contacts?

> And since I started blocking it I have had errors that delta chat is not responding and had to force close it almost as regularly. So there is definitely a issue there.

Okay, but _this_ issue may also come from the blocking (which, of course, could be prevented if Delta Chat would be more graceful)

> I understand most people use the built in contacts but then I know most people don't realize just what a security risk they are. [...]

I can imagine several risks, but which security risks are you talking about concretely?  Do you use the contact database at all? If not, you can easily give Delta Chat permissions to it (of course, you can also do so if you use it - Delta Chat, in contrast to most other messengers, does not send the contacts, or hashes from them, anywhere)

> (Btw is deltacrypt compatible with k9mails new autocrypt support(not that I can figure out how to turn it on in k9mail))

[...]

PR #251, opened 2018-02-11
Show the PermissionContact dialog only once

### LINKED ISSUE #83: Disable the "Give me permission to read contacts" dialog after it has been shown once
If the user denies the read-contacts permission, Delta Chat will show a dialog in which it explains what it needs the permission for. But this dialog is shown again after every reboot, which I think is annoying. 
I see no reason for which this could be useful because the user can still change his decision my enabling the permission in the system settings. (And if the user does not read the message properly, he probably won't do it if you show him again and again, either.)

Hocceruser:
Fixes #83 and partly #236. What do you think?

### DIFF
--- a/MessengerProj/src/main/java/com/b44t/messenger/LaunchActivity.java
+++ b/MessengerProj/src/main/java/com/b44t/messenger/LaunchActivity.java
@@ -767,8 +767,7 @@ public class LaunchActivity extends Activity implements ActionBarLayout.ActionBa
                             }
                             break;
                         case Manifest.permission.WRITE_EXTERNAL_STORAGE:
-                            // Do not show the StoragePermission dialog as this permission is not needed yet (only for chats)
-                            grantedCount++;
+                            msg += "- " + ApplicationLoader.applicationContext.getString(R.string.PermissionStorage) + "\n\n";
                             break;
                         case Manifest.permission.RECORD_AUDIO:
                             msg += "- " + ApplicationLoader.applicationContext.getString(R.string.PermissionNoAudio) + "\n\n";
--- a/MessengerProj/src/main/java/com/b44t/messenger/LaunchActivity.java
+++ b/MessengerProj/src/main/java/com/b44t/messenger/LaunchActivity.java
@@ -755,7 +755,7 @@ public class LaunchActivity extends Activity implements ActionBarLayout.ActionBa
                 else {
                     switch (permissions[i]) {
                         case Manifest.permission.READ_CONTACTS:
-                            SharedPreferences preferences = ApplicationLoader.applicationContext.getSharedPreferences("general", Context.MODE_PRIVATE);
+                            SharedPreferences preferences = ApplicationLoader.applicationContext.getSharedPreferences("mainconfig", Context.MODE_PRIVATE);
                             if (preferences.getBoolean("PermissionContactsDialogShown", false)) {
                                 // The dialog was already shown once
                                 grantedCount++;
--- a/MessengerProj/src/main/java/com/b44t/messenger/LaunchActivity.java
+++ b/MessengerProj/src/main/java/com/b44t/messenger/LaunchActivity.java
@@ -767,7 +767,8 @@ public class LaunchActivity extends Activity implements ActionBarLayout.ActionBa
                             }
                             break;
                         case Manifest.permission.WRITE_EXTERNAL_STORAGE:
-                            msg += "- " + ApplicationLoader.applicationContext.getString(R.string.PermissionStorage) + "\n\n";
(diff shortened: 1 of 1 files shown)


MessengerProj/src/main/java/com/b44t/messenger/LaunchActivity.java
```diff
@@ -753,7 +755,16 @@ public void onRequestPermissionsResult(int requestCode, String[] permissions, in
                 else {
                     switch (permissions[i]) {
                         case Manifest.permission.READ_CONTACTS:
-                            msg += "- " + ApplicationLoader.applicationContext.getString(R.string.PermissionContacts) + "\n\n";
+                            SharedPreferences preferences = ApplicationLoader.applicationContext.getSharedPreferences("general", Context.MODE_PRIVATE);
```
r10s: instead of "general", we should use "mainconfig" which is also used to store most of the other preferences.

MessengerProj/src/main/java/com/b44t/messenger/LaunchActivity.java
```diff
@@ -753,7 +755,16 @@ public void onRequestPermissionsResult(int requestCode, String[] permissions, in
...
+                                // The dialog was already shown once
+                                grantedCount++;
+                            } else {
+                                msg += "- " + ApplicationLoader.applicationContext.getString(R.string.PermissionContacts) + "\n\n";
+                                SharedPreferences.Editor editor = preferences.edit();
+                                editor.putBoolean("PermissionContactsDialogShown", true);
+                                editor.apply();
+                            }
                             break;
                         case Manifest.permission.WRITE_EXTERNAL_STORAGE:
```
r10s: WRITE_EXTERNAL_STOREAGE and RECORD_AUDIO seems not be a problem as they're queried only on explicit user actions?

Hocceruser: Actually, WRITE_EXTERNAL_STORAGE is also queried on every startup, and only RECORD_AUDIO is not (I didn't realize this).

r10s: Yes, but querying the WRITE_EXTERNAL_STORAGE permission comes after a user action, so I think this is always fine to ask here. Same as for RECORD_AUDIO - if the user _clicks actively_ on the microphone and the app has not the permission to record audio, it is fine to ask the user, even a thousand times - what else can the app do?

nutpantz:
The app freezes if contacts permission is not given and it fails to receive new messages

r10s:
> The app freezes if contacts permission is not given and it fails to receive new messages

can you provide more information? have you self-compiled the app? at least on my test devices I cannot reproduce this. is the problem new after this fix?

nutpantz:
What do you need to know? I using .13 and I block contacts when the permission is asked. It runs fine for a while (hours)but will freeze eventually and not check messages until restarted. I had no freezes until I stared blocking contracts.

r10s:
ahh, okay, so it does not have sth. to do with this PR.

nutpantz:
I'm on delta chat .16 and I'm still having delta chat freeze up and stop receiving messages because I don't allow access to android contacts. When I force close it and it restarts I get the old messages.

r10s:
@nutpantz might be that it does not have to do anything with the denied contact access. have to you tried to allow contact access - does the freeze _not_ happen then? (don't be afraid for testing this out - the contacts do not leave the device, there is not even code to contact other servers but the given smpt/imap server. alternatively, maybe you can also test it out with some fake contacts).

nutpantz:
Yes I have tested with access allowed. It only happens with access denied as far as I can tell.

r10s:
@nutpantz thanks, this is helpful. is there also an issue about this? if not, can you please create one and refer this thread (otherwise, i fear, it will be overseen here :)

nutpantz:
An issue other than it freezing and not reading new messages until it is closed and restated?

r10s:
@nutpantz i think, the description is just fine. 

however, currently these post are added as an _comment below a closed pull request_. i would suggest, if not yet done, to add it as an _real issue to the issue tracker_ here: https://github.com/deltachat/deltachat-android/issues . this way, it stays visible, can be flagged as a but etc.

nutpantz:
I have tried before. The issue was closed.
https://github.com/deltachat/deltachat-android/issues/236.

r10s:
the closed #236 issue was about not using the contact database at all. this is sth. different than this concrete reproducible bug imho. the result was eg. this PR, so i think this can stay closed.

i would suggest to open a new issue where the concrete crash is mentioned. you can reference this PR and the old issue from there.

Issue #168, opened 2017-09-17
Does not work with office 365 or outlook


Karkus476:
App refuses to authenticate with office 365 or outlook for smtp. (STARTTLS)
Server: smtp.office365.com/smtp-mail.outlook.com Port: 587
I am using my full email address as login name and the same password as for IMAP (which I think works).
For thunderbird, I have to set the authentication to "Normal Password" (and then I am able to send mail just fine.)
Thanks.

Wiredframe:
I can confirm that I can't login either. I'm using a Microsoft Exchange 2013 account. Any chance this will be fixed in a future release? 👍

r10s:
> Any chance this will be fixed in a future release?

Of course there are chances, it's just a thing of time and money ...

CitoplasmaX:
You can create al account for free (of charge).

CitoplasmaX:
An*

Wiredframe:
Thanks. I just added an alias adress and set it to IMAP. 👍

QuibblingAsh42:
Confirmed it's not working for me, either.  Hopefully you'll be able to find something.

Ampli-fier:
Performed test with own account and got:

`hh:mm:18.402  I com.b44t.messenger: No worthy mechs found`
`hh:mm:18.406  E DeltaChat: Error: SMTP-login failed for user xxxx@xxx.xxx (Login failed)`

Sent full logs (logcat zipped) to r10s.

Ampli-fier:
> `hh:mm:18.402 I com.b44t.messenger: No worthy mechs found`

I think this is related to missing configuration in the Android UI, which leads to wrong setting of the `server_flags`.
When I tested with **Delta Chat Core**, I set the value for `server_flags` to `66052`. With this value I was able to connect (to office365 server).

Here the calculation:

`4 (MR_AUTH_NORMAL) + 10000 -> (MR_SMTP_SOCKET_STARTTLS) + 200 -> (MR_IMAP_SOCKET_SSL)`
`=> 10204 (hex) => 66052 (dec)`

When I use Android and try to connect, I only see this configured in sqlite db:
`11|server_flags|66048`

After changing the value in sqlite manually to `66052`, the error message from above is gone (I am still not able to connect successfully, though. I receive a `Connection refused`).

Searching the web leads to some hints regarding authentication mechanism ("**mechs**") problems, too.

The authentication mechanism setting (flag) is not available in the UI and not used. I think this must be added.

[...]

PR #295, opened 2018-04-14
enable STATIC_LOGIN mech

Ampli-fier:
Enable STATIC_LOGIN mech to make it possible to use the authentication-type PLAIN

With this change it was possible to log in to Office365 SMTP server, see #168

### DIFF
--- a/MessengerProj/jni/cyrussasl/include/sasl/config.h
+++ b/MessengerProj/jni/cyrussasl/include/sasl/config.h
@@ -403,7 +403,7 @@
 /* #undef STATIC_LDAPDB */
 
 /* Link LOGIN Staticly */
-/* #undef STATIC_LOGIN */
+#define STATIC_LOGIN
 
 /* Link NTLM Staticly */
 /* #undef STATIC_NTLM */


r10s:

wondering if STATIC_LOGIN is set in the CLI-program that just uses the system lib.

Ampli-fier:
Searching the [github repo](https://github.com/cyrusimap/cyrus-sasl/search?q=STATIC_LOGIN&type=Code) shows a positive match in the [win32/include/config.h](https://github.com/cyrusimap/cyrus-sasl/blob/b6767cc9592c21ed42504ababa9e1a3b7c026fde/win32/include/config.h) file. No idea where header files for other OSs can be found.

In my distribution, I found a match in the development package:

```
id@VirtualBox ~ $ zcat /usr/share/doc/libsasl2-dev/examples/config.h.gz | grep _LOGIN
#define STATIC_LOGIN /**/
id@VirtualBox ~ $ 
```

**And** there is a LOGIN module in the Perl package included:

```
id@VirtualBox ~ $ dpkg-query -L libauthen-sasl-perl | grep LOGIN
/usr/share/man/man3/Authen::SASL::Perl::LOGIN.3pm.gz
/usr/share/perl5/Authen/SASL/Perl/LOGIN.pm
id@VirtualBox ~ $
```

Issue #293, opened 2018-04-12
Explain how to trigger a key exchange


holgerjakobs:
0.16

**Expected behavior**
When a contact deletes and reinstalls Delta chat, my app still has his old, now invalid key, so it doesn't decrypt his messages, which show as unencryptable messages in regular mail client.

Deleting the contact would probably lead to a new key exchange. But unfortunately, deleting a contact which is "in use" is disallowed.

How to trigger a new key exchange?

Please repair!

r10s:
> How to trigger a new key exchange?

the person who cannot read a message eg. because of a new device should send a message "hey i cannot read your message, please send again" - the answer should be readable and the keys will be exchanged.

r10s:
> Is the person who cannot read a message getting an error that explains this?

currently, there are no really clear advices, this could be improved.

testbird:
Is there a current error message in the code to improve?

r10s:
not yet, but i've just modified the code to have such an error message. the following text

```
The messsage was encrypted to a different setup.

Send any message to the contact to share your new
setup; eg. ask the contact to send the message again.
```

is currently shown in the message bubble if a message cannot be decrypted. Any suggestions and improvements welcome before we give this to the translators (@comradekingu :)

testbird:
Is there another regular cause for this then a re-installation? It could be more sensible to advice the local user towards key transfer instead of asking (all) contacts (and them get accustomed) to re-sending messages to with a new key.

```
The message can not be read, because it was not encrypted for your setup. Usually 
this only happens if you use a new installation without importing a preexisting setup.

You should now import the configuration from your preexisting installation. (Using the 
"setup message" feature or the key-file export and import.)

If you write to contacts without first importing the setup for which the message 
was encrypted, you will share your new setup. This would mean all your contacts 
would have to re-verify your new setup, and your preexisting installations would not 
be able to read messages encrypted for the new installation.
```

testbird:
I modified the text to also be shown whenever DC finds non-decryptable messages on the server (after installation), and not only when viewing a new message.

testbird:
Shorter, without any explanation:

```
The messsage was encrypted for a different setup.

Import the proper setup (e.g. through a "setup message"),
or send any message to the contact to share your new setup 
(e.g. ask the contact to send the message again).
```

r10s:
> Is there another regular cause for this than a re-installation?

[...]

PR #299, opened 2018-04-20
decrypt error: change advice order to: if... might have to import, otherwise...

testbird:
https://github.com/deltachat/deltachat-android/issues/293


r10s:
closing the pr as the wording differs from the discussions at https://github.com/deltachat/deltachat-android/issues/293#issuecomment-383378987 ; i think further discussion should be continued there to avoid fragmentation :)

PR #310, opened 2018-05-18
Small language rework

comradekingu:


MessengerProj/src/main/res/values/strings.xml
```diff
@@ -204,11 +204,11 @@
     <!--permissions-->
     <string name="PermissionContacts">Access to contacts needed to connect with your friends across all your devices.</string>
     <string name="PermissionStorage">Storage access needed to send and save photos, videos, music, and other media.</string>
-    <string name="PermissionNoAudio">Microphone access needed to record voice messages.</string>
```
r10s: i think "microphone" is more usual.
also android itself says "microphone".

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -413,7 +413,7 @@
     <string name="Never">Never</string>
     <string name="AutocryptKeyTransferInitiate">Send Autocrypt Setup Message</string>
     <string name="AutocryptKeyTransfer">Autocrypt Setup Message</string>
-    <string name="AutocryptKeyTransferMsgBeforeSend">An <![CDATA[<i>]]>Autocrypt Setup Message<![CDATA[</i>]]> securely shares your end-to-end setup with other Autocrypt-compliant devices.\n\nThe setup will be encrypted by a setup code which is displayed here and must be typed on the other device.</string>
+    <string name="AutocryptKeyTransferMsgBeforeSend">An <![CDATA[<i>]]>Autocrypt Setup Message<![CDATA[</i>]]> securely shares your end-to-end setup with other Autocrypt-compliant apps.\n\nThe setup will be encrypted by a setup code which is displayed here and must be typed on the other device.</string>
```
r10s: it's the app, that is Autocrypt-compiant, not the device.

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -204,11 +204,11 @@
     <!--permissions-->
     <string name="PermissionContacts">Access to contacts needed to connect with your friends across all your devices.</string>
     <string name="PermissionStorage">Storage access needed to send and save photos, videos, music, and other media.</string>
-    <string name="PermissionNoAudio">Microphone access needed to record voice messages.</string>
+    <string name="PermissionNoAudio">Mic access needed to record voice messages.</string>
     <string name="PermissionOpenSettings">Settings</string>
     <!--Intro view-->
     <string name="Intro1Headline">Delta Chat</string>
-    <string name="Intro1Message">The messenger with the <![CDATA[<b>broadest audience</b>]]> in the world. <![CDATA[<b>Free</b>]]> and <![CDATA[<b>independent</b>]]>.</string>
+    <string name="Intro1Message">The messenger with the <![CDATA[<b>broadest audience</b>]]> in the world. <![CDATA[<b>Libre</b>]]> and <![CDATA[<b>independent</b>]]>.</string>
```
r10s: i would leave this sentence untouched, for now,  this would require a broader discussion. "Free" is unprecise, of course, however, much more known as "libre".

MessengerProj/src/main/res/values/strings.xml
```diff
@@ -220,7 +220,7 @@
     <string name="Intro4Message"><![CDATA[<b>Unlimited</b>]]> chats, images, videos, audio messages and more. Multi-client capable.</string>
 
     <string name="Intro5Headline">Free</string>
-    <string name="Intro5Message"><![CDATA[<b>Delta Chat</b>]]> is free forever.<![CDATA[<br/>]]>Open-source. No ads. No subscription. No vendor lock-in.</string>
+    <string name="Intro5Message"><![CDATA[<b>Delta Chat</b>]]> is forever gratis.<![CDATA[<br/>]]>Copylefted libre software. No ads. No subscription. No vendor lock-in.</string>
```
r10s: same here, i would leave this sentence untouched, this would require a broader discussion:
the "free" here means more than "gratis" here (maybe the icon is not correct).
Also "Open-source" is more known than "Copylefted libre".

MessengerProj/src/main/res/values/strings.xml
comradekingu: Which is why I changed it to "app"  ┐(o.O)┌

MessengerProj/src/main/res/values/strings.xml
comradekingu: Which is when it gets localized into, and equivalents of "gratis", that duality is unambiguous in how it only translates into price.

MessengerProj/src/main/res/values/strings.xml
r10s: *sorry* :)

MessengerProj/src/main/res/values/strings.xml
comradekingu: Guessing you meant "than libre", which is true, it is also true equal capacity of "free" being more-so a price concern to the lesser known "free software.

MessengerProj/src/main/res/values/strings.xml
r10s: yes, but i have never meant only the price here.
it's free in many senses, including price, but not limited to that.
might be that translators will translate if to "gratis", which would be a wrong translation then, however, it's wrong in _one_ locale then - if we change the meaning of the english source, _all_ locales will have this error.
but we may add translator hint above the line.

as mentioned, the wording itself i would leave untouched for the moment.

MessengerProj/src/main/res/values/strings.xml
testbird: Maybe use a known term even partly understandable without full knowledge. `Fully public, independent Free Software.`?

r10s: might be.

but, as mentioned, i would prefer a broader discussion on this point. i would leave it "as is" for now and would be happy to merge the other changes. also i am not really free to dive in here just now. feel free to create another pr with this change or an issue so that it won't get forgotten.

comradekingu:
@r10s Reverted :)

PR #316, opened 2018-05-27
Fix typo and spacing in strings.xml

flortsch:
This commit fixes a typo and removes a wrong space from the german strings.xml file.


r10s:
thanks :)

PR #317, opened 2018-05-28
Spelling: A similie ate ion, redux

comradekingu:
Problematic things I was not able to solve

"system root"

"notifications" is repeated a lot
It could be

> **Notifications**
Type 1
Type 2
Condition 1
Condition 3

<string name="RepeatNotifications">Repeat notifications</string>
What is this?

"passcode"
I don't really like this word.

"Tap the reply-arrows"  
Not sure what this is
"Tap _arrowsymbol_ to reply"?

The claim for "business use" Should say something about transport-encryption?

"among other places" is not precise

### DIFF
--- a/MessengerProj/src/main/res/values/strings.xml
+++ b/MessengerProj/src/main/res/values/strings.xml
@@ -2,7 +2,7 @@
 
 
 
-<resources xmlns:tools="http://schemas.android.com/tools">
+<resources xmlns:tools="https://schemas.android.com/tools">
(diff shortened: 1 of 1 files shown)


r10s:
as this branch has conflicts in between, i picked the "clear" corrections and typos to https://github.com/deltachat/deltachat-android/commit/952eb60aff949b983e8d88f66aa8e4c0e10c672d#diff-dd26e77572b072235fca5a9bff505987 (sorry got to the wrong commit by accident :) and close this one.

Issue #319, opened 2018-06-02
Attaching Picture from Camera not working on Android 7.0


t-oster:
I just built from Source and when I try to attach something from Camera I get:
```
java.lang.SecurityException: Permission Denial: starting Intent { act=android.media.action.IMAGE_CAPTURE flg=0x3 cmp=com.mediatek.camera/com.android.camera.CaptureActivity clip={text/uri-list U:content://com.b44t.messenger.beta.provider/media/Pictures/Delta%20Chat/IMG_20180602_101413.jpg} (has extras) } from ProcessRecord{e7eb223 4748:com.b44t.messenger.beta/u0a234} (pid=4748, uid=10234) with revoked permission android.permission.CAMERA
```
maybe related to #259 and #86.
I will look into it and maybe send a PR if it's just missing permissions in the manifest file.

t-oster:
Also my bad. Somehow the permission was disabled in the app settings. No idea how this happened, but as long as I cannot prove it it may be my bad

PR #320, opened 2018-06-02
Request permissions to use the Camera before starting Capture

t-oster:
At least on my phone (Android 7.0) the ACTION_IMAGE_CAPTURE and ACTION_VIDEO_CAPTURE Intents silently fail if the app does not have the camera permission. (#319).
I use the same code like it is used on other places to request for that permission if it is not granted.
I tested it and it works as expected.

### DIFF
--- a/MessengerProj/src/main/java/com/b44t/messenger/ChatActivity.java
+++ b/MessengerProj/src/main/java/com/b44t/messenger/ChatActivity.java
@@ -1211,7 +1211,7 @@ public class ChatActivity extends BaseFragment implements NotificationCenter.Not
     private void processSelectedAttach(int which) {
         if (which == ChatAttachAlert.ATTACH_BUTTON_IDX_CAMERA ) {
             if (Build.VERSION.SDK_INT >= 23 && getParentActivity().checkSelfPermission(Manifest.permission.CAMERA) != PackageManager.PERMISSION_GRANTED) {
-                getParentActivity().requestPermissions(new String[]{Manifest.permission.CAMERA}, 4);
+                getParentActivity().requestPermissions(new String[]{Manifest.permission.CAMERA}, LaunchActivity.REQ_CAMERA_PERMISSION_ID);
                 return;
             }
             try {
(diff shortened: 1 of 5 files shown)


r10s:
always thought, the camera-permission is not needed when just firing an intent.

might be that this is an [android bug](https://stackoverflow.com/questions/32789027/android-m-camera-intent-permission-bug#32856112), however, the fix looks reasonable.

Not sure about the requestCode `4` here, this results in [creating directories here](https://github.com/deltachat/deltachat-android/blob/master/MessengerProj/src/main/java/com/b44t/messenger/LaunchActivity.java#L780) which are related to another permisson (sorry about the weired numbers, they come from the Telegram source ... wherever I came across them, i have converted them to a define, eg. REQ_CONTACT_N_STORAGE_PERMISON_ID).

Looks as if thee requestCode is not used outside [onRequestPermissionsResult()](https://github.com/deltachat/deltachat-android/blob/master/MessengerProj/src/main/java/com/b44t/messenger/LaunchActivity.java#L746), so maybe we should just use a new ID (possibly as a constant then) and add it to [this line](https://github.com/deltachat/deltachat-android/blob/master/MessengerProj/src/main/java/com/b44t/messenger/LaunchActivity.java#L748)

r10s:
@t-oster i've added a unique id for the camera permission and using it instead of the weired `4`

moreover, i've added an explanation string if the permission is not granted by the user; i am not 100% sure if this is really needed, however, it is done for the other permissions, so, for now we just do the same for the camera-permissions :)

as you mentioned via mail, currently, when the permissions are granted, the user has to tap the desired action again. 
this is not perfect, however, due to the complicated flow of the request code (Telegram routes everyting to a single activity), I would leave this "as is" for now ... when recording voice-messages and for other permissions, this is the same. i think there are more important things to fix :)

t-oster:
Successfully tested. I agree since we may switch to the signal codebase, this is enough for now.

r10s:
okay, then i'm happy to merge :)

Issue #55, opened 2017-03-20
Picture Viewer quality is not corresponding to resolution


bokf:
I have noticed, that the pictures a resized to ca. 1280x960 pixel. If I open the pictures by tapping on it, the quality is not corresponding to this resolution. If I open the same picture outside of Delta Chat, the quality is much better. Opening the picture on a tablet in Delta Chat the picture is not very clear and not nice to look at. I would like the internal viewer to show all the pixels or use an external picture viewer.

r10s:
Thank you for your report, I've also notices this issue some day - but forgot it in between.

I will fix this ASAP.

r10s:
NB: In between, you can display images in high-quality by **long-pressing** them and then select **Open** from the upper right menu.

angelo-fuchs:
Hypothesis: In ChatMessageCell:1428 -> The thumb image and the actual image are the same on first load.
Maybe the actual image gets thumbed, but stored under the original size of the image in the list of images (or the image in cache gets overwritten by the thumb)

angelo-fuchs:
Seems like "photoImage.setNeedsQualityThumb(true)" does it...

testbird:
Noticed a newly received photo first looked blured in chat, and clicking it opened if fine in zoomable in full view, but going back into the chat it switched to fine in the chat, unzoomable blurred in full screen.

PR #332, opened 2018-06-13
displaying a list of pictures instead of just one when opening a pic in chat

### LINKED ISSUE #10: Paging through images does not yet work
For now, you can only open one image at a time in the chat view.

We should add a functionality, where swiping left/right goes to the previous/next image.

angelo-fuchs:
This fixes #10.
Most action was in selecting a different segment of the openPhoto function in PhotoViewer (as indicated by method openPhotoList).
In ChatActivitiy I just gather together the pictures of this chat (currently all I can find, maybe there should be a limit...?)

change in line (original) 2670 (now 2679) should be reviewed: I suppress the blurring of thumbnails (thumbFilter "b" -> thumbFilter null). I did this because the list view kept showing the blurred thumbs for pictures that were to small to require a thumb in the first place.
I could not find the place in the code where the switch from "blurred thumb" to "clear image" would happen, but I assume there is a better solution then to just suppress the thumb blurring in the first place.


r10s:
great :) taggling #55 next makes sense (not sure if you got it on irc :)

PR #335, opened 2018-06-17
updated ISSUE_TEMPLATE.md

violoncelloch:
* commented out our statements so we don't have them visible in issues if ppl. don't remove them
* created a `details` section so long logs are expandable


r10s:
i do not know much about the possibilities in the ISSUE_TEMPLATE, but the changes look reasonable to me.

Issue #178, opened 2017-10-09
gmail not work


Suburbanno:

r10s:
Gmail works if you enable "Less secure apps".

danito:
And if you have 2fa enabled, you need to create an app id for DC see #20 
For me it works, but sometimes it takes several minutes before it synchronizes. And I don't know if this is a bug or a feature, but I get also replies for "non-chat" emails in DC.

r10s:
as mentioned, please refer to #20

Issue #20, opened 2017-02-03
Problem with two-factor sign-in verification


YoungNFoolish:
Hi,
first off, such a great idea man! surely it would be useful for everyone!
The two step sign-in verification is my problem to use Delta Chat! Actually when i tried to sign-in via my Gmail account, it ended with an error. Could you plz fix it?
Sorry for my bad english!

r10s:
We know about this issue, however, currently it is not clear to me how this can be fixed. Any help on this point would be welcome.

In between, you can enable support for "less secure apps"
https://support.google.com/accounts/answer/6010255 (the name is a little bit misleading, of course, Delta Chat uses all secure standards, however, there seems to be some proprietary google stuff)

I've also seen some Googlemail accounts where a simple "UnlockCaptcha" is sufficient, see https://gmailhelp.com/unlocking-googles-gmail-captcha/

Again, if anyone can help on these points, this would be very welcome.

PS: Your english seems to be perfect, sorry for mine.

petmos:
To my mind k9mail uses OAuth2 for IMAP, too.
https://github.com/k9mail/k-9/tree/master/k9mail-library/src/main/java/com/fsck/k9/mail/oauth
and
https://github.com/k9mail/k-9/commit/7774ebc788c8e33ac7994b6cea6614cbd4f68d81

Perhaps this helps you.

LordTyrius:
I am also using google 2fa. I have enabled "less secure apps", but to use mail clients like kmail or delta I need to generate and use a special "!app password" for just that application. (More information: https://security.google.com/settings/security/apppasswords)

danito:
I confirm that creating an application password works for Delta when you have 2fa enabled. (without enabling less secure apps). Thanks for the tip.

SaintDI:
App password works well, but I strongly suggest to write a little tip message when new account is configured, because it isn't clear for average users.

> e.g. "Please, generate an app password on your mail account if you're use 2FA autentication method."

ravilov:
Any chance of implementing OAuth2 in DeltaChat? I assume that's why we need to enable less secure apps in Gmail. I really hate having to enable that, plus it doesn't look well for an app that's trying to appeal to business users. Honestly having to enable that setting is right now the only reason I'm not using DC more seriously.

rizzopablo:
+1

r10s:
+1 :)
in fact, we just need someone who can do this resp. has the time to do so :)

[...]

PR #336, opened 2018-06-22
Add Gmail specific info in setup.

t-oster:
This PR adds a GMail specific text with links to `enable less secure apps` or `create an app password` to the setup dialog, if the E-Mail adress is recognized as GMail (`@gmail.` or `@googlemail.`).
It addresses #20 (#178).

screenshot_20180622-073030

### DIFF
--- a/MessengerProj/src/main/java/com/b44t/messenger/Cells/TextInfoCell.java
+++ b/MessengerProj/src/main/java/com/b44t/messenger/Cells/TextInfoCell.java
@@ -23,6 +23,7 @@
 package com.b44t.messenger.Cells;
 
 import android.content.Context;
+import android.text.method.LinkMovementMethod;
 import android.util.TypedValue;
 import android.view.Gravity;
 import android.widget.FrameLayout;
@@ -67,12 +68,25 @@ public class TextInfoCell extends FrameLayout {
 
     public void setText(CharSequence text)
     {
-        setText(text, null, true);
+        setText(text, null, true, false);
     }
 
     public void setText(CharSequence text, CharSequence icon, boolean borderBotton)
     {
+        setText(text, icon, borderBotton, false);
+    }
+
+    public void setText(CharSequence text, CharSequence icon, boolean borderBotton, boolean clickableHtml)
+    {
         textView.setText(text);
+
+        if( clickableHtml ) {
+            //allow for clickable Links in TextViews
+            //https://stackoverflow.com/questions/2734270/how-do-i-make-links-in-a-textview-clickable
+            //as this avoids the onClickListener to be called (used at least for the "expand" icon in the AccountSettings)
+            //we do this only explicitly
+            textView.setMovementMethod(LinkMovementMethod.getInstance());
+        }
 
         FrameLayout.LayoutParams lp = (FrameLayout.LayoutParams) textView.getLayoutParams();
         lp.width        = LayoutHelper.WRAP_CONTENT;
--- a/MessengerProj/src/main/java/com/b44t/messenger/SettingsAccountFragment.java
+++ b/MessengerProj/src/main/java/com/b44t/messenger/SettingsAccountFragment.java
@@ -31,10 +31,14 @@ import android.app.ProgressDialog;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.os.Bundle;
+import android.text.Editable;
+import android.text.Html;
 import android.text.InputType;
+import android.text.TextWatcher;
 import android.view.View;
 import android.view.ViewGroup;
 import android.widget.AdapterView;
+import android.widget.Button;
 import android.widget.FrameLayout;
 import android.widget.ListView;
 
@@ -150,6 +154,7 @@ public class SettingsAccountFragment extends BaseFragment implements Notificatio
         rowAddr          = rowCount++;
         rowMailPwHeadline= rowCount++;
         rowMailPw        = rowCount++;
+        rowInfoBelowSendPw = rowCount++;
         rowOpenAdvOpions = rowCount++;
 
         if( m_expanded ) {
@@ -183,7 +188,6 @@ public class SettingsAccountFragment extends BaseFragment implements Notificatio
             rowSendSecurity  = -1;
         }
 
-        rowInfoBelowSendPw = rowCount++;
     }
 
     @Override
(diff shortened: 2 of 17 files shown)


r10s:

i've just added a commit that allows the advanced options to be expanded (the onClickListener was no longer called due to LinkMovementMethod).

however, i would suggest to simplify this dialog a bit:
- merge MyAccountExplain2 into MyAccountExplain and show it directly after the password field (so that there is only one place for a hint)
- MyAccountExplain could be replaced _completely_ by MyAccountExplainGmail using the new logic
- instead of the "+" icon i would add a button  "Show more options..." (or a link or so) below the explaining text (i've seen more than one person not getting that the "+" is a button)

what do you think?

t-oster:
agreed

r10s:
great, if you run into troubles with the weird settings objects, please let me know.

as the "+" icon and the gmail-text would share the same item, we should replace it with a "more options ..:" link, so we won't get in trouble with link-click vs. onclick-handler.

t-oster:
yes

violoncelloch:
Wouldn't it be more clear to place the + plus on a white space to indicate it's a collapsed input area?

t-oster:
screenshot_20180623-083636
screenshot_20180622-073030
screenshot_20180623-083621
screenshot_20180623-083631

t-oster:
To-Do:
- [ ] make `advanced opations` (at least Look Like) a Button
- [ ] remove whitespace-changes
- [ ] fix german Translation
- [ ] squash everything to one commit

r10s:
great :)

> make advanced opations (at least Look Like) a Button

i would just made the field look like a link (same as the links on the second screenshot) and remove the "+".

r10s:
> I think it would be prettier and easier to understand if we place "advanced settings... +" on white background (input area) to make it clear it's an expandable input location

i would be fine with both. my general idea is that these "advanced settings" should not be needed for the majority of users, so it seems okay if it's not too disturbing. but a normal black-on-white button below the explaining text seems to be also very okay :) and would have the advantage that this is a know element.

violoncelloch:
Ok, if we see it this way a link or something less obvious seems ok too

t-oster:
While i'm at it, is there another provider needing special Info?

testbird:
Possible source: https://github.com/deltachat/deltachat-core/wiki/Provider-overview

testbird:
Don't remember the provider, but someone could not connect using DC right away.
Only later he found the provider had send him an email to confirm enabling email client access.
So maybe you could add to the general hint to check the emails in the usual way if login always fails despite the correct credentials.

violoncelloch:
@t-oster I think a problem could be that corporates using business domains but google or other services in the background are not detected as google mail through the actual algorythem. Are imap settings automatically detected through /.well-known or DNS Records for googles imap and smtp servers? So maybe it would be useful to provide these extra infos after the detection of a google imap server as well even though it's not `@gmail.` or `@googlemail.`

testbird:
Generally inform about leaking your contacts including their IPs to `$PROVIDER` ?

t-oster:
@testbird what exactly are you referring to?

t-oster:
@violoncelloch I don't think we currently get back the autoconfig values, so I am not sure if I can detect business domains with gmail backend. I could however also change the text if the user manually inputs a gmail smtp/imap server. However at this stage, the user is already below the `advanced settings` area and probably knows what he's doing(?).
Do you have an idea on how many people are using gmail with business accounts? Do you have an example in order to check if autoconfig even works?

t-oster:
The PR is now rebased and squashed.

20180625080843_image
20180625080844_image
20180625080842_image

violoncelloch:
@t-oster 
I agree, if the user is already in the advanced settings he most probably knows what he's doing and how to handle thes gmail things, so this doesn't make much sense...
 I don't have an account to test with. I only know there are quite a few corporates relying on google services.

r10s:
thanks, looks great :)

i've slightly modified the button in https://github.com/deltachat/deltachat-android/commit/8db7c9f45b512972138d125108962a4007188ff5 , must admit that a normal-looking button may be the better choice :)

screenshot_20180626-015824

testbird:
> what exactly are you referring to?

Was thinking about a GDPR complient notice. Something like
```
By using (not self-hosted) email providers, you transmit personal data 
(including IP addresses) to the email provider, and the GDPR requires 
you to disclose this to your contacts.

Your current email provider is: X operated by Y.
```

t-oster:
well. I think this is a bit off topic, because this issue was to help users getting DC to work and not to tell the user what he should know when using E-mail in general. Also I am not sure if the GDPR applies to private E-Mail users. At least I did not see such notice in any other MUA.

testbird:
Hi, it would sure be another issue. The comments / violoncelloch just touched on that more general issue.

When programing your own MUA it may not be that relevant. But if software is to be used by others, especially in business, users need to know what data is transfered.

It would be a good thing if DC would also be first to meet this. New issue?

t-oster:
You can open a new issue if you think this is relevant. I would not add this warning in the configuring section, because it would confuse users. Maybe we have to add a generic "Datenschutzerklärung" but since we are not the provider, I think this is not necessary and I don't think any other E-Mail client will add such a notice. It is obvious to me that if I send an E-Mail, it contains the E-Mail Address of the receipient.... I also don't add a notice to my snail-mail regular letters to inform the user that the postmaster will read it.
Also IMHO I do not send the IP address of the receipient to google, but the receipient does when answering the mail, so his MUA should warn him.

So long story short: IMHO we have much more important stuff to do than adding confusing and obvious information, but I may be wrong.

testbird:
I agree that coding wise there may be more important stuff, nevertheless, users have legal obligations, not being aware can easily happen, but is no excuse. DC could just be helpful and make a good example.

>  It is obvious to me that if I send an E-Mail, it contains the E-Mail Address of the receipient....

And yours, paired with your IP! Allowing any obvious intermediate providers (and those hidden under a different flag) to personalize your -analytics.

A comment above noted that the email domain does not always correspond with the actual provider.

testbird:
Thanks for the link, very good. I propose an attempt to sort the bullets into more easily comprehensible points https://github.com/deltachat/deltachat-pages/pull/89

It boils down to exactly the same crucial point: the email provider.

However, as users have no control about the email provider that others (incl. individual private users) use to contact them, and any reply would already transmit data, the only privacy-improving way I see would be to let the email client inform the user about the providers involved on the clients side, especially if they differ from the email domain.

testbird:
@hpk42 The messages above were the reason for me reading and creating https://github.com/deltachat/deltachat-pages/pull/89

PR #343, opened 2018-06-27
Code cleanup and fix of #55

angelo-fuchs:
- Now when you open a high-res image, after some seconds the proper
sized image appears.

### DIFF
--- a/MessengerProj/src/main/java/com/b44t/messenger/Cells/ChatMessageCell.java
+++ b/MessengerProj/src/main/java/com/b44t/messenger/Cells/ChatMessageCell.java
@@ -40,6 +40,7 @@ import android.text.TextPaint;
 import android.text.TextUtils;
 import android.text.style.ClickableSpan;
 import android.text.style.URLSpan;
+import android.util.Log;
 import android.view.MotionEvent;
 import android.view.SoundEffectConstants;
 import android.view.ViewStructure;
@@ -182,7 +183,7 @@ public class ChatMessageCell extends BaseCell implements SeekBar.SeekBarDelegate
 
     //private int TAG;
 
-    public boolean isGroupChat;
+    private boolean isGroupChat;
     private boolean isPressed;
     private boolean isHighlighted;
     private boolean mediaBackground;
@@ -223,6 +224,7 @@ public class ChatMessageCell extends BaseCell implements SeekBar.SeekBarDelegate
     private boolean drawName;
     private boolean drawNameLayout;
 
+    // TODO: figure out when this is false and why it is that way
     private boolean allowForwardedName;
     private static TextPaint forwardedNamePaint;
     private StaticLayout forwardedNameLayout;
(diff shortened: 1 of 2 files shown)


angelo-fuchs:
The code seems to be working like this (in rough outlines)
On creation of the ChatMessageCell, if this is an image: Create a Thumbnail.
There seems to be two separate ways of doing this: Basic and Quality (as noted by the param "needsQualityThumb").
The Basic way seems to override the imagelink in the source element. I could not find the place where that happens, so this is more speculation.
Now that I configured the function to create a Quality thumb for "photos" this doesn't happen any more.
But that means the underlying bug is not "fixed" it just doesn't occur any more. This problem is likely to reappear.

r10s:
thanks for pushing this forward. however, without understanding what is really going on in the code, this pr seems not to have the desired effect.

- taking a test image from the camera on my Nexus 4 with Lineage 14.1, the zoom of the image is worse as with out the fix (see images attached)

- moreover, i noticed a strange behaviour when trying to take more than one photo directly from the camera: in the chat view, always the first photo is shown - but the last take photo is really sent.

there may be more cornercase i have not noticed.

@angelo-fuchs in my opinion, the codebase is close to un-maintainable. maybe we should just close this pr, leave #55 open and look forward if we switch the codebase. if we decide _not_ to switch the codebase, we can try over.
Although not optimal, the current status seems to not be _that_ bad.

but maybe i am also wrong and the new problems can be fixed easily. as mentioned, i do not understand the code :)

zoom-with-pr-343

**^^^ this is _with_ the fix**

**vvv and this is the current status quo**

zoom-current-master

r10s:
> May I ask which alternatives you are considering?

We are considering signal-android. not sure if and when we switch, but we decided to have a closer look at it :)

testbird:
Hm ok, I don't know much about signal-android.
https://www.onli-blogging.de/1599/3-verschluesselnde-Messenger,-2-Desaster.html#c7196

But wouldn't it likely end up in an independent, concurrent fork again?

I'm not sure if there is a project that would allow to collaborate on shared code. The most pertinent I saw was probably GNU Ring. Its client is build upon the LibRing (distributed p2p messaging and direct p2p voice/video). The clients already support direct messages (using the telephony stream connection) and indirect messaging (through the transient p2p distributed hashtable network). An additional deltachat-core based email-chat messaging chanel might be a good fit: Deltachat-core would bring bulk store-and-foreward "offline" messaging and the immense "reach" of email to the table, while GNU Ring brings multi-platform clients and p2p voice and video calls.
https://ring.cx/en/about/technical

angelo-fuchs:
That's odd, I have exactly the opposite effect on my phone (but about the same image quality).
But given that I, too, don't actually understand the codebase I don't feel strongly about the fix.

I will give it a second shot after we decided to a) stay on this codebase and b) I have cleaned up that mess and actually understand what's going on.

r10s:
> I will give it a second shot after we decided to a) stay on this codebase 
> and b) I have cleaned up that mess and actually understand what's going on.

yeah. and i would treat the "and" between a) and b) as logical. only if a) is fulfilled, we check b) :)

PR #389, opened 2018-09-05
Do not try to update the permanent notification when the screen is on, fix #362

### LINKED ISSUE #362: "Connected to..." notification shown when unlocked for some time
**Delta Chat version**
0.18.2 (F-Droid); Android 7.1.2 (SDK 25)

**Expected behavior**
The notification should not be shown when the screen is on(if I understood everything correctly)

**Actual behavior**
When I use my phone after some time a "Connected to..." notification is shown. When I lock and unlock the screen, the notification disappears. **Update:** This is the case only sometimes, sometimes it stays there

**Steps to reproduce the problem**
1. use my phone while DC is running :-) **Update**: rotate the screen while DC running in background

I do not know if it works for you...

**Screen snapshots**

screenshot

Hocceruser:
Fix #362. You can reproduce #362 by changing the "configuration", e.g. by rotating the screen (Delta Chat does not need to be in foreground for this).

However, my fix adds a dependence from KeepAliveService to ApplicationLoader. Also, now, KeepAliveService "knows" that the permanent notification is only shown when the screen is off. Therefore, in case you want to change this, you will have to remember to change KeepAliveService as well. 

A better solution may be a "isForegroundTaskRunning()" method somewhere else, but I do not know where.

### DIFF
--- a/MessengerProj/src/main/java/com/b44t/messenger/KeepAliveService.java
+++ b/MessengerProj/src/main/java/com/b44t/messenger/KeepAliveService.java
@@ -111,8 +111,10 @@ public class KeepAliveService extends Service {
 
     public void updateForegroundNotification()
     {
-        // update the notification by simply creating a new notification with the same ID, see https://developer.android.com/training/notify-user/managing.html#Updating
-        NotificationManager nm = (NotificationManager)getSystemService(Context.NOTIFICATION_SERVICE);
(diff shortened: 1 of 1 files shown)


r10s:
great catch :)
thank you very much, i've overseen that :)

Hocceruser:
You're very welcome, DC is great work so far, so thank _you_! :grinning:

Issue #444, opened 2019-01-13
prepare for f-droid


r10s:
- [x] make a backup from the current `deltachat-android` by duplicating the repo ~~as described [here](https://help.github.com/articles/duplicating-a-repository/) (forking in the same orga is not possible)~~ EDIT: i think `--mirror` should be skipped, we want it just for backup and reference reasons

- [x] bring this repo in some way to `deltachat-android` ~~so that f-droid picks up this repo without additional changes~~ changes are required as f-droid will search in a subdirectory that does no longer exists so a pr to https://gitlab.com/fdroid/fdroiddata/blob/master/metadata/com.b44t.messenger.txt is needed. but this typically gets accepted fast by @BuBu

- [x] change the internal app name from `chat.delta.androidii` to `com.b44t.messenger.beta` - otherwise the users of the existing version cannot get the update - for ppl currently using android-dev, this means and export+import

- [x] change icon and name

- [x] increase version number to a larger value than the one currently used by f-droid

- [x] adapt maven repos and remove remaining google play dependencies

- [x] add a pr to the f-droid repo ([info](https://github.com/deltachat/deltachat-android/blob/master/docs/f-droid.md)) to remove the now unused subdir

r10s:
moved repo using
```
git remote set-url origin https://github.com/deltachat/deltachat-android
git push --force origin master
```

r10s:
the pr on the f-droid repo: https://gitlab.com/fdroid/fdroiddata/merge_requests/4354

PR #648, opened 2019-01-22
changes for f-droid

r10s:
f-droid does not allow random maven urls, only the following whitelisted: https://gitlab.com/fdroid/fdroidserver/blob/master/fdroidserver/scanner.py#L98 - so we have to remove the dependencies from the signalapp maven repos.

the remaining maven repos are: 

- [x] https://github.com/chrisbanes/PhotoView - replaced by https://jitpack.io
- [x] https://github.com/leolin310148/ShortcutBadger - replaced by maven std. repo (whatever this is)
- [x] https://github.com/dmytrodanylyk/circular-progress-button - we do not really use them and i do not like them, replaced by <Button>

i have no idea why signal uses this from their own repos.

also, if someone know about how the dependecyVerification works, we  could re-add this. not sure if this is needed,  at least the used repos are whitelisted and under some control.

- [x] finally, there are some references to google-play, these could probably be removed

tackles #444


angelo-fuchs:
jitpack is just a redirection to the signal codebase (afaics), not sure if this is an improvement.
We should probably find a way to remove the dependencies that come from this angle overall.
(continuing reviewing)

angelo-fuchs:
Shortcut Badger: Do we use that? https://github.com/leolin310148/ShortcutBadger - At least on my phone nothing like that is shown. I recommend pruning completely (unless someone objects)

angelo-fuchs:
PLAY_STORE_DISABLED should be removed completely from the code, not just the place where it might be set. (esp. as we don't actually use it)

r10s:
> jitpack is just a redirection to the signal codebase (afaics), not sure if this is an improvement.

it's just that jitpack.io is whitelisted by f-droid. in general, i agree, in the mid-term, we should get rid of as many maven repos as possible.

r10s:
> wrt PLAY_STORE_DISABLED [...]

k - i removed the comment and the surrounding dead code. the BuildConfig-thingie is just generated code. after a rebuild-all, it will be gone.

wrt issues - sure :)

Issue #656, opened 2019-01-24
Map and Reduce Maven Repositories


angelo-fuchs:
Currently we depend on several repositories. We should map out which one we need for what and reduce the amount if possible.
While we are at this we can remove dependencies overall in the build process (and map out the remaining deps so its clear why we need them)

This is a continuation from the conversation here https://github.com/deltachat/deltachat-android/pull/648

r10s:
these are the maven repos whitelisted by f-droid: https://gitlab.com/fdroid/fdroidserver/blob/master/fdroidserver/scanner.py#L98 - we should not use different ones, however, i think this list is mostly complete.

angelo-fuchs:
jitpack: We load PhotoViewer over this one (which does the zooming for the photos).

angelo-fuchs:
This dependency: com.soundcloud.android:android-crop:0.9.10@aar is only needed in Group Select Avatar. If we were to unify this with Profile: Select Avatar this dependency would be obsolete.

PR #661, opened 2019-01-25
Mapping Dependencies and Repositories

angelo-fuchs:
To ensure we only have what we need and that the things we need are in a way that we expect them.

After that we can decide if any of it should go.

For #656


angelo-fuchs:
@r10s I went through the dependencies and deleted what was unused. But this changes should be tested very thoroughly.

r10s:
removed the targetSdk, merging

Issue #666, opened 2019-01-25
Signature text cleared in profile settings


adbenitez:
If I change my signature text and later I decide to set it back to the original value, when I save the changes the signature text is removed/cleared like if it is not possible for an user to set the default text as a value for it.

- Platform (android/blackberry/anbox): Android 6.0
- Device:BLU
- Delta Chat Version: 0.100

r10s:
thanks for reporting, changed the behavor in the core as this makes things also easier for desktop/ios at a later point

PR #667, opened 2019-01-25
[wip] adapt signature handling to new core behavior

r10s:
simplifies the status text handling, the defaults are handled in the core now, no more need to worry about this.

can be merged after the next update of the core.

tackles #666


r10s:
ready to merge, core is updated.

angelo-fuchs:
@r10s I leave the merge to you. Don't remember if something needs to be done explicitly if a change that depends on core is merged.

PR #682, opened 2019-01-29
Notification issue 451

angelo-fuchs:
Please test if the reduced importance also reduces the annoyance.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -79,7 +79,6 @@ import org.thoughtcrime.securesms.components.camera.QuickAttachmentDrawer;
 import org.thoughtcrime.securesms.components.camera.QuickAttachmentDrawer.AttachmentDrawerListener;
 import org.thoughtcrime.securesms.components.camera.QuickAttachmentDrawer.DrawerState;
 import org.thoughtcrime.securesms.components.emoji.EmojiDrawer;
-import org.thoughtcrime.securesms.components.location.SignalPlace;
 import org.thoughtcrime.securesms.components.reminder.ReminderView;
 import org.thoughtcrime.securesms.connect.ApplicationDcContext;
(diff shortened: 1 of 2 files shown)


r10s:
will continue testing, however, the list is still in reverse order on my device:

sending the messages a, b, c, d, e shows the following notification when receiving e:

image-36

clicking the notification enlarges to the following view then:

image-35

angelo-fuchs:
@r10s Actually now that I look into the order: That is by design :) The newest message on top (as all notifications in the top bar are ordered newest -> oldest).
Can we make a different issue for this?
Did you test @cyBerta's commit? Did it help?

r10s:
> Can we make a different issue for this?

sure. we can merge this pr and see how it works.

however, i still believe the order should be inverse.
it does not make sense when the oldest message always pops up.

violoncelloCH:
+1 @angelo-fuchs
It's also the same order as it would be in the chat view and therefore feals much more natural! So I would leave it as it is in the screenshot...

angelo-fuchs:
@violoncelloCH But the conversations are the opposite direction. Maybe we should have a talk about directions in general and where do we want which.
For future reference: The place where the order is maintained is:
org/thoughtcrime/securesms/notifications/MessageNotifier.java:322 currently
```
    while(iterator.hasPrevious()) {
      NotificationItem item = iterator.previous();
      builder.addMessageBody(item.getRecipient(), item.getIndividualRecipient(), item.getText());
    }
```
must change to (untested)
```
    while(iterator.hasNext()) {
      NotificationItem item = iterator.next();
      builder.addMessageBody(item.getRecipient(), item.getIndividualRecipient(), item.getText());
    }
```

violoncelloCH:
@angelo-fuchs 
I would go for showing the latest message on top and on expanding show the new messages in the order they would be in the chat conversation. This would also make sense, because you swipe down to expand (it feels like you pull down the newest message and get the older ones visible above).
Btw. this is also how other messengers I know do it...

PR #686, opened 2019-01-31
allowing the LED color to be purple.

angelo-fuchs:
This should be tested by people whose phone supports purple LED. (Mine does not)


r10s:
image-1548981812

PR #693, opened 2019-02-01
disable autocorrection on server input fields

### LINKED ISSUE #441: auto-correction in server input field should be disabled
not sure if this can be disabled or if it is only a behavior on my moto-g:

when entering a server in the RegistrationActivity, after typing eg. `imap.` the input field changes it to `IMAP.` - for whatever reason ...

violoncelloCH:
fixes #441


r10s:
great, thanks :)

violoncelloCH:
@r10s Thank you for merging :)
What about the login-name fields? Should I apply this to those as well, as I asked in the issue?

PR #695, opened 2019-02-02
re-add qr-code overlay logo

r10s:


r10s:
@violoncelloCH thanks for testing, great!

PR #696, opened 2019-02-02
First attempt to display initial duration for audio attachements.

### LINKED ISSUE #688: indicate audio message duration initially
If you receive a voice message, you want to know the duration of the audio. Currently deltachat shows 00:00 and starts counting upwards from there. I think it should show the whole duration initially and only after clicking play, it should be set to 00:00 and start counting...

Platform (android/blackberry/anbox): android
- Device: honor 8
- Delta Chat Version: 0.100.0
- Expected behavior: show full duration of audio message initially
- Actual behavior: 00:00 is shown; you don't know the length of the audio before playing it

chklump:
This is a trial to fix #688.

To get the duration for an attachment, one has to create a new `AttachmentServer` and `MediaPlayer` and wait for the attachment to be loaded into the player. This btw is also done each time one toggles between play and pause during playback. While it seemed very fast to me during testing, it could be a problem during fast scrolling through a chat history with many audios.

Therefore, it is maybe not a completely clean approach from a technical perspective.

A different approach would be to somehow save the duration with the attachment already on downloading, but then that would rather be a core problem to my understanding (and the core should probably be file format agnostic and shouldn't deal with that stuff, either?).


r10s:
in fact, storing duration in the core is already possible using [dc_msg_latefiling_mediasize()](https://c.delta.chat/classdc__msg__t.html#a7687ff969841f3d00c3a212f9ad27861)

r10s:
so i would suggest to 
- call requestDuration() only if msg.getDuratation() returns 0
- storing the calculated durations using [msg.lateFilingMediaSize()](https://c.delta.chat/classdc__msg__t.html#a7687ff969841f3d00c3a212f9ad27861) then

if there are other things that can be optimized due to play/pause, of cause, this would also be great :)
@chklump thanks for pushing this forward, anyway :)

chklump:
@r10s I've tried to implement your suggestions. Hope, that it's okay this way? If not, please tell me, so I can improve. Also fixed a bug where durations of over a minute were displayed as for example "2:124" instead of "2:04".

r10s:
@chklump cool :)
looking forward to test this :)

r10s:
tested & works as expected :)

if there are issues left, i think we can fix them in the next iteration :)

PR #702, opened 2019-02-07
change targetSdkVersion from 25 to 26

### LINKED ISSUE #436: Update target SDK Version
Our targetSdkVersion is currently 25.
For deployment in Googles Play store we will need at least 26 at the moment but should consider doing more then that.

Also our current gradle version is lagging behind, we should check if we can get more up-to-date on that as well.

While we are at it, we should check if we can upgrade our dependency versions as well, many seem dated and it would be beneficial to have an overview of what we have, why we need it and what the best version for us is.

This should not happen as long as we still have code in the codebase that is scheduled to be deleted (like SmsDatabase or similar).

I'd target Mid-January for it.

r10s:
this pr not just increases the targetSdkVersion but also does changes required wrt background execution, see https://developer.android.com/about/versions/oreo/android-8.0-changes

closes #436

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -142,7 +142,7 @@ android {
         applicationId "com.b44t.messenger"
 
         minSdkVersion 14
-        targetSdkVersion 25
+        targetSdkVersion 26
 
         vectorDrawables.useSupportLibrary = true
         project.ext.set("archivesBaseName", "Deltachat");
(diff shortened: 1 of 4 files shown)


r10s:
just installed lineage with android o resp. sdk 26 and test this a bit :)

btw. incredible that oreo is available for my 7yo nexus s :)

r10s:
merging this now so that problems will arise soonish (we have to upgrade to 26 in any case)

PR #706, opened 2019-02-08
Lock pattern security issues #649

### LINKED ISSUE #649: Lock pattern security issues
When user press cancel/back on the pattern screen sometimes the Delta Chat screen (with the sensitive content) is briefly shown. 
Also when the app is put in background/minimized and user go to the app again, the chats list is shown (all chats and their last message are visible) when the user tries to open a chat, then the pattern screen is shown.

- Device: BLU
- Delta Chat Version: 0.98.2

Boehrsi:
Should fix #649:

- Cleanups 
- Call super.onActivityResult(requestCode, resultCode, data) where needed to avoid problems
- Hide UI on stop and only show it if no screen lock is active or if the screen lock was successfully unlocked
- Fixed some cases where the screen lock wasn't applied
- Fixed one case where the screen lock was applied after initial activation, but it shouldn't

I tried to test on new and old devices and hope every edge case is now fixed. But it would be great if some tests on new and old device would be performed by the reviewer before merging.

### DIFF
--- a/src/org/thoughtcrime/securesms/BaseActionBarActivity.java
+++ b/src/org/thoughtcrime/securesms/BaseActionBarActivity.java
@@ -39,13 +39,21 @@ public abstract class BaseActionBarActivity extends AppCompatActivity {
 
     @Override
     protected void onStart() {
-        super.onStart();
         if (ScreenLockUtil.isScreenLockEnabled(this) && ScreenLockUtil.getShouldLockApp() && !isWaitingForResult) {
-            ScreenLockUtil.applyScreenLock(this);
+          ScreenLockUtil.applyScreenLock(this);
+        } else {
+          findViewById(android.R.id.content).setVisibility(View.VISIBLE);
         }
+        super.onStart();
     }
 
-    @Override
+  @Override
+  protected void onStop() {
+    findViewById(android.R.id.content).setVisibility(View.GONE);
+    super.onStop();
+  }
+
+  @Override
     protected void onActivityResult(int requestCode, int resultCode, Intent data) {
       super.onActivityResult(requestCode, resultCode, data);
       isWaitingForResult = false;
@@ -139,6 +147,8 @@ public abstract class BaseActionBarActivity extends AppCompatActivity {
   @Override
   public void startActivityForResult(Intent intent, int requestCode) {
     super.startActivityForResult(intent, requestCode);
-    isWaitingForResult = true;
+    if (requestCode != -1) {
+      isWaitingForResult = true;
+    }
   }
 }
--- a/src/org/thoughtcrime/securesms/BaseActionBarActivity.java
+++ b/src/org/thoughtcrime/securesms/BaseActionBarActivity.java
@@ -49,7 +49,9 @@ public abstract class BaseActionBarActivity extends AppCompatActivity {
 
   @Override
   protected void onStop() {
-    findViewById(android.R.id.content).setVisibility(View.GONE);
+    if (ScreenLockUtil.isScreenLockEnabled(this)) {
+      findViewById(android.R.id.content).setVisibility(View.GONE);
+    }
     super.onStop();
   }
 
--- a/src/org/thoughtcrime/securesms/BaseActionBarActivity.java
+++ b/src/org/thoughtcrime/securesms/BaseActionBarActivity.java
@@ -28,6 +28,7 @@ public abstract class BaseActionBarActivity extends AppCompatActivity {
   private Timer timer;
 
   private boolean isWaitingForResult;
+  private boolean isHiddenByScreenLock;
 
   @Override
   protected void onCreate(Bundle savedInstanceState) {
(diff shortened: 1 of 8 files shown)


r10s:
on my nexus4 with lineage-oreo, the screen is still shown shortly when the app is exited and started again without switching the device off in between.

when the device is switch off in between, i did not see the content shortly.

however, not sure if this is fixable at all.

@Boehrsi can you explain a bit more what this pr is doing?  Call super.onActivityResult() where needed to avoid problems -> which problems?
Hide UI on stop -> which command does this? and did you expected it to work in the case above?

this security-lock seems to cause much more problems as initially expected ...

Boehrsi:
@r10s 

>  Call super.onActivityResult() where needed to avoid problems -> which problems?
The lock screen activity provided by the systems gets back to the app by using the onActivityResult 
callback. To handle this one correctly the the our base activity class needs to know if the callback was triggered by the mentioned lock screen activity. To get the call down to the parent the super.onActivityResult() is needed. If this is not done it's possible that unlocking won't work in some specific activities.

> Hide UI on stop -> which command does this? and did you expected it to work in the case above?
Hide is at https://github.com/deltachat/deltachat-android/pull/706/commits/fa27115594edb1a245f3e4a1ac5edec611c0b717#diff-f9dfd94364d950ec18ae828e20ea3f51R52
Showing it again is at https://github.com/deltachat/deltachat-android/pull/706/commits/fa27115594edb1a245f3e4a1ac5edec611c0b717#diff-f9dfd94364d950ec18ae828e20ea3f51R45
In general I simply hide the main content view and make it visible again.

> this security-lock seems to cause much more problems as initially expected ...
Yep, that's true. It's kind of annoying that no "default" way is provided by the Android framework. But sadly we have to handle that all our self.

Boehrsi:
Regarding the "mentioned case above" i replied in the actual issue: https://github.com/deltachat/deltachat-android/issues/649#issuecomment-463955780

src/org/thoughtcrime/securesms/BaseActionBarActivity.java
```diff
@@ -39,13 +39,21 @@ protected void onCreate(Bundle savedInstanceState) {
 
     @Override
     protected void onStart() {
-        super.onStart();
         if (ScreenLockUtil.isScreenLockEnabled(this) && ScreenLockUtil.getShouldLockApp() && !isWaitingForResult) {
-            ScreenLockUtil.applyScreenLock(this);
+          ScreenLockUtil.applyScreenLock(this);
+        } else {
+          findViewById(android.R.id.content).setVisibility(View.VISIBLE);
```
r10s: this does not alway help to hide the content in a reliable way, eg. on my lineage/oreo it does only work sometimes.

also, i fear, this could have side-effects, esp. as the state-change is always done, also when the screen lock is not used.

i would suggest to skip the hiding therefore. with the screen securiy, we have a more reliable option that seems to work better; we should just add a note to recommend the user to enable it.

src/org/thoughtcrime/securesms/BaseActionBarActivity.java
```diff
@@ -39,13 +39,21 @@ protected void onCreate(Bundle savedInstanceState) {
...
         }
+        super.onStart();
     }
 
-    @Override
+  @Override
+  protected void onStop() {
```
r10s: same here, i would drop this again.

src/org/thoughtcrime/securesms/BaseActionBarActivity.java
Boehrsi: Does screen security on without the hiding is enough? As far as I remember I was able to see content if no hiding was applied, but I could be wrong. If screen security on is enough I would totally agree to drop the hiding part.

Boehrsi: @r10s in my test it works most of the time when the UI hiding is removed. I saw content one or two times when performing actions pretty fast. How do we want to proceed, should I remove the hiding anyway and we continue testing "in the wild"?

r10s: > Does screen security on without the hiding is enough?

can you check this out again on your device?

if hiding is still necessary, however, i would do it only if lock-pattern is really enabled, not always.

Boehrsi: I did, my last comment form an hour ago was referring to that. The "hide only if lock is enabled part" makes sense, I think that's a good trade.

r10s: /me did not refresh the page before replying :)

r10s: > The "hide only if lock is enabled part" makes sense, I think that's a good trade.

cool, i think this is a good way to go.

Boehrsi: Nice, I will wrap the onStop + setVisibility(View.GONE) part, but not the setVisibility(View.VISIBLE) part to avoid problems when enabling / disabling the function via the settings. Calling setVisibility(View.VISIBLE) shouldn't be a problem as it would not change anything if the view is already visible.

Boehrsi: > Nice, I will wrap the onStop + setVisibility(View.GONE) part, but not the setVisibility(View.VISIBLE) part to avoid problems when enabling / disabling the function via the settings. Calling setVisibility(View.VISIBLE) shouldn't be a problem as it would not change anything if the view is already visible.

Done @r10s

r10s:
@Boehrsi i rebased the branch to master, please pull.

testing now :)

Issue #685, opened 2019-01-30
Image shown in wrong day in gallery(all media)


adbenitez:
in a group chat, there was only one picture posted that day, (under the label "Today" as shown in the first screenshot) but in the gallery (all media) two images are shown under the header "today" one of them was from yesterday and isn't shown in the "yesterday" section (see the second screenshot)

- Platform (android/blackberry/anbox): Android 6.0
- Device:BLU
- Delta Chat Version:0.100
- Screenshots:

chat view:
![screenshot_20190126-115158](https://user-images.githubusercontent.com/24558636/52014634-0f3bd680-24ae-11e9-9f90-5e7eec2d1ec3.png)

all media:
![screenshot_20190126-115218](https://user-images.githubusercontent.com/24558636/52014674-2b3f7800-24ae-11e9-8442-e8afc012784f.png)

adbenitez:
will do, it is happening again for other two pictures

adbenitez:
this screenshots are all from 2 new pictures, the messages details are NOT from the old ones:
today pic:
![screenshot_20190131-082713](https://user-images.githubusercontent.com/24558636/52077259-96488780-255e-11e9-9063-e38bd0ddc469.png)
yesterday pic is shown under today header:
![screenshot_20190131-082629](https://user-images.githubusercontent.com/24558636/52077263-98124b00-255e-11e9-8c52-22735440c55b.png)
today pic is info:
![screenshot_20190131-133547](https://user-images.githubusercontent.com/24558636/52077272-9ea0c280-255e-11e9-86f5-b75ccd8ecc60.png)
yesterday pic is info:
![screenshot_20190131-082849](https://user-images.githubusercontent.com/24558636/52077274-a06a8600-255e-11e9-85c6-3699f3abaf6b.png)

angelo-fuchs:
Thanks! I will investigate this more, Pic Info helped.

angelo-fuchs:
Problem is somewhere here:
org/thoughtcrime/securesms/database/loaders/BucketedThreadMediaLoader.java:88

angelo-fuchs:
Okay, now I have it figured out: 
"Today" means "Last 24 hours"
"Yesterday" is the 24 hours before that
"This week" means "In the last 7*24 hours until yesterday"
"This month" is the same for 30 days.

I have to think about this.

r10s:
i would not change the wording, this is just fine. but we can define
- "today" as "today since midnight" 
[...]

PR #718, opened 2019-02-13
#685 date alignment in media gallery

angelo-fuchs:
@r10s I could use some unpedantic pragmatism here. Current wording is confusing (at least to me & @adbenitez ).

Problem: In the Media Gallery "Today" means "The last 24 hours". In the Conversation view it means "Since 00:00.00" Similar for "Yesterday". Then they diverge even more: Conversation stays in the "Day" segmentation, Gallery changes to "This week" which means 7*24 h - 5*24 h timespan, and "This month" which is everything after that until 30*24h.

How would you name the tags & what timeframe do you think is good for it?

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -36,6 +36,8 @@
     <string name="yesterday">Yesterday</string>
     <string name="this_week">This week</string>
     <string name="this_month">This month</string>
+    <string name="last_week">Last week</string>
+    <string name="last_month">Last month</string>
     <!-- Translators: show beside messages that are "N minutes old". prefer a short string, prefer abbreviations if they're well-known -->
     <plurals name="n_minutes">
         <item quantity="one">%d min</item>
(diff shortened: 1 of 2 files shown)


angelo-fuchs:
I personally would go this route:
"Today" -> Since 00:00.00
"Yesterday" -> The 24 hours before "Today"
"This week" -> Everything before "Yesterday" that is also after the closest start of week (Monday/Sunday depending on current Timezone)
"Last week" -> Everything that happened in the 7 days preceding the last start of week.
"This month" -> Everything between "Last week" and the closest 1st of a month.
And before that keep it as it is: Display month and year and be done.

angelo-fuchs:
Or use discordian date, and call it a day.

violoncelloCH:
@angelo-fuchs 
+1 because it makes sense and is consistent -> improves UX

r10s:
@angelo-fuchs i agree not to change the wording but use the corresponding timeframes for "Today" and "Yesterday" in the same way as it is done in the chatview and as you've suggested it above.

"this week" is a bit weird as it overlaps with "today" and "yesterday", however, it's seems to be a handy term and your suggestion is reasonable. same for "this month".

i think as long as there are no gaps, it is fine :)

hpk42:
gaps in time, hard to avoid, sometimes.  
time is evasive despite all control and capture attempts ;) 
/me trolls off ...

angelo-fuchs:
If someone is reviewing this: Please also test quirky days (like start of week, start of month and start of year. For the truly pedantic even DST days are interesting)
@r10s We really need a test suite for this kind of scenario.

PR #719, opened 2019-02-14
save drafts reliable in onPause()

### LINKED ISSUE #420: Draft sometimes lost on incoming call
If you receive a call while you are typing a message, the draft is not saved when Delta Chat's activity is stopped and replaced with the incoming call activity.

- Platform (android/blackberry/anbox): Android
- Device: BLU
- Delta Chat Version: 0.97.1
- Expected behavior: the draft is saved when the view lost the focus
- Actual behavior: the draft is lost when the view lost the focus due to an incoming call
- Steps to reproduce the problem: with device A start typing a message on a chat, with device B make a call to device A, and wait the current activity (Delta Chat) is stopped and the phone call activity is shown.

r10s:
closes #420 

@adbenitez if you like to test it out, here is an APK including this pr: https://share.riseup.net/#k_xtF-Dqe_YCjhUvi7UYVA

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -290,6 +290,14 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
   @Override
   protected void onPause() {
     super.onPause();
+
+    // theoretically, saving the draft in ConversationActivity.onDestroy()
+    // should be sufficient, however, seems as if it RAM gets low, Activities/Apps are terminated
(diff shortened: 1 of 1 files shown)


adbenitez:
@r10s I am downloading the app, thanks!!!

adbenitez:
it worked in one test, I will test this several times latter, but at first glance seems good!! +1

r10s:
not merging, this does not help really on the issue, it's just that the app gets killed due to low memory, if we want to work against this, this would be some more effort, see #420

PR #721, opened 2019-02-14
fix crash wrt notifications

### LINKED ISSUE #720: GlideApp...getI() must not be called from main thread
got two crashes wrt notifications on nexus4/android 8.1/oreo

the crashes were assertions from the Glide module; saying "You must call this method on a background thread":
https://gist.githubusercontent.com/r10s/28ef216fad00e1e1fd467cac71daa804/raw/84bc8797f71e3aeda23083197302ff157b6a3378/gistfile1.txt

tracing back the log, it's probably [SingleRecipientNotificationBuilder.setThread()](https://github.com/deltachat/deltachat-android/blob/master/src/org/thoughtcrime/securesms/notifications/SingleRecipientNotificationBuilder.java#L75) which calls GlideApp...get(), probably from the main thread and not from a background thread.

wondering why the app does not crash all the time.

r10s:
closes #720

### DIFF
--- a/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
+++ b/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
@@ -76,23 +76,23 @@ public class MessageNotifier {
 
   static final  String EXTRA_REMOTE_REPLY = "extra_remote_reply";
 
-  public  static final long   NO_VISIBLE_THREAD         = -1L;
+  public  static final long   NO_VISIBLE_CHAT_ID        = -1L;
   private static final  int   SUMMARY_NOTIFICATION_ID   = 1338;
   private static final int    PENDING_MESSAGES_ID       = 1111;
   private static final String NOTIFICATION_GROUP        = "messages";
   private static final long   MIN_AUDIBLE_PERIOD_MILLIS = TimeUnit.SECONDS.toMillis(20);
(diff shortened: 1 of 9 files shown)


src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
```diff
@@ -174,7 +180,7 @@ public static void updateNotification(@NonNull  Context context,
                                         int       threadId,
                                         boolean   signal)
   {
-    boolean    isVisible  = visibleThread == threadId;
+    boolean    isVisible  = visibleChatId == threadId;
```
angelo-fuchs: should changename of threadId here too.

angelo-fuchs:
I just tested it and the following issue bugs me:
When I already have Delta open and there are notifications. When I open a chat the notification stays until I leave that chat again.

adbenitez:
yes! I was preparing screenshots for a bug report for this issue!

adbenitez:
in my case not only until I go back but until I go into the chat again?!

r10s:
> When I already have Delta open and there are notifications. When I open a chat the notification stays until I leave that chat again.

yip, i think there are still lots of tiny issues wrt notificatrions, however, for now merging this pr so that the crashes stop :)

angelo-fuchs:
@adbenitez I opened #725 for further treatment

PR #723, opened 2019-02-15
Use Googles repositories first

Boehrsi:
We should use the Google repositories at first in the order to resolve dependencies, as sometimes the sync between Googles repos and Maven.org / MavenCentral ones fails and builds will fail. E.g. right now for me with _Could not find manifest-merger.jar (com.android.tools.build:manifest-merger:26.0.1)_).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,12 +1,12 @@
 buildscript {
     repositories {
         maven {
-            url "https://repo1.maven.org/maven2"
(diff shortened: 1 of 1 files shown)


angelo-fuchs:
lgtm and we are using the google repos anyway, so I see no harm here.

r10s:
okay, let's go for that :)

Issue #716, opened 2019-02-13
mailto:-links without address should work as "share"


r10s:
while mailto-links with e-mail addresses work line a charme since https://github.com/deltachat/deltachat-android/issues/473, mailto:-links without e-mail addresses (sth. as `mailto:?subject=bla&body=blub`, used eg. on golem.de) currently do not work as expected.

these links open the contact list; once a contact is selected, however, the proposed text is gone.

i think these links should work in the same way as sharing: offering the chatlist and adding the text as draft. the user can click "send" or modify the text before sending then.

angelo-fuchs:
[Testlink](mailto:?subject=foo&body=bar) for easier testing.

Boehrsi:
Will have a look :).

r10s:
great :)

Boehrsi:
Should be fixed by https://github.com/deltachat/deltachat-android/pull/724

r10s:
should be fixed by https://github.com/deltachat/deltachat-android/pull/724

PR #724, opened 2019-02-15
mailto:-links without address should work as "share" #716

Boehrsi:
Fix for #716.

- Subject and body are now handled
- To avoid bigger changes the created share text is added as extra to the intent, as this is already handled by the app
- Tested for existing and new contacts
- Should also work for a combination of mail + subject + body

### DIFF
--- a/src/org/thoughtcrime/securesms/NewConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/NewConversationActivity.java
@@ -32,6 +32,9 @@ import com.b44t.messenger.DcContext;
 
 import org.thoughtcrime.securesms.connect.DcHelper;
 
+import java.util.HashMap;
+import java.util.Map;
+
 /**
  * Activity container for starting a new conversation.
  *
@@ -43,6 +46,10 @@ public class NewConversationActivity extends ContactSelectionActivity {
   @SuppressWarnings("unused")
   private static final String TAG = NewConversationActivity.class.getSimpleName();
   private static final String MAILTO = "mailto";
+  private static final String SUBJECT = "subject";
+  private static final String BODY = "body";
+  private static final String QUERY_SEPARATOR = "&";
+  private static final String KEY_VALUE_SEPARATOR = "=";
 
   @Override
   public void onCreate(Bundle bundle, boolean ready) {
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/NewConversationActivity.java
```diff
@@ -121,8 +125,6 @@ public void onClick(DialogInterface dialog, int which) {
   }
 
   private void openConversation(int chatId) {
-    final DcContext dcContext = DcHelper.getContext(this);
```
Boehrsi: Wasn't used anymore, so cleaned up that one.

src/org/thoughtcrime/securesms/NewConversationActivity.java
```diff
@@ -103,16 +110,13 @@ public void onContactSelected(int specialId, String addr) {
...
-                .setPositiveButton(android.R.string.ok, new DialogInterface.OnClickListener() {
-                  @Override
-                  public void onClick(DialogInterface dialog, int which) {
-                    int contactId = dcContext.createContact(null, addr);
-                    if(contactId==0) {
-                      Toast.makeText(NewConversationActivity.this, R.string.bad_email_address, Toast.LENGTH_LONG).show();
-                      return;
-                    }
-                    openConversation(dcContext.createChatByContactId(contactId));
+                .setPositiveButton(android.R.string.ok, (dialog, which) -> {
```
Boehrsi: More lambdas = less code

r10s:
@Boehrsi thanks for pushing this forward.

i've just tested it, for the testlink from the issue, it worked - but the shave-via-mail-links from golem do not. do you have an idea what the problem is there?

also, i would prefer to see the chatlist and not the contactlist to select a chat. can we change this? then, this share-functionality would be on pair with the normal android share-function and the forward-function. also, this way, things can be shared to groups.

Boehrsi:
I will have a look regarding the Golem links. The change from contact list to chat list is a bit more work, as I have to redirect the intent to another activity then. I will also have a look into that one.

Boehrsi:
Tested and working now with

- [Golem Link](mailto:?subject=Intel: Apple soll ab 2020 auf ARM-Prozessoren umschwenken&body=Artikel auf Golem.de lesen:%0Ahttps://glm.io/139561?m%0A)
- [Our test Link](mailto:?subject=foo&body=bar)
- [Mail to with recipient](mailto:test@testing.de)

src/org/thoughtcrime/securesms/NewConversationActivity.java
```diff
@@ -43,6 +46,10 @@
   @SuppressWarnings("unused")
   private static final String TAG = NewConversationActivity.class.getSimpleName();
   private static final String MAILTO = "mailto";
+  private static final String SUBJECT = "subject";
```
Boehrsi: Constants are good

src/org/thoughtcrime/securesms/NewConversationActivity.java
```diff
@@ -61,22 +68,23 @@ private void handleIntent() {
...
-            }
-            if (textToShare != null && !textToShare.isEmpty()) {
-              getIntent().putExtra(ConversationActivity.TEXT_EXTRA, textToShare);
-            }
             String recipientsList = mailto.getTo();
             if(recipientsList != null && !recipientsList.isEmpty()) {
               String[] recipientsArray = recipientsList.split(",");
               if (recipientsArray.length >= 1) {
                 String recipient = recipientsArray[0];
+                if (textToShare != null && !textToShare.isEmpty()) {
```
Boehrsi: If we have a recipient and also content, append it to the new chat.

src/org/thoughtcrime/securesms/NewConversationActivity.java
```diff
@@ -61,22 +68,23 @@ private void handleIntent() {
...
+                  getIntent().putExtra(ConversationActivity.TEXT_EXTRA, textToShare);
+                }
                 onContactSelected(DcContact.DC_CONTACT_ID_NEW_CONTACT, recipient);
               }
+            } else {
+              Intent shareIntent = new Intent(this, ShareActivity.class);
```
Boehrsi: If no recipient is given delegate the text to the ShareActivity

src/org/thoughtcrime/securesms/NewConversationActivity.java
```diff
@@ -87,6 +95,29 @@ private void handleIntent() {
...
+    Map<String, String> mailtoQueryMap = getMailtoQueryMap(uri);
+    String textToShare = mailtoQueryMap.get(SUBJECT);
+    String body = mailtoQueryMap.get(BODY);
+    if (body != null && !body.isEmpty()) {
+      textToShare += "\n" + body;
+    }
+    return textToShare;
+  }
+
+  private Map<String, String> getMailtoQueryMap(Uri uri) {
```
Boehrsi: Manual query "parsing" and "handling" as the MailTo stuff given by the systems fails ...

Boehrsi:
@r10s added the delegation to the ShareActivity if no recipient is given, if one is given the old way is used.

r10s:
cool, i will check this out :)

r10s:
works in my tests now :+1:

PR #726, opened 2019-02-15
[wip] MessageNotifier: Cleanup and restructuring for #725

### LINKED ISSUE #725: Notifications are not cleared on display of chat
Since #721 notifications are not cleared properly, it fixes #720 which was a high prio (crash) but we should make sure that notifications go away anyway.

Behavior:
1. A notification appears
2. I open the chat from chat overview
3. The notification stays

Expected behavior:
1. A notification appears
2. I open the chat from chat overview
3. The notification is cleared up.

angelo-fuchs:
Two objectives:
1. Reduce the amount of entry points (and therefore complexity) of the Message Notifier.
2. fix #725 so that notifications are properly cleaned up.


hpk42:
guess this needs to be restarted to be useful and this is not happening soon, so closing this.

PR #729, opened 2019-02-18
remove dead references in gradle

r10s:
i think the files and references removed in this pr are no longer needed.


angelo-fuchs:
lgtm. If we ever bring back unit testing we should revert this commit, but until then its better to have it gone.

PR #737, opened 2019-02-22
tweak onboarding

r10s:
- add "no delta servers" hint
- move "advanced" button more to the top, hint below (this way it should not happen that you click the button and nothing is changed on the screen without scrolling)
- left align text
- tweak spacing

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -258,7 +258,7 @@
     <string name="login_title">Login</string>
     <string name="login_header">Login to your server</string>
     <string name="login_subheader">For known email providers additional settings are setup automatically. Sometimes IMAP needs to be enabled in the web frontend. Consult your email provider or friends for help.</string>
-    <string name="login_no_servers_hint">Delta Chat has no servers, your data stays on your device!</string>
+    <string name="login_no_servers_hint">There are no Delta Chat servers, your data stays on your device!</string>
     <string name="login_inbox">Inbox</string>
     <string name="login_imap_login">IMAP login-name</string>
     <string name="login_imap_server">IMAP server</string>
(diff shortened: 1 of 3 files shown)


r10s:
cc @hpk42

violoncelloCH:
how about saying something that your data is stored on your *mail providers server and* your device?
and maybe also say: "*e-mail password*" so it's really clear the mail account password has to be used?

r10s:
@violoncelloCH hm, not sure. imho these parts are rather clear. however, when user complain that this is unclear, we'll change this :)

this pr is mainly about the "no servers"; we got concrete feedback several times of people being scared about giving their password away to a "delta server" that they may not trust (and looking at several other fancy new mail apps, this is a reasonable concern)

res/values/strings.xml
```diff
@@ -258,6 +258,7 @@
     <string name="login_title">Login</string>
     <string name="login_header">Login to your server</string>
     <string name="login_subheader">For known email providers additional settings are setup automatically. Sometimes IMAP needs to be enabled in the web frontend. Consult your email provider or friends for help.</string>
+    <string name="login_no_servers_hint">Delta Chat has no servers, your data stays on your device!</string>
```
hpk42: on delta.chat the first half sentences was marked as a germanism and fixed to read: Delta Chat doesn’t have their own servers. 

It's a bit long though and i wonder if we could say "There are no Delta Chat servers." both here and on the website ;) 
 
FWIW I also a user on twitter use "delta chat requires no sign up" -- thought it makes sense because "sign up" is what is in user's faces.  Then again, someone could claim "no sign up" and still create some cloud account using the same credentials as your e-mail account.

r10s: so bad that "Delta Chat has no servers" is no english :)

but i would also go for "There are no Delta Chat servers." then. doing a pr

hpk42:
nice progress ;)

Issue #740, opened 2019-02-23
Check how unrecognized attachments are handled


r10s:
see https://support.delta.chat/t/unrecognized-attachments/279

iirc some older version (0.9somwhat) of  Delta Chat sent wrong mimetypes, might be related to that.

chklump:
I think, the #741 at least fixes the first part of the referenced forum thread.

r10s:
that was quick :)

adbenitez:
check https://support.delta.chat/t/unrecognized-attachments/279/5?u=adbenitez

r10s:
i think this very generic issue can be closed, in general, attachments work.

PR #741, opened 2019-02-23
Open files via intent didn't work for old android versions.

chklump:
For android SDK versions 23 and below `startActivity` throws a RuntimeException if called from a context other then an activity. To circumvent this, the flag FLAG_ACTIVITY_NEW_TASK has to be added to the intent.

Edit: I've just seen there has just been #740 opened at last partly on this. The corresponding [forum thread](https://support.delta.chat/t/unrecognized-attachments/279) describes the exactly same behaviour a friend of mine had and the mentioned devices (Samsung Note 2..5) all run android versions older than SDK 23, so I think this one fixes this.


r10s:
cool, looking forward to check and to merge this :)

r10s:
@chklump do you know why FLAG_ACTIVITY_NEW_TASK is not needed on SDK>23?

however, as the base problem seems to be common, (found eg. https://github.com/twitter/twitter-kit-android/issues/130 ) and this pr is already an improvement anyway, i'll merge it now. we can improve or change it later as needed.

PR #742, opened 2019-02-25
email interactions

r10s:
this pr will improve the interaction between chats and email

todo:

- [x] fix some annoyances with handling the popping up contact requests
- [x] add an advanced option "show emails" with the choices "no", "for active chats", "all".
- [x] if not "all" is selected as "show emails" option, the contact-request-activity and -menu-entry is completely hidden
- [x] rename the contact-request-activity to "emails"
- [x] wait for [core-part](https://github.com/deltachat/deltachat-core/pull/588) being ready

the option should be handled by the core, proposed name is `show_emails`.

@hpk42 wrt the default, meanwhile i also tend to not show emails but replies to chats by default.

### DIFF
--- a/res/values/arrays.xml
+++ b/res/values/arrays.xml
@@ -153,6 +153,18 @@
       <item>none</item>
   </string-array>
 
+  <string-array name="pref_show_emails_entries">
+    <item>@string/pref_show_emails_no</item>
+    <item>@string/pref_show_emails_accepted_contacts</item>
+    <item>@string/pref_show_emails_all</item>
+  </string-array>
+
+  <string-array name="pref_show_emails_values">
+    <item>0</item>
+    <item>1</item>
+    <item>2</item>
+  </string-array>
+
   <!-- discrete MIME type (the part before the "/") -->
 
 
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -206,6 +206,8 @@
     <string name="chat_archived_label">Archived</string>
     <string name="chat_no_messages">No messages.</string>
     <string name="chat_self_talk_subtitle">Messages I sent to myself</string>
+    <string name="chat_contact_request">Contact request</string>
+    <string name="chat_no_contact_requests">No contact requests.\n\nIf you want classic emails to appear here as contact requests, you can change the corresponding setting in the app settings.</string>
 
     <!-- search -->
     <string name="search">Search</string>
@@ -382,6 +384,11 @@
     <string name="pref_watch_mvbox_folder">Watch DeltaChat folder</string>
     <string name="pref_auto_folder_moves">Automatic Moves to DeltaChat folder</string>
     <string name="pref_auto_folder_moves_explain">Chat conversations are moved to avoid cluttering the Inbox folder</string>
+    <string name="pref_email_interaction_title">Email interaction</string>
+    <string name="pref_show_emails">Show classic emails</string>
+    <string name="pref_show_emails_no">No, chats only</string>
+    <string name="pref_show_emails_accepted_contacts">For accepted contacts</string>
+    <string name="pref_show_emails_all">All</string>
 
 
     <!-- autocrypt -->
@@ -611,6 +618,5 @@
     <string name="qrshow_join_verified_group_title">QR invite code</string>
     <!-- see qrshow_join_group_hint -->
     <string name="qrshow_join_verified_group_hint">Scan this to join the group \"%1$s\".</string>
-
 
 </resources>
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -599,7 +599,7 @@ JNIEXPORT void Java_com_b44t_messenger_DcContext_setConfig(JNIEnv *env, jobject 
 }
 
 
-JNIEXPORT jstring Java_com_b44t_messenger_DcContext_getConfig(JNIEnv *env, jobject obj, jstring key, jstring def/*may be NULL*/)
+JNIEXPORT jstring Java_com_b44t_messenger_DcContext_getConfig(JNIEnv *env, jobject obj, jstring key)
 {
 	CHAR_REF(key);
 		char* temp = dc_get_config(get_dc_context(env, obj), keyPtr);
--- a/res/layout/preference_divider.xml
+++ b/res/layout/preference_divider.xml
@@ -8,7 +8,7 @@
 
     <View
             android:layout_width="match_parent"
-            android:layout_height="10dp"
+            android:layout_height="20dp"
             android:background="?pref_divider"/>
 
 </LinearLayout>
--- a/res/layout/preference_right_summary_widget.xml
+++ b/res/layout/preference_right_summary_widget.xml
@@ -10,6 +10,6 @@
               android:layout_gravity="right|center_vertical"
               android:gravity="right|center_vertical"
               android:textSize="16sp"
-              android:textColor="@color/delta_primary_dark"/>
+              android:textColor="@color/delta_accent"/>
 
 </FrameLayout>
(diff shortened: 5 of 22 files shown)


hpk42:
what is "for active chats" as option value?

r10s:
quote from irc: "show emails for active chats" shall mean you want to see root-non-chat mails from users you have a chat with

res/values/arrays.xml
```diff
@@ -153,6 +153,18 @@
...
+  <string-array name="pref_show_emails_entries">
+    <item>@string/off</item>
+    <item>@string/pref_show_emails_accepted_contacts</item>
+    <item>@string/pref_show_emails_all</item>
+  </string-array>
+
+  <string-array name="pref_show_emails_values">
+    <item>0</item>
+    <item>1</item>
+    <item>3</item>
```
hpk42: why not 2? ;)

res/values/strings.xml
```diff
@@ -382,6 +383,10 @@
     <string name="pref_watch_mvbox_folder">Watch DeltaChat folder</string>
     <string name="pref_auto_folder_moves">Automatic Moves to DeltaChat folder</string>
     <string name="pref_auto_folder_moves_explain">Chat conversations are moved to avoid cluttering the Inbox folder</string>
+    <string name="pref_email_interaction_title">Email interaction</string>
+    <string name="pref_show_emails">Show classic emails</string>
```
hpk42: huh, shouldn't this be the "no" option and is more like "only show chat mails"?

res/values/arrays.xml
r10s: i'm using bits to be prepared when we might want to separate "show mua-replies in chats" (bit#0) from "show all emails in an activity" (bit#1). in the core headers, these are named values, it's a bit ugly here, not sure if the xml stuff can use values from the java/native part.

res/values/strings.xml
r10s: "off" is no separate string but a standard string currently.

do you suggest to name the option sth. as "Show classic emails? - No, chats only|For accepted contacts|all" ?

hpk42:
On Tue, Feb 26, 2019 at 02:56 -0800, björn petersen wrote:
> r10s commented on this pull request.
> 
> 
> 
> > @@ -382,6 +383,10 @@
>      <string name="pref_watch_mvbox_folder">Watch DeltaChat folder</string>
>      <string name="pref_auto_folder_moves">Automatic Moves to DeltaChat folder</string>
>      <string name="pref_auto_folder_moves_explain">Chat conversations are moved to avoid cluttering the Inbox folder</string>
> +    <string name="pref_email_interaction_title">Email interaction</string>
> +    <string name="pref_show_emails">Show classic emails</string>
> 
> "off" is no separate string but a standard string currently.
> 
> do you suggest to name the option sth. as "Show classic emails? - No, chats only|For accepted contacts|all" ?

yes, that sounds good.

r10s:
this is how it looks like after rewording:

testbird:
Attempt to make the language more consistent:
```
Show classic email messages in addition to chats?
* No
* Only from accepted chat contacts
* Yes (INBOX browser)
```
It could also make sense to avoid the middle advanced option entirely, and just always show email messages from accepted chat contacts **as icons** within the contact's chat. They could always provide a reference to any ongoing email communication.

res/values/arrays.xml
r10s: as "show emails in an activity" is now always on again, there seems to be very few reasons left not to use `2` here. doing a commit.

adbenitez:
the use case for people on slow/paid networks and people that want to use Delta Chat as their main email client:
1- user receives an email from a classic MUA
2- user receives that email in Delta Chat is contact requests and accepts it, the emails from the accepted contact is moved when chat is accepted
3- any subsequent email received from this accepted contact is moved to DeltaChat folder since it belongs to an active chat, it doesn't matter it is sent from a classic MUA, user expect Delta Chat to be compatible with other MUA, so chatting with one shouldn't fill they INBOX.

adbenitez:
have this use case any chance to get implemented at some point in the future??

r10s:
> have this use case any chance to get implemented at some point in the future??

everything has chances :)

iiuc, the basic idea is that emails from class MUAs are treated as if they're always sending a `Chat-Version`-header? apart from the move, this is sort of the "Accepted contacts" option above.

however, it's not clear to me how this explicitly helps users to on slow/paid networks.

adbenitez:
users on slow/paid networks, don't want to download messages twice, if they receive them in an active chat in DeltaChat, they don't want to go to their classic email client and redownload the messages again, also it is a waste of time to go checking what message you already received in DeltaChat and delete them.
Also users in this situation don't expect messages to be downloaded if they didn't accepted the contact.
Real Life Story:
A friend who accepted to use DeltaChat, received a big(some MB) message in DeltaChat (in contacts requests list) which drained his quota/money a lot of time before he figured it out.

r10s:
i see.

however, this would probably add some additional ux questions and situations; not sure if we should enter this field just _now_.

wrt to avoid double downloading larger messages: maybe an option to delete a dedicated message _only_ for the server could already help a lot.

r10s:
> A friend who accepted to use DeltaChat, received a big(some MB) message in DeltaChat (in contacts requests list) which drained his quota/money a lot of time before he figured it out.

it's planned that the contact requests messages are not always downloaded, at least not with the "chats only setting".
however, before optimizing on that, we'd like to check how the settings above as such work out wrt interaction with normal clients.

messages should then pop up only in delta _or_ in normal MUAs - not in both by default - so that this would also avoid double data (when the downloading above is optimized)

adbenitez:
yup, the idea is that messages from accepted contacts are moved to Delta Chat folder even if they are sent from a classic MUA, for better compatibility with classic MUA

> the basic idea is that emails from class MUAs are treated as if
>they're always sending a `Chat-Version`-header?

adbenitez:
don't worry I am not talking about just now, I am looking the far horizon ;)
about UX, this may be an option users opt-in or as part of a "lite mode" option that minimizes consumption as much as possible.

adbenitez:
also options to prevent automatic  download of attachment, makes a lot of sense to me, and will mitigate this issues.

r10s:
> also options to prevent automatic download of attachment, makes a lot of sense to me

yip, these sort of optimisations are much more straight-forward and do no required long ux explanations with probably undesired side-effects.

adbenitez:
actually if I reply from delta chat to a classic MUA message, all subsequent messages from that chat (all replies from classic MUA to a Delta Chat message) are moved to the DeltaChat folder, seems like only the messages sent from a classic MUA that are not in reply to a deltachat message stay in INBOX, this is good enough, my point was not to leave replies that belong to a chat in INBOX, which deltachat already does!(at least this 0.101.2 preview r10s shared on IRC does)

testbird:
> I reply from delta chat to a classic MUA message, all subsequent messages from that chat (all replies from classic MUA to a Delta Chat message) are moved to the DeltaChat folder, seems like only the messages sent from a classic MUA that are not in reply to a deltachat message stay in INBOX, this is good enough

Ok, good if this works for you. There are also many people, however, who use email to exchange rather longer and formal messages, say about "serious work", and in parallel they have some daily chatting coordination stuff thread.

In this case, if accepting a short message as a chat makes every subsequent longer email (intendend for the classic MUA) appear in the chat is disturbing. And removing it from the INBOX (MUA) is a no-go blocker.

Disabling the backwards compatibility as in "No" from this pull request is not an option (and blocking new chat initiations from classic MUAs), if some users aonly have "desktop email".

testbird:
Err, and my point was to allow non-chatclient contacts to re-start chat, by considering _short-text_ classic emails as chat messages.

PR #746, opened 2019-02-27
disable glide disk cache

r10s:
the disk cache of the glide module seems to use some hashes or so.

just got the situation that a completely different image was shown in the chat. enlarging the image, the correct one was shown.

by disabling the disk cache (the memory cache is still active), the problem gone away.

i think we do not need the disk cache as all images are already locally, nothing gets downloaded from the ui side.

therefor this pr disables the disk cache (as already done for most other load processes) and avoids such problems.

might be related to https://github.com/bumptech/glide/issues/3544

caching is considered harmful.


hpk42:
+1 disable all the caches!

PR #747, opened 2019-03-01
[wip] using rpgp in android

r10s:
todo

- [x] add rpgp as submodule, similar to deltachat-core
- [x] add at lest armeabi-v7a and x86; [armeabi"no-v7" seems to be deprecated](https://developer.android.com/ndk/guides/abis#sa))
- [ ] create librpgp.a's by automatically when calling ndk-build, gradle or so
- [ ] check, what to do for f-droid (maybe, this sould go to another pr. however)

### DIFF
--- a/jni/Android.mk
+++ b/jni/Android.mk
@@ -883,7 +883,6 @@ LOCAL_MODULE     := native-utils
 LOCAL_C_INCLUDES := $(JNI_DIR)/utils/ \
 $(JNI_DIR)/messenger-backend/libs/openssl/include \
 $(JNI_DIR)/messenger-backend/libs/libetpan/include \
-$(JNI_DIR)/messenger-backend/libs/netpgp/include \
 $(JNI_DIR)/messenger-backend/libs/sqlite
 
 LOCAL_LDLIBS 	:= -ljnigraphics -llog -lz -latomic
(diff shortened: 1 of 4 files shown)


jni/Android.mk
```diff
@@ -851,6 +851,19 @@ LOCAL_SRC_FILES     := \
...
 
+
+################################################################################
+# rpgp
+################################################################################
+
+include $(CLEAR_VARS)
+
+LOCAL_MODULE    := librpgp
+LOCAL_SRC_FILES := /home/bpetersen/projects/rpgp/target/armv7-linux-androideabi/release/libpgp_ffi.a
```
hpk42: possible to specify this less bjoern-dependent?

hpk42: :)

r10s: just create a user account with the name `bpetersen` ;)

however, there is much more missing, the whole call of rustc crosscompiling and a bunch of other things ...

r10s:
the current state of the pr requires the commands

```
cross build --release --target armv7-linux-androideabi -p pgp-ffi --features nightly
cross build --release --target i686-linux-android -p pgp-ffi --features nightly
```

being executed in the rpgp directory which must be located at `../rpgp`. preparations for the commands to work are 

```
cargo install cross
sudo apt install docker.io
sudo usermod -a -G docker $USER  # may be needed, shell restart or rebook required afterwards
```

changing this to the submodule already located at `jni/rpgp` is probably easy.

however, i have no concrete idea how to change the tooling so that rust is called whenever ndk-build is executed (which then calls the makefile `Android.mk`.

and there is also f-droid, however, this could be a next step then :)

@hpk42 @flub @dignifiedquire any help on this would be warm welcome :) (however, next week i am in holidays :)

dignifiedquire:
Maybe this could help https://github.com/bbqsrc/cargo-ndk

r10s:
thanks for the link. however, i do not get how this can build the current [Android.mk](https://github.com/deltachat/deltachat-android/blob/master/jni/Android.mk) as well. but maybe i will habe a closer look next week when im am back from vacation :)

dignifiedquire:
https://github.com/Dushistov/rust_swig/tree/master/android-example might help as well, this integrates using gradle directly

hpk42:
isn't this PR outdated now that we go for rust-core?

r10s:
yip, also think so.

PR #751, opened 2019-03-02
add an experimental feature for high image compression

r10s:
this pr adds an experimental feature that introduces a much higher compression for jpeg images to save some bandwidth.

 

while the compression is still okay-ish for many images, eg. for photos containing text, the text typically gets less readable so that i would not change the defaults for "balanced" currently.

the real resulting image size always depends a bit on the content, however, the example above is a typical one; most images i've tried with the new settings are about 50K, however, there may be some that are still larger, jpeg does not give you an guarantee here.

### DIFF
--- a/res/values/arrays.xml
+++ b/res/values/arrays.xml
@@ -165,6 +165,16 @@
     <item>2</item>
   </string-array>
 
+  <string-array name="pref_compression_entries">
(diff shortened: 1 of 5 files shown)


cyBerta:
straight forward. looks good.

PR #752, opened 2019-03-04
fix formatting for gplay store

hpk42:
@r10s please upload it -- i didn't change content, just formatting which doesn't look good on gplay.

### DIFF
--- a/store/text.md
+++ b/store/text.md
@@ -4,67 +4,25 @@ Chat over e-mail. Elegantly Retro, Decentralized and Open. Privacy by Design.
 
 # Full description (max. 4000 characters)
 
-Delta Chat looks and feels like other popular messenger apps, 
-but does not involve centralized control, tracking or selling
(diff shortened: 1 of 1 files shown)


r10s:
> Shouldn't we translate the play store description?

yip, in general, this makes probably sense, however, not sure if it is really needed for the beta version.

r10s:
@hpk42 thanks for the update, uploaded it to gplay (will take some hours or so until really updated there, however)

webratte:
> in general, this makes probably sense, however, not sure if it is really needed for the beta version.

You have more then 1000 downloads. And I think the description for
version 1.0 will be the same.
And it would be great to start v.1.0 with much translations as possible.
That makes DC more international ;-)
Isn't it?

r10s:
not sure if the release will have the same description.
however, i thing sooner or later, we'll have translations, however, it seems to be some effort, not only for the translators.

webratte:
OK, probably your right.

PR #755, opened 2019-03-06
Location streaming

cyBerta:
should be merged only after merging https://github.com/deltachat/deltachat-core/pull/601


angelo-fuchs:
works on first attempt, superficial checks accepted.

Issue #744, opened 2019-02-25
log should be reachable during onboarding


r10s:
i would suggest to add the "view log" button (as a blue link) to the end of the "advanced" section.

adbenitez:
this would be useful since server registration is a common place for errors, I had needed access to this log in the registration view more than once.

florianhaar:
I will tackle this today

r10s:
@florianhaar thanks again, this can be closed by https://github.com/deltachat/deltachat-android/pull/758 now :)

PR #758, opened 2019-03-08
log should be reachable during onboarding #744

florianhaar:
Added link to Logview at the end of advanced settings on the registration page


r10s:
great, thanks for pushing this forward.
i only tweaked the layout a bit (things are left-aligned, not centered) and re-used the existing string (we're already sharing strings with desktop and ios, so in general, the strings are not really dialog-centered; we add explicit strings as needed)

Issue #434, opened 2019-01-17
sharing files does not work


r10s:
sharing images works, however, when trying to sharing random files (just tried with *.bak from the amaze file manager) delta is not offered in the list. this worked with the old f-droid-version, however.

also, sharing URLs still lack the title, this also worked with the old f-droid-version.

angelo-fuchs:
@r10s Should we add this to Jan2019?

r10s:
i think jan2019 is done so far :)
but we can handle this with high priority. should not be that hard.

violoncelloCH:
something for the project playOne?

r10s:
> something for the project playOne?

added it there as "optional", we'll see if someone is targeting it :)

Boehrsi:
I would look into that one on Friday :).

r10s:
great!

Boehrsi:
The general problem is that the Signal code base isn't able to handle multiple shared files at once and that's what some of the messengers do. I'm adjusting that.

r10s:
cool. as we do not have a method to stage multiple files, i think, as a quick-fix, it would be okay to ask the user "do you want to sent 3 files to _chat?_" after selecting a chat.

Boehrsi:
I'm nearly done with that one. Just the permission handling for required cases and some string adjustments are missing.

Boehrsi:
Did the last adjustments and will now test if everything is working fine. This ticket should definitely be ready this week.

r10s:
cool :)

r10s:
closed by https://github.com/deltachat/deltachat-android/pull/759 - @Boehrsi great job! and of course, my initial comment - "should not be that hard." - was wrong in the end :)

adbenitez:
@Boehrsi  +1

PR #759, opened 2019-03-08
sharing files does not work #434

Boehrsi:
Bigger changes than expected ahead:

- Added an intent filter for sharing of multiple files to the manifest
- Added strings for the required dialogs ("permissions" and "should send multiple files")
- Added handling for sharing via file paths (that's what e.g. Amaze does @r10s)
- Added permission handling if sharing via file paths is used, instead of content providers (otherwise the app would crash here)
- Added sharing of multiple files to the ShareActivity (a dialog comes up, asking the user if it's fine to send the files right away, as we can't show a preview / draft for multiple files at once)
  - General separation between one / multi file handling added
  - Logic added for message creation and message sending
- Cleanup for the resolving of attachements

In general i tried to keep all the existing logic of the "single file sending" and wrapped the "multi file sending" and "file paths sending" around it. I tested every case I could imagine extensively, so I hope no existing features broke.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -81,6 +81,18 @@
             <data android:mimeType="*/*"/>
         </intent-filter>
 
+        <intent-filter>
+            <action android:name="android.intent.action.SEND_MULTIPLE" />
+            <category android:name="android.intent.category.DEFAULT" />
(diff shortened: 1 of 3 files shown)


r10s:
@Boehrsi very cool, testing this now :)

Issue #753, opened 2019-03-04
[UI] Selecting all contacts but counter still not updated


adbenitez:
When I press one contact to start selecting them the selection mode is enabled, the contact is selected and the counter is in 1, then I go to the options and press "select all" all contacts are selected but the counter is not updated and remains '1' if I deselect a contact then the number is updated properly.

- Platform (android/blackberry/anbox): Android 
- Device:BLU
- Delta Chat Version: 0.101.1
- Screenshots:
![screenshot_20190301-215400](https://user-images.githubusercontent.com/24558636/53747149-a4950680-3e70-11e9-86e0-58d7e76a54bf.png)

Boehrsi:
Will look into that one right now :).

r10s:
fixed by https://github.com/deltachat/deltachat-android/commit/959e007e391f57fc4046fe17b63e8916d4dbe495 - thanks :)

PR #767, opened 2019-03-15
[UI] Selecting all contacts but counter still not updated #753

Boehrsi:
Added the updating of the counter also when selecting all entries via the menu.


r10s:
works like a charme :)

PR #790, opened 2019-03-26
disable invite activity for now

r10s:
via the "invite" activity, mainly phone-number-based-apps will be reached - there is no need to access email-based-apps as these can be contacted with delta directly.

however, as users may not have their email-credentials in mind, pushing them to switch from phone- to mail-based messengers directly, may cause more harm than success.

also, there is also the footer that already defaults to "hey, let's use delta chat".


r10s:
google-oauth also has some issues currently, maybe this is related, see https://status.cloud.google.com/incident/cloud-iam/19001
Screenshot_2019-04-18 Google Cloud Status Dashboard

PR #791, opened 2019-03-26
Map filtering 2

cyBerta:


r10s:
looks good to me, really looking forward to see the slider in action with data :)

if i have enough data, would the map already being updated? not sure at the moment if i just have no data of if the slider is no yet linked to dc_get_locations()

@cyBerta although there is no [wip] in the title, i am not sure if you want it to being merged, so i leave it up to you :)

Issue #748, opened 2019-03-01
received .webp image are not shown


adbenitez:
Some friend started sending me some stickers images with extension .webp, the received images are not shown as images in Delta Chat but as attachments, but if I take one of those images and send it, they are shown as normal images (with preview)

- Platform (android/blackberry/anbox): Android
- Device: BLU
- Delta Chat Version: 0.101.1

Boehrsi:
I will have a look why this behaves wrong.

Boehrsi:
I tried to reproduce that and on both devices (Android 8.0 and 4.4) I wasn't able to reproduce that problem. Does that still happen in version 0.200.0?

adbenitez:
just sent one image to a friend and he sent back an screenshot with the image shown as attachment, while I sent it and from the sender side looks normal, both use Delta Chat 0.200.0
if you wanna test, send a webp to me: adbenitez@nauta.cu

adbenitez:
preferible something small like the image I used for the test: https://github.com/chai2010/webp/blob/master/testdata/2_webp_a.webp

Boehrsi:
Okay, thanks for the information. I will continue testing with your test images on Friday. :+1:

Boehrsi:
I'm now able to reproduce the behavior and I'm on it to fix it.

PR #798, opened 2019-03-29
received .webp image are not shown #748

Boehrsi:
The sending of the message went totally fine, but on the receiving side the type "file" (instead of "image") is returned. To avoid that I set the correct mime type for WebP images. 
We can also consider adding the WebP format to the dc_msg_guess_msgtype_from_suffix method of DCC (https://github.com/deltachat/deltachat-core/blob/master/src/dc_msg.c#L904).


r10s:
i think it is better handled in the core, and probably https://github.com/deltachat/deltachat-core/pull/635 is sufficient

@Boehrsi would be great if you can give it a try and merge the core pr and close this one :)

PR #816, opened 2019-04-04
reword 'enter email address'

r10s:
this pr simplifies the hint in the 'add chat' activity from `enter name or email address` to just `enter email address`.

the reason for this is that entering a name only works for email-addresses already known (through system address book or received mails) while entering an email address _always_ works.
so, the hint that entering a _name_ always works is misleading.

also the text is shorter and won't be truncated.

### DIFF
--- a/res/layout/contact_filter_toolbar.xml
+++ b/res/layout/contact_filter_toolbar.xml
@@ -21,7 +21,7 @@
                 android:contentDescription="@string/email_address"
                 android:fontFamily="sans-serif"
                 android:gravity="center_vertical"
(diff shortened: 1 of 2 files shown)


hpk42:
hum -- i usually try to enter the name and if that doesn't work i try
to remember the e-mail address.  "E-mail address" i typically associate here
with the routable address part even though i know that the "display name" 
also belongs to it.  Curious what others say?

r10s:
well, maybe this is a non-issue and everything with `enter name or email address` is just fine.

it just came into my mind that this might be a bit unclear. i'm happy to close this and leave things as they are until we really run into real problems here.

PR #819, opened 2019-04-05
Update release-checklist.md

adbenitez:
just some little play with case and fixed 2 typos


r10s:
great, thank you :)
the test-instructions, however, was only a quick writedown, feed free to refine that as well. i think the most important, however, is, that it remains "doable" - if a tests needs 1 hour, it probably won't be done :)

PR #822, opened 2019-04-05
rework inbox-hello and inbox-zero screens

r10s:
this pr modifies the image that is shown on the first startup when the chatlist is typically empty and the image that is shown when all chats are archived ("inbox zero").

the old images were from the signal source and we want something own :)

preview:

 &nbsp; &nbsp; 

<details>
<summary>old versions</summary>
 

 
</details>


violoncelloCH:
I think that doesn't really fit into the design schema... the rest of the design is material design, isn't it? So why not use something that fits and is more attracting?

r10s:
there is no need that such graphics fit into the "material design".
and in various offline meetups, in general, there was a consens to go for something handdrawn.

r10s:
however, more important, the texts also need some adaption.
but this should go to a different pr/issue.

adbenitez:
I am "offline" and I can't see the image, but I am attracted by the idea that it is handdrawn :+1:

violoncelloCH:
so at least do the whole app in hand drawn style then (all UI elements)...
othewise I think it just looks like it's randomly combined stuff (found in a quick internet image search or something)
but of course it's all your decision... 
However if design doesn't matter then don't expect future contributions by me (and probably quite a few others) and consider the popularity of Delta Chat to depend quite a bit on design too...

r10s:
i appreciate the feedback :) - and of course, design matters.
it's not that the whole ui shall be handrawn, however, some graphics, why not at least experiment with that a bit? this is not that uncommon. 

however, please note that this is probably not the "final" state that will be "there forever".
it's all subject to change :)

esp. the "inbox zero" image may disappear at all, and i think this is also the image that will be more in the face of the user that uses the "archive" function regulary.
the other graphic can be regarded more as a side-graphic as in a website, it is shown only _once_ to the user.
however, also that one will might be replaced by some more explaining texts and my some _really_ helping graphic.

it's just a first step to get rid of the old signal-screen - the signal-blue did not fit well to the rest of the app btw.

also it was really too close to signal and we had to go for something else here anyway :)

r10s:
i think for the "inbox zero", we should just go for a short explaining text as "All your chats have been archived. Click on **+** to start a new one"; an image is not needed at all in this case.

r10s:
updated the graphics and the text.
the drawing is still subject to change and could be more explanatory.

hpk42:
@violoncelloCH introducing a hand-drawn graphic in am empty chat list screen does IMO not require having hand-drawn graphics everywhere.  Hand-drawn is also starting to get used here https://delta.chat/en/help 
Do you think it gives a random impression as if collected from some internet-search? :)

hpk42:
the image is a bit big, isn't it? 91KB. :)

r10s:
@hpk42 it's 58k now, including the phones and laptops, see image atop

i think this can be merged now :)

violoncelloCH:
thank you for considering!

I see the point that you quickly wanted something new to get rid of the signal graphics...

Of course hand drawn graphics are not 'bad design' as such. However I think you can't isolate design to a single element; you always have to keep the whole picture in mind... and as the App itself is kept in material design, it just doesn't make sense to me to introduce a completely different style for a specific aspect of it; I would rather go for something in the material design inside of the app...
On the other hand on the webpage the hand drawn graphics fit much better :)

also the newer hand drawn graphic is already much better than the postbox one with this circle around it... (as it's more neutral)

I don't know if it's worth the effort because I fear it could just be dumping around for weeks like the other design related suggestions/PRs I made, but I really like the idea of reproducing your hand drawn info graphic (which I would keep using on the website) in a more materialistic style (inspired by the opensource graphics at https://undraw.co/illustrations)...

adbenitez:
Finally I see the images, I don't anything about design but the graphics look good, to me they don't look out of place at all, and are much more useful than the previous one

hpk42:
thanks for your suggestions.  I am not sure yet about it.
it's likely we are going to have more discussions around
this at the upcoming in-person gathering end april in Kyiv. 
I'll see to post back what i gather from these discussions. 
Ideally, you would be part of these discussions directly ;)

PR #828, opened 2019-04-10
Dark Theme improvements

Ampli-fier:
- Set a Conversation list item divider
- Change the Color for the round button at the bottom of the Conversation list (Start Chat/Group)
- Add a default background image for the dark theme


cyBerta:
@Ampli-fier there's one little code issue you could clean up. I've commented it

Ampli-fier:
@cyBerta Thanks, just I do not manage to find any comment. Where is it?

cyBerta:
src/org/thoughtcrime/securesms/ConversationActivity.java  l. 756:
please replace `else if (!dynamicTheme.isDarkTheme(this))` with `else`

r10s:
looks good! thanks for pushing this forward!

some nitpicking:

- i would expect the light in the background image coming from the upper left (as for the light theme) - as a quick hack, maybe rotate the image just by 180 degrees?

- i think the color of the bubbles could be a bit _darker_ than the background then, not black, but close to

however, this can be done in a later pr, i'm very happy to merge this pr now :)

Ampli-fier:
I agree, rotated image looks better ... the initial version was for left handed people ;-)

The bubbles ... you mean the chat messages from the 'other' chat member?
Messages from myself have the Delta-Colour as background, which fits to the Attachment-button and should probably not be changed ?

r10s:
> Messages from myself have the Delta-Colour as background, which fits to the Attachment-button and should probably not be changed ?

there is not need to bind the self-bubbles-color to the attachment-button-color. i did this some time ago just to have "some color that fits" :) in the light-theme there is also no relation between these colors.

maybe, after rotating, we should check how it looks with "full black" for the other-bubbles in the dark theme. this would make some sense as in the light-theme, we're using "full white" there.

not sure for the self-bubbles then. maybe they're already fine :)

PR #842, opened 2019-04-15
Dark Theme improvements part 2

Ampli-fier:
Cleanup of color usages (use names instead of hex values, where names were already defined).
Better colors for dark theme (plus added new color names).
Rotate dark background by 180 deg.


r10s:
cool, thanks for pushing this foward.
i am not sure if introducing an additional color (purple) is needed, maybe i would prefer to go for a neutral gray instead.

but already now it's a great improvement and i think the dark theme is fine for now :)

adbenitez:
@r10s in the test version you shared 0.250 the main colour of the app seems to be the default blue from Google?? in dark mode it looks fine but in light mode it looks pretty bad contrasting with the main colour of DC

adbenitez:
(I am talking that the colour used for the text cursor and horizontal line of text fields, for switchs, and for touch effects, is now blue)

r10s:
@adbenitez at least on my phone this looks nice now :)
we've changed many things the last days, so, i would recommend having a look at the next release :)

adbenitez:
I'm craving it :)

PR #848, opened 2019-04-16
Tweak dark three

r10s:
some more tweaking, simplifying and cleanup


r10s:
@Ampli-fier some final little things i noticed, feel free to merge.
but i think the dark theme is really great now, thank you so much for your help and for pushing this forward!

Ampli-fier:
The scroll to bottom **button** (circle) is not visible when there is only the background behind. Only when there is a sent message or a sent picture behind it, it is visible. I'd prefer a lighter background for the button. Maybe similar to the attachment button.

r10s:
@Ampli-fier thanks for checking. you're right, both, fab and scroll-to-bottom could be more visible. 

- i've removed the commit that changed the fab-color, the old one was already nice
- i've made the scroll-to-bottom more visible in dark-mode

things get really nice :)

feel free to merge :)

Ampli-fier:
Thanks, happy to merge that :)

PR #878, opened 2019-04-30
Poi improvements

cyBerta:
* delete pois (DC_LOCATION_UPDATE event on core side missing yet)
* add labels below POIs if they have a longer description than 1 character
* add labels below last position marker showing the user's display name
* fix disappearing markers when info windows appears
* slight differentiation between POI markers and message markers


r10s:
cool :) looking forward to check this together with #881

r10s:
lgtm :)

PR #881, opened 2019-04-30
show own device location independent of location streaming

cyBerta:
* a little dot and a bearing indicator on the map show the user's current device position
* update of mapbox sdk required


webratte:
Now if this is implemented, maybe it's a step closer to "send current position once"!?

I would like it.

r10s:
very cool. in fact this is an important feature :)
@webratte thank you very much for pushing this forward :)
i will check this out as soon as possible (currently traveling without real network ...)

r10s:
very cool, seems to work prefectly as expected!

PR #884, opened 2019-05-01
do not check location-provider we've not requested

r10s:
as we've stopped to get locations from the NETWORK_PROVIDER a few days ago, it's probably also useless to check them.

@cyBerta however, i'm not sure, just came over that.
feel free to close or merge :)


cyBerta:
Right!

PR #886, opened 2019-05-01
init mapbox only as needed and disable mapbox-telemetry

r10s:
the current calls to mapbox seems to send some data to the mapbox-server as reported at https://support.delta.chat/t/end-of-privacy/393/7

this pr targets this issue by:

- initializing Mapbox only as needed
- calling functions to tell Mapbox not to send telemetry data

seems as if this is respected by Mapbox, however, if not, it might be needed to swich totally. there are a bunch of options, eg. at https://wiki.openstreetmap.org/wiki/Frameworks#Displaying_interactive_maps however, probably not all are _better_ options.


cyBerta:
Great!

PR #895, opened 2019-05-08
search during forwarding, forward to archives, forward to new contacts

r10s:


r10s:
one more thing:

- the three-point-menu right in the titlebar should be hidden as well

cyBerta:
@r10s all mentioned issues have  been addressed

r10s:
cool, will test a bit :)

r10s:
great. some comments:

- it's already better that the system-back-key does not close the app, however, i'd expect the last activity (the chat) to appear, not the chatlist

- bikeshedding: is there some additional space right of the "search" button? looks to be as if it should be more to the right.

however, it's a great progress, so i merge this, we can iterate from there and also target https://github.com/deltachat/deltachat-android/issues/901 then.

Issue #756, opened 2019-03-07
show a hint that qr-verification cannot be done without network


Hocuri:
- Platform (android/blackberry/anbox): Android
- Device: Moto G3
- Delta Chat Version: 0.101.1
- Expected behavior: The verified key should be transferred on the next internet connection
- Actual behavior: DC tries to transfer the key instantly, which doesn't work. To continue using Delta Chat you need to press "Cancel".
- Steps to reproduce the problem: Scan a QR invite code without internet

r10s:
i'd say this is by design, at least for now.

the verification is not only about transferring a key but also about receiving the key of the other side and about knowing that the other side knows that you have the key. this requires a network connection.

even if we could delay this in a way, there would be no clear verification result after scanning and it would be confusing ux-wise when and if you can add the contact eg. to a verified group.

however, there are rough ideas to allow verifications also through other channels that do not required a in-person-meeting and can be used when both sides have actually network.

Hocuri:
Well a dialog "No Internet connection. The contact will be verified when you are connected again" could appear.

r10s:
might be that the qr code is not valid "forever", so this is not really an option.

so, at least for now, network is required during out-of-band-verification :)

Hocuri:
> might be that the qr code is not valid "forever", so this is not really an option.

Do you mean that the keys change, e.g. after reinstalling DC? Or that they dynamically change?

r10s:
currently, the qr codes are more or less static to make testing easier (it's still an experimental feature!), however, there are ideas that the qr-code shown is only valid for a short period of time, say one minute.

in this case the out-of-band-verification needs network anyway.

to keep us the option in the future and to make things easier, it's seems to be appropriate to keep the network requirement for now. 

also, as mentioned, imho an out-of-band-verification is a "thing" that is done by two people at a given time with an immediate outcome. ux-wise everything else would be more complicated.

BachUhr:
> even if we could delay this in a way, there > would be no clear verification result after > scanning and it would be confusing ux-    > wise when and if you can add the contact   > eg. to a verified group.

But at the moment there is no notification or sth. like this, so it is not obvious that the verification cannot be done.
This is confusing too, maybe there could be a notification so that the user knows that an internet connection is necessary.

r10s:
> But at the moment there is no notification or sth. like this, so it is not obvious that the verification cannot be done.

yip, this is something we should target.

Hocuri:
[...]

PR #904, opened 2019-05-10
show a hint that qr-verification cannot be done without network #756

Boehrsi:
- Added a toast when entering / reentering the view
- Changed the hint text to inform about the network requirement (only if no internet connection is available)
- Added a connection listener to adjust the hint text if the connection type changes while in the view 
- Tested for general and group QR view

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -479,6 +479,8 @@
     <string name="qrshow_join_group_hint">Scan this to join the group \"%1$s\".</string>
     <string name="qrshow_join_contact_title">QR invite code</string>
     <string name="qrshow_join_contact_hint">Scan this to set up a contact with %1$s</string>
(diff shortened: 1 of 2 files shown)


r10s:
great pr :) works like a charme - even with network changes while the activity is shown.

however, i am wondering if the "toast" is needed at all. it shows basically the same information as already shown under the toast and results in a bad ui as the has to wait for the toast to disappear until he can read the _same_ information underneath.

therefore, i suggest to remove the toast completely. instead, we can show the text in red (0xff0000 should be fine for dark mode as well as for light mode) or add a warning-symbol or so beside the text (might be a bit more tricky).

r10s:
merging this, thinking it over, it might be fine as it is and we can iterate from here later as needed.

so, again, thanks for pushing this forward :)

Issue #901, opened 2019-05-09
fix sharing of multiple images


cyBerta:
Currently the PersistentBlobProvider seems to be buggy so that images are not always copied or loaded completely during the share process of multiple files at the same time. It results in half-drawn images getting shared.

r10s:
i did some recent cleanup there, maybe this is releated? https://github.com/deltachat/deltachat-android/pull/890

however, we should not revert the changes; they are needed as there were bugs in the (unused!) crypto parts which were some signal-relicts.

r10s:
just an idea: as this seems to happen for multiple images only - may it be that the PersistentBlobProvider gives that same id-or-whatever to all of them?

r10s:
closed by https://github.com/deltachat/deltachat-android/pull/913

PR #907, opened 2019-05-11
use cacheDir if externalCacheDir is not available

r10s:
just had this situation in an emulator, this commit fixes the issue. i think it cannot harm much as it is executed only if the first attempt fails (instead of throwing an error).

however, apart from that, the PersistenBlobProvider needs some more love anyway, see eg. #901


hpk42:
Looks legit ;)

PR #913, opened 2019-05-14
Fwd and share fixes

cyBerta:


cyBerta:
@r10s ready for testing/review

r10s:
merging, however, we should keep the removed `android:launchMode="singleTask"` in mind if weird things happen in the activity stack ;)

Issue #451, opened 2018-12-20
Notifications are too much intrusive


adbenitez:
Notification messages are too much intrusive/annoying. They are showing up while the user is writing in a chat or while seeing the chat list. In v0.20 the notification icon is shown in the "status bar" and sound is played but users aren't disturbed with a notification dialog showing up on the screen while they are chatting, this can be really annoying if the user start to receive several messages from other chats while in a conversation.
 
- Platform (android/blackberry/anbox): Android 6.0
- Device: BLU
- Delta Chat Version: 0.92

angelo-fuchs:
@adbenitez Could you post a screenshot of such a notification "showing up"? When I just tested it, the vibration went off, the sound was heard and the message scrolled through the status window, but nothing that would stop me from continuing writing. (Maybe my phone can't display that kind of notification which would explain why I didn't notice it earlier)

I will work on the frequency of the vibration / sound thing

adbenitez:
will start testing this, screenshots coming soon...

adbenitez:
At the chats list:
if user is at the chat list, and receives a message, do he really needs the notification??
As an user my UX with this is a total disaster I feel really bad with this since I just want to tap in the chat in the list, but the notification feels intrusive, when a chat have a message it is shown at the top of the list and the notification is blocking the view, all the info I need is already in the chat list, I think there are no need for a notification here...
Screenshot:
![screenshot_20190122-114145](https://user-images.githubusercontent.com/24558636/51552258-335c3f80-1e3e-11e9-9796-ced1e710ab9c.png)

r10s:
> if user is at the chat list, and receives a message, do he really needs the notification??

i think, if the user is in a chat or in a chatlist, notifications should be shown by an icon in the statusbar only.
and notifications for an opened chat can be skipped at all.

adbenitez:
(note: all this tests are in 0.98.2)
In-chat notifications (part 1):
When I am in a private (1x1) chat and the other user sends me a message, the app now works fine, no notification is created at all, since I am in the chat and can see the message, if the cell is in vibrate-mode it works fine and do not vibrate, but otherwise still the notification sound is played, mine is soft one, but a friend have set a notification sound that is quite noisy (rock music) and I can imagine other people with loud notifications also, all them will hate that the notification sound is played for messages from the current chat, 0.20 used a soft "bubble pop", we don't have to use the same, but at least something that is audible but not annoying like the old one.

adbenitez:
[...]

PR #915, opened 2019-05-15
Notification fixes

cyBerta:


hpk42:
while you are doing notification fixes -- would/could the behaviour
improve for when a lot (say 30 messages, some of them bigger)
are pending?  currently, it's vibrating/making a sound every
few seconds, for a long time.  This happens when travelling
or otherwise being offline for longer times but still participating
in chatty chats.

r10s:
@hpk42 i think, at least part of the problem is that on program start, _all_ unread mails are notified, not only the "fresh" ones. this will be changed. 

i think we'll see how this works out and already mitigates the problem. if not, we can iterate from there.

the problem are a bit the new "notification channels" where you have few control about a single message that is notified ("notification channels" are a MUST to be used when targeting the most recent android-versions :/

cyBerta:
> while you are doing notification fixes -- would/could the behaviour
> improve for when a lot (say 30 messages, some of them bigger)
> are pending?

Sure, I'll work on that.

hpk42:
On Thu, May 16, 2019 at 03:03 -0700, cyBerta wrote:
> > while you are doing notification fixes -- would/could the behaviour
> > improve for when a lot (say 30 messages, some of them bigger)
> > are pending?
> 
> Sure, I'll work on that.

no big worries, not a high prio. It only occurs if you have very 
busy chats and offline times in between, so i wouldn't classify it 
as "immediate annoyance when starting to use the app".

cyBerta:
@hpk42 I silenced all notifications within the first minute after app start. I tried it out and it helps a lot regarding the described issue.

hpk42:
On Thu, May 16, 2019 at 10:54 +0000, cyBerta wrote:
> @hpk42 I silenced all notifications within the first minute after app start. I tried it out and it helps a lot regarding the described issue. 

but if the app is killed (from Android) wouldn't it then be silence, not notify? 
or are you talking about the sound?

cyBerta:
> but if the app is killed (from Android) wouldn't it then be silence, not notify? or are you talking about the sound?

I'm talking about the sound. All messages appear in the notification bar as expected.

r10s:
> I'm talking about the sound. All messages appear in the notification bar as expected.

"All messages" are "all new incoming messages" then, right?
"old" messages that are already downloaded are not notified again then? i think that was part of the issue - and the main difference between signal-ui and telegram-ui.

cyBerta:
> "All messages" are "all new incoming messages" then, right?

Exactly.

r10s:
some comments:

- cleaning up notifications when entering a chat works as expected

- if there are two chats with unread notifications, say ChatA (1 unread) and ChatB (2 unread) and a new message comes in to Chat A, the app rings 3 times - 1 time for ChatA and again the 2 times for ChatB. The app should only for new messages propagaged by DC_EVENT_INCOMING_MSG - nb: just had a look: the old telegram-ui did rely completely on that event and uses dc_get_fresh_msgs() only to remove items from the notification bar. maybe this is the better approach to not notify too often.

- if we're in ChatA and a new messages comes in for ChatB this is not notified until the chatlist is opened again.  
i am not deep in the logic, however, also this should not happen when reacting mainly on DC_EVENT_INCOMING_MSG and using "current chat", "app is active" only as secondary signals to decide about show/dontShow, priority, sound on/off.  
i think entering/leaving an activity should trigger any notifications at all.

- for in-chat-sounds i'd use the sound-files i mentioned at https://github.com/deltachat/deltachat-android/issues/451#issuecomment-492772270

r10s:
i think the clear-notification-code also causes the other chats to be notified again ...

cyBerta:
>     * if we're in ChatA and a new messages comes in for ChatB this is not notified until the chatlist is opened again.
>       i am not deep in the logic, however, also this should not happen when reacting mainly on DC_EVENT_INCOMING_MSG and using "current chat", "app is active" only as secondary signals to decide about show/dontShow, priority, sound on/off.
>       i think entering/leaving an activity should trigger any notifications at all.

This behavior is intended. The idea behind it is: As long as people write in a chat, they should not be disturbed by incoming notifications from different chats. Those incoming notifications are in a pending state. they will be deliviered as soon, as the the user is not in a chat view anymore. It addresses the issue described here: https://github.com/deltachat/deltachat-android/issues/451#issuecomment-456504011

cyBerta:
>     * if there are two chats with unread notifications, say ChatA (1 unread) and ChatB (2 unread) and a new message comes in to Chat A, the app rings 3 times - 1 time for ChatA and again the 2 times for ChatB. The app should only for new messages propagaged by DC_EVENT_INCOMING_MSG - nb: just had a look: the old telegram-ui did rely completely on that event and uses dc_get_fresh_msgs() only to _remove_ items from the notification bar. maybe this is the better approach to not notify too often.

uff, thanks for the hint!!!

cyBerta:
>     * for in-chat-sounds i'd use the sound-files i mentioned at [#451 (comment)](https://github.com/deltachat/deltachat-android/issues/451#issuecomment-492772270)

The app has an option to set individual notification sounds per chat. Shall I remove that option?

cyBerta:
>     * if there are two chats with unread notifications, say ChatA (1 unread) and ChatB (2 unread) and a new message comes in to Chat A, the app rings 3 times - 1 time for ChatA and again the 2 times for ChatB. The app should only for new messages propagaged by DC_EVENT_INCOMING_MSG - nb: just had a look: the old telegram-ui did rely completely on that event and uses dc_get_fresh_msgs() only to _remove_ items from the notification bar. maybe this is the better approach to not notify too often.

I cannot reproduce that. Can we have a short direct chat session?

cyBerta:
Oh, I oversaw that comment: https://github.com/deltachat/deltachat-android/issues/451#issuecomment-492772270. Will implement it

PR #917, opened 2019-05-16
simplify profile

### LINKED ISSUE #428: combine profile, group editing, map, media+document gallery
not the most urgent issue, but in the mid-way, it is probably a good idea to combine these views in a single activity and navigate through these info with tabs.

maybe a single activity that takes a contactId _or_ an chatId as a parameter will be a simple approach.

cc @angelo-fuchs , we talked about this in kyiv

related, an probably more important is #546

### LINKED ISSUE #448: refine document messages
- [x] display in blue as for the old t'gram-version
- [x] the file-attach icon should also be in blue
- [x] staging area should be improved

see also https://github.com/deltachat/deltachat-android/issues/522

### LINKED ISSUE #846: Documents not shown in all media view
I received a pdf, and also a txt in a chat, I went to the "all media" view to see how this will look in the documents tab, but the tab is empty, is this a bug, or what is considered a document in DC?

- Delta Chat Version: 0.200.0

r10s:
- [x] create ProfileActivity with all needed tabs
- [x] show correct tab-names depending on self/group/single
- [x] open ProfileActivity as needed
- [x] make "profile" tab work
- [x] make "gallery" tab work
- [x] make "document" tab work

the links and the map tab however, should go into another pr.

closes #428, closes #846, closes #448

EDIT: all tabs and desired options are there, gallery, docs are working, shared-chats added as a completely new feature.

TODO: making all settings-buttons work and copy the functionality from the old activities. however, this should be done in a subsequent pr; for now the old activities are still available and functional in the chat-menu


r10s:
another impression:

hpk42:
really looking forward to trying this out ;)

webratte:
@hpk42 #metoo ;-)

Issue #910, opened 2019-05-14
close search-keyboard on click on "ok"


csb0730:
v0.200.0

When search result is presented keyboard is still there and covers result list.

![Screenshot_2019-05-14-22-23-27](https://user-images.githubusercontent.com/30560716/57730338-8c274080-7698-11e9-8add-4c340c1fc00a.png)

Keyboard should be closed when first results are coming up.

csb0730:
Maybe this is not an issue at standard devices, but at mini devices it is.

r10s:
> Keyboard should be closed when first results are coming up.

well, i do not think so. there is no "search" button, so we cannot really say when the user is finished with typing. the user may want to refine the search or remove characters to become broader.

> Maybe this is not an issue at standard devices, but at mini devices it is.

but for these small devices you can hide the keyboard by the back button. this is one extra tap, however, to really know when the search is done (see above), we'd need a search button which would also require an extra tap.

so, one extra tap on small devices, no extra tap on normal devices (most devices are larger than the screenshot nowadays)

the discussion can be continued in the forum, however, i'd suggest to close the issue here.

r10s:
just had a closer look: there is a "search" button on the keyboard that is currently unused.

- as the search is done incremental and automatically (normally fast ...), the "search" button should be converted to an "ok" button (as in the "new chat" activity where contacts can be search with an incremental search)

- a click on this "ok" should just hide the keyboard then. this is also the behavior in the "new chat" activity and also in many stock-android apps

i think this is a good compromise.

csb0730:
Good proposal :+1: 
In addition to that, search could be started when this button is pressed and not before? This would make debouncer unnecessary and speed up search process again :+1:

r10s:
> In addition to that, search could be started when this button is pressed and not before? This would make debouncer unnecessary and speed up search process again +1

well, i'd prefer making the additive search usable and/or check if #868 it is really a general issue :)

Boehrsi:
I will add the suggested behavior today :).

r10s:
cool :)

r10s:
fixed by #918

Issue #868, opened 2019-04-23
Search function very slow on dev-version, Android 4, S3 mini


csb0730:
v0.200.0, Android 4.1.2, S3 mini

Search function in chat view is very slow and unresponsive. IMHO much much slower than in old 0.20.0 !

Example:
I'm searching for a chat in chat list by entering first four characters in search input field.
Then:
1. after approx 10s (!) first results appear which shows 65 chats affected (strange!), then
2. after another approx 20s last results appear (2 chats and 4 mails; end result).

While searching time UI and phone is quite unusable.

Are there any changes in search functions since 0.20.0? I mean there was an issue in 0.20.0 regarding unresponsive UI but in 0.20.0 search was not done in separate thread.
With current 0.200.0 again search function get's much slower again :-(

Any idea? Are there several searches started in parallel with every next character typed?

csb0730:
Have a first approach:
Identified search functions in _chatlist.c_ `dc_chatlist_load_from_db()` and _context.c_ `dc_search_msgs()`.
Activated `dc_log_info()` in `dc_chatlist_load_from_db()` again.
In both functions a text query is used which starts search query with - every - char typed. Even with first char! 

My first approach is to start search with 3rd char, not with 1st. This shows a significant speedup of search function :)

csb0730:
See https://github.com/deltachat/deltachat-android/issues/296

The problem itself about overloading the device with unnecessary searches is still existant.

My 2nd approach now is: start search in chatlist with 3rd char (`dc_chatlist_load_from_db()`) and full text search (`do_search_msgs()`) with 4th char typed.

This makes search function much more responsive, gives response for contacts quite immediately and starts slow full text search later.
IMHO this makes sense because my standard use case for search function is finding contacts in huge chat list. Not full text search in msgs.

Maybe is would be an approach too, to separate full text search from contact and chat search.

csb0730:
@adbenitez 
You know many users with old mobiles. Are there no experiences about?

adbenitez:
Dear Christian Schneider,
I have  a pretty close friend that have a crappy phone: myself ;) 
my phone has like 450MB of RAM and android 6.0
I have more than 360 contacts, currently more than 190 chats, and more than 27000 chat messages,
If I search the letter "a" I get matching:
194 chats
284 contacts
23684 messages
I get te result instantaneously, maybe in 1 second.
[...]

PR #918, opened 2019-05-17
close search-keyboard on click on ok #910

Boehrsi:
- Changed the icon (from "search" to "ok")
- Keyboard is closed if "ok" is pressed
  - The current search stays intact
- Tested on Android 6 and 9


r10s:
cool, tested, lgtm :)

@Boehrsi as you're a bit deeper into search now: do you have a stance on #868 - is it also slow for you? or can you image why it is significantly slower on other devices? maybe it's just a broken index on just that device, not sure, however.

Issue #817, opened 2019-04-04
Wrong group picture shown in group settings


adbenitez:
I was added to a group, someone set an image for the group, which is shown in the chat list, and in-chat(inside the chat of the group) and when I touch the group name in that view it is also fine, but when I tap on the group icon to add/remove someone, an old picture (that once was the image of another group) is shown instead of the right group avatar.

- Platform (android/blackberry/anbox): Android 6.0
- Device: BLU
- Delta Chat Version: 0.200.0
- Steps to reproduce the problem: I don't know how this is happening, so it seems random, but has happened before with other groups, I just can't figure it out why.
- Screenshots:
### Real group avatar:
![Screenshot_20190404-124844](https://user-images.githubusercontent.com/24558636/55580246-25515780-56e8-11e9-878e-0c51bea571ee.png)
### Wrong avatar:
![Screenshot_20190404-124837](https://user-images.githubusercontent.com/24558636/55580235-1f5b7680-56e8-11e9-9242-fbf320fdfd66.png)

r10s:
we've disabled some glide-caches in the past, might be that this does not affect images already present on the device.

might be related to https://github.com/bumptech/glide/issues/3544

adbenitez:
had the same issue again yesterday, with the same group, pattern:
- lets call the buggy group X and the other unrelated group from where the images are taken group Y
- as I reported, I changed the image y1 of group Y for image y2 and in group X image y1 was shown in group X is settings
- now again, I changed the image y2 of group Y for image y3 and in group X image y2 was shown in group X is settings

adbenitez:
also happened to another group just now, after a member was added, the picture was a really old picture that once was the picture of the same group "Y"

Boehrsi:
I will try to reproduce and fix the problem.

r10s:
@Boehrsi cool. i have also noticed that.
it happens also with the self-avatar that can be set i settings/profile - changing the avatar here should also change the avatar shown in the "Me"-chat, however, the update does not arrive there.

r10s:
fixed by https://github.com/deltachat/deltachat-android/pull/924

PR #924, opened 2019-05-24
Wrong group picture shown in group settings #817

Boehrsi:
Sadly I wasn't able to reproduce neither the group image problem nor the "me" image problem. But I found some Glide caching stuff (https://bumptech.github.io/glide/doc/caching.html#skipping-the-cache) which could produce those problems. I disabled the disk caching as DCC provides the images anyway "on disk" and we don't need additional caching there. I kept the in memory caching as this definitely speeds up the repeated loading of images in our use case. @r10s could you check if these adjustments are fixing the problem?


r10s:
@Boehrsi thanks - i will have a look.
but it sounds reasonable.
destroy all caches!

r10s:
seems to do the job, let me cite the _two major programming problems: naming things, cache invalidation, off-by-one errors._

in fact, we had a similar issue https://github.com/deltachat/deltachat-android/pull/746 - and iirc, there is also an ongoing discussion if there is a bug in glide.

Issue #838, opened 2019-04-15
disable gps device when done with location streaming


r10s:
not totally sure, but i think i've notived several times that the gps sensor is not disabled even if location streaming was paused.

in general, the gps-sensor can be disabled when dc_set_location() returns 0. 

this issue might also be releated to the core, however.

r10s:
just had a closer look: 

- it is very easy to reproduce on my device when no gps positions come in

- one problem is that stopLocationEngine() is only called when gps-positions are retrieved - but when you are in a building and no more positions come in, stopLocationEngine() never gets called. i fixed that in https://github.com/deltachat/deltachat-android/pull/932/commits/46c80731602e4e3595fded6e4ec13c1ddec5420b

however, even with the fix, the location icon does not disappear (it disappears when the app is restarted,  however, so the internal core-state is fine)

- to see things clearer, i streamlined the init-functions in DcLocationManager() a bit: there are two start-functions but only one stop-function which confuses me a bit, i fixed that in https://github.com/deltachat/deltachat-android/pull/932/commits/46240ef0948cf361495a3c3a2bcffa08d97c99fa

@cyberta: you are deeper in the location-stuff, maybe you can take over if you find some time (it's not the most urgent thing. you find my commits in the pr you're https://github.com/deltachat/deltachat-android/pull/932

cyBerta:
I think, this ticket can be closed for now.

r10s:
@cyBerta have you done anything in addition to my pr #932 ? becasue #932 does not fix the problem, it just cleans up.

r10s:
after a talk with @cyBerta we decided to close this issue. one part is fixed and it is not reproducible on other devices. might also be that it is "just my" device and/or that the gps-closing is delayed for $reasons (or, without the fix from above, never took place)

so, closing this for now, there is no concrete actionable item left.

PR #932, opened 2019-05-26
 try to stop location-engine more gracefully

r10s:
see #838 for some comments


cyBerta:
looks straight forward and good to me

PR #935, opened 2019-05-28
Notification refactoring

### LINKED ISSUE #453: visual feedback missing when changing contact notifications
Changing the notification sound and vibrate options for a contact have no visual effect in the settings, they are set in the background but not visually in the contact is settings, in the case of the vibrate option if you go back an open the settings again it is updated with the user selection but the sound option still don't show the selected value.

- Device: BLU
- Delta Chat Version: 0.92
- Steps to reproduce the problem: Go to a contact is profile and try to change the the notification sound  and vibrate options, they are changed internally but not visually.

cyBerta:
* tested with android 4.1, 7.0 and 8.0
* show images in notifications
* reduce the noise on Android 0+
* fixed the summary notification layout for pre API 23: latest chats above
* removed dead signal code
* fixed  #453 (missing visual feedback for last selected notification sound)


cyBerta:
@r10s I think I've finally made it... ready for review

src/org/thoughtcrime/securesms/ProfileActivity.java
```diff
@@ -385,13 +385,16 @@ public void onSoundSettings() {
   }
 
   public void onVibrateSettings() {
+    int checkedItem = Prefs.getChatVibrate(this, chatId).getId();
+    int[] selectedChoice = new int[]{checkedItem};
     new AlertDialog.Builder(this)
```
ralphtheninja: Can we change this? It should be a function instead.

ralphtheninja: Nevermind, I thought this was JavaScript ;)

ralphtheninja: Just .. ignore me. It's for the best :wink:

r10s: :)

r10s:
lgtm, merging. i think if there are still problems, they should be targeted in a new pr.

Issue #447, opened 2019-01-04
show "new message marker"


r10s:
when entering a chat, there should be a marker above the first unread message.

this marker can disappear when sending messages and is also not important while chatting (unless the last messages is not visible due to scrolling away, cmp #731)

however, when the screen goes off and on again and new messages arrive in between, the marker should appear.

the last parameter for [dc_get_chat_msgs() (marker1before)](https://c.delta.chat/classdc__context__t.html#a51353ff6b85fa9278d2ec476c3c95eda) can help on this issue.

adbenitez:
is this implemented??? it was weird, but yesterday I have seen a dark marker and if I remember well a text saying the number of unread messages, but that is the first and only time I have seen this marker, it was kind of an illusion, please tell my I am not driving crazy!!!!

r10s:
afaik it's not yet implemented ;)

but maybe there is a hidden view that pops up only every 42 days or so for a fragment of a second :)

violoncelloCH:
![img_20190301_131250](https://user-images.githubusercontent.com/22591354/53657313-23622780-3c56-11e9-8da0-f8d6d8dd3797.png)
@adbenitez are you talking about this thing?
It's strange, because it appeared above a message I sent myself... and I sadly don't know how to reproduce, but at least I've been fast enough to take a screenshot when it appeared the second time...

adbenitez:
violoncello I am "offline" could you send the image to adbenitez@nauta.cu ??
I was so amazed that I could not take the screenshot, also this happened only once to me.

r10s:
@violoncelloCH thanks, interesting, so it's already there, but just not displayed accordingly

@adbenitez so you were not dreaming or driving crazy but you have very good eyes :)

r10s:
so, basically, the "new message marker" is there but not shown.

r10s:
closed by https://github.com/deltachat/deltachat-android/pull/940

PR #940, opened 2019-06-04
show new message marker #447

cyBerta:
* shows the 'x unread messages' header if 
1. the user has entered a chat, but the app is in background / the screen went off / another activity like the profile is on top of the chat
2. and a new message arrived

* the header disappears if:
1. the user sends a new message
2. a new message arrives while the user has opened the chat and the chat is visible
3.  the user goes back to the chat list

* no header appears if:
1. a new message arrived while the user has not entered the chat yet (a notification is visible in the notification bar and the unread message badge in the chat list though)

### DIFF
--- a/res/layout/conversation_item_last_seen.xml
+++ b/res/layout/conversation_item_last_seen.xml
@@ -4,27 +4,26 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
+    android:gravity="center"
     android:orientation="vertical"
-    android:paddingTop="24dp"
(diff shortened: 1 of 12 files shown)


r10s:
wrt "the header disappears if: 2. a new message arrives while the user has opened the chat and the chat is visible":

thinking it over, removing the "new message" marker on incoming messages may be a bit unexpected, esp. when a series of message is coming in and after the first, the screen is turned on, the "new message" marker will flash for a short moment and disappear.

i did removed this behaviour in the commit above 

(btw. just checked it: the old delta-ui also did not remove the marker on incoming messages)

PR #943, opened 2019-06-06
do not scroll down on list updates

### LINKED ISSUE #731: if not at the end of the chat, do not scroll down on new messages
When I am on a group and people are actively posting, it is really hard to read at my peace while this new post come out, since the view is scrolled.
IMO if the scroll frame is not at the end of the chat, the frame should not be scrolled, since this makes messages really hard to read if they are constantly moving, it is stressing for users.

- Platform (android/blackberry/anbox): Android
- Delta Chat Version: 0.101.1

r10s:
- [X] keep item offset
- [x] keep pixel position

closes #731

this pr also restores the just disabled scrollToLastSeenPosition() function as this did not get the desired effect. in fact, the app never _really_ scroll down on incoming messages, but the position was disturbed what, of course, is also annoying :)

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -450,10 +450,9 @@ public class ConversationFragment extends Fragment
         int pixelOffset = 0;
         if (!firstLoad) {
             oldCount = adapter.getItemCount();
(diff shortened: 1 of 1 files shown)


r10s:
@cyBerta thanks for figuring that out. works like a charm :) also, the snpping when removing messages is gone this way. great!

Issue #885, opened 2019-05-01
let manually downloaded and not updated APKs expire after some time


r10s:
we allow downloading and installing APKs direcly from github, sometimes we also provide testing APKs via other channels.

however, as these APKs are not updates they may to cause problems after a while. eg. we have APKs that cause group-splits still around although some around this bug are already fixed for a while.

an auto-update might be possible, however, this would not work for APKs given on special channels for testing purposes. also this leaks some information, requires some server work etc.

an easy to implement alternative might be to just check if the APK does  not come from a channel that provides updates after. if so, the APK expires eg. after 2 month. 

i think there is no need to avoid starting and using the APK, a clear informational message that the APK is probably too old. the message, however, has to  be annoying enough, i think about sth. in the chatlist (similar to the request to allow battery-usage) that cannot be clicked away

to detect APKs that are not updatable, see
https://stackoverflow.com/questions/11072079/how-to-know-if-an-app-has-been-downloaded-from-google-play-or-amazon
(while f-droid has its own applicationId, however, playstore and github downloads are equal. dev-versions always have `.beta` in their name, however. there are already a bunch of possibilities)

csb0730:
I think too:
- a clear informational message is ok
- prevent using is not ok

Timing should be extended (6 month?)

There should be an easy way to switch off this feature in sources.

Boehrsi:
If this is not that urgent I would have a look in the next weeks. :)

r10s:
next week is great :)
from the ui-side, i would suggest to go the same way as the DozeReminder that is shown from [ConversationListFragment::updateReminders()](https://github.com/deltachat/deltachat-android/blob/7c8ef694676e748b5068bdb396b5468445ac79b4/src/org/thoughtcrime/securesms/ConversationListFragment.java#L199) (and maybe add an function that makes the reminder non-discardable so that it is annoying enough to update at some time)

Boehrsi:
@r10s due to holidays / vacation the earliest date for the completion of this task is next week, not this week. Sorry for the confusion. Would this be also fine? Otherwise someone else should take over this issue I would say.

r10s:
@Boehrsi nice holidays i'd say :)
the issue is not urgent and i am pretty sure, it is still open when you're back - so, have a good time!

hpk42:
On Wed, May 29, 2019 at 09:30 -0700, Daniel wrote:
[...]

PR #945, opened 2019-06-07
Let manually downloaded and not updated APKs expire #885

Boehrsi:
Added the mentioned feature:

- A reminder (not dismissable) is shown in the chat list. It's marked with a special error color defined by the reminder class
- The reminder is only added if no other reminders are applied. I implemented it this way to avoid blocking of other reminders, as this one isn't dismissable
- It's shown only for apps where no installerPackageName is set and if the app wasn't updated within the last 90 days (I think this is a good value inbetween the suggested two vs. six months)
- If the user taps on the reminder a dialog is shown which allows the user to directly go to F-Droid or the Google Play store

Feel free to change the time frame value or the added strings. Please see the screenshots for more information.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -103,6 +103,11 @@
     <string name="message_delivered">Message delivered</string>
     <string name="message_read">Message read</string>
 
+    <string name="information_outdated_app_title">Your app version is outdated.</string>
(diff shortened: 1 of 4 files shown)


Boehrsi:
Screenshot_20190607-103912_Delta Chat
Screenshot_20190607-103902_Delta Chat

PR #947, opened 2019-06-08
rust core

r10s:


r10s:
nb: when it comes to supporting 64bit android as required for gplay since 1st august, in gradle, the abiFilters should be changed to "arm64-v8a", "armeabi-v7a", "x86_64", "x86" and the Application.mk also needs "x86_64"

Issue #937, opened 2019-05-29
mailto links with a body but no subject


Simon-Laux:
`mailto:email@example.com?body=content`
- Expected behavior: `content` in the text input field
- Actual behavior: `null\ncontent` in the text input field
- Steps to reproduce the problem: open `mailto:email@example.com?body=content`

adbenitez:
also support for mailto links inside delta chat is missing

Boehrsi:
I will try to fix that one today.

r10s:
great to see this fixed :)

PR #950, opened 2019-06-14
mailto links with a body but no subject #937

Boehrsi:
Adjusted our checks to avoid the wrong handling of the input.


r10s:
lgtm

PR #953, opened 2019-06-17
[wip] try to record video from quick-view

r10s:
experiment with recording video directly from the quick-camera-view with a long-tap.

it does not work for now, but this pr does some steps into the (hopefully) right direction. however, there is some more work needed ...


r10s:
closing for now, we can do a new pr as needed.

Issue #825, opened 2019-04-09
Date labels has wrong position after changing screen orientation


adbenitez:
If you change the screen orientation while in a chat, the "date labels" are not aligned correctly (see the screenshot)

- Platform (android/blackberry/anbox): Android 6.0
- Device: BLU
- Delta Chat Version: 0.200.0
- Steps to reproduce the problem: enter a chat, change screen orientation.
- Screenshots:
![Screenshot_20190407-161812](https://user-images.githubusercontent.com/24558636/55816497-5a7ef080-5ac0-11e9-9da3-3c7ef7a6cb6b.png)

r10s:
closed by https://github.com/deltachat/deltachat-android/pull/958

PR #958, opened 2019-06-18
fix layout issues with list item decorations (e.g. date labels) on...

cyBerta:
screen orientation change #825
(fixes layout issues of that kind in different places in the app)


cyBerta:
reminder to myself: minor performance improvement - only invalidate on orientation change (e.g. not on keyboard/keyboardHidden)

PR #963, opened 2019-06-20
Added saving the log to file

Ampli-fier:
Not sure about the file name and what is the best place to define it.

### DIFF
--- a/src/org/thoughtcrime/securesms/LogViewActivity.java
+++ b/src/org/thoughtcrime/securesms/LogViewActivity.java
@@ -7,6 +7,7 @@ import android.view.MenuInflater;
 import android.view.MenuItem;
 import android.widget.Toast;
 
+import org.thoughtcrime.securesms.database.NoExternalStorageException;
 import org.thoughtcrime.securesms.util.DynamicLanguage;
 import org.thoughtcrime.securesms.util.DynamicTheme;
 import org.thoughtcrime.securesms.util.Util;
@@ -57,9 +58,21 @@ public class LogViewActivity extends BaseActionBarActivity {
   public boolean onOptionsItemSelected(MenuItem item) {
     super.onOptionsItemSelected(item);
     Float newSize;
+
     switch (item.getItemId()) {
       case android.R.id.home:
         finish();
+        return true;
+      case R.id.save_log:
+        try {
+          if (logViewFragment.saveLogFile())
+            Toast.makeText(getApplicationContext(), R.string.pref_saved_log, Toast.LENGTH_LONG).show();
+          else
+            Toast.makeText(getApplicationContext(), R.string.pref_save_log_failed, Toast.LENGTH_LONG).show();
+        } catch (NoExternalStorageException e) {
+          e.printStackTrace();
+          Toast.makeText(getApplicationContext(), R.string.pref_save_log_failed, Toast.LENGTH_LONG).show();
+        }
         return true;
       case R.id.copy_log_to_clipboard:
         Util.writeTextToClipboard(this, logViewFragment.getLogText());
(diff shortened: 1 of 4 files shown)


r10s:
great to see that coming :)

wrt to the filename: I would avoid mixed upper and lower characters, sth. as `delta-chat-log.txt` seems to fit better. the txt-extension might be better than .log for some external processing.

maybe we can also add a date/time to the log, sth. as `dc-log-20190602-231233.txt` (with YYYYMMDD and HHMMSS) this would not make easier to save multiple logs.

csb0730:
Should be merged with priority as this helps saving effort in daily bugfixing :-)

Ampli-fier:
This is how the new format looks like: 

```
deltachat-log-20190624-185949.txt
```

r10s:
great, really great!

src/org/thoughtcrime/securesms/LogViewFragment.java
```diff
@@ -87,14 +94,49 @@ public String getLogText() {
...
+  public boolean saveLogFile() throws NoExternalStorageException {
+
+    File             outputDir   = null;
+    SimpleDateFormat dateFormat  = new SimpleDateFormat("yyyyMMdd-HHmmss");
+    Date             now         = new Date();
+    String           logFileName = "deltachat-log-" + dateFormat.format(now) + ".txt";
+
+    try {
+      outputDir = StorageUtil.getDownloadDir();
+      try {
```
cyBerta: there's no need for the nested try catch clause 
catch both exceptions in one place: `... } catch (IOException | NoExternalStorageException e) { ...`

src/org/thoughtcrime/securesms/LogViewFragment.java
```diff
@@ -87,14 +94,49 @@ public String getLogText() {
...
-    logPreview.setTextSize(TypedValue.COMPLEX_UNIT_PX, Textsize);
+  public void setLogTextSize(Float textSize) {
+    logPreview.setTextSize(TypedValue.COMPLEX_UNIT_PX, textSize);
   }
 
   public void scrollDownLog() { logPreview.setSelection(logPreview.getText().length()); }
 
   public void scrollUpLog() { logPreview.setSelection(0); }
 
+  public boolean saveLogFile() throws NoExternalStorageException {
```
cyBerta: Is NoExternalStorageException thrown for real? It looks like you caught it in the try catch clause

src/org/thoughtcrime/securesms/LogViewActivity.java
```diff
@@ -57,10 +58,22 @@ public boolean onPrepareOptionsMenu(Menu menu) {
...
       case android.R.id.home:
         finish();
         return true;
+      case R.id.save_log:
+        try {
+          if (logViewFragment.saveLogFile())
+            Toast.makeText(getApplicationContext(), R.string.pref_saved_log, Toast.LENGTH_LONG).show();
+          else
+            Toast.makeText(getApplicationContext(), R.string.pref_save_log_failed, Toast.LENGTH_LONG).show();
+        } catch (NoExternalStorageException e) {
```
cyBerta: This try -  catch seems to be redundant (see my comments in LogViewFragment)

src/org/thoughtcrime/securesms/LogViewFragment.java
Ampli-fier: You may be right. I tried different approaches and this might be a leftover. I'll give it a try.

src/org/thoughtcrime/securesms/LogViewActivity.java
Ampli-fier: Android Studio forced me to use this here (as well).

cyBerta: The reason is that you've defined the method like this `.. saveLogFile() throws NoExternalStorageException`. Remove `throws NoExternalStorageException` and Android Studio will be your friend again.

Ampli-fier:
Thanks @cyBerta. The code looks much better now. PR is coming ...

Issue #91, opened 2017-05-18
Text input field size


bokf:
On my device when I write a message it shows only 4 lines of my message. After that the text input field does not expand any more and I have to scroll to read my full text before sending. I would appreciate this text input field size to be a little bigger, for example two more lines, to read and correct the written message before sending more comfortably.
Vielen Dank!

r10s:
Well, I'm not sure if _just more_ lines fit on smaller displays.

bokf:
OK, I thought it could be a short thing. I tested with Kontalk and Conversations and both seem to have no limit.
![screenshot_2017-05-19-08-17-23_cr](https://cloud.githubusercontent.com/assets/25562831/26236081/d64985e4-3c6e-11e7-84b3-94c2a87abd8d.png)
![screenshot_2017-05-19-08-18-27_cr](https://cloud.githubusercontent.com/assets/25562831/26236086/da4da4d6-3c6e-11e7-90c0-0ac88dc77f40.png)

r10s:
Yeah, I would also vote for it :-) Any PR are welcome :-)

kskarthik:
AFAIK even WA & Telegram has same problem...

r10s:
closing this for now here, of course, the discussion can go on in the new support forum at https://support.delta.chat

csb0730:
@bokf 

Tried to tweak max input lines to 10 and it works well on my small S3 mini.

The only change which is needed is in
`deltachat-android/res/values/styles.xml`
where the line `<item name="android:maxLines">4</item>` needs to be changed to `<item name="android:maxLines">10</item>`

That's all.

@r10s - maybe You could do this directly without a PR because I need some time to get my repo updated again with local sources.

csb0730:
@r10s - as long as the change is not done this issue should be reopened.

csb0730:
Easier than thought. See PR.

PR #970, opened 2019-06-23
use larger input-view and smaller chat-view

csb0730:
Extend size of message text input field from 4 to 10 lines max.

### DIFF
--- a/res/values/styles.xml
+++ b/res/values/styles.xml
@@ -71,7 +71,7 @@
     <style name="ComposeEditText" parent="@style/Signal.Text.Body">
         <item name="android:padding">2dp</item>
         <item name="android:background">@null</item>
-        <item name="android:maxLines">4</item>
+        <item name="android:maxLines">10</item>
         <item name="android:maxLength">64000</item>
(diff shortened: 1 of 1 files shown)


csb0730:
This PR tackles #91

r10s:
@csb0730 thanks for pushing this forward :)

some additional lines are welcome, however, the text input field should not be larger than, say, 50% of the available height so that the user can scroll around in the message list while typing.
not sure if this is achieved by this change.

this probably needs some testing on different devices.

as we plan a release this week, I would tend to merge this _after_ this release so that some larger testing "in the wild" can be done before.

Ampli-fier:
I tested on my phone (larger display) and my personal feeling was that 10 is too much. 8 fits better.

Ampli-fier:
Is it possible to have this field resizable? That the user can move the upper border to a preferred position?

adbenitez:
Maybe not to have it resizable but alternating between the default mode and a fullscreen mode, where the text area fills the screen, may be easier and good enough

r10s:
replaced by https://github.com/deltachat/deltachat-android/pull/976

PR #971, opened 2019-06-24
share to single contacts (#900)

### LINKED ISSUE #900: Share content from share menu to single chat
It would be nice to get the option to share content to a certain contact/chat directly from the share menu instead of opening delta chat first and then select the chat the content should be shared with.

cyBerta:
closes #900

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -92,10 +92,9 @@
             <data android:mimeType="text/*"/>
             <data android:mimeType="*/*"/>
         </intent-filter>
-
         <meta-data
                 android:name="android.service.chooser.chooser_target_service"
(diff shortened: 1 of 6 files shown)


r10s:
cool! looking forward to test this later!

r10s:
i rebased that to master.

cyBerta:
> would be great if the group- and chat-icons are shown in the direct-share dialog.
> 
> currently, the fallback-character-icons are shown which is a bit strange as they are not used on other places in the app if a group-image is set for a group.

https://github.com/deltachat/deltachat-android/pull/971/commits/978c4b4f0b1a91623ac12002690834e11c1babb4 fixes that

PR #973, opened 2019-06-25
don't allow to remove yourself from the multiselect contact list, 

### LINKED ISSUE #833: Do I need to add myself when creating a group?
When creating a group, I can check or uncheck "Me" in the list of group members.

group_creation
("(Me)" added as translation for the german "Ich" in the screenshot)

I am not sure, what happens, when I uncheck myself, but usually I would expect to be part of any group I create.

So either the entry is redundant and confusing, or I do not really understand why there is a function to create a group without being part of the group.  
(There may be use cases for this, but they are not obvious for an average user)

cyBerta:
fixes #833
!!! This PR doesn't allow to remove yourself from any multiselect contact list !!!
- this means using the "add members" item in profile -> group tab will not allow the user anymore to leave a group
- leaving a group is restricted to the corresponding menu item "Leave group"

### DIFF
--- a/src/org/thoughtcrime/securesms/GroupCreateActivity.java
+++ b/src/org/thoughtcrime/securesms/GroupCreateActivity.java
@@ -19,6 +19,7 @@ import android.widget.ImageView;
 import android.widget.ListView;
 import android.widget.Toast;
 
+import com.b44t.messenger.DcContact;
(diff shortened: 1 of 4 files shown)


r10s:
seems to work as expected :)

PR #976, opened 2019-06-26
make input field 25% larger :)

r10s:
replaces #970 

we cannot use more lines without getting problems or limited functionality on displays smaller 4 inches.

so using 6 or even 10 lines or resizable field (i think i would not really like that, too fiddly) would require some larger refactoring. maybe we can also use smaller padding atop and abottom of the field, however, also this is out of scope for a "fast fix today"

### DIFF
--- a/res/values/styles.xml
+++ b/res/values/styles.xml
@@ -71,7 +71,7 @@
     <style name="ComposeEditText" parent="@style/Signal.Text.Body">
         <item name="android:padding">2dp</item>
         <item name="android:background">@null</item>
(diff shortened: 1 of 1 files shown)


csb0730:
Small step but better than before :)
I would test with max 6 lines as an ideal compromise (I'm coming from 10)  ;-)

r10s:
6 is already too much on a nexus s or similar screens.

csb0730:
Ok, I've only my one device. Maybe the nexus is smaller than mine ;-)

PR #987, opened 2019-07-01
 update image editing functionality to latest signal code 

### LINKED ISSUE #664: crop images before sending
All users I have shared Delta Chat v0.100 with, complain about the new image editor, the old image editor (I guess it is the system build-in one) has options to crop the image, this is not possible now with this new image editor.
They mostly want to crop the image to reduce the size, so options for scaling, or selecting image quality/compression (ex. high, medium, low) would be really appreciated.

- Delta Chat Version: 0.100

cyBerta:
- crop images before sending
closes #664

- requires at least android 4.4 kitkat


r10s:
looks great.

some nitpicking:

- there is still a weird animation when the editor is opened

- I think the aspect-ratio-lock is more confusing than really helpful: (a) it is initially enabled, this prevents easy and fast cropping (b) typically some things are cut from an image, in the majority of cases, this changes the aspect ratio (c) when enabled, the corners lock the ratio but the borders do not, which is even more confusing. I think, in sum this is too complicated and not needed at all.  
I would disable locking by default and remove the aspect-ratio-lock-button at all.

- it is great that all controls are at the top now. if easily possible, it would be great if we can left-align the tools and right-align the checkbox. the leftmost tools should be the "undo" or "delete" function.

- in landscape the checkbox collides with the system controls. a simple fix would be to use just a normal system-button-control in manifest.xml - maybe it would be a bit smarter to leave some border at the right in landscape, however, if this gets too complicated, the fist solution is just fine :)

r10s:
as this pr introduces [webp images that are supported only partly in minTargetSdK=14](https://developer.android.com/studio/write/convert-webp), we should check how this behaves on old phones; i am checking this in the emulator just now :)

r10s:
i [played](https://stackoverflow.com/questions/15380257/status-bar-flicker-when-exiting-fullscreen-activity) [aroud](https://stackoverflow.com/questions/5431365/how-to-hide-status-bar-in-android) to get rid of the flickering status bar, without real success.

however, i found out that if we start fullscreen using setFullscreen() and using a titleBar, everything is fine. this is basically the same as we do in the preview window.

maybe this is the way to go: move the buttons to the title bar as totally normal icons (they would be again right-aligned then, however) we could also use a normal back-button then. and: on smaller screens, the icons would go to the overflow menu.

however, i'd say it is also fine as it is now. i think the flickering is more an android bug than something we did wrong ;)

Issue #996, opened 2019-07-06
Sharing from special app crashed


webratte:
It crashed by sharing from this app: https://play.google.com/store/apps/details?id=de.badischezeitung.smart

There are also other News Apps who using (probably) the same software as Badische Zeitung.
So I think this issue also concerns also other Apps.

- Platform (android/blackberry/anbox): Android 7
- Device: Huawei nova
- Delta Chat Version: 0.500.0
- Expected behavior: Sharing links to DC
- Actual behavior: Chrash of DC

Boehrsi:
I would have a look on Friday, if the issue isn't fixed until then.

Boehrsi:
The PR should fix the problem. The actual problem was, that the mentioned app attached a SpannableString instead of a String. Kind of uncommon and "special", but the Delta shouldn't crash. Both cases are working now and I added a check to avoid further crashes, if no valid text is given.

r10s:
great that you figured that out!

r10s:
fixed by https://github.com/deltachat/deltachat-android/pull/1003

PR #1003, opened 2019-07-12
Sharing from special app crashed #996

Boehrsi:
Fix for #996

- Changed the extra type to char sequence to cover String and SpannableString
- Added a null check to avoid crashes

### DIFF
--- a/src/org/thoughtcrime/securesms/ShareActivity.java
+++ b/src/org/thoughtcrime/securesms/ShareActivity.java
@@ -213,7 +213,14 @@ public class ShareActivity extends PassphraseRequiredActionBarActivity implement
 
   private Intent getBaseShareIntent(final @NonNull Class<?> target) {
     final Intent intent = new Intent(this, target);
-    CharSequence text = getIntent().getCharSequenceExtra(Intent.EXTRA_TEXT);
+    String text = getIntent().getStringExtra(Intent.EXTRA_TEXT);
+    if (text==null) {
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/ShareActivity.java
```diff
@@ -213,7 +213,10 @@ private void handleResolvedMedia(Intent intent) {
 
   private Intent getBaseShareIntent(final @NonNull Class<?> target) {
     final Intent intent = new Intent(this, target);
-    setSharedText(intent, getIntent().getStringExtra(Intent.EXTRA_TEXT));
+    CharSequence text = getIntent().getCharSequenceExtra(Intent.EXTRA_TEXT);
```
r10s: after all this mess, i would be a bit more conservative and call getCharSequenceExtra() only if getStringExtra() returned null before.

i would not be surprised if there are implementation that (willI) return null on getCharSequenceExtra() and the text on getStringExtra().

r10s:
@Boehrsi i created a pr for this pr that implemented what is suggested above (https://github.com/deltachat/deltachat-android/pull/1004)

maybe you can check this out with the apps you're already using for testing.

with this change, the flow is more like in the past and uses the common getStringExtra(), so if google decides that getCharSequenceExtra() behaves differently, we're save.

the null-pointer check, however, is not changes.

Boehrsi:
@r10s I reviewed and retested your changes made in #1004. Works as expected and can get merged. :+1:

r10s:
@Boehrsi thanks for testing 👍 merging that then :)

PR #1010, opened 2019-07-23
bring back Shortcut badge

### LINKED ISSUE #962: add badge counter
Since upgrade from v0.200.0 to v0.304.0 I see no new message counter any more beside DC icon at base screen.
In the version before I saw a small orange bubble which indicates the number of new messages.
This is gone away.

cyBerta:
fixes #962
* implements unread message counter for launcher icons of several launchers
* unread messages != new notification count

I think we should rethink the approach of having two different counters - one that reflects the notification state (representing new incoming messages since last app start) and unread messages (representing the number of not marked seen messages) since this might cause confusion.
I've implemented the launcher icon badge as unread message counter instead of a new notification counter since the former is a much more reliable information. For example, if the app gets killed by the system all notifications get automatically removed but there's no guaranteed hook (onDestroy() is not necessarily called)  to reset the badge counter. In this scenario there would be chances that the badge counter is inconsistent with the notification counter but also with the overall unread message counter.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,7 +45,7 @@ dependencies {
 
     implementation 'com.journeyapps:zxing-android-embedded:3.4.0' // QR Code scanner
 
-    implementation "me.leolin:ShortcutBadger:1.1.16" // display messagecount on the home screen icon.
+    implementation "me.leolin:ShortcutBadger:1.1.22@aar" // display messagecount on the home screen icon.
     implementation 'com.jpardogo.materialtabstrip:library:1.0.9' // used in the emoji selector for the tab selection.
     implementation 'com.github.chrisbanes:PhotoView:2.1.3' // does the zooming on photos / media
     implementation 'com.github.bumptech.glide:glide:4.5.0'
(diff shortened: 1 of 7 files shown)


cyBerta:
This is also the behavior I can see in signal on my test device btw.

csb0730:
Integrated pr code in my local version 0.500.0:
-> works as expected :-)

@cyBerta: Good job :+1:

r10s:
i close this for now. we should solve the actual notification problems instead of adding new features on top of unreliable things.

we might come back to this, once, the situation is settled a bit, however, in general, i would vote for "we just notify the system" and the system cares for the rest. everything else seems to be a call for trouble.

csb0730:
Hi @r10s 
I hope Your idea is working as expected, but with v0.510.1 (Google release) I got many complaints from my friends that DC is not reliable: means that icon shows no new mails (missing badge!).
The mini delta sign from system most of times is overseen! :-/

r10s:
@csb0730 as said, it's not about never doing this, but we should solve the _basic_ problems about notification first - before opening the doors for new ones :)

csb0730:
Hi Björn,

yes, I see. There seem to be many bugs to fix in new version =:-|

I cannot assist in the moment :-(



Am 16.10.19 um 01:31 schrieb björn petersen:

PR #1017, opened 2019-08-08
c-core-64bit experiment

r10s:
this pr tries to build the c-core at 64bit.

in general, this works, the c-core as such is 64 bit compatible, however, it is not clear if the used openssl library is.

while testing, however, opening the "new chat" activity leads to a crash "Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR)" which seems to be releated to openssl used by MessageDigest, see:                                     
https://stackoverflow.com/questions/17840521/android-fatal-signal-11-sigsegv-at-0x636f7d89-code-1-how-can-it-be-tracked <- **so it might be an android issue, not related to core-c**

might also be related to emulator-only, searching around, i found some hints to that.

further tests:

- check if the arm64 works, however, this would require a emulator executing arm, this should be possilble, but i did not manage to create one (startup took forever)

in general, all this might not be required as we'll probably switch to rust. but it would be good to know if the issue is relate to the core or to android/emulator.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -113,7 +113,7 @@ android {
         buildConfigField "String", "MAP_ACCESS_TOKEN", '"pk.eyJ1IjoiZGVsdGFjaGF0IiwiYSI6ImNqc3c1aWczMzBjejY0M28wZmU0a3cwMzMifQ.ZPTH9dFJaav06RAu4rTYHw"'
 
         ndk {
-            abiFilters "armeabi", "armeabi-v7a", "x86"
+            abiFilters "arm64-v8a", "armeabi-v7a", "x86_64", "x86"
         }
     }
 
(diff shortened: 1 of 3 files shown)


csb0730:
Hi @r10s 
I thought all activities to c-core are stopped? You said all activities are done in rust core now?
Isn't it?

r10s:
@csb0730 yip, that's true. activities to c-core are more of less stopped. but this pr is not about modifying core-c. it's about building core-c in 64bit on android.

csb0730:
Ok, I ask because I'm still waiting to get a go to implement new import/export backup procedure. You know?!

hpk42:
i think it's rather about using the new core for 64 bit installs so guess we can close this experiment.

PR #1037, opened 2019-09-22
Permanent location streaming with alternate location sources

ganskef:
Please have a look at my ideas heading README.md. I'm using it every day since weeks on my Samsung S3 with LineageOS, but this Location DRAFT is not intended to integrate directly. I'm looking for your comments...


hpk42:
I like the README and the thoughts behind it :) Also trying to integrate with GPSLogger seems like a good idea. We don't want to bloat deltachat too much itself ;) 

If introducing permanent location streaming we probalby need to think about how we can appropriately warn -- to prevent abuse.  A simple "confirm-dialog" is not enough.  It's already good that each chat has a location-streaming symbol both in the chat and the chatlist view.  Maybe it'd be good to be disable location streaming more easily?

cyBerta:
https://github.com/mendhak/gpslogger/

cyBerta:
Hey, thanks for the draft, and sorry for the looong delay. We've been leaving Android more or less aside for a while. I think working on this feature has no priority for the core team but it seems like you're an experienced Android developer who might be able to push this topic forward. 
In my personal opinion I would rather look how gps logger is optimizing the battery usage and either use parts of it or rewrite parts, stripping the necessary code down to the bare minimum. 
My gut feeling tells me also that it is rather unusual to depend on an app that has not even a AIDL API. I will write some more comments in the diff.

cyBerta:
I also remember some discussions about the social implications writing an app that allows permanently location streaming. How do you prevent abuse of this feature? What about people stalking their partners by putting pressure on them to turn this feature permanently on?

cyBerta:
I also remember some discussions about the social implications writing an app that allows permanently location streaming. How do you prevent abuse of this feature? What about people stalking their partners by putting pressure on them to turn this feature permanently on?

README.md
```diff
@@ -1,3 +1,22 @@
...
+
+While I'm not using Google Play Services on my device `NetworkLocationProvider`s aren't available. Permanently using GPS Sensor is draining the battery very fast. **GPSLogger for Android** is a battery efficient GPS logging application. It's better to use its data than resolving it in Delta Chat.
+
+GPSLogger can be started and stopped with an Intent triggered by Delta Chat. The LocationBackgroundService in Delta Chat is disabled if GPSLogger is installed on the device.
+
+It makes no sense to monitor GPS and send a bunch of mails if you're connected with an **well known WiFi** access point so it's stopped at home for example. A Point Of Interest (POI) sent to yourself with the name of the router and suffix ' *' is used to define its location. On using such POI it will be recreated if it is older than a day to be shown on the map.
+
+Additionally if the device is laying on your desktop or the night table in a hotel room GPSLogger is disabled too by **detecting missing movements** of compass heading and location distances.
```
cyBerta: This is a very discrete improvement +1

README.md
```diff
@@ -1,3 +1,22 @@
...
+
+Therefore contacts with name ending with ' *' will enable location streaming any time if disabled. TODO: A regular feature should implement a contact property instead of a name extension of course.
+
+While I'm not using Google Play Services on my device `NetworkLocationProvider`s aren't available. Permanently using GPS Sensor is draining the battery very fast. **GPSLogger for Android** is a battery efficient GPS logging application. It's better to use its data than resolving it in Delta Chat.
+
+GPSLogger can be started and stopped with an Intent triggered by Delta Chat. The LocationBackgroundService in Delta Chat is disabled if GPSLogger is installed on the device.
+
+It makes no sense to monitor GPS and send a bunch of mails if you're connected with an **well known WiFi** access point so it's stopped at home for example. A Point Of Interest (POI) sent to yourself with the name of the router and suffix ' *' is used to define its location. On using such POI it will be recreated if it is older than a day to be shown on the map.
```
cyBerta: I like also here the general idea of turning off the location when entering a known wifi. UX wise I think there needs some more brainstorming how to allocate a location to a wifi

ganskef:
Hi, today I've merged upstream tracking to commit 5e1f2793 before committing some notes and screenshots. Unfortunately the experimental Location feature is broken now:

* My existing locations aren't visible on the map anymore.

* New location mails are flooding: 5 mails in s few seconds with one minute delay, subject: "Chat: Location", encrypted, but no attachments.

I'm pretty sure to have a proper (new) tool chain since every thing is successfully built.

I'll try to dig into it but I argue it's in the core somewhere. Any ideas?

r10s:
> I'll try to dig into it but I argue it's in the core somewhere.

yip, it is very probable a rust-core-issue. i think it is not really tested after the rust-move - so it would be great if you can have a look at that.

ganskef:
Hm, I've answered https://github.com/deltachat/deltachat-android/pull/1037#issuecomment-534603689 from @hpk42 with a mail reply at Sat, 28 Sep 2019 14:59:06 +0200 but it's not in this thread (?), sorry:

Permanent location streaming is enabled for very special contacts named with "NAME *" (DRAFT) only. Every 24 hours you send two text messages: Location streaming disabled and Location streaming enabled to the contact. The sent text messages are a noisy warning already. I think you
can't forget it.

The location symbol is visible in the chat(s). This is not modified.

To disable streaming modify the contact (strip " *") and hit the location button in the chat(s) attachments. Good idea to implement a enable/disable setting action in the contact.

link2xt:
@ganskef 
Location streaming fixes were merged into rust core recently: https://github.com/deltachat/deltachat-core-rust/pull/782

ganskef:
@hpk42 @link2xt thanks for investigating! I've pulled the core submodule to the master branch including deltachat/deltachat-core-rust#782 and built deltachat-android successfully. Locations will be sent now again, great.

But locations are still not be displayed on the map. (This behavior was/is on the desktop version too which is been ported to Rust before. It is built with Arch Linux makepkg using the deltachat-desktop/master branch.)

I've noticed delta-chat is blocking the main thread (Android UI) now during a bad network connection, it not happens if network is off. This I had not seen before the new version.

Sorry, I'm not very familiar with Rust development at the moment but I'm thrilled a bit...

hpk42:
On Sun, Nov 03, 2019 at 15:45 -0800, Frank Ganske wrote:
> @hpk42 @link2xt thanks for investigating! I've pulled the core submodule to the master branch including deltachat/deltachat-core-rust#782 and built deltachat-android successfully. Locations will be sent now again, great.
> 
> But locations are still not be displayed on the map. (This behavior was/is on the desktop version too which is been ported to Rust before. It is built with Arch Linux makepkg using the deltachat-desktop/master branch.)

are locations not displayed on Android compiled with the recent core? 
if so, could you please file a separate issue? 
I'd put it to the "release-critical" list. 
thanks for following and giving feedback!

simonsan:
> I also remember some discussions about the social implications writing an app that allows permanently location streaming. How do you prevent abuse of this feature? What about people stalking their partners by putting pressure on them to turn this feature permanently on?

You could probably use a fixed maximum time which is enough for most travels (around 1 hour) and authenticate against the phone password for using longer times. For longer times it would be viable to show a warning inside deltachat as a small red infobar and under notifications as well (Something basic like "Long-term notification streaming is enabled"). Otherwhise there would be also something like coffeine from LineageOS, an option for the screen display to stay awake for a longer period of time. It's basically a countdown and as soon as you tap on it it adds minutes to the countdown in a certain scale. minimum `5 Minutes - 10 - 15 - 30 - 60 - unlimited`. With this you could as every hour of location streaming for a small feedback (tap on point in dialog box or so) and if that is not fulfilled you deactivate it. 

I find the awareness really good here, that something like this can be abused by toxic persons. Thank you for even thinking about that.

r10s:
added this pr to the new "Project Resurrection" project where we track prs that are not handled just now but where we can access them easily and revive them as needed.

i am closing this pr for now; not because we do not want to go to this direction but because we want to have some overview about what is currently happening i nthe pr/issue overview.

i think we will retarget location streaming sometime next year, currently all effort goes to getting stable releases out.

PR #1042, opened 2019-09-29
Changed Notification level for Android8+

darkestDark:
Enable heads-up notification and fix notification delay on android 9 (if device is muted)

### DIFF
--- a/src/org/thoughtcrime/securesms/notifications/AbstractNotificationBuilder.java
+++ b/src/org/thoughtcrime/securesms/notifications/AbstractNotificationBuilder.java
@@ -158,7 +158,7 @@ abstract class AbstractNotificationBuilder extends NotificationCompat.Builder {
         // we cannot change the settings, however, this is handled by using different values for chId
         if(!channelExists) {
           NotificationChannel channel = new NotificationChannel(chId,
-                  "New messages", NotificationManager.IMPORTANCE_DEFAULT);
+                  "New messages", NotificationManager.IMPORTANCE_HIGH);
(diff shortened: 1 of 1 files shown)


r10s:
> Enable heads-up notification and fix notification delay

what exactly is achieved by this pr? what is the "notification delay"? do you get notifications that you did not get before? or does it just tweak some display? i think, in general, the priority "default" is a good compromise for most situations - and, depending on the os, the exact "default" may be also changed globally in the system settings

wrt notifications, our primary goal is currently to fix [these issues](https://github.com/deltachat/deltachat-android/issues?q=is%3Aissue+is%3Aopen+label%3Anotify).

so, if this pr does not solve one of these issues, i would tend to close it for now to avoid adding even more noise to the whole notification area :)

some time later, we can come back to that. in general, i think it would be good to have a static notification channel for delta, maybe chat-wise.

darkestDark:
it has only one goal... on api level 28+ the notifications apears even if device is muted... else it will apear after running app...

firebase cloud messaging does the same for messages like this so i think its the right level...

for the other problem... for publishing on the playstore the way with the foreground service wont work... goggle will deny it because you could (but i wouldnt like) use fcm...

in the fdroid build it is possible with some lines of code for foreground service and permission for foreground service in manifest...

r10s:
> it has only one goal... on api level 28+ the notifications apears even if device is muted... 

okay, this makes some sense, i assume, the notification do not make some noise, however :)

we give it a try, just had a look, signal does the same (not saying signal does everything right, though :)

PR #1055, opened 2019-10-06
Some language fixes

OzancanKaratas:
- Compliance for spelling rules.
- Compliance with each other the strings.

May I add the iOS strings before merging?

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -24,6 +24,7 @@
     <string name="main_menu">Main menu</string>
     <string name="show_password">Show password</string>
     <string name="hide_password">Hide password</string>
+    <string name="continue">Continue</string>
     <string name="not_now">Not now</string>
     <string name="never">Never</string>
     <string name="one_moment">One moment...</string>
@@ -550,8 +551,6 @@
 
     <!-- permissions -->
     <string name="perm_required_title">Permission required</string>
-    <string name="perm_continue">Continue</string>
-    <string name="perm_not_now">Not now</string>
     <string name="perm_explain_need_for_camera_access">To capture photos and video, allow Delta Chat access to the camera.</string>
     <string name="perm_explain_need_for_mic_access">To send audio messages, allow Delta Chat access to your microphone.</string>
     <string name="perm_explain_need_for_storage_access">Delta Chat needs access to your files in order to export files or to create backups.</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -26,7 +26,7 @@
     <string name="hide_password">Hide password</string>
     <string name="not_now">Not now</string>
     <string name="never">Never</string>
-    <string name="one_moment">One moment…</string>
+    <string name="one_moment">One moment...</string>
     <string name="done">Done</string>
     <string name="undo">Undo</string>
     <string name="error">Error</string>
@@ -341,7 +341,7 @@
 
     <!-- share and forward messages -->
     <!-- Translators: Title shown above a chat/contact list; the user selects the recipient of the messages he wants to forward to -->
-    <string name="forward_to">Forward to …</string>
+    <string name="forward_to">Forward to...</string>
     <string name="share_multiple_attachments">Do you want to send %1$d files to the selected chat?\n\nThe files are sent unmodified in their original size, eg. images and videos are not recoded.</string>
     <string name="share_abort">Sharing aborted due to missing permissions.</string>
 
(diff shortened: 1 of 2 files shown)


res/values/strings.xml
```diff
@@ -26,7 +26,7 @@
     <string name="hide_password">Hide password</string>
     <string name="not_now">Not now</string>
     <string name="never">Never</string>
-    <string name="one_moment">One moment…</string>
+    <string name="one_moment">One moment...</string>
```
r10s: "…" is the unicode character 'HORIZONTAL ELLIPSIS'; following most recommendations and style guides, this should be favored over the three dots afaik.

res/values/strings.xml
```diff
@@ -441,17 +441,17 @@
     <string name="pref_watch_inbox_folder">Watch Inbox folder</string>
     <string name="pref_watch_sent_folder">Watch Sent folder</string>
     <string name="pref_watch_mvbox_folder">Watch DeltaChat folder</string>
-    <string name="pref_auto_folder_moves">Automatic Moves to DeltaChat folder</string>
+    <string name="pref_auto_folder_moves">Automatic moves to DeltaChat folder</string>
```
r10s: thanks for pointing that out, i corrected this in another commit.

OzancanKaratas:
@Jikstra I want to discus with you. I want to help and language problems is very important for me. This pull request corrects the problems I noticed.

res/values/strings.xml
```diff
@@ -441,17 +441,17 @@
...
     <string name="pref_auto_folder_moves_explain">Chat conversations are moved to avoid cluttering the Inbox folder</string>
     <string name="pref_email_interaction_title">Email interaction</string>
     <string name="pref_show_emails">Show classic emails</string>
     <string name="pref_show_emails_no">No, chats only</string>
     <string name="pref_show_emails_accepted_contacts">For accepted contacts</string>
     <string name="pref_show_emails_all">All</string>
     <string name="pref_empty_server_title">DANGER ZONE: Delete e-mails from server</string>
-    <string name="pref_empty_server_msg">This function helps to free space on your IMAP server by deleting ALL e-mails, including chat-messages, in the given folders. Messages on this device will not be deleted.  The deletion cannot be undone!</string>
-    <string name="pref_empty_server_inbox">Delete ALL e-mails in Inbox folder</string>
-    <string name="pref_empty_server_mvbox">Delete ALL e-mails in DeltaChat folder</string>
```
r10s: i think the intention of the upper-case ALL is some emphasis. but i agree that this is a bit odd as it is not used somewhere else. changing this in another pr.

res/values/strings.xml
```diff
@@ -556,10 +556,10 @@
...
+    <string name="perm_explain_access_to_camera_denied">Delta Chat requires the Camera permission in order to take photos or videos, but it has been permanently denied. Please go to the app settings menu, select \"Permissions\", and enable \"Camera\".</string>
+    <string name="perm_explain_access_to_mic_denied">Delta Chat requires the Microphone permission in order to send audio messages, but it has been permanently denied. Please go to app settings menu, select \"Permissions\", and enable \"Microphone\".</string>
+    <string name="perm_explain_access_to_storage_denied">Delta Chat requires the Storage permission in order to attach or export photos, videos, or audio, but it has been permanently denied. Please go to the app settings menu, select \"Permissions\", and enable \"Storage\".</string>
+    <string name="perm_explain_access_to_location_denied">Delta Chat requires Location permission in order to attach a location, but it has been permanently denied. Please go to the app settings menu, select \"Permissions\", and enable \"Location\".</string>
     <string name="perm_enable_bg_reminder_title">Tap here to receive messages while Delta Chat is in background.</string>
```
r10s: not sure if "go to" is better than "continue to" here, my english is not good enough to decide that. btw, the wording is not from delta chat but from the signal source - who should at least native speakers, if that means anything :)

OzancanKaratas:
@r10s I think you should handle each commits individually.

r10s:
ups, i've just noticed there are more commit, i will go to them as well.

wrt keys: we cannot change them easily as they're also used on ios, windows, mac etc. and changing them produces a lot of follow-up work that is usually not worth the effort.

res/values/strings.xml
```diff
@@ -305,7 +306,7 @@
     <string name="welcome_intro6_message">Encrypted with all common algorithms. Messages stay on your servers.</string>
     <string name="welcome_intro7_headline">Trustworthy</string>
     <string name="welcome_intro7_message">Delta Chat is safe for business use, compatible and standards-based.</string>
-    <string name="welcome_start_messaging">Start Messaging</string>
+    <string name="welcome_start_messaging">Start messaging</string>
```
r10s: i correct that in another pr.

src/org/thoughtcrime/securesms/permissions/Permissions.java
```diff
@@ -156,8 +156,8 @@ private void executePreGrantedPermissionsRequest(PermissionsRequest request) {
     @SuppressWarnings("ConstantConditions")
     private void executePermissionsRequestWithRationale(PermissionsRequest request) {
       RationaleDialog.createFor(permissionObject.getContext(), rationaleDialogMessage, rationalDialogHeader)
-                     .setPositiveButton(R.string.perm_continue, (dialog, which) -> executePermissionsRequest(request))
-                     .setNegativeButton(R.string.perm_not_now, (dialog, which) -> executeNoPermissionsRequest(request))
+                     .setPositiveButton(R.string.continue, (dialog, which) -> executePermissionsRequest(request))
+                     .setNegativeButton(R.string.not_now, (dialog, which) -> executeNoPermissionsRequest(request))
```
r10s: i will remove that double string, thanks for the hint!

r10s:
> May I add the iOS strings before merging?

ios just uses the same strings as android, nothing more to do :)

PR #1058, opened 2019-10-08
use strdup() more carefully

r10s:


r10s:
stringToData() is no longer needed as it was only needed for DC_EVENT_GET_STRING that was replaced by dc_set_stock_translation() just now

PR #1072, opened 2019-10-20
fix utf8 issues for android4

r10s:
targets https://github.com/deltachat/deltachat-android/issues/1065 - while the crash is already fixed with https://github.com/deltachat/deltachat-core-rust/pull/728, this pr makes sending emojis & co. _possible_ :)


cyBerta:
great! works perfectly

PR #1077, opened 2019-10-22
apply a hack that removes the padding from the settings screens

r10s:
found at https://stackoverflow.com/questions/51518758/preferencefragmentcompat-has-padding-on-preferencecategory-that-i-cant-get-rid/52960668#52960668


cyBerta:
we should migrate to [AndroidX](https://developer.android.com/jetpack/androidx/migrate) sooner than later and keep in mind to remove part of this workaround.

r10s:
don't know much about androidX, but, yes.
i also removed some more or less useless special preference-stuff, so, all in all, we're closer to the standard now.

in general, the alignment of the options with the title is nice. might also be that we have some more icons at some point, this would also discard this issue.

however, for now i am happy to have the same look as the old version.

PR #1078, opened 2019-10-22
[wip] revert utf 8 fixes

r10s:
the changes from #1072 seems to lead to crashes on android 64bit, not sure if it is related to strndup, at least this function is suspicious :)

cc @link2xt


link2xt:
Works for me

r10s:
not needed, @link2xt found the real problem at #1079

PR #1080, opened 2019-10-23
jni/Android.mk: enable compiler warnings

link2xt:


link2xt:
This is to future-proof against bugs such as #1079

PR #1091, opened 2019-10-29
Add strings for certificate checks configuration

link2xt:


r10s:
i also pushed the translations to transifex so that the new strings can be translated and used on ios/desktop

Issue #1099, opened 2019-11-01
Freezing and Force Close on Android 10


darkestDark:
- Android version: 10
- Device: lavender
- Delta Chat version: 0940.2
- Actual behavior: App freezing sometimes
- Logs:
```
chat.delta not qualified for android.app.role.HOME due to missing RequiredComponent{mIntentFilterData=IntentFilterData{mAction='android.intent.action.MAIN', mCategories='[android.intent.category.HOME]', mDataScheme='null', mDataType='null'}, mPermission='null', mMetaData=[]}
chat.delta not qualified for android.app.role.SMS due to missing RequiredComponent{mIntentFilterData=IntentFilterData{mAction='android.provider.Telephony.SMS_DELIVER', mCategories='[]', mDataScheme='null', mDataType='null'}, mPermission='android.permission.BROADCAST_SMS', mMetaData=[]}
Receiver during timeout of BroadcastRecord{fef2cd4 u0 null} : ResolveInfo{9514c63 chat.delta/org.thoughtcrime.securesms.connect.TimerReceiver m=0x0}
channel '5d0819e chat.delta/org.thoughtcrime.securesms.ConversationActivity (server)' ~ Consumer closed input channel or an error occurred.  events=0x9
channel '5d0819e chat.delta/org.thoughtcrime.securesms.ConversationActivity (server)' ~ Channel is unrecoverably broken and will be disposed!
Attempted to unregister already unregistered input channel '5d0819e chat.delta/org.thoughtcrime.securesms.ConversationActivity (server)'
```

r10s:
> chat.delta not qualified for android.app.role.HOME

what is this HOME?

cyBerta:
related to:

> CATEGORY_HOME
> Added in API level 1
> public static final String CATEGORY_HOME
> This is the home activity, that is the first activity that is displayed when the device boots.
> Constant Value: "android.intent.category.HOME" 

https://developer.android.com/reference/android/content/Intent.html#CATEGORY_HOME

cyBerta:
`Receiver during timeout of BroadcastRecord{fef2cd4 u0 null} : ResolveInfo{9514c63 chat.delta/org.thoughtcrime.securesms.connect.TimerReceiver m=0x0}`

it shows that too much is processed on the main thread during onReceive() method call.
The timeout is 10s though!
https://developer.android.com/reference/android/content/BroadcastReceiver.html#onReceive(android.content.Context,%20android.content.Intent)

[...]

PR #1102, opened 2019-11-01
[WIP] implement TimerIntentService 

cyBerta:
to work around possible broadcast receiver timeouts

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -275,6 +275,8 @@
             <action android:name="android.service.chooser.ChooserTargetService" />
         </intent-filter>
     </service>
+    <service android:name=".connect.TimerIntentService"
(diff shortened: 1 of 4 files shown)


cyBerta:
This doesn't fix https://github.com/deltachat/deltachat-android/issues/1099, but prevent's the described force close. 
The actual bug is in the core's triggered by interruptImapIdle()

r10s:
@cyBerta do you think this would fix https://github.com/deltachat/deltachat-android/issues/1104 ?

webratte:
If you would offer a apk witch contains this PR I would make a factory reset of my test phone.

This way I could simulate a fresh install on a new, not optimized, phone like non tech users do.

r10s:
i think we should got for #1108, WorkManager instead.

PR #1108, opened 2019-11-07
streamline background notifications

### LINKED ISSUE #1099: Freezing and Force Close on Android 10

### LINKED ISSUE #1104: No notifications after a few minutes standby of the phone
- Android version: Android 9
- Device: Samsung Galaxy S8
- Delta Chat version: 0.940.2
- Expected behavior: Notification for incoming messages
- Actual behavior: No notification if the phone is in standby for a few minutes. I have played around with all battery saving options and turned all off.

1. set the phone in standby
2. wait 15-30 minutes
3. receive a message
4. you got no notification
- Screenshots:
- Logs:
I will create a log and post it here later

### LINKED ISSUE #1105: check out options for a reliable background fetch
the current approach to have a socket instantly open, becomes harder and harder, the user may tweak their system, however, it is a mess, see eg. the latest issues here or https://dontkillmyapp.com

so, i think, we should have a new look at Timer, AlertManager, JobScheduler and so  on. not sure what is the hottest-api in 2019

for reference: issue from 2018 discussing the same thing: https://github.com/deltachat/deltachat-android/issues/269

r10s:
this pr is a try to get reliable background notifications.

- if the user has allowed delta to "consume battery", the background threads are running always. depending on the os, there may be several switches the user has to tweak, see https://dontkillmyapp.com . a  WorkManager [1] timer is used in this case to keep the socket alive by calling dc_interrupt*() very few minutes (for whatever $reasons, but this seems was an important tweak at some point)

- if we're not allowed to consume battery, we use a WorkManager [1] timer to check every some minutes realiable, moreover, we use a hidden foreground-service-notification, so, at best, we get instant notifications even in this case.

it's still a bit hackish and as some things in the core does not yet work reliable, we do not really rely on interrup() or close(). however, these type of things are typically not very long-living anyway ...

unfortunately, some devices implement killing the app from the recents menu  as a force stop (cmp eg. [stackoverflow](https://stackoverflow.com/questions/50682061/android-is-workmanager-running-when-app-is-closed)), (stock android does not do this), so even with that, it may happen that the app is killed completely. maybe we should just show the foreground-notification, however, this is annoying for users in another way. also this issue is true for whatsapp etc. as well.

closes #1104, closes #1105, closes #1099

[1] [WorkManager guide](https://developer.android.com/topic/libraries/architecture/workmanager/), [WorkManager reference](https://developer.android.com/reference/androidx/work/WorkManager)


src/org/thoughtcrime/securesms/connect/ApplicationDcContext.java
```diff
@@ -293,28 +288,22 @@ public ThreadRecord getThreadRecord(DcLot summary, DcChat chat) { // adapted fro
    **********************************************************************************************/
 
   private final Object threadsCritical = new Object();
+  private final Object incLoopsCritical= new Object();
 
-  private boolean imapThreadStartedVal;
-  private final Object imapThreadStartedCond = new Object();
   public Thread imapThread = null;
   private PowerManager.WakeLock imapWakeLock = null;
+  private int inboxLoops = 0;
```
cyBerta: you could also use some AtomicInt's here to get rid of some synchronized blocks (except the one at line 402)

r10s: didn't know that even exists :)
but i do no think this will block things?

cyBerta:
nice! I didn't know WorkManager before.

src/org/thoughtcrime/securesms/connect/ApplicationDcContext.java
cyBerta: yes, AtomicIntegers have blocking read/write access: https://developer.android.com/reference/java/util/concurrent/atomic/AtomicInteger and can be used for programming concurrenency.

PR #1110, opened 2019-11-10
Android localizable string editions

OzancanKaratas:
Hello @r10s,

Please do not close this. Comment the file lines you want to change.

I used Google service names, remove spaces between texts and 3 dots, such as in Android.

Also the uppercase strings was converted lowercase like others.

Finally the desktop string “Choose Language...” should not be 3 dots due to sequence of the other strings. But I didn't want to do anything without asking you.


res/values/strings.xml
```diff
@@ -169,7 +169,7 @@
     <string name="menu_zoom_out">Zoom out</string>
     <string name="menu_save_log">Save log</string>
     <string name="title_share_location">Share location with all group members</string>
-    <string name="device_talk">Device Messages</string>
+    <string name="device_talk">Device messages</string>
```
r10s: i think names can be left in uppercase completely, so "Device Messages" should be fine here.

res/values/strings.xml
```diff
@@ -319,17 +319,17 @@
...
+    <string name="login_smtp_password">SMTP password</string>
     <string name="login_smtp_server">SMTP server</string>
     <string name="login_smtp_port">SMTP port</string>
-    <string name="login_smtp_security">SMTP Security</string>
+    <string name="login_smtp_security">SMTP security</string>
     <string name="login_auth_method">Authorization method</string>
     <string name="login_info_oauth2_title">Continue with simplified setup?</string>
     <string name="login_info_oauth2_text">The entered e-mail-address supports a simplified setup (OAuth2).\n\nIn the next step, please allow Delta Chat to act as your Chat with E-Mail app.\n\nThere are no Delta Chat servers, your data stays on your device!</string>
-    <string name="login_info_gmail_text">For GMail Accounts you need to create an App-Password if you have 2FA enabled. If this setting is not available, you need to enable Less Secure Apps.</string>
+    <string name="login_info_gmail_text">For Gmail Accounts you need to create an App-Password if you have 2-Step Verification enabled. If this setting is not available, you need to enable less secure apps.</string>
```
r10s: - 2FA is "Two-Factor Authentication", however, maybe even the abbreviation is known enough.
- maybe write "Less secure apps" in quotes to make more clear that it is the name of an option.

res/values/strings.xml
```diff
@@ -459,7 +459,7 @@
     <string name="pref_show_emails_accepted_contacts">For accepted contacts</string>
     <string name="pref_show_emails_all">All</string>
     <string name="pref_empty_server_title">Delete e-mails from server</string>
-    <string name="pref_empty_server_msg">This function helps to free space on your IMAP server by deleting ALL e-mails, including chat-messages, in the given folders. Messages on this device will not be deleted.  The deletion cannot be undone!</string>
+    <string name="pref_empty_server_msg">This function helps to free space on your IMAP server by deleting ALL e-mails, including chat-messages, in the given folders. Messages on this device will not be deleted. The deletion cannot be undone!</string>
```
r10s: maybe also replace "ALL" by "all", i think we removed these all-caps already at other places.

res/values/strings.xml
OzancanKaratas: All strings lowercase at the moment. Regularity and consistency are more important.

res/values/strings.xml
OzancanKaratas: I wrote as it is in Google's documents and Android interface.

OzancanKaratas: I wrote it like Google's documents and Android interface.

r10s: you're right, google does not speak of 2FA at all. but i would put it in quotes, both "2-Step Verification" and "Less secure apps".

and if maybe bring both terms to the same form, so, first character uppercase, rest lowercase, independingly how google writes that (there is often marketing involved, no need to pick that up :)

res/values/strings.xml
r10s: fair enough, i do not have a strong opinion in that.

res/values/strings.xml
r10s: but maybe it is just fine :)

PR #1113, opened 2019-11-11
fix map

### LINKED ISSUE #1112: map broken
the map, that is available as an experimental feature, is currently broken, see https://github.com/deltachat/deltachat-android/pull/1109#issuecomment-552207204

cyBerta:
closes #1112 

deltachat/deltachat-core-rust#798 needs to be fixed still.

### DIFF
--- a/res/layout/bottom_map_sheet.xml
+++ b/res/layout/bottom_map_sheet.xml
@@ -7,7 +7,7 @@
     android:id="@+id/bottom_sheet"
     app:behavior_hideable="false"
     app:behavior_peekHeight="35dp"
-    app:layout_behavior="android.support.design.widget.BottomSheetBehavior"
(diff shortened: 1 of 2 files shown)


r10s:
great, tested that on my androd7 motog this fixes the crash :)

there may still be issues in location streaming as such with the new rust-core, but this is out of scope of this pr :)

@cyBerta is the "wip" thingie still correct?

cyBerta:
I've found a bug while debugging https://github.com/deltachat/deltachat-core-rust/issues/798.

r10s:
> I've found a bug while debugging deltachat/deltachat-core-rust#798.

only found? or fixed by the commit above?

cyBerta:
found and fixed :)

PR #1115, opened 2019-11-13
adapt to ndk r20b

### LINKED ISSUE #1116: failed to run custom build command for `ring v0.16.9`
this issue is about the feat/async-imap pr https://github.com/deltachat/deltachat-core-rust/pull/814

trying to build it on android results in the error message `failed to run custom build command for ring v0.16.9`, full log at https://gist.github.com/r10s/2ef6e2dbe5674bfe5507d20f2eabda90

i tried cargo and rustup updates and re-installed the toolchains following the guides at https://github.com/deltachat/deltachat-android/blob/master/ndk-make.sh#L10

maybe i have missed something that is required to build on android. 

nb: building android on current master and feat/async-imap is also working on mac and on ios.

r10s:
because of issues as https://github.com/deltachat/deltachat-android/issues/1116 , but also to not use too outdated libraries, we have to change the used ndk from the used r14b or so to the most recent r20b.

this requires some adaptions:

- r20b does support a minimal sdk16 only, so we can no longer got for minSdkVersion=14
- 64bit supports minSdkVersion=21, not sure if that can be mixed
- there is no longer a script to install a toolchain, instead r20b comes with some predefined toolchains
- r20b seems to look in AndroidManifest.xml for minSdkVersion, not sure, however.
- clang is used as linker, c++_static as standard library

so, currently, we're using minSdkVersion=21 everywhere - which makes things compiling and working. this would drop support for android4 (ice cream sandwich, jelly bean, kitkat) completely, according to https://developer.android.com/about/dashboards/ , in sum 10.4% of android-devices :/

but maybe we can bring back support for kitkat (6.9%) and jelly bean (3.2%) - the ndk supports compiling down to sdk16 =  jelly bean. however, due to the needed 64 bit support, this may become tricky. not sure if and how this is doable.

EDIT: support for kitkat 4.4 and partly jelly bean 4.3 is back :). 
also, tested: we could support jelly bean fully by lowering sdk from 18 to 16, however, [the rust "ring" crate does not support that currenty](https://github.com/briansmith/ring/blob/2a3230cb6346a1249354a07395246a28b4904c5b/BUILDING.md#supported-toolchains).

**in summary, this pr changes the supported user base from 99.7% to 96.7%.** the advantage is having a supported and maintained ndk base and better support for libs (eg. ring did not work out of the box on ndk14). if we manage to target sdk16 again, we can again reach 99.4%. getting higher is only doable by using outdated libs.

closes #1116

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -3,8 +3,6 @@
     xmlns:tools="http://schemas.android.com/tools"
     package="org.thoughtcrime.securesms">
 
-    <uses-sdk tools:overrideLibrary="com.amulyakhare.textdrawable,com.astuetz.pagerslidingtabstrip,pl.tajchert.waitingdots,com.h6ah4i.android.multiselectlistpreferencecompat,android.support.v13,com.davemorrissey.labs.subscaleview,com.tomergoldst.tooltips,com.klinker.android.send_message,com.takisoft.colorpicker,android.support.v14.preference"/>
-
     <uses-feature android:name="android.hardware.camera" android:required="false" />
     <uses-feature android:name="android.hardware.bluetooth" android:required="false" />
     <uses-feature android:name="android.hardware.location" android:required="false"/>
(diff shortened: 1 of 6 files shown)


cyBerta:
nice!

PR #1120, opened 2019-11-17
simplify routing to welcome-activity

r10s:
i came over that while finding the correct places to add welcome-device-messages.


cyBerta:
nice :)

PR #1121, opened 2019-11-18
add device msgs

r10s:
todo:
- [x] add initial messages to the correct place
- [x] for some reason, the first added message is shown in the chatlist summary, it should be the second imho
- [x] the battery-dialog should maybe appear if the first message is sent or so
- [x] there should be no system-notification just when the app is started
- [x] maybe add saved-mesages-chat on new installation

once things are settled on android, it should be pretty easy to port them to ios and desktop

later on, we can add also move the the old-version-thingie as well as the battery-question to the device-msg.

needs https://github.com/deltachat/deltachat-core-rust/pull/854 and https://github.com/deltachat/deltachat-core-rust/pull/856 and https://github.com/deltachat/deltachat-core-rust/pull/939 to be merged first.

### DIFF
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -674,6 +674,12 @@ JNIEXPORT jboolean Java_com_b44t_messenger_DcContext_wasDeviceMsgEverAdded(JNIEn
 }
 
 
+JNIEXPORT void Java_com_b44t_messenger_DcContext_updateDeviceChats(JNIEnv *env, jobject obj)
+{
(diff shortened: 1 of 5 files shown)


r10s:
turned this into wip, we decided yesterday that the welcome messasges etc. are added by the core, the ui can add messages on their own, however, just now this seems not to be needed.

so, we'll use sth. as dc_update_device_messages() that will do most of the things, see https://github.com/deltachat/deltachat-core-rust/issues/937

PR #1122, opened 2019-11-19
Language reworked

comradekingu:

PR #1125, opened 2019-11-22
Language updates

OzancanKaratas:
This merges #1122 and changes the strings according to the recommendations there.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -427,7 +427,7 @@
     <string name="pref_read_receipts_explain">If read receipts are turned off, you won\'t know if your messages were displayed.</string>
     <string name="pref_manage_keys">Manage keys</string>
     <string name="pref_use_system_emoji">Use system emoji</string>
-    <string name="pref_use_system_emoji_explain">Disable Delta Chat\'s built-in emoji support</string>
+    <string name="pref_use_system_emoji_explain">Turn off Delta Chat\'s built-in emoji support</string>
     <string name="pref_app_access">App access</string>
     <string name="pref_communication">Communication</string>
     <string name="pref_chats">Chats</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <!-- common strings without special context -->
-    <string name="app_name">Delta Chat</string>
+    <string name="dc_app_name" translatable="false">Delta Chat</string>
+    <string name="app_name" translatable="false">Delta Chat</string>
     <string name="ok">OK</string>
     <string name="cancel">Cancel</string>
     <string name="yes">Yes</string>
@@ -44,7 +45,7 @@
     <string name="copied_to_clipboard">Copied to clipboard.</string>
     <string name="contacts_headline">Contacts</string>
     <string name="email_address">Email address</string>
-    <string name="bad_email_address">Bad email address.</string>
+    <string name="bad_email_address">Incorrect email address.</string>
     <string name="password">Password</string>
     <string name="now">Now</string>
     <!-- Translators: used as a headline in sections with actions that cannot be undone. could also be "Caution" or "Cave" or so. -->
@@ -383,7 +384,7 @@
     <string name="pref_trim_now_ask">Delete all old messages now? This will immediately trim all chats to the %d most recent messages.</string>
     <string name="pref_blocked_contacts">Blocked contacts</string>
     <string name="pref_profile_photo_remove_ask">Remove profile photo?</string>
-    <string name="pref_help_url">https://delta.chat/en/help</string>
+    <string name="pref_help_url" translatable="false">https://delta.chat/help</string>
     <string name="pref_password_and_account_settings">Password and account</string>
     <string name="pref_who_can_see_this_information">Who can see this?</string>
     <string name="pref_who_can_see_profile_explain">Your profile picture and name will be shown alongside your messages when communicating with other users. Already sent information can not be deleted or removed.</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -636,7 +636,7 @@
     <string name="download_attachment_desktop">Download attachment</string>
     <string name="export_backup_desktop">Export backup</string>
     <string name="transfer_key_desktop">Transfer key</string>
-    <string name="show_key_transfer_message_desktop">Your key has been sent to yourself. Switch to the other device and open the Autocrypt Setup Message. You should be prompted for a setup code. Once the following digits are typed into your other device, it will be ready to use Autocrypt.</string>
+    <string name="show_key_transfer_message_desktop">Your key has been sent to yourself. Switch to the other device and open the Autocrypt Setup Message. You should be prompted for a setup code. Once the following digits are typed into on other device, it will be ready to use Autocrypt.</string>
     <string name="new_message_from_desktop">New message from</string>
     <string name="unblock_contacts_desktop">Unblock contacts</string>
     <string name="none_blocked_desktop">No blocked contacts yet</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -257,8 +257,8 @@
     <string name="chat_self_talk_subtitle">Messages I sent to myself</string>
     <string name="chat_contact_request">Contact request</string>
     <string name="chat_no_contact_requests">No contact requests.\n\nIf you want classic emails to appear here as contact requests, you can change the corresponding setting in the app settings.</string>
-    <string name="retry_send">Retry to send message</string>
-    <string name="send_failed">Sending message failed</string>
+    <string name="retry_send">Try to send message again?</string>
+    <string name="send_failed">Could not send message</string>
 
     <!-- map -->
     <string name="filter_map_on_time">Show locations in time frame</string>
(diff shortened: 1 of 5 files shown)


res/values/strings.xml
```diff
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <!-- common strings without special context -->
-    <string name="app_name">Delta Chat</string>
+    <string name="dc_app_name" translatable="false">Delta Chat</string>
```
cyBerta: why was this string  added?

res/values/strings.xml
```diff
@@ -207,13 +208,13 @@
     <string name="ask_export_attachment">Export attachment? Exporting attachments will allow any other apps on your device to access them.\n\nContinue?</string>
     <string name="ask_block_contact">Block this contact? You will no longer receive messages from this contact.</string>
     <string name="ask_unblock_contact">Unblock this contact? You will once again be able to receive messages from this contact.</string>
-    <string name="ask_delete_contacts">Delete contacts? This will permanently delete the selected contacts.\n\nContacts with ongoing chats and contacts from the system\'s address book cannot be deleted permanently.</string>
-    <string name="ask_delete_contact">Permanently delete contact %1$s?\n\nContacts with ongoing chats and contacts from the system\'s address book cannot be deleted permanently.</string>
+    <string name="ask_delete_contacts">Delete selected contacts?\n\nContacts with ongoing chats and contacts from the system\'s address book cannot be deleted permanently.</string>
+    <string name="ask_delete_contact">Delete %1$s?\n\nContacts with ongoing chats and contacts from the system\'s address book cannot be deleted permanently.</string>
```
cyBerta: I would be happy to not be deleted personally ;) ("Delete cyBerta?")
I think `Delete contact %1$1?` is better.

res/values/strings.xml
OzancanKaratas: I deleted untranslated.xml. There was only this. So I moved those over here.

res/values/strings.xml
```diff
@@ -629,32 +628,32 @@
...
+    <string name="initiate_key_transfer_desktop">An Autocrypt Setup Message securely shares your end-to-end setup with other Autocrypt-compliant apps. The setup will be encrypted by a setup code displayed here and must be typed on the other device.</string>
     <string name="reply_to_message_desktop">Reply to message</string>
     <string name="select_group_image_desktop">Select group image</string>
     <string name="imex_progress_title_desktop">Backup progress</string>
     <string name="download_attachment_desktop">Download attachment</string>
     <string name="export_backup_desktop">Export backup</string>
     <string name="transfer_key_desktop">Transfer key</string>
-    <string name="show_key_transfer_message_desktop">Your key has been sent to yourself. Switch to the other device and open the setup message. You should be prompted for a setup code. Type the following digits into the prompt. Once you\'re done, your other device will be ready to use Autocrypt.</string>
+    <string name="show_key_transfer_message_desktop">End-to-end setup transferred. This device is now ready to use Autocrypt with the same setup as the other device.</string>
```
cyBerta: This translation does not match the intended meaning. It's a desktop client's message stating that the Autocrypt message has been send from the desktiop client and further autocrypt setup steps needs to be done on the new device. I think this change should be reverted completely.

res/values/strings.xml
cyBerta: Ok, I've found it in the untranslated.xml. Instead of adding this entry here, iOS needs to use the right string "app_name"

cyBerta:
Overall it looks good to me! Thanks for your contributions @OzancanKaratas  and @comradekingu!
Please revert the changes that alter the initially intended meaning (that I mentioned).

res/values/strings.xml
OzancanKaratas: Is this good?

res/values/strings.xml
OzancanKaratas: It does not hurt the iOS string. I did it to get out of the crowd. Would you like undone?

cyBerta: Yes, you can delete the entry for dc_app_name but keep app_name. dc_app_name is not used in iOS (anymore)

OzancanKaratas: Does Android use it? I did it for Android.

res/values/strings.xml
```diff
@@ -635,7 +635,7 @@
     <string name="download_attachment_desktop">Download attachment</string>
     <string name="export_backup_desktop">Export backup</string>
     <string name="transfer_key_desktop">Transfer key</string>
-    <string name="show_key_transfer_message_desktop">End-to-end setup transferred. This device is now ready to use Autocrypt with the same setup as the other device.</string>
+    <string name="show_key_transfer_message_desktop">Your key has been sent to yourself. Switch to the other device and open the setup message. You should be asked for a setup code. Type the following digits:</string>
```
cyBerta: Proposal from the delta chat gathering:
`You should be asked for a setup code, enter the following digits`

res/values/strings.xml
cyBerta: see my comment here: https://github.com/deltachat/deltachat-android/pull/1125#discussion_r351018095

res/values/strings.xml
cyBerta: yes it is used there, but in a separate untranslated.xml file. In addition to that, I'll remove that and use only app_name instead on Android as well.

OzancanKaratas:
I got your commits. If you are going to make any more changes, I will also pick them up and collect them in this pull request.

res/values/strings.xml
```diff
@@ -308,34 +308,32 @@
...
     <string name="welcome_start_messaging">Start messaging</string>
-    <string name="login_title">Login</string>
-    <string name="login_header">Login to your server</string>
-    <string name="login_explain">Login with an existing e-mail account.</string>
-    <string name="login_subheader">For known email providers additional settings are setup automatically. Sometimes IMAP needs to be enabled in the web frontend. Consult your email provider or friends for help.</string>
-    <string name="login_no_servers_hint">There are no Delta Chat servers, your data stays on your device!</string>
+    <string name="login_title">Log in</string>
+    <string name="login_header">Log in to your server</string>
+    <string name="login_explain">Log in with an existing email account</string>
+    <string name="login_subheader">For known email providers additional settings are set up automatically. Sometimes IMAP needs to be enable in the web settings. Consult your email provider or friends for help.</string>
```
cyBerta: a 'd' is missing in enabled

res/values/strings.xml
```diff
@@ -484,18 +482,18 @@
...
+    <string name="autocrypt_send_asm_explain_after">Your setup has been sent to yourself. Switch to the other device and open the setup message. You should be asked for a setup code. Enter the following digits:\n\n%1$s</string>
     <string name="autocrypt_prefer_e2ee">Prefer end-to-end encryption</string>
     <string name="autocrypt_asm_subject">Autocrypt Setup Message</string>
-    <string name="autocrypt_asm_general_body">This is the Autocrypt Setup Message used to transfer your end-to-end setup between clients.\n\nTo decrypt and use your setup, open the message in an Autocrypt-compliant client and enter the setup code presented on the generating device.</string>
-    <string name="autocrypt_asm_click_body">This is the Autocrypt Setup Message used to transfer your end-to-end setup between clients.\n\nTo decrypt and use your setup, tap or click on this message.</string>
+    <string name="autocrypt_asm_general_body">This is the Autocrypt Setup Message used to transfer your end-to-end setup between clients.\n\nOpen the message in an Autocrypt-compliant app and enter the setup code presented on the device it was sent from to use it.</string>
```
cyBerta: after consulting @r10s we came to the conclusion to revert the changes for the keys: autocrypt_asm_general_body, autocrypt_asm_click_body

res/values/strings.xml
```diff
@@ -505,9 +503,9 @@
     <string name="systemmsg_member_added">Member %1$s added.</string>
     <string name="systemmsg_member_removed">Member %1$s removed.</string>
     <string name="systemmsg_group_left">Group left.</string>
-    <string name="systemmsg_read_receipt_subject">Read receipt</string>
-    <string name="systemmsg_read_receipt_body">This is a read receipt for the message \"%1$s\".\n\nIt means that message was displayed on the recipient\'s device, not necessarily that the content was read.</string>
-    <string name="systemmsg_cannot_decrypt">This message cannot be decrypted.\n\n• It might already help to simply reply to this message and ask the sender to send the message again.\n\n• In case you re-installed Delta Chat or another email program on this or another device you may want to send an Autocrypt Setup Message from there.</string>
+    <string name="systemmsg_read_receipt_subject">Message opened</string>
```
cyBerta: Other clients use these terms, too. so please revert it

res/values/strings.xml
```diff
@@ -505,9 +503,9 @@
...
+    <string name="systemmsg_read_receipt_body">The \"%1$s\" message you sent was displayed on the screen of the recipient.\n\nThis is no guarantee the content was read.</string>
+    <string name="systemmsg_cannot_decrypt">This message cannot be decrypted.\n\n• Reply to this message and ask the sender to try sending it again.\n\n• In case you have re-installed Delta Chat or another email app on this or another device, send it an Autocrypt Setup Message.</string>
```
cyBerta: The string now suggests that the solution for the failing decryption is a reply. But there might be other reasons why the decryption can fail. That's why we're in favor of a more vague statement.

cyBerta:
ok, @OzancanKaratas this was our last last review :)
@r10 and me discussed some of the strings.
If you want you can also merge https://github.com/deltachat/deltachat-android/pull/1133 into your PR

res/values/strings.xml
OzancanKaratas: I reverted this. Is it good?

OzancanKaratas:
There is nothing to solve. You can now merge, @cyBerta.

PR #1129, opened 2019-11-25
fix qr-code scanning on 64bit andoird

### LINKED ISSUE #1128: qr code scanning crashes on 64bit
scanning any qr code crashes on 64bit android.

probably an issue somewhere in the wrapper.

r10s:
the issue was that the java-part of the wrapper function was declared as `int` where it should be `long`.

moreover, this pr moves `checkQrCPtr()` to the private part together with the other CPtr functions.

closes #1128


cyBerta:
lgtm

PR #1144, opened 2019-12-03
adapt locale things

r10s:
make device messages translatable, adapt empty-chat-hints

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -159,6 +159,7 @@ public class ConversationFragment extends Fragment
     }
 
     private void setNoMessageText() {
+        DcChat dcChat = getListAdapter().getChat();
         if(chatId == DcChat.DC_CHAT_ID_DEADDROP) {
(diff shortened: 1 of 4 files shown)


src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -167,22 +168,23 @@ private void setNoMessageText() {
                 noMessageTextView.setText(R.string.chat_no_messages);
             }
         }
-        else if(getListAdapter().isGroupChat()){
+        else if(dcChat.isGroup()){
             if(dcContext.getChat((int) chatId).isUnpromoted()) {
```
cyBerta: I'm not sure, but is `dcContext.getChat((int) chatId)` the same here as `dcChat` from line 162?

r10s: i think so. the ListAdapter contains a DcChat object and some getters that are using that.
i've added a ListAdapter.getChat() function to avoid duplicating the whole DcChat api - and also we can avoid a database roundtrip this way.

so getListAdapter().isGroupChat() calls isGroup() on the internal chat object - this is the same as calling getChat().isGroup(),

PR #1147, opened 2019-12-04
tweak welcoming screen

r10s:
this pr removes the pager with weird descriptions and images from the welcoming screen.

moreover, the whole layout is a bit tweaked and more elegant ;)

Screen Shot 2019-12-05 at 00 09 50

closes https://github.com/deltachat/interface/issues/25


OzancanKaratas:
Please type something better into device_talk_welcome_message. Remove emojis and dashes. They make the text incomprehensible. The current source may be good for English and German, but it is bad for languages without auxiliary verbs.

r10s:
> Remove emojis and dashes

whats wrong with them?

> The current source may be good for English and German, but it is bad for languages without auxiliary verbs.

the welcome message is translatable on transifex and can be adapted.

OzancanKaratas:
If you will you will 👻

This is just an example. I really don't know how to translate it. I'm sure many people won't know.

r10s:
if in doubt, i would just leave the emoticons where they are - the "ghost" beside the "new dress" and the pointing finger beside the url. this is what the translators of the other languages did.

PR #1162, opened 2019-12-16
move doze reminder to device-chat

### LINKED ISSUE #1152: move doze-reminder to device-chat
- the current doze-reminder is more sth. the just just clicks away. it is easier and better to add a device message for that - this way, there is nothing the user has to click away and he can come back to the issue later as needed.

- ~~the current outdated-reminder is not working well, see #1141, we should remove it completely from here. the core can add a device-message if a version is older than 6 month or so (maybe even more ;)~~ this is another issue, for now, we leave it. it is unlikely, we will run into these issues again

r10s:
this pr moves the doze-reminder from a system-thing that users typically click away fast to a device-message.

this way, the user can come back to the information when the problem actually arise for them.

also, things are re-doable this way (the user can change the setting several times)

maybe the text needs some adaption, however, as the current text is already translated in most languages and we're close to a release, i do not change that.

also, it may make some sense to link to https://dontkillmyapp.com if the message is clicked although background-things are already allowed.

closes #1152

Screen Shot 2019-12-16 at 14 13 16

### DIFF
--- a/src/com/b44t/messenger/DcContact.java
+++ b/src/com/b44t/messenger/DcContact.java
@@ -3,7 +3,8 @@ package com.b44t.messenger;
 public class DcContact {
 
     public final static int DC_CONTACT_ID_SELF               = 1;
-    public final static int DC_CONTACT_ID_DEVICE             = 2;
+    public final static int DC_CONTACT_ID_INFO               = 2;
+    public final static int DC_CONTACT_ID_DEVICE             = 5;
     public final static int DC_CONTACT_ID_LAST_SPECIAL       = 9;
     public final static int DC_CONTACT_ID_NEW_CONTACT        = -1; // used by the UI, not valid to the core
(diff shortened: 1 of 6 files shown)


src/com/b44t/messenger/DcContact.java
```diff
@@ -3,7 +3,8 @@
 public class DcContact {
 
     public final static int DC_CONTACT_ID_SELF               = 1;
-    public final static int DC_CONTACT_ID_DEVICE             = 2;
+    public final static int DC_CONTACT_ID_INFO               = 2;
+    public final static int DC_CONTACT_ID_DEVICE             = 5;
```
r10s: we changed this in the core and forgot to adapt this in android. was no issue in the past as DC_CONTACT_ID_DEVICE was just never used in android up to now.

cyBerta:
lgtm

PR #1164, opened 2019-12-17
update changelog

r10s:


CHANGELOG.md
```diff
@@ -1,5 +1,41 @@
...
+2019-12-17
+
+Finally, after months of coding and fixing bugs, here it is: Delta Chat 1.0 :)
+An overview over the changes since v0.510:
+
+* a new core; for better stability, speed and future maintainability,
+  the core is written completely in the Rust programming language now
+* for end-to-end-encryption, rPGP is used now;
+  the rPGP library got a first independent security review mid 2019
+* improved behavior of sending and receiving messages in flaky networks
```
hpk42: i suggest to put this part further down -- not sure how much we really are better -- i think the move to async has, for now, rather lead to some regressions.

r10s: moved it down.

CHANGELOG.md
```diff
@@ -1,5 +1,41 @@
...
+An overview over the changes since v0.510:
+
+* support for user avatars: select your profile image
+  at "My profile info" and it will be sent out to people you write to
+* introduce a new "Device Chat" that informs the user about app changes
+  and, in the future, problems on the device
+* new "Saved messages" chat
+* add "Certificate checks" options to "Login / Advanced"
+* if "Show classic emails" is set to "All",
+  emails pop up as contact requests directly in the chatlist
```
missytake: afaict this is already built into 0.510.1 (at least)? My F-Droid installation has this feature already.

r10s: the option was there, but the emails did not pop up as "contact requests" in the chatlist.

PR #1165, opened 2019-12-17
fix builds for MacOSX Catalina

cyBerta:
CPATH needs to be set in order to compile deltachat iOS whareas deltachat android builds fail if CPATH is set

@r10s can you please verify that this change has no effect on mojave?


r10s:
i will, however, after we released 1.0 - i do not want to do changes on the make-script before a final release (i have no good idea what CPATH does)

PR #1176, opened 2019-12-19
call dc_maybe_network() from a worker thread.

r10s:
theoretically, dc_maybe_network() can be called from the main thread and returns at once,
however, in reality, it does currently halt things for some seconds.
this is a workaround that make things usable for now.

closes #1176


cyBerta:
I forced blocking within the Thread fo see if the broadcast receiver will complain (has a short lifecycle) and it works perfectly.

PR #1181, opened 2019-12-20
use Intent.ACTION_VIEW from an Activity-context

### LINKED ISSUE #1154: opening PDF directly not possible sometimes
It isn't possible to **open** an attached **PDF** directly inside the App.  I have to save it in order to be able to open it.
Device:
Oneplus 6
OS :
Lineage 16 , latest build
( Android 9 )

r10s:
fixes #1154 by force using an Activity-context. this seems to be needed for Android 9, tried opening a pdf in an emulator with Android 9 and this works - without this fix it fails as described in #1154

i am unsure if FLAG_ACTIVITY_NEW_TASK is still needed for SDK <= 23 (Android 6, Marshmallow), the documentation is unclear to me here, https://developer.android.com/reference/android/content/Intent.html#FLAG_ACTIVITY_NEW_TASK, and it was introduced for some reasons at https://github.com/deltachat/deltachat-android/pull/741 ,  however, probably it is fine. but i will try to test this explicitly.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -648,7 +648,7 @@ public class ConversationItem extends LinearLayout
 
         context.startActivity(intent);
       } else if (slide.getUri() != null) {
-        ApplicationDcContext.openForViewOrShare(context, dcContext, slide.getDcMsgId(), Intent.ACTION_VIEW);
+        dcContext.openForViewOrShare(context, slide.getDcMsgId(), Intent.ACTION_VIEW);
       }
     }
   }
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -648,7 +648,7 @@ public class ConversationItem extends LinearLayout
 
         context.startActivity(intent);
       } else if (slide.getUri() != null) {
-        dcContext.openForViewOrShare(slide.getDcMsgId(), Intent.ACTION_VIEW);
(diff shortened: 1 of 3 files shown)


src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -648,7 +648,7 @@ public void onClick(final View v, final Slide slide) {
 
         context.startActivity(intent);
       } else if (slide.getUri() != null) {
-        dcContext.openForViewOrShare(slide.getDcMsgId(), Intent.ACTION_VIEW);
+        ApplicationDcContext.openForViewOrShare(context, dcContext, slide.getDcMsgId(), Intent.ACTION_VIEW);
```
cyBerta: I would use an static import here (less distracting text, and with a click I can see where it origins from), but that's only a matter of style.

src/org/thoughtcrime/securesms/ProfileDocumentsFragment.java
```diff
@@ -158,7 +158,7 @@ private void handleMediaPreviewClick(@NonNull DcMsg dcMsg) {
       return;
     }
 
-    dcContext.openForViewOrShare(dcMsg.getId(), Intent.ACTION_VIEW);
+    ApplicationDcContext.openForViewOrShare(getActivity(), dcContext, dcMsg.getId(), Intent.ACTION_VIEW);
```
cyBerta: here as well

src/org/thoughtcrime/securesms/ConversationItem.java
cyBerta: wait. Actually I agree with hugotty here. I think it would be cleaner to keep it an object function. With the newimplemented instance check, passing the Context will be obvious, too

cyBerta:
lgtm

PR #1190, opened 2019-12-22
Tabbed activity for qr code scanning

Hocuri:
To be done:
- actually test scanning with the new activity
- bug: When the camera permission is requested and then granted or denied, nothing happens (`handleQrScanWith*` is not called)
... both done.


r10s:
@Hocuri great that you're pushing that forward, just tried it out in an emulator, looks great!

what i've seen so far is a very well written and readable pr - very nice!

i know this is still `W I P`, so, not sure if my comments are already welcome, but anyway, the following came into my mind just now :)

- i would swap the two tabs - first show, then scan. 
- "show" will be opened by default - this is also some implicit guidance as it is more clear what "scan" means if you've seen a huge qr-code before
- the app can remember the last open tab and open it again, maybe it is not needed to persist this state
- if the camera is unavailable for some reasons (emulator, not present, rights not given, broken), the whole activity should still be able to be opened - i think this is also easier to achive when "show" is the default and "scan" needs another tab - that might crash things then :)

Hocuri:
- on the other hand, you scan more often than you show (because you also scan group qr codes)
- 
- why would this improve the user experience? Users will often change between "show" and "scan" role anyway.
- that's the bug I mentioned

Anyway, I need to wait until https://github.com/deltachat/deltachat-core-rust/issues/1100 is solved.

r10s:
> on the other hand, you scan more often than you show (because you also scan group qr codes)

sure, this is because it makes sense to remember the selected tab

> why would this improve the user experience? Users will often change between "show" and "scan" role anyway.

it is about the _first_ time a user taps around in the app without reading/thinking much about it.
when the user sees a big qr-code, it is pretty clear what he can "scan" here - without reading instructions etc.  
when he sees a "camera", he might wonder, what he can do here. esp. as eg. whatsapp offers a camera a a similar position (on ios)

but my other point is that showing a qr-code, there are much fewer things that can go wrong, if the camera is not working, the user has to click things away, is confonted with errors etc - even if she/he may want to only scan a qr-code.

however, these are just my thoughts, i might be wrong :)

Hocuri:
@hpk42 @Simon-Laux I also discussed about this with you, what do you think? 

Also, can so help me with https://github.com/deltachat/deltachat-core-rust/issues/1100?

hpk42:
"first usage" UX-driven design sounds and @r10s' points make sense with me. 

FYI, 36c3 is starting, people are building up or tearding down things,
traveling etc. -- so right now there is not much hacking going on
and it's hard to predict the next days :)

Hocuri:
Done.

Issue #743, opened 2019-02-25
Draft lost if app is swiped out from the recent list


adbenitez:
If you are writing some message, then stop writing and minimize Delta Chat, then go to the recent list and swipe/close the app from there, the draft is lost!! this may be the cause of https://github.com/deltachat/deltachat-android/issues/420, seems like the draft is not saved in `onPause()`????

- Platform (android/blackberry/anbox): Android 6.0
- Device: BLU (RAM<500MB, low mem device)
- Delta Chat Version: 0.101.1

adbenitez:
in this situation, k9 do save the draft.

r10s:
for me, the draft is saved.
again, it seems that onDestroy() is not called because of low memory.

r10s:
wrt k9, well might be that the draft is saved under different circumstances or that the app was just not killed, maybe it consumes less memory in total or so.

adbenitez:
DeltaChat isn't killed! if I minimize DeltaChat and go back to it the app is resumed from where it was in the chat with the draft still there, for some reason dc isn't saving the draft when I minimize it, the app isn't destroyed until I do it manually swiping out the app from the recent list, k9 save the draft as soon as I minimize the app (a toast is shown saying that the draft was saved)
(my system kill the minimized app if I open another memory consuming app, but I am not doing it)

adbenitez:
if it is easy to do, I can test a debug apk which print a toast on screen when the draft is saved to test this on my device, if you think this isn't worthy I am ok if you close this issue :)

adbenitez:
I mean, if it easy to you to provide such apk to me, I can't compile it by myself atm :(

r10s:
i think it's worth some tries, however, just now, me personally is not up for that :)

adbenitez:
ok, tested this on a tablet from my work with GBs of RAM, and it also happens, the tablet has Android 6.0.1 so this is not a low RAM problem, seems like it is something realated to Android versions (Android 6 in this case)
can someone else with Android 6 confirm this??
Steps to reproduce:
1. Open Delta Chat and open a Chat
2. Write something, but don't send it.
3. Minimize Delta Chat and go to your recent apps list (at this point the message you typed should be saved as a draft)
4. Swipe out Delta Chat from the recent apps list
5. Open Delta Chat again
6. The draft is lost, it wasn't saved

r10s:
[...]

PR #1192, opened 2019-12-28
Fix743lost draft on pause

angelo-fuchs:
This change causes a save of the current draft in the 'onPause' event.

I can't see an unwanted side effect.


angelo-fuchs:
reference #743

cyBerta:
Cool! Looks good to me codewise. I can test the pr tomorrow and merge it afterwards.

Issue #934, opened 2019-05-27
Search function is not able to search for special chars


csb0730:
v0.200.0

With text search function it's not possible to search for characters like "_-+*/\()". This is only an excerpt of all possible special chars.

I tried to get it into the code but didn't achieve it.

It's possible to enter these chars in input field and if no result is found You can see the correct spechial chars in a toast.
But on the way to the core they are filtered somewhere in the code. They never appear in core search functions.

Tried to modify/find:
- android:inputType="textPersonName"
- androd:digits="abcdefg ....."
- search for some filter functions.

cyBerta:
This is happening on purpose: https://github.com/deltachat/deltachat-android/blob/5efe4deceeb46f11ae28ca056d6bcd481da70fee/src/org/thoughtcrime/securesms/search/SearchRepository.java#L25

I didn't checked the code more in detail for now, but to me it smells like a security measurement to avoid sql injections.

hpk42:
huh, can't you do special chars in search without security problems? 
I am guessing one just needs to properly parse/escape special chars -- 
probably best done in core.  Maybe @r10s can comment if this restriction is 
neccessary?

r10s:
i do not think, this restriction is necessary. 
it seemsto be a relict from the the signal-ui, maybe it was necessary there.

r10s:
here is a little more detailed explanation: https://github.com/deltachat/deltachat-android/blob/5efe4deceeb46f11ae28ca056d6bcd481da70fee/src/org/thoughtcrime/securesms/search/SearchRepository.java#L80 - i think this needs a bit more investigation at some point.

angelo-fuchs:
A quick test with just removing the filter produces a very good result. I searched with backspaces, braces, Quotation marks... Everything turned out well (e.G. searching exactly the character I typed, no Exceptions raised etc.) Seems like there is already a proper input sanitation in place

r10s:
fixed by https://github.com/deltachat/deltachat-android/pull/1193

PR #1193, opened 2019-12-28
#934 removed filter for special chars. Core seems to be handling this…

angelo-fuchs:
… just fine.


cyBerta:
Have you tested searching for things like "I'm" (phrases including an apostroph)? Also the other banned characters, did you test them? I think I can test that tomorrow, too. Codewise it looks good to me. 👍

r10s:
yip, if there are some characters core cannot handle, core should filter them.

Issue #990, opened 2019-07-04
voice message can't be exported to file


csb0730:
- Platform: Android 8.0.0:
- Device: Huawei P10
- Delta Chat Version: 0.500.0 Google Play

- Expected behavior:
When saving a voice message by disk symbol it should be saved to a .aac file either in folder "music" or "downloads".
 
- Actual behavior:
It seems that export is done but file cannot be found.

- Steps to reproduce the problem:
Mark a voice message by long tap, tap at disk symbol to export, do it, then no file can be found.

csb0730:
At export dialog there is no hint to what folder the file is exported.

r10s:
fixed by #1195

PR #1195, opened 2019-12-28
Fix990 saving voice message

angelo-fuchs:
This fix introduces a message to be displayed to the user that says where a file has been stored.

I'm uncertain if I got  the strings.xml right or if I need to do anything more so that translations can pick this up.


angelo-fuchs:
Reference #990

r10s:
> I'm uncertain if I got the strings.xml right or if I need to do anything more so that translations can pick this up.

everything fine :) from time to time we upload new strings to transifex, let them translate there and pull the result.

r10s:
i this still wip? otherwise it can be merged from my point of view.

PR #1198, opened 2019-12-30
Fix not trying to play a sound when -no sound- has been selected.

### LINKED ISSUE #1197: Notifications are not shown sometimes (and sound not played, even if note shows)
Sometimes (I think always when I pause DC while on the contact list) my Notifications are not shown.

During debugging this I changed the sound that should be played when a new message arrives and since then it doesn't happen anymore, so if somebody can still reproduce this, please tell.

Most likely not related to #1033 as the problem here is within the handling of URIs during the playing, which could not cause the problems described there.

angelo-fuchs:
fix #1197 
When the uri comes in with a `null` state the ringtone sometimes is build but with an internal ­­­­broken uri.
So this fix removes the creation and call to the ringtone altogether if it is null to begin with.

Should have no side effects.


r10s:
> Should have no side effects.

i like this statement ;)

PR #1204, opened 2020-01-02
add verified group hint

r10s:
this pr adds a small text that explains what is a "verified group". up to now, this was completely missing.

moreover, as there may be os-stores that do not allow beta/experimental stuff, we removed the string "experimental" from the main option, it lives on, a bit tuned down, in the explanation.

cc @hpk42 

Screen Shot 2020-01-02 at 21 45 16

### DIFF
--- a/res/layout/group_create_activity.xml
+++ b/res/layout/group_create_activity.xml
@@ -6,71 +6,49 @@
               android:layout_height="match_parent"
               android:orientation="vertical">
 
-    <RelativeLayout
+    <LinearLayout
         android:layout_width="fill_parent"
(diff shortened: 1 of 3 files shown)


cyBerta:
nice clean up in group_create_activity.xml. What was the rationale behind having a progressbar etc. there? As you cc'ed @hpk42  I'm going to wait before I merge.

r10s:
>  What was the rationale behind having a progressbar etc. there?

the progressbar is removed - and was always hidden before. i assume it is a relict of the signal code, it is not needed for delta, there is no central server where we have to announce the group or so :)

PR #1210, opened 2020-01-08
[wip] Get iOS and desktop strings

OzancanKaratas:
Please clean untranslated.xml in iOS repository and untranslated_en.json in Desktop repository.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -385,7 +385,7 @@
     <string name="pref_trim_now_button">Trim all chats now</string>
     <string name="pref_trim_now_ask">Delete all old messages now? This will immediately trim all chats to the %d most recent messages.</string>
     <string name="pref_blocked_contacts">Blocked contacts</string>
-    <string name="pref_profile_photo_remove_ask">Remove profile photo?</string>
+    <string name="pref_profile_photo_remove_ask">Remove profile picture?</string>
     <string name="pref_help_url">https://delta.chat/help</string>
     <string name="pref_password_and_account_settings">Password and account</string>
     <string name="pref_who_can_see_this_information">Who can see this?</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -37,6 +37,7 @@
     <string name="one_moment">One moment…</string>
     <string name="done">Done</string>
     <string name="undo">Undo</string>
+    <!-- Translators: used eg. for the next view, could also be "continue" or so. as used in ios headers, the string should be as short as possible, though. -->
     <string name="next">Next</string>
     <string name="error">Error</string>
     <string name="error_x">Error: %1$s</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -579,11 +579,11 @@
     <string name="perm_required_title">Permission required</string>
     <string name="perm_continue">Continue</string>
     <string name="perm_explain_need_for_camera_access">To capture photos and videos, allow Delta Chat to access the camera.</string>
-    <string name="INFOPLIST.NSCameraUsageDescription">This is used to take photos and videos or scan QR codes.</string>
-    <string name="INFOPLIST.NSPhotoLibraryUsageDescription">This is used to share your photos and videos or set up a profile picture.</string>
+    <string name="INFOPLIST.NSCameraUsageDescription">Delta Chat uses your camera to take and send photos and videos and to scan QR-codes.</string>
+    <string name="INFOPLIST.NSPhotoLibraryUsageDescription">Delta Chat will let you choose which photos from your library to send.</string>
     <string name="perm_explain_need_for_mic_access">To send audio messages, allow Delta Chat to access the microphone.</string>
-    <string name="INFOPLIST.NSMicrophoneUsageDescription">This is used to send your voice.</string>
-    <string name="INFOPLIST.NSContactsUsageDescription">This is used to find your contacts who using Delta Chat.</string>
+    <string name="INFOPLIST.NSMicrophoneUsageDescription">Delta Chat uses your microphone to record and send voice messages and videos with sound.</string>
+    <string name="INFOPLIST.NSContactsUsageDescription">Delta Chat uses your contacts to show a list of email addresses you can write to. Delta Chat has no server, your contacts are not sent anywhere.</string>
     <string name="perm_explain_need_for_storage_access">Delta Chat needs access to your files in order to export files or to create backups.</string>
     <string name="perm_explain_need_for_storage_access_share">Delta Chat needs access to your files to share them.</string>
     <string name="perm_explain_access_to_camera_denied">To take photos or capture videos, go to the app settings, select \"Permissions\", and enable \"Camera\".</string>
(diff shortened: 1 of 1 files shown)


res/values/strings.xml
```diff
@@ -665,10 +684,12 @@
     <string name="a11y_remove_account_btn_label">Remove account</string>
     <string name="a11y_message_context_menu_btn_label">Message actions</string>
     <string name="a11y_background_preview_label">Background preview</string>
+    <string name="a11y_profile_image_label">Your profile image</string>
```
r10s: we call it "profile picture".

in general, i would just use the string `pref_profile_photo` (i know ;) and do no use a separate a11y for that.

res/values/strings.xml
```diff
@@ -610,6 +625,7 @@
     <string name="contact_request_title_desktop">Contact request</string>
     <string name="delete_message_desktop">Delete message</string>
     <string name="more_info_desktop">More info</string>
+    <string name="switch_account_desktop">Switch account</string>
```
r10s: we should not add the _desktop suffix where it is not really desktop-specific.

so, the _desktop suffix should go only to strings that exist eg. on android and desktop, but differ in details.

i would not see this as a blocker, however, maybe it makes sense to adapt the desktop first.

res/values/strings.xml
```diff
@@ -627,6 +643,7 @@
     <string name="initiate_key_transfer_title_desktop">Initiate key transfer</string>
     <string name="initiate_key_transfer_desktop">An Autocrypt Setup Message securely shares your end-to-end setup with other Autocrypt-compliant apps. The setup will be encrypted by a setup code displayed here and must be typed on the other device.</string>
     <string name="reply_to_message_desktop">Reply to message</string>
+    <string name="select_profile_image_desktop">Select your new profile image</string>
```
r10s: also here, the desktop suffix is not needed.

res/values/strings.xml
```diff
@@ -639,12 +656,14 @@
     <string name="autocrypt_correct_desktop">Autocrypt setup transferred.</string>
     <string name="autocrypt_incorrect_desktop">Incorrect setup code. Please try again.</string>
     <string name="create_chat_error_desktop">Could not create chat.</string>
-    <string name="ask_delete_chat_desktop">Delete this chat?</string>
     <string name="email_validation_failed_desktop">Email address required.</string>
     <string name="forget_login_confirmation_desktop">Delete this login? Everything will be deleted, including your end-to-end setup, contacts, chats, messages and media. This action cannot be undone.</string>
     <string name="me_desktop">me</string>
     <string name="in_this_group_desktop">Group members</string>
     <string name="not_in_this_group_desktop">Potential group members (not in group)</string>
+    <string name="ask_delete_chat_desktop">Are you sure you want to delete \"%1$s\"?</string>
```
r10s: also here, the desktop suffix is not needed.

res/values/strings.xml
```diff
@@ -538,8 +549,8 @@
     <string name="qrshow_join_group_hint">Scan this to join the group \"%1$s\".</string>
     <string name="qrshow_join_contact_title">QR invite code</string>
     <string name="qrshow_join_contact_hint">Scan this to set up a contact with %1$s.</string>
-    <string name="qrshow_join_contact_no_connection_hint">QR code setup requires an internet connection. Please connect to a network before proceeding.</string>
```
r10s: please no changes the english strings in the same pr - this would force all translations to be invalidated, so fixing typos in the english strings should be handled differently in a separated pr.

res/values/strings.xml
```diff
@@ -568,13 +579,18 @@
     <string name="perm_required_title">Permission required</string>
     <string name="perm_continue">Continue</string>
     <string name="perm_explain_need_for_camera_access">To capture photos and videos, allow Delta Chat to access the camera.</string>
+    <string name="INFOPLIST.NSCameraUsageDescription">This is used to take photos and videos or scan QR codes.</string>
```
r10s: the INFOPLIST-strings are just updated on ios, https://github.com/deltachat/deltachat-ios/commit/f3d00176497b51a5bb86b8c13e21d17695b2565b. 

i would wait if they are accepted by apple before pushing them to the translators.

res/values/strings.xml
```diff
@@ -37,7 +37,6 @@
     <string name="one_moment">One moment…</string>
     <string name="done">Done</string>
     <string name="undo">Undo</string>
-    <!-- Translators: used eg. for the next view, could also be "continue" or so. as used in ios headers, the string should be as short as possible, though. -->
```
r10s: why remove this hint?

res/values/strings.xml
```diff
@@ -125,9 +124,10 @@
 
     <string name="message_delivered">Message delivered</string>
     <string name="message_read">Message read</string>
-
+    <string name="message_not_verified">The message was sent with non-verified encryption. See \"Info\" for details.</string>
```
r10s: this message comes from a hack in https://github.com/deltachat/deltachat-desktop/blob/master/src/main/deltachat/chatlist.js#L110 - there are good chances that the original string will change.

r10s: i filed an issue for that, https://github.com/deltachat/deltachat-desktop/issues/1319 , so maybe remove the string for now.

res/values/strings.xml
OzancanKaratas: Because it was not necessary.

res/values/strings.xml
OzancanKaratas: Should I remove this?

res/values/strings.xml
r10s: i think it is helpful as "Next" can be translated to different words in different languages. the hint is shown eg. in transifex and other tools and gives some context

res/values/strings.xml
OzancanKaratas: <string>"Delta Chat uses your contacts to show a list of email addresses you can write to. Delta Chat has no server, your contacts are not sent anywhere.</string>

Delete the quotation mark after the greater. This may cause the application to crash.

res/values/strings.xml
OzancanKaratas: They are so small that it will not be a problem to be handled again.

res/values/strings.xml
OzancanKaratas: Desktop developers did this. All I'm doing is just add the ones in the untranslated_en.json file.

res/values/strings.xml
r10s: but there is no need to delete >20 translations just because of a typo fix in the english source. therefore, this pr must only contain adding strings. changes to the english sources must be removed completely.

OzancanKaratas: You're right, but if corrections are made with new translations, won't we get something more complete?

res/values/strings.xml
r10s: thanks, good catch, i will fix that. however, en.lproj/InfoPlist.strings has precedence over Info.plist so the wrong string is at least not displayed.

OzancanKaratas:
@cyBerta, do you have an idea?

res/values/strings.xml
```diff
@@ -327,7 +332,7 @@
     <string name="login_smtp_server">SMTP server</string>
     <string name="login_smtp_port">SMTP port</string>
     <string name="login_smtp_security">SMTP security</string>
-    <string name="login_auth_method">Authorization method</string>
+    <string name="login_auth_method">Authentication method</string>
```
r10s: what is the reason for this chage?

options are "auto" and "oauth2" - the latter is about "authorization",  https://oauth.net/2/ . so, this is hard. if in doubt, i would just leave it unchanged.

OzancanKaratas: Changes was undone.

PR #1213, opened 2020-01-11
[WIP] Refactor Internal Setter smell in ProgressDialog class

emaiannone:
Hi, I'm Emanuele Iannone, a master student at University of Salerno.
Since my bachelor's thesis I have been working on a **code smell refactoring plugin** called *aDoctor*, which is able to identify and fix energy-related problems in Android apps.
I launched it on your project, finding different instances of code smells. I chose one of them and let the plugin automatically fix it.
In this case I chose Internal Setter, that is present when non static method calls a setter method of an instance variable instead of directly changing its value. These kind of smell may have a non trivial impact on energy consumption, as shown in this paper: `https://www.sciencedirect.com/science/article/pii/S0950584918301678`.
Besides, this kind of refactoring does not impact on the functionalities of your app, so it is totally safe. Let me know if you are interested in this refactoring proposal.

### DIFF
--- a/src/org/thoughtcrime/securesms/util/views/ProgressDialog.java
+++ b/src/org/thoughtcrime/securesms/util/views/ProgressDialog.java
@@ -51,7 +51,7 @@ public class ProgressDialog extends AlertDialog {
         if (progressBar != null) {
             progressBar.getIndeterminateDrawable()
                     .setColorFilter(ContextCompat.getColor(getContext(), R.color.delta_accent), PorterDuff.Mode.SRC_IN);
-            progressBar.setIndeterminate(indeterminate);
+            this.indeterminate = indeterminate;
         }
     }
 


src/org/thoughtcrime/securesms/util/views/ProgressDialog.java
```diff
@@ -51,7 +51,7 @@ private void setupProgressBar() {
         if (progressBar != null) {
             progressBar.getIndeterminateDrawable()
                     .setColorFilter(ContextCompat.getColor(getContext(), R.color.delta_accent), PorterDuff.Mode.SRC_IN);
-            progressBar.setIndeterminate(indeterminate);
+            this.indeterminate = indeterminate;
```
r10s: i do not understand why progressBar.setIndeterminate() is superfluous. imu, this swaps drawables, starts animations etc. it is just a system call that tells the used object how it shall appear.

but even if the call is not needed, `this.indeterminate = indeterminate` is a void statement, why does this avoid some smell?

but, tbh, i did not dived into details :)

r10s:
@emaiannone thanks for checking Delta Chat out, the thing with "identify and fix energy-related problems" sounds very interesting, however, not sure of the thing above is related. also as the code is not even run on normal program execution afaik

angelo-fuchs:
This change will have (most likely) side effects on the call graph. The ProgrgessBar seems not properly set up if called this way.

While I would find it cool to detect situations where we can remove calls that could be avoided this would need an explanation on why it is side-effect-free.

angelo-fuchs:
I was curious and went looking at the indeterminate value here (toggling if a progress bar is displayed (during registration) or a rotating spinner icon (when mass deleting contacts from the list))

**The optimizer is wrong, this change is dangerous.**

The indeterminate value in the ProgressBar is `true` in `dialog_progress.xml` and needs to be set to `false` during setup.

Compare callgraph from `RegistrationActivity:407` to the one from `ConversationListFragment:286`.

r10s:
closing this as not mergeable. but i am happy to continue the discussion.

PR #1216, opened 2020-01-17
some cleanup

### LINKED ISSUE #1211: remove image from "no chats" view
as we show the device chat by default, the whole "empty" screen is shown very rarely.

we have such images in the device chat now.

r10s:
this pr removes some 100k of unused resources, code and several strings, the latter mainly to avoid unneeded work for the translators.

to check, if a string is in used in any ui, this pr adds a little script, ./tools/grep-string.sh

closes #1211


hpk42:
removing code is the best :) 
but you didn't want to merge it before releasing 1.1 to playstore?

r10s:
tbh, i just forgot about this pr :)

PR #1221, opened 2020-01-23
fix saving drafts

### LINKED ISSUE #1220: error in draft saving in test version 1.1
~~maybe~~ introduced by https://github.com/deltachat/deltachat-android/pull/1192

the issue is that the pr calls processComposeControls() twice on leaving the activity - but the function also cleans up the controls for $reasons, so the second call saves an empty draft.

not totally sure about the call of the livetime functions, however, i assume onPause() is always called, so maybe we should remove the call from onDestroy() (which is not called eg. when the app terminates or is slided away - this was the issue the #1192 tries to fix)

EDIT: here is a nice page about activity lifecyles, including onNewIntent() which is used by us: https://www.nickagas.com/using-onnewintent-and-activity-lifecycle/

r10s:
this is an addition to #1192:

by calling processComposeControls() twice, in onPause() and on Destroy(),
the seccond call gets an empty draft.

as onDestroy() is not guarateed to be executed anyway -
this was the intent of the move in pr #1192 -
we just remove this call and rely on onPause().

see https://developer.android.com/reference/android/app/Activity.html
for lifecycles and killable events.

closes #1220

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -305,7 +305,6 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
   @Override
   protected void onPause() {
     super.onPause();
-    // save draft on pause
     processComposeControls(ACTION_SAVE_DRAFT);
     MessageNotifierCompat.updateVisibleChat(MessageNotifierCompat.NO_VISIBLE_CHAT_ID);
     if (isFinishing()) overridePendingTransition(R.anim.fade_scale_in, R.anim.slide_to_right);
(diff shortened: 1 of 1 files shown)


angelo-fuchs:
Looks good to me. I actually didn't manage to get onDestroy() called anyway in my tests, so if you saw an error there its a good idea to have it gone. I can see no scenario where we first have a draft, then onDestroy gets called but onPause not.

([Technically](https://www.xkcd.com/1475/)) Design wise it would most likely be correctest if we shift the draft to the onSaveInstanceState function instead but I would need to test if that actually works the way its designed to.
So then onPause and onResume would not handle the chat state at all and we would rely on the system to call onSaveInstanceState as it is supposed to.

r10s:
@angelo-fuchs thanks for checking :)

> Design wise it would most likely be correctest if we shift the draft to the onSaveInstanceState function instead

i think onSaveInstanceState() is more for transient state, state of the ui etc. it is not always called.

from https://developer.android.com/reference/android/app/Activity.html: 

_"Note that it is important to save persistent data in onPause() instead of onSaveInstanceState(Bundle) because the latter is not part of the lifecycle callbacks, so will not be called in every situation as described in its documentation."_

this livecycle stuff is way too complicated.

PR #1256, opened 2020-02-21
sign-in vs. sign-up

r10s:
improve wording to make more clear, this is sign-in, not-sign up.

this pr changes the text "password" to "existing password".

Screen Shot 2020-02-21 at 18 31 13

the idea was also to change the text below the field to "Please login to your e-mail server using your existing e-mail address and password" or - however, this would at least partly only repeat the headline and the labels. also the existing hint seems to be important and, we're talking about ppl that do not really read hints, so adding more of them is maybe not helpful

so, i think, we can maybe just go for the small change and see if this already helps. it is also not that we get tons of complains to that - but i think the "no delta servers" actually helped to avoid scaring people.

### DIFF
--- a/res/layout/registration_activity.xml
+++ b/res/layout/registration_activity.xml
@@ -69,7 +69,7 @@
                 android:id="@+id/password_text"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:hint="@string/password"
+                android:hint="@string/existing_password"
                 android:inputType="textPassword" />
 
         </com.google.android.material.textfield.TextInputLayout>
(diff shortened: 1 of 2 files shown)


Ampli-fier:
I have another idea: The server is normally not the user's server. But the account is his/her own.

Regarding the _Delta Chat server_ it might sound a little confusing what is meant by  the **data**. Is it also the chats, media, etc?

Maybe even state:

`Please log in to an existing email account that you want to use for Delta Chat.`

As _existing_ is already mentioned the hint, I removed the word from the Password field.

Screenshot_20200221-220017

angelo-fuchs:
> Maybe even state:
> `Please log in to an existing email account that you want to use for Delta Chat.`
> As _existing_ is already mentioned the hint, I removed the word from the Password field.

I would keep the _existing_ in both places. Otherwise, I love this suggestion.

r10s:
i think, to move forward, we should just go for the initial idea of `Exiting password`and see if this already helps. also on sign-up, you need an existing email address, so, maybe this is already more clear.

PR #1257, opened 2020-02-25
autodelete messages from server

r10s:
this pr is about the ui of https://github.com/deltachat/deltachat-core-rust/pull/1310 (delete-from-server) and can be easily enhanced for targeting https://github.com/deltachat/deltachat-core-rust/issues/1309 (delete-from-device) later. both will be created as independent options.

Screen Shot 2020-03-12 at 15 22 37 Screen Shot 2020-03-12 at 13 45 04

Screen Shot 2020-03-12 at 13 47 28 Screen Shot 2020-03-12 at 13 47 38

previous thoughts:

~~i am currently thinking about two options:~~

1. ~~showing two checkboxes and one time-selectors (1hour, 1day, 1week etc, maybe a slider)~~
2. ~~showing two time-selectors~~

~~2 seems easier to implement, also as the core does not have an on/off setting preserving the time. also ui-wise, 2 seems to be easier to understand. 2 also has the advantage, that it has some build-in-protection against accidentally enabling, for 1., we would probably needs some extra-dialog or so, checkboxes are pretty easy hit on android.~~

~~however, 1 has the advantage that the user can temporarily disable the deletion (but will he? it is always about _old_ messages). moreover, 1 has the advantage that the technical limitation delete-from-server-time<=delete-from-device-time is no issue as there is only one time-selectors - in 2 this needs some additional ui. another advantage of 1. is that we can add further conditions as "keep N messages", however, it is unclear if this will be needed in this way. also we can adapt things as needed then.~~

~~i would probably prefer 2 - what do you think? cc @hpk42 @link2xt~~

### DIFF
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -605,6 +605,13 @@ JNIEXPORT jint Java_com_b44t_messenger_DcContext_getFreshMsgCount(JNIEnv *env, j
 }
 
 
+JNIEXPORT jint Java_com_b44t_messenger_DcContext_estimateDeletionCount(JNIEnv *env, jobject obj, jboolean from_server, jlong seconds)
+{
+	return dc_estimate_deletion_cnt(get_dc_context(env, obj), from_server, seconds);
+}
+
+
+
 JNIEXPORT jlong Java_com_b44t_messenger_DcContext_getMsgCPtr(JNIEnv *env, jobject obj, jint id)
 {
 	return (jlong)dc_get_msg(get_dc_context(env, obj), id);
--- a/res/layout/autodel_confirm.xml
+++ b/res/layout/autodel_confirm.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<FrameLayout
+    xmlns:android="http://schemas.android.com/apk/res/android"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <CheckBox
+        android:id="@+id/i_understand"
+        android:text="@string/pref_autodel_confirm"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_marginLeft="20dp"
+        android:layout_marginRight="20dp"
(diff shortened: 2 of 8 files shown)


link2xt:
> 1 has the advantage that the technical limitation delete-from-server-time<=delete-from-device-time is no issue

There is no such limitation, actually. Deleting message locally without deleting it on the server moves it into trash chat and clears its context, so it is just a tombstone to prevent accidentally redownloading it.

Core is going to support all combinations of timers, so I think 2 is better option overall. While local message deletion is not implemented, we can release a version with remote message deletion only, and add another item later.

r10s:
ahh, cool. i though deleting-from-server=1month is not possible if eg. delete-from-device=1day.

link2xt:
I'd replace "1 month" with "4 weeks" or "30 days", because month duration is not fixed.

And probably remove all options with higher limits. 1 month is enough to synchronize all your devices. On longer timespan database, DC version etc. are bound to rot and cause compatibility problems.

Simon-Laux:
what does "At once" mean here? immediately after receiving?

r10s:
> what does "At once" mean here? immediately after receiving?

yip, i think so.

Simon-Laux:
I believe we should find a better term than. If its confusing for me it will likely be confusing for other folks too. How about "Immediately"?

r10s:
afaik, these terms are quite synonym. what's the difference between "Immediately" and "at once"?

Simon-Laux:
I associate "at once" more with "auf einmal"(german) than with "sofort"(german). But your right in english they are kind of synonyms - atleast dict.cc says so.

Am 26. Februar 2020 19:12:17 MEZ schrieb bjoern <notifications@github.com>:
>afaik, these terms are quite synonym. what's the difference between
>"Immediately" and "at once"? 
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/deltachat/deltachat-android/pull/1257#issuecomment-591565473

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

PR #1258, opened 2020-02-28
tweak help colors

r10s:
use sth. as the accent-color for links in the help.


angelo-fuchs:
Should we also define the color for visited links? The blue seems to be the default, is it changed somewhere and should we maybe just unset that instead?

r10s:
hm, i do not really think, this is important for now. eg. it does not get the visited by "scrolling through", so i do not think it is that helpful. i would not go for that now.

PR #1266, opened 2020-03-09
less clicks on save-message and start-chat

### LINKED ISSUE #1007: Don't show prompt "do you want to chat with x" if a chat with x already exists
with the new bot when an user taps in a mailto link in the form `mailto:BOT-ADDR?body=some-text`
they can open that link with Delta Chat, but every time the tap such a link, Delta ask if you want to chat with that contact as it was the first time you try to chat with that contact even when you already has an active chat wth the contact this should not be asked and the chat with that contact should be opened directly with the text prefilled.
This issue is not especific of mailto links, if you press the "+" then select to chat with a contact you already has a chat, a prompt is shown instead of opening directly, for the bot UX this is really annoying since you has to confirm this prompt quite often

- Platform (android/blackberry/anbox):Android 6
- Delta Chat Version: 0.510.1
- Steps to reproduce the problem: click a mailto link that points to an already acepted contact

r10s:
this pr streamlines the ui, by removing some superfluous alerts and by re-ordering the forward-dialog so that "saved messages" is always one of the first entries.

the pr relies on https://github.com/deltachat/deltachat-core-rust/pull/1336 , so that should be merged before and the core should be updated afterwards.

closes #1007

Screen Shot 2020-03-11 at 13 28 31

### DIFF
--- a/src/com/b44t/messenger/DcContext.java
+++ b/src/com/b44t/messenger/DcContext.java
@@ -38,6 +38,7 @@ public class DcContext {
     public final static int DC_GCL_ARCHIVED_ONLY    = 0x01;
     public final static int DC_GCL_NO_SPECIALS      = 0x02;
     public final static int DC_GCL_ADD_ALLDONE_HINT = 0x04;
+    public final static int DC_GCL_FOR_FORWARDING   = 0x08;
 
     public final static int DC_GCM_ADDDAYMARKER = 0x01;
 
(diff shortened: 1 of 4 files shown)


cyBerta:
UI-wise I think having the saved messages below the pinned chats would be nicer. 
Besides that it's a really nice feature.

cyBerta:
ups... wrong MR, I wanted to comment that on the iOS couterpart

Issue #1253, opened 2020-02-18
Use Rust Gradle plugin to build native library


link2xt:
Instead of our ad-hoc scripts, such as https://github.com/deltachat/deltachat-android/blob/master/ndk-make.sh, we could use "official" (from Mozilla) Rust plugin for gradle: https://github.com/mozilla/rust-android-gradle

angelo-fuchs:
the rust-android-gradle still requires the dev to perform some activity (like the rustup commands) that we need to somehow communicate to new devs. Currently they are in a comment in ndk-make.sh.

When we switch to rust-android-gradle we should check if we can include a relevant check into the gradle to see if the setup is already present and good and if not fail with a meaningful error message.

r10s:
closing this for now, there is no real need, also f-droid is happy with our ad-hoc script.

PR #1268, opened 2020-03-15
Run ndk-build from Gradle

link2xt:
Now changes to JNI should result in recompilation.

ndk-make.sh still has to be run to compile libdeltachat.a

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -148,6 +148,12 @@ android {
         release
     }
 
+    externalNativeBuild {
+        ndkBuild {
+            path "./jni/Android.mk"
+        }
+    }
+
     buildTypes {
         debug {
             minifyEnabled false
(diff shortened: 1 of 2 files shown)


r10s:
what is the reasoning for this change?

because, on mac, build stuck for more than 10 minutes, did not finish, no error. might be that it does not pick up the correct ndk or so, might be fixable, i did not dived into that. however, not sure if it is worth the effort, things are working in general.

Also, as changes that are picked up automatically do not include the rust-thing and also many devs are familiar with the "old" way and this does not simplify things much. 

finally,  i have a good idea about which ndk is picked up by the ndk-make - but i have to learn these things with this change again :)

iirc, btw, in the very beginning, i played with this externalNativeBuild but it turned out to cause problems.

link2xt:
> what is the reasoning for this change?

Trying to move all native building to Gradle so #1253 can be fixed and new developers don't have to configure anything manually (following ndk-make.sh instructions) besides installing rustup. F-Droid build could also be simplified, removing the need to configure cargo with echo commands.

r10s:
hm, i think the hard part is to set up rust. so, i am not sure if it is worth to put much energy here currently, also changing f-droid is some effort, tbh, i am pretty happy it works - never change a running system (without reasons :)

it seems not to be needed for docker, or?

link2xt:
Moved to project resurrection

PR #1270, opened 2020-03-16
Fix Android codename in the changelog

link2xt:
4.1 codename is "Jelly Bean", not Lollipop


r10s:
thanks!

where is the _red pill?_

JellyBellyBeans

PR #1273, opened 2020-03-30
Fix #1028, make sharing directly to a chat work always

### LINKED ISSUE #1028: Share to chat from share menu doesn't work
Share to a specific (recently contacted) contact/chat was proposed and implemented with #900,
But it doesn't work as expected: I can click on share with this contact and it presents the select contact to share menu thingie anyways and ignores the contact I actually clicked on in the first place.

If its hard to reproduce I will share more device info and such things, just answer to this post.

Hocuri:
I found it out.
How to reproduce:
- uninstall and re-install Delta Chat
- login
- click "Share" in another app
- select "Saved Messages"

We now have a code duplicate (in onNewIntent and in onCreate), but I was unsure if both codes should go into another method and how it should be called. `refresh`? `init`?(also, the codes of these were already very similar before this PR).

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -98,20 +98,19 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
     initializeSearchListener();
 
     TooltipCompat.setTooltipText(searchAction, getText(R.string.search_explain));
-    if (isRelayingMessageContent(this)) {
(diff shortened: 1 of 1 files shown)


r10s:
@Hocuri thanks a lot! looks good and reasonable to me

@cyBerta can you have a look at that, you're much deeper in the "share code" part than me.

Hocuri:
So, I did introduce a new method `refresh` for code dedup. Now `handleOpenpgp4fpr();` is called before
```
    conversationListFragment.onNewIntent();
    invalidateOptionsMenu();
```
but this should not make any difference.

PR #1277, opened 2020-04-01
Fix #999 swiping a notification away should not mark things as "noticed"

### LINKED ISSUE #999: swiping a notification away should not mark things as "noticed"
just tested that with wa that has exactly this behavior.  also this is the behavior of the old tg-ui.

Hocuri:
I looked around in the code a bit to see what implications my change might have.

The only one I found is that `NotificationState.removeNotificationsForChat(int)` is not called and `NotificationState.notificationCount` is not decreased. This should be just fine because `notificationCount` is used to show an unread message count. Here, it is expected behavior that the unread message count does not decrease just because the message was swiped away.

### DIFF
--- a/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
+++ b/src/org/thoughtcrime/securesms/notifications/MessageNotifier.java
@@ -253,7 +253,6 @@ abstract class MessageNotifier {
                 firstItem.getText(""), firstItem.getSlideDeck());
         builder.setContentIntent(firstItem.getPendingIntent(context));
         builder.setGroup(NOTIFICATION_GROUP);
-        builder.setDeleteIntent(notificationState.getMarkAsReadIntent(context, chatId, notificationId));
 
         long timestamp = firstItem.getTimestamp();
         if (timestamp != 0) builder.setWhen(timestamp);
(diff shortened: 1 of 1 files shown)


cyBerta:
can you please check, if the old notification that has been swiped away reappears if a new notification arrives? I think we should avoid that behavior.

Hocuri:
> can you please check, if the old notification that has been swiped away reappears if a new message arrives? I think we should avoid that behavior.

I don't know anymore if I tested this but I am pretty sure that they do reappear. This also is the behavior of WhatsApp (if a notification arrives, I swipe it away without marking it as read, another notification appears, then the first notification will be shown together with the second one). I find this pretty useful because I get reminded of old unread messages. Also I think that we should stick to the behavior of messengers users already know.

cyBerta:
> then the first notification will be shown together with the second one

That's the point - are they reappearing as single messages right after each other or stacked?

Hocuri:
Ok, I tested it again now. They are shown as one message group ("3 messages in 2 chats") as if the user had never swept it away at all. The message group can be swept away as all at once or expanded to show the individual chats. (on Android 7)

r10s:
unfortunately, this fix came with a side-effect, #1353 , not sure, if this is better than the original issue :) however, this whole notification thing is not easily manageable meanwhile, i will do some general rework and cleanup.

PR #1279, opened 2020-04-01
Fix #1208 Back button on the Unlock screen allows to read the chat list

### LINKED ISSUE #1208: Back button on the Unlock screen allows to read the chat list
Hello,
I've noticed that a quick press of the the back button on the Unlock screen allows to read the chat list of Delta Chat. Please see https://youtu.be/mxhMDceLySc

A third person could read the contacts and the preview of the messages from a video like this just pausing it.

- Android version: /e/OS (Android 9)
- Device: Google Pixel
- Delta Chat version: 1.0.3

Hocuri:
I could reproduce something similar to #1208: Starting DC the first time after booting up, the apps list briefly showed up. This PR should fix both problems.

### DIFF
--- a/src/org/thoughtcrime/securesms/BaseActionBarActivity.java
+++ b/src/org/thoughtcrime/securesms/BaseActionBarActivity.java
@@ -38,18 +38,28 @@ public abstract class BaseActionBarActivity extends AppCompatActivity {
       forceOverflowMenu();
     }
     super.onCreate(savedInstanceState);
+
+    if (shouldLock()) {
+      // Already hide everything here to prevent sensible data from popping up shortly:
+      findViewById(android.R.id.content).setVisibility(View.INVISIBLE);
+      isHiddenByScreenLock = true;
+    }
   }
 
-    @Override
-    protected void onStart() {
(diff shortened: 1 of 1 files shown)


Hocuri:
Now, what about these pending PRs? Am I allowed to merge them myself if nobody had concerns?

r10s:
@Hocuri patience :) in general, someone else should review the prs. exceptions may be minor changes (readme, version, translatinos ...) or if some subsequent work depends on it or other $goodReasons :)

r10s:
hm, i could reproduce #1208 without and with this fix:
- enable screen lock, exit app, maybe force quit by swiping away from the recent-app-list
- start app and hit "back" several times per second - the chatlist flashes up shortly. this is doable even when waiting after the app-start

r10s:
added some findings to #1208, closing this for now. it is available as a reference from the issue,

r10s:
just noticed another issue with the pr: once the lock is disabled again in the system (via Delta Chat/Settings/App access/Change secret), all Delta Chat screen stays white (probably INVISIBLE) unrecoverable, also after restarts. back to master, things are visible again.

PR #1280, opened 2020-04-03
in chat search

### LINKED ISSUE #1217: add search to chatview and to help
i filed the same issue for both as this seems to be very similar.

- add "Search" as the first menu item  - not: a magnifier-icon in the titlebar as on the chatlist - (reason is that there is few place in the chatview titlebar and we want to be consistent)

- a click opens a search-bar, similar to the one from the chatlist

- in the search-bar, right of the search field, there should be up/down arrows that allow jumping to all results (if the input field is empty, this can be used to jump to the beginning/end, a bit hidden, but well, #892)

- the view should jump immediately to the first occurence found, no need to tap on search or so

to get all matching messages in a chatlist, there is [dc_search_msgs()](https://c.delta.chat/classdc__context__t.html#a777bb1e11d7ea0288984ad23c2d8663b).  
to get all matching text in the help, [WebView.findAllAsync()](https://developer.android.com/reference/android/webkit/WebView.html#findAllAsync(java.lang.String)) and [WebView.findNext()](https://developer.android.com/reference/android/webkit/WebView.html#findNext(boolean)) should do the job.

### LINKED ISSUE #1284: device-chat has menu entry "add attachment"
the device chat cannot send, the "add attachment" should be hidden (input bar is already hidden)

r10s:
closes #1217 and closes #1284 (not really related, but as the menu is reworked in this pr, it makes sense here to avoid merge conflicts)

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -461,13 +461,13 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
       searchItem.setOnActionExpandListener(new MenuItem.OnActionExpandListener() {
         @Override
         public boolean onMenuItemActionExpand(final MenuItem item) {
-          ConversationActivity.this.makeSearchMenuVisible(menu, searchItem, false);
+          searchExpand(menu, item);
           return true;
         }
 
         @Override
         public boolean onMenuItemActionCollapse(final MenuItem item) {
-          ConversationActivity.this.makeSearchMenuVisible(menu, searchItem, true);
+          searchCollapse(menu, item);
           return true;
         }
       });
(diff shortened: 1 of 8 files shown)


src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1499,4 +1537,82 @@ public void handleEvent(int eventId, Object data1, Object data2) {
       invalidateOptionsMenu();
     }
   }
+
+
+  // in-chat search
```
Hocuri, line 1542: Actually this whole file is getting pretty large (it is very large already), maybe all the search-related things could go into a new class `InChatSearchManager`

src/org/thoughtcrime/securesms/ConversationAdapter.java
```diff
@@ -170,6 +170,14 @@ public long getItemId(int position) {
     return fromDb;
   }
 
+  public int msgIdToPosition(int msgId) {
```
Hocuri, line 173:
```suggestion
  /**
   * Returns the position of the message with msgId in the chat list, counted from the top
   */
  public int msgIdToPosition(int msgId) {
```
because it took me a while to understand what this does.

src/org/thoughtcrime/securesms/mms/AttachmentManager.java
```diff
@@ -592,4 +592,18 @@ public void onClick(View v) {
...
 
   }
+
+  public int getVisibility() {
+    int ret = View.GONE;
+    try {
+      ret = attachmentViewStub.get().getVisibility();
+    } catch(Exception e) {}
+    return ret;
+  }
```
Hocuri, line 602:
```suggestion
  public int getVisibility() {
    try {
      return attachmentViewStub.get().getVisibility();
    } catch(Exception e) {
      return View.GONE;
    }
  }
```

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -456,6 +456,43 @@ public boolean onPrepareOptionsMenu(Menu menu) {
...
+      Log.e(TAG, "cannot set up in-chat-search: ", e);
+    }
+
+    if (!dcChat.canSend()) {
+      MenuItem attachItem =  menu.findItem(R.id.menu_add_attachment);
+      if (attachItem!=null) {
+        attachItem.setVisible(false);
+      }
+    }
+
```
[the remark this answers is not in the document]
r10s: i agree on more but short methods when this results in a better overview. i cases, wehre things are just initialized sequentially, long methods does not add complexity, so, i would not change that. it is coarse overview vs fine overview, but this is a longer discussion, better to be done one to one :)

src/org/thoughtcrime/securesms/ConversationActivity.java
r10s: yip, the file was already large before, however, splitting search, not sure if this makes sense, in general, i do not think, this section adds much complexity.

PR #1282, opened 2020-04-04
Make the links in the dialog clickable (needed for Yandex)

Hocuri:
See https://github.com/deltachat/deltachat-core-rust/issues/1377.

### DIFF
--- a/src/org/thoughtcrime/securesms/RegistrationActivity.java
+++ b/src/org/thoughtcrime/securesms/RegistrationActivity.java
@@ -20,6 +20,7 @@ import androidx.appcompat.app.AlertDialog;
 import android.text.Editable;
 import android.text.TextUtils;
 import android.text.TextWatcher;
+import android.text.util.Linkify;
 import android.util.Patterns;
 import android.view.Menu;
 import android.view.MenuInflater;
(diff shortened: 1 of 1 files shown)


Hocuri:
Because Yandex sends links in its error messages.

src/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -536,10 +537,13 @@ public void handleEvent(int eventId, Object data1, Object data2) {
...
                 if (dcContext.hasCapturedError()) {
-                    new AlertDialog.Builder(this)
+                    AlertDialog d = new AlertDialog.Builder(this)
                             .setMessage(dcContext.getCapturedError())
                             .setPositiveButton(android.R.string.ok, null)
-                            .show();
+                            .create();
+                    d.show();
+                    //noinspection ConstantConditions
+                    Linkify.addLinks((TextView) d.findViewById(android.R.id.message), Linkify.WEB_URLS|Linkify.EMAIL_ADDRESSES);
```
r10s, line 546: from the logic - i would find it nicer, if we prepare the links before we create the dialog, see https://stackoverflow.com/questions/1997328/how-can-i-get-clickable-hyperlinks-in-alertdialog-from-a-string-resource for an example.

Hocuri: Why? Because this could be done in the core? I looked at https://stackoverflow.com/questions/1997328/how-can-i-get-clickable-hyperlinks-in-alertdialog-from-a-string-resource, too but I didn't find any where this could be in the core.

The solution I chose was the shortest one.

r10s: my comment was not about the core.

i just think that the order of _first_ showing the dialog and then, after it is displayed, changing the text, a bit uncommon.

if there are no good reasons for that, i would do it the other way round, so, first creating the text including the hiliting (still, on android, core is not related) and the setting the created string to the alert.

however, having a closer look at the linked stackoverflow post, they are working with a custom control. so, seems as if your solution is at least shorter :)

so, yes, just leave it this way.

but i maybe surround the whole "Linkify ... d.findViewById()" by a try/catch block then - might be that some android versions use different ids, and/or Linkify is not working as expected or so, i would not risk a crash on startup for that.

Hocuri:
Both done.

PR #1298, opened 2020-04-15
multi account

r10s:
well, finally, this pr brings multi account to android:

Screen Shot 2020-04-16 at 14 16 38

adding a new account brings you back to the "welcome screen" from where you can also scan a new account :)

in my tests, this works out pretty nice, however, it is an "exeperimental feature", EDIT: you find it at  "Settings / Advanced / Experimental features / Switch account" - if you have more than one account, you can also long-press the "+" icon ;)

### DIFF
--- a/src/org/thoughtcrime/securesms/connect/AccountManager.java
+++ b/src/org/thoughtcrime/securesms/connect/AccountManager.java
@@ -126,11 +126,11 @@ public class AccountManager {
 
     public void switchAccount(Context context, Account account) {
         SharedPreferences sharedPreferences = PreferenceManager.getDefaultSharedPreferences(context);
-        String oldDbName = sharedPreferences.getString("curr_account_db_name", DEFAULT_DB_NAME);
+        String prevDbName = sharedPreferences.getString("curr_account_db_name", DEFAULT_DB_NAME);
 
         // we remember prev_account_db_name in case the account is not configured
         // and the user want to rollback to the working account
-        sharedPreferences.edit().putString("prev_account_db_name", oldDbName).apply();
+        sharedPreferences.edit().putString("prev_account_db_name", prevDbName).apply();
         sharedPreferences.edit().putString("curr_account_db_name", account.dbName).apply();
         resetDcContext(context);
     }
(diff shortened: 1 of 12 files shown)


link2xt:
Once you click "switch account" there is no indication that there are other accounts already. It was not obvious for me that I can press "back", maybe also add a link/button somewhere?

r10s:
> Once you click "switch account" there is no indication that there are other accounts already.

not sure what you mean :) if you click "switch account" you get a list of accounts as shown in the right screenshot.

cyBerta:
1. If I tap on "add account" the Welcome screen opens. 2. If I then tap the back button, the whole app closes. 3. If I then reopen the app the welcome screen appears and there's no way to switch back to the account that was already set up. 
I think for 2. it would be nice to come back to the main screen. 
For 3. maybe a check might make sense that verifies if an account is already set up?

cyBerta:
interstingly 2. worked when I have already setup 2 accounts and try to create a third one

cyBerta:
When I try to delete an active account, I get an error message "Cannot delete account". I think the error message could be more specific here (Cannot delete an active account?) or the app should switch to the first remaining account in the list before an account gets deleted.

r10s:
i think, the error message already says "Cannot delete _current_ account"? however, i am already working on making this a bit nicer anyway.

cyBerta:
Having add account in the radio button list and delete account not looks a little bit unusual :)

But overall a great thing 🛸

r10s:
> 1. If I tap on "add account" the Welcome screen opens. 2. If I then tap the back button, the whole app closes. 3. If I then reopen the app the welcome screen appears and there's no way to switch back to the account that was already set up.

i can reproduce this with a fresh account, i will do a fix.

r10s:
we should not merge this, until we have figured out https://github.com/deltachat/deltachat-core-rust/issues/1402 - or define the issue as irrelevant for merging :) but we should keep it in mind.

src/org/thoughtcrime/securesms/connect/AccountManager.java
```diff
@@ -0,0 +1,293 @@
...
+        new AlertDialog.Builder(activity)
+                .setTitle(R.string.delete_account)
+                .setSingleChoiceItems(menu.toArray(new String[menu.size()]), -1, (dialog, which) -> selection[0] = which)
+                .setNegativeButton(R.string.cancel, (dialog, which) -> handleSwitchAccount(activity))
+                .setPositiveButton(R.string.ok, (dialog, which) -> {
+                    if (selection[0] >= 0 && selection[0] < accounts.size()) {
+                        AccountManager.Account account = accounts.get(selection[0]);
+                        if (account.isCurrent()) {
+                            new AlertDialog.Builder(activity)
+                                    .setMessage("To delete the currently active account, switch to another account first.")
```
cyBerta, line 272: This needs to be translated at some point!

r10s: yip, however, it is unclear if the ui stays this way.

dumblob:
Hm, does this mean I won't get notified about new messages arriving to other than the currently selected account?

Hocuri:
> Hm, does this mean I won't get notified about new messages arriving to other than the currently selected account?

Yes.

dumblob:
Thanks @Hocuri . If there are any plans to support multiple concurrently online accounts, could you point me to related discussion/issue/PR? If not, I'll make a bug report.

r10s:
@dumblob if you mean by "any plan" a concrete timetable - no.  
if if you mean by "any plan" i we have ideas about that - of course :)

in general feature requests are discussed in the forum, i just had a look, there is eg. https://support.delta.chat/t/multi-accounts-profiles-support/179

PR #1302, opened 2020-04-21
permanent foreground notification

r10s:
this pr changes the behavior of the foreground notification (not to be confused with the new-message-is-there notifications) so that it is ways shown unconditionally.

the foreground notification is an important element that avoids the system from killing the app.

in the past, we added the foreground notification when the screen turns off and removed it when the screen turns on again - the reason was that the notification does not annoy users this way.

meanwhile, however, it seems that this approach is no longer as reliable as in the past. might be that we are killed before the screen turns off - or that some handlers are not executed by some systems.

also meanwhile, even the "allow to run in background" resp. "allow battery consumption" seems to be unreliable on various systems, there are several reports eg. on stackoverflow that apps are killed anyway - but a foreground notification is not known to be killed yet - also as the api contract is about to avoid exactly that.

finally, this notification is also a good method for users to notice if the app gets killed -
so that they can blame the system then.

therefore, the annoying of the permanent notification is probably the smaller drawback compared to received messages delayed.


solidboredom:
So very cool the things might be moving! thanks for opening it here, i also think the gui option to hide it is not essential since it can be hidden though android anyways.
the most important thing is to make deltachat usable for the most users now imho. as for the notification, i think clicking on it should open the Device Message in deltachat explaining why it is there, and hinting not to disable the app when the battery optimization: "app is running in the background and eating up your batt " comes up . thanks again!

r10s:
@solidboredom do you know if the app-is-eating-your-battery comes because of foreground-notification or because of wakelocks? or both?

solidboredom:
i am not sure actually,  it has been long since i have sent this warning off, no i cannot tell. the only hint i have, is that riot(android) did not produce it (if i remember correctly). 
By the way, would my review/approval help? it seems to me i would have the privilege( somehow strangely), but i cannot compile check the code myself, and i am not familiar with the deltchats codebase anyway, so i can only visually check the code changes and weather they make sense to me (and they do), by looking at the files that affected.or should i rather leave it to some of the commiters?

csb0730:
Hi @r10s 
is there an uptime counter available into DC? This would be suitable too to determine if app is being killed.
It could be shown somewhere in extended settings.

r10s:
there is no uptime-counter afaik, maybe this would be sth, for the core, could be returned by [dc_get_info()](https://c.delta.chat/classdc__context__t.html#a2cb5251125fa02a0f997753f2fe905b1) that is shown on all os.

r10s:
@solidboredom your help and your comments are very welcome! there is also a testing group, if you are not yet there, you can be added, see https://support.delta.chat/t/who-wants-to-help-with-testing-new-releases/697

solidboredom:
@r10s and CyBerta, so glad it went so incredibly quick! that is awesome! i've mailed the tester group, let's see if i can help.

PR #1303, opened 2020-04-21
updated docker toolchain

missytake:
In https://github.com/deltachat/deltachat-core-rust/pull/1398, the rust toolchain was changed back to 2020-03-12, but the Android Dockerfile even missed the update to 2020-04-10 in between.

We realized this when the Android Nightlies stopped building.

On this branch, the build on b1.delta.chat works again.


missytake:
After this, the nightly build script needs to be changed back to follow master.

PR #1304, opened 2020-04-22
Ask the core for the selfavatar directory, fix "backup export/import loses self-avatar"

Hocuri:
Ask the core for the selfavatar directory, fix https://github.com/deltachat/deltachat-core-rust/issues/1381 backup export/import loses self-avatar

The problem was: DC-Android did not care about what the core said about the self-avatar, it just put a file with the own email address into a folder called "avatars". Of course, this folder was not backed up. Also, we had the self avatar multiple times, once in the blobs directory and once in the directory called "avatars" (probably a leftover of the Signal code).

### DIFF
--- a/src/org/thoughtcrime/securesms/profiles/AvatarHelper.java
+++ b/src/org/thoughtcrime/securesms/profiles/AvatarHelper.java
@@ -72,9 +72,8 @@ public class AvatarHelper {
     }
 
     public static File getSelfAvatarFile(@NonNull Context context, @NonNull String address) {
-        File avatarDirectory = new File(context.getFilesDir(), AVATAR_DIRECTORY);
-        avatarDirectory.mkdirs();
-        return new File(avatarDirectory, address+".jpg");
+        String dirString = DcHelper.getContext(context).getConfig(DcHelper.CONFIG_SELF_AVATAR);
+        return new File(dirString);
     }
 
     public static void setSelfAvatar(@NonNull Context context, @NonNull String address, @Nullable byte[] data) throws IOException {
--- a/src/org/thoughtcrime/securesms/profiles/AvatarHelper.java
+++ b/src/org/thoughtcrime/securesms/profiles/AvatarHelper.java
@@ -24,7 +24,7 @@ public class AvatarHelper {
             dcContext.setChatProfileImage(chatId, null);
         } else {
             try {
-                File avatar = File.createTempFile("groupavatar", "jpg", context.getCacheDir());
+                File avatar = File.createTempFile("groupavatar", ".jpg", context.getCacheDir());
                 FileOutputStream out = new FileOutputStream(avatar);
                 bitmap.compress(Bitmap.CompressFormat.JPEG, 85, out);
                 out.close();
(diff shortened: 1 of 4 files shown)


src/org/thoughtcrime/securesms/profiles/AvatarHelper.java
```diff
@@ -61,35 +61,29 @@ public static void setGroupAvatar(Context context, int chatId, Bitmap bitmap) {
         }
     }
 
-    public static InputStream getInputStreamFor(@NonNull Context context, @NonNull String address)
+    public static InputStream getInputStreamFor(@NonNull Context context)
```
Hocuri, line 64: What cought my eye was this method here. It is called `getInputStreamFor`, which sounds as if it could also be used to get a "foreign" avatar, like `getInputStreamFor(context, "bob@testrun.org").` (if `self` is Alice). Of course, before this PR, this worked just fine because `getSelfAvatarFile(context, "bob@testrun.org") returned the avatar of Bob, not the self-avatar. 

I looked around both usages actually meant to get the self-avatar, so no risk here. I think that it should be inlined, however, because this method does not do anything actually useful and the name is very confusing.

r10s:
thanks for diving into that.

the core should make a copy when dc_set_config("selfavatar", file) resp. dcContext.setConfig("selfavatar", file) is called, i just checked that, the copy is actually made in all cases i've checked, see eg. https://github.com/deltachat/deltachat-core-rust/pull/1419

so i am not sure, if this is really related.

src/org/thoughtcrime/securesms/profiles/AvatarHelper.java
```diff
@@ -1,95 +1,58 @@
...
-        String avatarPath = dcContext.getChat(chatId).getProfileImage();
-        if (avatarPath != null && !avatarPath.isEmpty()) {
-            File oldImage = new File(avatarPath);
-            if (oldImage.exists()) {
+
+        if (bitmap == null) {
+            dcContext.setChatProfileImage(chatId, null);
+        } else {
+            try {
+                File avatar = File.createTempFile("groupavatar", "jpg", context.getCacheDir());
```
r10s, line 27: the extension needs to be `.jpg` - otherwise, at least the control showing it has problems.

src/org/thoughtcrime/securesms/profiles/AvatarHelper.java
```diff
@@ -1,95 +1,58 @@
...
-    }
-
-    public static void setSelfAvatar(@NonNull Context context, @NonNull String address, @Nullable byte[] data) throws IOException {
-        File avatar = getSelfAvatarFile(context, address);
+    public static void setSelfAvatar(@NonNull Context context, @Nullable byte[] data) throws IOException {
         if (data == null) {
-            avatar.delete();
             DcHelper.set(context, DcHelper.CONFIG_SELF_AVATAR, null);
         } else {
+            File avatar = File.createTempFile("selfavatar", "jpg", context.getCacheDir());
```
r10s, line 49: same here: the extension needs to be `.jpg` - otherwise, at least the control showing it has problems.

PR #1310, opened 2020-04-24
stop threads when switching between mupltiple accounts

r10s:
this pr stops threads and call close() on switching account.

the imap-/mvbox-/sentbox-/smtp-threads needs to be terminated before dc_close() can be called successfully.

if this is not done, the threads are hanging around, and, even worse, if an account is opened _again_, there are now two threads working on the database which lead to duplicated messages.

note, that this issue is only related to duplicated messages if ever an account was selected, which, however, might be the same as the current.

fixes https://github.com/deltachat/deltachat-core-rust/issues/1404

EDIT: the last few commits move the account switching to an async task

### DIFF
--- a/src/org/thoughtcrime/securesms/connect/ApplicationDcContext.java
+++ b/src/org/thoughtcrime/securesms/connect/ApplicationDcContext.java
@@ -404,21 +404,26 @@ public class ApplicationDcContext extends DcContext {
   }
 
   public void stopThreads() {
-    Log.i(TAG, "!!!!!!!!!!!! Stopping threads ...");
     run = false;
     synchronized (threadsCritical) {
-      if (imapThread!=null) { interruptImapIdle(); }
-      if (mvboxThread!=null) { interruptMvboxIdle(); }
-      if (sentboxThread!=null) { interruptSentboxIdle(); }
-      if (smtpThread!=null) { interruptSmtpIdle(); }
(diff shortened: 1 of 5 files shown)


src/org/thoughtcrime/securesms/connect/ApplicationDcContext.java
```diff
@@ -397,6 +403,27 @@ public void waitForThreadsExecutedOnce() {
...
+      if (sentboxThread!=null) { interruptSentboxIdle(); }
+      if (smtpThread!=null) { interruptSmtpIdle(); }
+      while (true) {
+        if ( (imapThread==null || !imapThread.isAlive())
+          && (mvboxThread==null || !mvboxThread.isAlive())
+          && (sentboxThread==null || !sentboxThread.isAlive())
+          && (smtpThread==null || !smtpThread.isAlive())) {
+          break;
+        }
+        Util.sleep(100);
```
link2xt, line 421: It is probably better to keep interrupting all alive threads in this loop until they stop. Interrupt may not work the first time due to https://github.com/deltachat/deltachat-core-rust/issues/925.

r10s: good hint, great! i do a commit targeting that. i also have one other issue found ...

PR #1319, opened 2020-04-26
Fix #1208 Back button on the Unlock screen allows to read the chat list - second try

### LINKED ISSUE #1208: Back button on the Unlock screen allows to read the chat list

Hocuri:
Ok. 

In order to actually fix the original issue, I moved the screen locking from onStart() to onResume(). The problem had been that applyScreenLock uses startActivityForResult() to show the screen lock. Therefore, isWaitingForResult was set to true and shouldLock() returned false when called from onPause(). 

As onActivityResult() sets isWaitingForResult to false, and onResume() is called after onActivityResult(), this problem does not exist anymore. (now applyScreenLock is called twice sometimes, from onActivityResult() and from onResume(), but this should be fine because as soon as one screen lock gets the right credentials, onActivityResult() is called again with RESULT_OK and the second screen lock is never shown to the user).

I also fixed the issue of the previous commit ("once the lock is disabled again in the system (via Delta Chat/Settings/App access/Change secret), all Delta Chat screen stays white unrecoverable, also after restarts."):

The problem had been that shouldLock() does call isScreenLockAvailable() in the end but isScreenLockAvailable() only checked the system version. Therefore, the screen was whitened in onCreate() because it thought that the app should be locked. Now it also checks keyguardManager.isKeyguardSecure() i.e. if there is a keyguard that could be called. To unify things, AppProtectionPreferenceFragment also calls this method now instead of checking this same thing itself.

I hope that I finally thought of everything now ;-)

And sorry for the long explanation but I do not think there is a shorter way to explain all this.

Fixes #1208


angelo-fuchs:
I can still bypass screen lock with this. But it became much harder. Will try to investigate some more.

src/org/thoughtcrime/securesms/BaseActionBarActivity.java
```diff
@@ -38,26 +38,16 @@ protected void onCreate(Bundle savedInstanceState) {
...
+    if (shouldLock()) {
+      // Already hide everything here to prevent sensible data from popping up shortly:
+      findViewById(android.R.id.content).setVisibility(View.INVISIBLE);
       isHiddenByScreenLock = true;
     }
-    super.onStop();
+  }
+
+  private boolean shouldLock() {
+    return ScreenLockUtil.isScreenLockEnabled(this) && ScreenLockUtil.getShouldLockApp() && !isWaitingForResult;
```
angelo-fuchs, line 50: This line causes the screen lock on my phone to fail and show ConversationList briefly.

The suggestion inverts the logic of "isWaitingForResult" here so that shouldLock returns true if it is already locking.

This seems to cause the lock screen to appear more then once if it failed and then got entered correctly, so this might not be the best solution.

```suggestion
    return isWaitingForResult || (ScreenLockUtil.isScreenLockEnabled(this) && ScreenLockUtil.getShouldLockApp());
```

src/org/thoughtcrime/securesms/BaseActionBarActivity.java
```diff
@@ -76,6 +66,12 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
 
   @Override
   protected void onResume() {
+    if (shouldLock()) {
+      ScreenLockUtil.applyScreenLock(this);
```
angelo-fuchs, line 70: Only here do we need to make sure not to apply the screen lock twice.
This must be an inner selection clause because the `else if` below does not check on `isWaitingForResult` again.

```suggestion
      if(!isWaitingForResult)
        ScreenLockUtil.applyScreenLock(this);
```

src/org/thoughtcrime/securesms/BaseActionBarActivity.java
```diff
@@ -38,26 +38,16 @@ protected void onCreate(Bundle savedInstanceState) {
...
       isHiddenByScreenLock = true;
     }
-    super.onStop();
+  }
+
+  private boolean shouldLock() {
+    return ScreenLockUtil.isScreenLockEnabled(this) && ScreenLockUtil.getShouldLockApp() && !isWaitingForResult;
   }
 
   @Override
```
angelo-fuchs, line 53: I would move `isWaitingForResult = false;` in `onActivityResult` inside the `if (resultCode == RESULT_OK) {` block. But I didn't actually measure any difference when doing so. Maybe this is just the race-condition paranoia of mine.

angelo-fuchs: I should note that I am aware that you did not modify `onActivityResult`.

src/org/thoughtcrime/securesms/BaseActionBarActivity.java
angelo-fuchs: Note that the suggested change below fixes the "multiple screen locks" issue, so together they worked fine on my phone.

angelo-fuchs:
Low Prio, but I also noticed that one has to enter the code if one changes the background image. If this is trivially to avoid, would be nice.

Hocuri:
> I could still bypass the lock.

Nooooo! :cry: :wink: 

> I made some suggested changes to fix that, but am not 100% certain they do not have side effects I haven't seen.

Let's see...

src/org/thoughtcrime/securesms/BaseActionBarActivity.java
Hocuri: Actually, why check for isWaitingForResult here at all? Because, `isWaitingForResult || (ScreenLockUtil.isScreenLockEnabled(this) && ScreenLockUtil.getShouldLockApp());` will return true if isWaitingForResult is true and getShouldLockApp() is false (i.e. the user did not even enable screen locking at all).

OTOH, currently this is not a problem because in onCreate() the screen is only blanked. Then in onActivityResult() isWaitingForResult is set to false anyway and in onResume() the screen is set to visible again.

You understand what I mean?

By the way, thanks for digging into this confusing issue with me!

src/org/thoughtcrime/securesms/BaseActionBarActivity.java
Hocuri: But in `onActivityResult()` `isWaitingForResult` is always set to `false`, so that this new condition will always be true! Or am I missing something?

r10s:
tbh, i think, for now, it would not throw endless time on that. i'd fix some more low-hanging fruits before.

i would also be _very_ careful in mergin in a fix that fixes some cornercases - it is so easy, that afterwards things are worse for other cases. we had that several times in the past.

src/org/thoughtcrime/securesms/BaseActionBarActivity.java
Hocuri: But we could remove `ScreenLockUtil.applyScreenLock(this);` from `onActivityResult()`. The screen is currently white anyway and in `onResume()`  `applyScreenLock()` will be called.

r10s:
otoh, yeah, maybe the latest commit fix the issue, @angelo-fuchs will find out :)

Hocuri:
Ok, I pushed one last commit with what I think are improvements to what @angelo-fuchs proposed. Seems to work fine, but this issue surprised me already multiple times.

angelo-fuchs:
With your latest change I can no longer bypass screenlock and no double-entry was required when I entered it wrong multiple times.
Good work!
As you are no longer using `isWaitingForResult` in any situation you might want to remove the variable altogether. If you do, maybe consider auto-indenting / auto-formatting the class as well.

angelo-fuchs:
@r10s This works well on my phone and I have it active to try to provoke any corner cases where it might not work, so I think we can merge it so it gets delivered to the nightlies.
I would prefer if we can have this change out productively rather sooner then later (it nothing speaks against it) as we advocate DC to be privacy centered and having this issue open feels not so cool in this regard.

Hocuri:
I agree, if there is a possibility to circumvent the screenlock, then
the screenlock has no point at all.

r10s:
in fact, i could not bypass the screenlock now. but unfortunately, this comes at cost of a flashing between activity changes, which is not really nice. i did not checked the code for side-effects yet.

i am really wondering, how signal solves that issue.

angelo-fuchs:
The flashing seems to be there in the old code as well, just so much shorter that its hard to see.
I would merge this and then figure out where its from (I think its the animation).

r10s:
we tentatively decided to drop support of this lock-screen at all, see recent discussion on Delta Chat Dev and in real life.

angelo-fuchs:
Closing this PR as the solution will be to remove the feature completly.

PR #1322, opened 2020-04-28
remove dead code

r10s:


r10s:
we should leave the code for now, might become handy.

PR #1323, opened 2020-04-28
show permanent notification unconditionally and add a switch to override this

r10s:
successor of https://github.com/deltachat/deltachat-android/pull/1302 - in fact, there were still some conditions in the code left.

now, the permantent-notification is really always shown, however, we also add a switch to override this. the wording of the switch, however, makes pretty clear that this is not encouraged:

### DIFF
--- a/src/org/thoughtcrime/securesms/connect/KeepAliveService.java
+++ b/src/org/thoughtcrime/securesms/connect/KeepAliveService.java
@@ -10,32 +10,29 @@ import android.content.Context;
 import android.content.Intent;
 import android.os.Build;
 import android.os.IBinder;
-import android.os.PowerManager;
 import androidx.core.app.NotificationCompat;
 import android.util.Log;
 
 import org.thoughtcrime.securesms.ConversationListActivity;
 import org.thoughtcrime.securesms.R;
+import org.thoughtcrime.securesms.util.Prefs;
 
 public class KeepAliveService extends Service {
 
     static KeepAliveService s_this = null;
 
+    public static void maybeStartSelf(Context context) {
+        // note, that unfortunately, the check for isIgnoringBatteryOptimizations() is not sufficient,
+        // this checks only stock-android settings, several os have additional "optimizers" that ignore this setting.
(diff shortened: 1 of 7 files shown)


src/org/thoughtcrime/securesms/connect/KeepAliveService.java
```diff
@@ -10,32 +10,29 @@
...
-                    return; // fine, the user has disabled the battery optimisations for us
-                }
-            }
-            else {
-                return; // android <= lollipop does not have a doze mode
-            }
-
-            // if we're here, we're on a system with doze mode and battery optimisations enabled.
-            // add foreground notification to stay alive.
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
cyBerta, line 36: The Keep-Alive Service gets now always started even on lollipop devices and earlier. I think this is not necessary.

src/org/thoughtcrime/securesms/preferences/NotificationsPreferenceFragment.java
```diff
@@ -64,6 +66,19 @@ public void onCreate(Bundle paramBundle) {
     initializeListSummary((ListPreference) findPreference(Prefs.NOTIFICATION_PRIORITY_PREF));
 
     initializeRingtoneSummary(findPreference(Prefs.RINGTONE_PREF));
+
+
+    CheckBoxPreference reliableService =  this.findPreference("pref_reliable_service");
```
cyBerta, line 71: Same here: I think this setting is on pre-M devices not necessary

src/org/thoughtcrime/securesms/connect/KeepAliveService.java
r10s: wrt android 4+5 and foreground notification: well, i do not think, it will hurt to use it, and it is also the gist there.

not doing the notification there is again one thing more to explain, ppl may wonder why it is not there, where the switch is etc. 

so, i would prefer to do that unconditionally 

also we can just say on notification probems: make sure, the notification "Background notification enabled" is there and visible. (btw. i like to have some other wording there, but this is another topic)

PR #1325, opened 2020-04-29
use wording 'Reliability service' for the permantent notification

r10s:
this pr is about the name used for the permanent-notification switch and about unifying the wording, so that the user gets and idea that the switch belongs to the permanent-notification.

the idea is:

- have a more positive wording, sth. the user _wants_ to have

- the old "Background connection enabled" is very abstract - and also misleading as it may be understood as if the IMAP-server is working, network connected or so

- the wording is chosen so that the user gets the idea that disabling the notification has some drawbacks 

- the name of the switch using should be similar to the line displayed in the notification

screenshots:

 

i think, we should also change the icon of the permantent notification, so that there are no two delta-chat icons, maybe we should just use something very common for that, eg. the arrows typically used to indicate "transfer". we cannot have _no_ icon.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -468,9 +468,10 @@
     <string name="pref_background_default_color">Default color</string>
     <string name="pref_background_custom_image">Custom image</string>
     <string name="pref_background_custom_color">Custom color</string>
-    <!-- disabling "Reliable service" will hide a the maybe annoying permanent-notification with the drawback that new-message-notifications get potentially unreliable -->
-    <string name="pref_reliable_service">Reliable service</string>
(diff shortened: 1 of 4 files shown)


hpk42:
i think "reliability service" is a bit generic and and very hard to web-search and also we have no FAQ entry on this.  I wonder if using more technical language is all that bad like "Background Message-fetch enabled" or so.

r10s:
closing this for now, the current "Background connection enabled" is maybe not that bad, also, it is the same wording used as for other apps (at least signal)

r10s:
but the switch stays as "Reliable service", but i think this in context of "notifications" and the hint is fine

PR #1331, opened 2020-05-02
tweak device-msg that explains the permantent-notification

r10s:
this pr tweaks the device-message that explains the permantant-notification, esp. it point out that it also acts as an indicator that things may not work as expected and make the link to dontkillmyapp.com clearer.

if we can merge this is quickly, there are changes that it gets translated :)

Screen Shot 2020-05-02 at 14 02 02

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -692,7 +692,7 @@
 
     <!-- android specific strings, developers: please take care to remove strings that are no longer used! -->
     <!-- this message is shown in the device chat, starting in version 1.6, it should explain the need for the permanent-notifiation and also use the same wording as used there (see "Background connection enabled") -->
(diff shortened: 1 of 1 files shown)


Hocuri:
> If you think, this is weird - we totally agree - but many manufacturers do exactly that.

There are two grammar mistakes:
1. It's `If you think this is weird` (without comma) or `If you think that this is weird`
2. `we totally agree` is inserted, so that the sentence should make sense without. But `If you think this is weird but many manufacturers do exactly that.` is not a correct sentence.

Maybe `If you think this is weird, we totally agree - but many manufacturers do exactly that.` or `If you think, this is weird - we totally agree - have a look at dontkillmyapp.com`

Hocuri:
Also, I could not find the workaround you talked about at https://dontkillmyapp.com. Maybe I did not search long enough but users will probably not search longer that I did.

r10s:
> Also, I could not find the workaround you talked about at https://dontkillmyapp.com

if you select a manufactor on the site, you get detailed information about what you can do if _this_ manufactor kills an app, eg. https://dontkillmyapp.com/samsung#user-solution-title for samsung

PR #1346, opened 2020-05-09
Do not show the devicetalk when forwarding or sharing.

Hocuri:
Do not show the devicetalk when forwarding or sharing.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationListFragment.java
@@ -83,7 +83,6 @@ public class ConversationListFragment extends Fragment
   implements LoaderManager.LoaderCallbacks<DcChatlist>, ActionMode.Callback, ItemClickListener, DcEventCenter.DcEventDelegate
 {
   public static final String ARCHIVE = "archive";
-  public static final String FORWARDING = "for_forwarding";
 
   @SuppressWarnings("unused")
   private static final String TAG = ConversationListFragment.class.getSimpleName();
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -361,8 +361,9 @@ private void handleCreateConversation(int chatId) {
     int listflags = 0;
     if (archive) {
       listflags |= DcContext.DC_GCL_ARCHIVED_ONLY;
-    } else if (forwarding) {
-      listflags |= DcContext.DC_GCL_FOR_FORWARDING;
+    } else if (RelayUtil.isRelayingMessageContent(getActivity())) {
+      listflags |= DcContext.DC_GCL_NO_SPECIALS;
```
r10s, line 365: why not just use `DcContext.DC_GCL_FOR_FORWARDING` for both, forwarding and sharing?

r10s:
on forwarding, the device-talk is already removed by DC_GCL_FOR_FORWARDING, for sharing i would suggest to use just the same flag.

src/org/thoughtcrime/securesms/ConversationListFragment.java
Hocuri: > also, the `forwarding` argument can be probably removed when using the RelayUtil, or?

What do you mean with "The forwarding argument"? The one I already removed?

Hocuri: > why not just use DcContext.DC_GCL_FOR_FORWARDING for both, forwarding and sharing?

Yes, makes sense, I changed this.

> also, the `forwarding` argument can be probably removed when using the RelayUtil, or?

What do you mean with "The forwarding argument"? The one I already removed?

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -75,6 +76,7 @@
 
 import static org.thoughtcrime.securesms.util.RelayUtil.REQUEST_RELAY;
 import static org.thoughtcrime.securesms.util.RelayUtil.acquireRelayMessageContent;
+import static org.thoughtcrime.securesms.util.RelayUtil.isForwarding;
```
cyBerta, line 79: at a first glance this looks like an unnecessary import

Hocuri: I optimized the imports.

PR #1349, opened 2020-05-10
prep 1.8

r10s:
this pr changes the default of the permantent-notification to "off" for the reasons explained in the first commit.

the other commit streamline the ui a bit around this new default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,13 @@
 # Delta Chat Android Changelog
 
 ## v1.8.0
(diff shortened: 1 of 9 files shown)


hpk42:
probably a good plan -- but no change to device messages?  or do you want to do the device message when actually preparing the release?

r10s:
there is a device-message added, R.string.device_talk_background_connection2_android, however, only for users of 1.6

for other users, things do not really have changed (things may or may not work)

PR #1350, opened 2020-05-11
Fix #1347 Draft file name "Unknown".

### LINKED ISSUE #1347: Draft file name "Unknown"
- Android version: 9
- Device: Fairphone 3
When sharing a file to Delta Chat and then selecting a chat, "Unknown" is shown instead of the file name. What I found out so far: The filename is still there at the beginning of `AttatchmentManager.setMedia()` (in the "uri" variable) but vanishes somewhere before `documentView.setDocument((DocumentSlide) slide);` is called. Reported at https://github.com/deltachat/deltachat-android/pull/1342.

Hocuri:
Fix #1347 Draft file name "Unknown".

getManuallyCalculatedSlideInfo() in AttatchmentManager sometimes failed to get a file name. Now, in this case, it is just extracted from the uri.

### DIFF
--- a/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
+++ b/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
@@ -354,7 +354,10 @@ public class AttachmentManager {
         }
 
         if (fileName == null) {
-          fileName = new File(uri.getPath()).getName();
+          String path = uri.getPath();
+          if (path != null) {
+            fileName = new File(path).getName();
+          }
         }
 
         Log.w(TAG, "local slide with size " + mediaSize + " took " + (System.currentTimeMillis() - start) + "ms");
--- a/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
+++ b/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
@@ -60,6 +60,7 @@ import org.thoughtcrime.securesms.util.concurrent.SettableFuture;
 import org.thoughtcrime.securesms.util.guava.Optional;
 import org.thoughtcrime.securesms.util.views.Stub;
 
+import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/mms/AttachmentManager.java
```diff
@@ -352,6 +353,10 @@ public void onReceivedDuration(int millis) {
           height = dimens.second;
         }
 
+        if (fileName == null) {
+          fileName = new File(uri.getPath()).getName();
```
r10s, line 357: `uri.getPath()` may return null, in this case, `new File()` will throw an error.

i suggest to add a try/catch block around this statement.

Hocuri: Ough, you're right. I'm not used to looking for this kind of problem anymore after programming Rust and Kotlin a lot :wink:

Does a try/catch block have any advantage over just checking for null?

Hocuri:
> @Hocuri so, you could you reproduce #1347 in general?

Yes.

src/org/thoughtcrime/securesms/mms/AttachmentManager.java
r10s: in this case, checking for `null` would probably do the job - esp. as we just checked the possible return values.

but one is not always that accurate, called function may be adapted, implementations can change - and `getPath()` and `getName()` might also throw an error in the future. what, if getPath() returns a path `new File()` cannot handle?

so, assuming, that there are many things that can go wrong in uri to file conversion, i would go for a try/catch block for exactly this added line (although i am not the biggest fan of try/catch :) 

also, try/catch on that line makes sure we do not get a regression by this fix.

Hocuri: Ok.

PR #1351, opened 2020-05-11
clarify that newer ndk versions will likely work

### LINKED ISSUE #1316: Question about NDK r21 support
I wanted to be sure if NDK r21 is also allowed to build Delta Chat for Android as "ndk-make.sh" had r20 as the minimum requirement.

r10s:
closes #1316


r10s:
thanks for reviewing, @EchedeyLR!

Issue #1354, opened 2020-05-12
Flaky PeriodicWorkRequest (?)


Hocuri:
Reportedly, sometimes no messages arrive and @r10s guessed that something might not be right with the PeriodicWorkRequest that is supposed to pull every 15 minutes.

r10s:
probably closed by #1355, otherwise reopen

PR #1355, opened 2020-05-12
Update WorkManager.

Hocuri:
This might already fix issue #1354 because recently they "Fixed a bug where if a Worker was interrupted, it would not get rescheduled correctly.", see https://developer.android.com/jetpack/androidx/releases/work#version_233_3.


r10s:
nice!

PR #1356, opened 2020-05-13
Use the mute functions of the core

### LINKED ISSUE #1338: "mute" does not work on multi-account
"mute" stores only chat-ids in the android SharedPreference object - with multi-account, this get mixed up - if a chat is muted on AccountA, the random chat with the same ID will be muted on AccountB.

if chat-muting is not used or only one account is used, things are fine.

to fix that, we should just switch to the new core apis as [dc_set_chat_mute_duration()](https://c.delta.chat/classdc__context__t.html#a6460395925d49d2053bc95224bf5ce37) or [dc_chat_is_muted()](https://c.delta.chat/classdc__chat__t.html#ae93eee9385d1d2c74b4de62619cfd954)

when switching, we should migrate the old values (for one account, multi-account can be ignored on this issue, it is experimental anyway)

Hocuri:
Fix #1338.

Muting and unmuting now takes about 0.1s while it was instantly before, but I guess it's not worth improving this. And maybe it is improved by https://github.com/deltachat/deltachat-core-rust/pull/1492.

Du you think that I should inline the Prefs.*mute*() functions?

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationListItem.java
@@ -70,7 +70,7 @@ public class ConversationListItem extends RelativeLayout
 
   private DcLot              dcSummary;
   private Set<Long>          selectedThreads;
-  private long               chatId;
+  private int                chatId;
   private int                msgId;
   private GlideRequests      glideRequests;
   private TextView           subjectView;
@@ -130,7 +130,7 @@ public class ConversationListItem extends RelativeLayout
     this.dcSummary        = dcSummary;
     this.selectedThreads  = selectedThreads;
     Recipient recipient   = thread.getRecipient();
-    this.chatId           = thread.getThreadId();
+    this.chatId           = (int) thread.getThreadId();
     this.msgId            = msgId;
     this.glideRequests    = glideRequests;
     this.unreadCount      = thread.getUnreadCount();
@@ -173,7 +173,7 @@ public class ConversationListItem extends RelativeLayout
     }
 
     fromView.setCompoundDrawablesWithIntrinsicBounds(
-        Prefs.isChatMuted(getContext(), (int) chatId)? R.drawable.ic_volume_off_grey600_18dp : 0,
+        Prefs.isChatMuted(dcContext.getChat(chatId))? R.drawable.ic_volume_off_grey600_18dp : 0,
         0,
         thread.isVerified()? R.drawable.ic_verified : 0,
         0);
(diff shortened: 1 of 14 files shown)


src/org/thoughtcrime/securesms/ConversationListItem.java
```diff
@@ -173,7 +173,7 @@ public void bind(@NonNull ThreadRecord thread,
     }
 
     fromView.setCompoundDrawablesWithIntrinsicBounds(
-        Prefs.isChatMuted(getContext(), (int) chatId)? R.drawable.ic_volume_off_grey600_18dp : 0,
+        Prefs.isChatMuted(dcContext.getChat(chatId))? R.drawable.ic_volume_off_grey600_18dp : 0,
```
r10s, line 176: this database roundtrip looks unnecessarily expensive to me, esp. as it is executed for every line in the chatlist. we have the chat-record as legacy thread-object, we can copy the state there without using the database again.

but nm, i have already a fix, did that during testing/review :)

src/org/thoughtcrime/securesms/ConversationTitleView.java
```diff
@@ -66,7 +66,7 @@ public void setTitle(@NonNull GlideRequests glideRequests, @Nullable DcChat dcCh
       setComposeTitle();
     } else {
       setRecipientTitle(dcChat, showSubtitle);
-      if (Prefs.isChatMuted(getContext(), dcChat.getId())) {
+      if (Prefs.isChatMuted(dcChat)) {
```
r10s, line 69: nice, here we save one access :)

src/org/thoughtcrime/securesms/MuteDialog.java
```diff
@@ -15,24 +15,24 @@ public static void show(final Context context, final @NonNull MuteSelectionListe
     builder.setItems(R.array.mute_durations, (dialog, which) -> {
       final long muteUntil;
 
+      // See https://c.delta.chat/classdc__context__t.html#a6460395925d49d2053bc95224bf5ce37.
       switch (which) {
-        case 0:  muteUntil = System.currentTimeMillis() + TimeUnit.HOURS.toMillis(1);    break;
-        case 1:  muteUntil = System.currentTimeMillis() + TimeUnit.HOURS.toMillis(2);    break;
-        case 2:  muteUntil = System.currentTimeMillis() + TimeUnit.DAYS.toMillis(1);     break;
-        case 3:  muteUntil = System.currentTimeMillis() + TimeUnit.DAYS.toMillis(7);     break;
-        case 4:  muteUntil = System.currentTimeMillis() + TimeUnit.DAYS.toMillis(36500); break;
```
r10s, line 23: oh - our old "forever" was 10 years only :o

src/org/thoughtcrime/securesms/ConversationListItem.java
r10s: did a thing: https://github.com/deltachat/deltachat-android/pull/1356/commits/8b19b81c5b4078e2ad54612eeb1de3d860987b06

PR #1357, opened 2020-05-13
Voice message requires foreground

### LINKED ISSUE #1022: Voice message stops when screen locks
IMO it should continue to play as the screen turns off.
What do you think?

angelo-fuchs:
This fixes #1022 by requesting the system to stay active during playback.

While at it I noticed that the AudioSlidePlayer has possibilities for race conditions and is generally badly structured. So, commits 3 and after are cleanup code.

To understand the synchronized(MONITOR) lines one has to know that synchronized on a static method and a non-static method does not synchronize on the same element.

### DIFF
--- a/src/org/thoughtcrime/securesms/audio/AudioSlidePlayer.java
+++ b/src/org/thoughtcrime/securesms/audio/AudioSlidePlayer.java
@@ -116,28 +116,29 @@ public class AudioSlidePlayer {
   }
 
   public void play(final double progress) throws IOException {
-    // TODO: synchronized (MONITOR) {
-    if (this.mediaPlayer != null) {
-      return;
+    synchronized (MONITOR) {
+      if (this.mediaPlayer != null) {
+        return;
+      }
+
+      if (slide.getUri() == null) {
+        throw new IOException("Slide has no URI!");
+      }
+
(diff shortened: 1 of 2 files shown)


angelo-fuchs:
I just noticed that when one has two players and starts them after each other, the second does not keep the screen on as the "start" for the second is called BEFORE the "stop" of the second.

src/org/thoughtcrime/securesms/audio/AudioSlidePlayer.java
```diff
@@ -99,87 +125,128 @@ public void onPlayerStateChanged(boolean playWhenReady, int playbackState) {
...
-    });
+    }
+  }
+
+  private void stopKeepingScreenOn() {
+    synchronized (MONITOR) {
+      Log.d(TAG, "stopKeepingScreenOn " + name + " list: " + CURRENTLY_PLAYING);
+      if (context instanceof Activity) { // should always be true
+        // check on string equality is valid as it should be the same instance.
+        if(CURRENTLY_PLAYING.get(CURRENTLY_PLAYING.size()-1) == name) {
```
cyBerta, line 243: string comparison: use equals() instead of ==

angelo-fuchs:
closed as the issue is being worked on with the better solution of displaying an audio player permanent notification instead.

PR #1368, opened 2020-05-17
Rework notifications

### LINKED ISSUE #999: swiping a notification away should not mark things as "noticed"

### LINKED ISSUE #1033: Sound notifications problems
A problem occurred to me, I have a Xiaomi Redmi 6A Android 9 Delta works fine, but when a new messages arrives 2 sounds plays at the same time, and when unlock phone and open Delta (was in background) another sound notification (too many for my taste) even the sounds aren't the same. sorry by my english.

### LINKED ISSUE #1034: delta chat not respecting do not disturb mode
Android 7.0 Samsung s7 
Latest delta chat from fdroid.
Delta chat still makes new mail notification sounds when in do not disturb mode.

### LINKED ISSUE #1296: notification icons appears and disappears
when there are unread messages in different chats, there are several appearing and disappearing notification icons in the system-bar.

for fixing that, we should probably move towards using notification-channels as they are supposed to.

this issue is a bit older, thought, it was filed already :)

### LINKED ISSUE #1337: clear all notifications on account switching
the used chat-ids are no longer correct

r10s:
this pr moves the notification system closer to what the system intends, it is a complete rewrite, things are not handled mostly in NotificationCenter that does things pretty straight forward.

new features:

- **create and use user-editable notification-channels** as needed if a chat has an separate sound/vibrate setting - the user can edit these channel in-app or in the system-settings (which may offer more specific options)
- unique notification-channels across **multi-account**

fixes/details:

- do not play sounds or vibrate on our own, fixes #1034, fixes #1033 and similar issues
- use setOnlyAlertOnce() to inform the system about sounds not worth playing
- cleanup unused notification channels (also all channels from previous installations :)
- notifications are now only added as they come in, no sync issues, fixes #1296
- bind notification object to context, not to app to avoid messing up with chat-ids on multi account (the channels use the chatIds, however, on incoming notifications, the names are updated)
- remove notifications on account switching, fixes #1337
- swiping a notification away does not mark messages as being read, fixes #999
- add db-name to notification-hash to allow use different channel settings on different databases 

while the inbox-style, avatars etc. are preserved, the preview of images in the notification area is removed - for simplicity, but also as it clashes with the inbox-style and is only of limited use.


src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -0,0 +1,485 @@
...
+                        int wh = context.getResources().getDimensionPixelSize(R.dimen.contact_photo_target_size);
+                        Bitmap bitmap = BitmapUtil.createFromDrawable(drawable, wh, wh);
+                        if (bitmap != null) {
+                            builder.setLargeIcon(bitmap);
+                        }
+                    }
+                } catch (Exception e) { Log.w(TAG, e); }
+            }
+
+            // add a reply-button that allows answering without opening Delta Chat.
```
Hocuri, line 398: The "Mark read" button in the notifcations is gone :disappointed:

r10s:
> When you receive a message while you are in a chat then it will not be marked as read. At least in 1.8.1 this was better (did not test on Master yet)

yip, might be i've overseen that completely, i will have a look at that, thanks for the hint!

> [...] notifications from multiple senders come as single notifications and can't be swiped away as one (IIRC this was better a long time ago but it actually has nothing to do with this PR. Maybe I should just open an issue for it.)

well, this is the way, the android system wants the notifications to be used, we cannot/shall not do much on that - otherwise we have to leave the channel track again or give up per-chat settings. each chat has its notification channel, thats it. depending on the operating system, however, there might be an option to swipe all away. with the notification group, at least, we give all informations needed for that.

res/xml/preferences_notifications.xml
```diff
@@ -65,4 +51,9 @@
                 android:entries="@array/pref_notification_priority_entries"
                 android:entryValues="@array/pref_notification_priority_values"/>
 
+        <org.thoughtcrime.securesms.components.SwitchPreferenceCompat
+            android:key="pref_key_inthread_notifications"
+            android:title="@string/pref_in_chat_sounds"
```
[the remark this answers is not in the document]
r10s: with this setting, you can enable/disable the plipp-plopp send when you are chatting, inside an open chat.  
the option is not directly related to system-notifications, but fits best here (it was also here before)

src/com/b44t/messenger/DcContext.java
```diff
@@ -84,6 +84,12 @@ protected void finalize() throws Throwable {
         contextCPtr = 0;
     }
 
+    public String getDbName() {
+        String ret = getBlobdir();
+        ret = ret.substring(ret.lastIndexOf('/') + 1);
+        return ret.replace("-blobs", "");
+    }
```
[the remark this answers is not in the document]
r10s: +1, good idea!

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -294,7 +293,7 @@ protected void reverseAction(@Nullable Void parameter) {
 
   @SuppressLint("StaticFieldLeak")
   private void handleDeleteAllSelected() {
-    final DcContext     dcContext          = DcHelper.getContext(getActivity());
+    final ApplicationDcContext dcContext = DcHelper.getContext(getActivity());
```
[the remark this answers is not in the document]
r10s: i formatted the change according to the rest of the file; changing formatting the whole file would mess up this pr.

src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -0,0 +1,485 @@
...
+                    channel.setShowBadge(true);
+
+                    if (!ledColor.equals("none")) {
+                        channel.enableLights(true);
+                        channel.setLightColor(getLedArgb(ledColor));
+                    } else {
+                        channel.enableLights(false);
+                    }
+
+                    if (!TextUtils.isEmpty(ringtone.toString())) {
```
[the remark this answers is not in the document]
r10s: good catch!

src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
r10s: yip, this is by design, it was never really correct the old way.

"markread" means we send out an mdn, this should really only happy when the whole message is at least displayed once to the user,  this is not true for an excerpt shown in the notification area. 

you can swipe the notification away if you want to dismiss it locally.

src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -0,0 +1,485 @@
...
+        return GRP_MSG;
+    }
+
+    private String getNotificationChannel(NotificationManagerCompat notificationManager, DcChat dcChat) {
+        int chatId = dcChat.getId();
+        String channelId = CH_MSG_PREFIX;
+
+        if(notificationChannelsSupported()) {
+            try {
+                // get all values we'll use as settings for the NotificationChannel
```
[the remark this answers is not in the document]
r10s: :)

res/xml/preferences_notifications.xml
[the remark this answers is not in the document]
Hocuri: So, is it possible to rename `pref_key_inthread_notifications` to `pref_key_inthread_sounds`? If not, maybe we should put a comment.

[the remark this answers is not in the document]
r10s: well, we cannot rename easily as this would break settings with existing users. but i have added a comment and also added a section to make this clearer to the user.

r10s:
thinking over the in-chat-notification: i think it is better to treat them the same as other notification, this way eg. muting a chat, a plipp-plopp does not cause surprises.
in fact, before it was weird: for muted chats, in-chat-notifications were played for outgoing messasges but not for incoming.

res/xml/preferences_notifications.xml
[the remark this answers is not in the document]
r10s: i finally made in-chat-sounds respecting global notification on/off as well as mute. an independent setting would rise the question "what is about 'mute'"? cmp https://github.com/deltachat/deltachat-android/pull/1368#issuecomment-633025655

r10s:
> When you receive a message while you are in a chat then it will not be marked as read. At least in 1.8.1 this was better (did not test on Master yet)

i just tried that out in 1.8.1 - this is not a new bug, it also did not worked before, markseen was only triggered when scrolling takes place or the chat is entered, not on incoming messages, filed #1380 for that, i'll do a fix

Hocuri:
> i just tried that out in 1.8.1 - this is not a new bug, it also did not worked before, markseen was only triggered when scrolling takes place or the chat is entered, not on incoming messages, filed #1380 for that, i'll do a fix

Ah ok, maybe I scrolled a bit when testing this.

src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
Hocuri: Maybe we could at least have a markread button when the whole message is shown in the notification?

res/xml/preferences_notifications.xml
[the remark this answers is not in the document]
Hocuri: :+1:

src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
r10s: hm, then i'd do it unconditionally. do you know if whatsapp/telegram mark things already as read one the sender side if the user hits such a button in the notification?

having the button sometimes only would introduce even more questions.

but if this is the last open question of this pr, i'd like to discuss/change this in a follow up, so that these changes can already be tested in the nightlies.

r10s: but wait: the old "Mark read" button did not really sent out read-receipts, it only removed the unread-counter, i'd say this is a things we can do, you're right.

Hocuri:
Yes, from my side thats the last open point.

r10s:
@Hocuri please do not merge yet, i'll re-add mark-seen first

src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
r10s: re-added mark-read in https://github.com/deltachat/deltachat-android/pull/1368/commits/e8a3bae004f730ee5499cca657dd5f90b0faf079

res/xml/preferences_notifications.xml
```diff
@@ -65,4 +51,10 @@
                 android:entries="@array/pref_notification_priority_entries"
                 android:entryValues="@array/pref_notification_priority_values"/>
 
+        <org.thoughtcrime.securesms.components.SwitchPreferenceCompat
```
angelo-fuchs, line 54: The send 'tok' is present even if this is "off". So while not a "notification" sound we should keep this together.

(Receive "Tok" is working as expected)

src/org/thoughtcrime/securesms/connect/ApplicationDcContext.java
```diff
@@ -578,13 +581,12 @@ public long handleEvent(final int event, long data1, long data2) {
         break;
 
       case DC_EVENT_INCOMING_MSG:
-        MessageNotifierCompat.updateNotification((int) data1, (int) data2); // updateNotification() makes sure to run in the correct thread
+        notificationCenter.addNotification((int) data1, (int) data2);
```
angelo-fuchs, line 584: I think it would be cleaner to handle the notification here as an EventObserver.
Currently we have this one event and event handler that are outside the existing Event segment. That would need a good explanation

r10s: well, we just have not "global" event handler, and i did not want to change this in this pr. apart from that, i think, it is fine - we also handle errors here, and we all the things in a file that also starts threads etc. not sure, an additional indirection helps here for better overview or so.

res/xml/preferences_notifications.xml
r10s: ahh, i though i've fixed that.

r10s: i have a fix, one second ...

Issue #1358, opened 2020-05-13
show a clear error message, if video recording is not available


r10s:
there are various reasons, that video-recording or -recoding is not working as expected:

- file to send is too large (may happen after some seconds if recoding is not available, iirc recoding is not there on android4)
- no camera, camera not working
- camera access denied
- disk full
- operating system issue, eg. we got a report from gplay from a sailfish device using Delta Chat

we probably cannot make video recording available in all these situations, but we should check what the current state for an error message is and improve things as needed.

angelo-fuchs:
Situation "Camera access denied" is handled properly.

angelo-fuchs:
Trying to share a .webm File from my disk failed.

Log entry:

    VideoRecoder: recoding for /data/user/0/com.b44t.messenger.beta/files/messenger.db-blobs/2020-05-14-14-56.webm failed: cannot get info

The message was also shown very briefly as a toast.

angelo-fuchs:
When I create a video and the after-compression size is too large this is in the log:

    AttachmentManager: local slide with size 1393650202 took 20698ms
    AttachmentManager: File too large.

The "File too large" toast is displayed very shortly and then the video is gone.
We should not just drop it, but instead give the user the option to at least store it to disk and then see if he can transmit it somehow else.

angelo-fuchs:
When the inbuilt camera app is disabled and no other app is available to provide the vid feature nothing happens when one clicks the 'video' icon.

The log contains nothing about the situation and no toast or anything is displayed. Taking photos works though, so this could lead to users who (somehow) forgot that they had suppressed video to think that "DC can't to video"

angelo-fuchs:
Attaching file near limit (17MB) does not succeed due to SMTP timeout (on my network, probably no issue on fibered location)

angelo-fuchs:
When attaching an .mp4 (of acceptable size) as "File" it is displayed as "File" until send is hit. Then it vanishes for 20 seconds (while being converted) and reappears as video.
This might not be what a user expects.

AndreaEnzoLattmann:
Is this bug still present?

PR #1370, opened 2020-05-18
bump exoplayer version to 2.9.6 becaus of some fixes that apply to us

angelo-fuchs:
Changes exoplayer Version from 2.9.1 to 2.9.6. Most current is 2.11.4, but I didn't invest time to figure out if we need any of the new features.
exoplayer Release notes contain also this line:
2.9.3 Fix regression where some audio formats were incorrectly marked as being
  unplayable due to under-reporting of platform decoder capabilities

There are also some other fixes that might be beneficial to us, even if we didn't encounter them yet.


angelo-fuchs:
This also fixes https://github.com/deltachat/deltachat-android/issues/1358#issuecomment-628618516

PR #1371, opened 2020-05-19
Fix video recording error messages

### LINKED ISSUE #1358: show a clear error message, if video recording is not available

angelo-fuchs:
This contains the changes to make sure the video recording either works properly, or displays an understandable error message and tries to enable the user to recover where possible.

I tried to make only one kind of change per commit so the modifications are clear in each step.

fixes #1358


angelo-fuchs:
I think this is as far as we can go without either writing a proper video recording library and using that or fixing the whole VideoRecorder class.
The changes should produce at least an error log that contains what went wrong. With this we can then start debugging and fixing more concrete errors.
I suggest closing #1358 after merging this and opening concrete issues for individual bugs as they occur.

src/org/thoughtcrime/securesms/video/recode/VideoRecoder.java
```diff
@@ -847,31 +862,35 @@ public static boolean prepareVideo(Context context, int chatId, DcMsg msg) {
...
+//          vei.resultVideoBitrate, vei.originalDurationMs, vei.originalAudioBytes);
 
-      if (vei.estimatedBytes > MAX_BYTES+MAX_BYTES/4) {
-        logNtoast(context, String.format("recoding for %s failed: resulting file probably too large", inPath));
-        return false;
-      }
+//      if (vei.estimatedBytes > MAX_BYTES+MAX_BYTES/4) {
+//        logNtoast(context, String.format("recoding for %s failed: resulting file probably too large", inPath));
+//        return false;
+//      }
```
r10s, line 871: why is that commented out?

iiuc, it is a rough check not even start recoding of a 1gig file.

however, vei.estimatedBytes could be a local variable - or removed if the check is really superfluous.

angelo-fuchs: This causes the video to be dropped.
So, if one makes a Video in DC about something and it was too large it was just gone permanently.
Currently it is present as a Draft and can be saved from there.
Alternative path here would be to store the raw version as a Draft instead and skip the recoding if its too large.

Hocuri:
@r10s what about this pr? You commented and @angelo-fuchs answered. Shall I review this? Or is it actually ready to merge because you only had this one question?

BTW, this needs a rebase.

angelo-fuchs:
rebasing on master didn't work out, I'll recreate the branch

adbenitez (approved):

r10s:
@angelo-fuchs iirc, you wanted to change a detail wrt handling recoding always and unconditionally. keeping a "huge" file is unexpected to the user, so we should always try to recode, even if later sending is not possible. at least this is what i remember from the discussion with @hpk42 :)

r10s:
closing stale pr

PR #1372, opened 2020-05-20
Background image no longer distorted.

### LINKED ISSUE #466: Fix aspect-ratio of custom background images

angelo-fuchs:
I mirrored the activity of the ImageView on the Background setting view (as it was working there properly) in the conversation activity.

Then I modified the transformation during Background setting to create an image that is scaled down to the larger sides of the screen (so the scaling is indifferent at this stage to portrait/landscape).

When active the users will need to set his background image once so that the old image (that was cropped too much) is removed and a new one is set.
It could be a good idea to inform the user in the system messages about that change.
fixes #466

### DIFF
--- a/src/org/thoughtcrime/securesms/preferences/ChatBackgroundActivity.java
+++ b/src/org/thoughtcrime/securesms/preferences/ChatBackgroundActivity.java
@@ -136,13 +136,15 @@ public class ChatBackgroundActivity extends PassphraseRequiredActionBarActivity 
             Display display = ServiceUtil.getWindowManager(context).getDefaultDisplay();
             Point size = new Point();
             display.getSize(size);
+            // resize so that the larger side fits the screen accurately
+            int largerSide = (size.x > size.y ? size.x : size.y);
             Bitmap scaledBitmap = GlideApp.with(context)
                     .asBitmap()
                     .load(imageUri)
-                    .centerCrop()
+                    .fitCenter()
                     .skipMemoryCache(true)
                     .diskCacheStrategy(DiskCacheStrategy.NONE)
-                    .submit(size.x, size.y)
+                    .submit(largerSide, largerSide)
                     .get();
             FileOutputStream outStream = new FileOutputStream(destinationPath);
             scaledBitmap.compress(Bitmap.CompressFormat.JPEG, 85, outStream);
--- a/res/layout/conversation_activity.xml
+++ b/res/layout/conversation_activity.xml
@@ -1,10 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
     android:layout_height="match_parent" android:layout_width="match_parent">
+
+    <ImageView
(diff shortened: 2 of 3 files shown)


src/org/thoughtcrime/securesms/preferences/ChatBackgroundActivity.java
```diff
@@ -136,13 +136,15 @@ private void scaleAndSaveImage(Context context, String destinationPath) {
             Display display = ServiceUtil.getWindowManager(context).getDefaultDisplay();
             Point size = new Point();
             display.getSize(size);
+            // resize so that the larger side fits the screen accurately
+            int largerSide = (size.x > size.y ? size.x : size.y);
```
cyBerta, line 140: the parenthesis can be removed

src/org/thoughtcrime/securesms/preferences/ChatBackgroundActivity.java
```diff
@@ -136,13 +136,15 @@ private void scaleAndSaveImage(Context context, String destinationPath) {
...
             Bitmap scaledBitmap = GlideApp.with(context)
                     .asBitmap()
                     .load(imageUri)
-                    .centerCrop()
+                    .fitCenter()
                     .skipMemoryCache(true)
                     .diskCacheStrategy(DiskCacheStrategy.NONE)
-                    .submit(size.x, size.y)
+                    .submit(largerSide, largerSide)
```
cyBerta, line 147: the aspect ratio is kept intact by .fitCenter()?

src/org/thoughtcrime/securesms/preferences/ChatBackgroundActivity.java
Hocuri: I'd actually leave it there because it improves readability

src/org/thoughtcrime/securesms/preferences/ChatBackgroundActivity.java
angelo-fuchs: @cyBerta Yes, it is. fitCenter scales the smaller side of the source to match the given dimensions and scales the larger side to match the aspect ratio.

a 100:200 landscape image on a 50:20 screen will be scaled -> 50:100.
Result on Landscape (20:50): scale (during display) down to 25:50 and cut off 5 on the edges.
Result on Portrait (50:20): no further scaling needed, display 50:20 and cut off 80.

a 200:100 portrait image on that 50:20 screen will be scaled -> 100:50.
Result on Landscape (20:50): no scaling (during display): Display 20:50 and cut off 80
Result on Portrait (50:20): scale (during display) down to 50:25 and cut off 5.

That way the image is never distorted, just cut.

src/org/thoughtcrime/securesms/preferences/ChatBackgroundActivity.java
angelo-fuchs: @cyBerta Yes, I personally like them though and would prefer to keep them.
Its clearer that way that something is done here (instead of just an assignment), but it is just a personal style preference.

r10s:
unfortunately, this pr adds another issue: when opening/closing the keyboard, the background image jumps up and down.

PR #1374, opened 2020-05-20
fix string placeholder in strings.xml

cyBerta:
fixes an issue described here https://github.com/deltachat/deltachat-ios/pull/726


r10s:
i think, this is not sufficient to fix the issue on ios, the placeholder will reappear in translations.

so, although `%s` should not be used in strings.xml and the correct format is `%1$s` there, i think we should be robust against such tiny formatting details, esp. as it is correct in printf().

so, i suggest to be more accepting in ios.

i am closing this issue, to avoid getting errors in non-english only :)

PR #1376, opened 2020-05-21
fix typo

r10s:
closes https://github.com/deltachat/deltachat-ios/issues/738 

requires a source-update on transifex after merging.


angelo-fuchs:
I'm surprised we fix a typo on iOS with a commit on Android. Otherwise lgtm.

r10s:
@angelo-fuchs we're using the same strings on all platforms. android, ios and desktop all pull from the same transifex-repo, but the upstream is typically handled here.

PR #1381, opened 2020-05-23
fix marking incoming message as being seen

### LINKED ISSUE #1380: incoming messages not marked as "seen" immediately when in chat
when inside a chat, a new message comes in, the message is only marked as "seen" when the chat is left and entered again or when some (minimal) scrolling takes place.

reason is that markseenMsgs() is only triggered in onScrolled() https://github.com/deltachat/deltachat-android/blob/86e5e16bcd3ef94a1edc36aa8f49a1686792e7c6/src/org/thoughtcrime/securesms/ConversationFragment.java#L577 - but onScrolled() seems not to be fired when a message is added to the recycler-view.

r10s:
closes #1380 

before, messages were only maked as seen after the chat was entered
or after some scrolling takes place.

if a user was inside a chat, reading the last incoming message,
that did not sent out markseen (unless, some scrolling was done,
which, honestly, is often the case)

this commit starts the markseen-bouncer also when the chatlist is reloaded
(true eg. on incoming messages) - but this is not done when the
list is paused (eg. screen is off or some other fragment/dialog above)

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -111,6 +111,7 @@ public class ConversationFragment extends Fragment
     private TextView                    noMessageTextView;
     private ApplicationDcContext        dcContext;
 
+    public boolean isPaused;
     private Debouncer markseenDebouncer;
 
     @Override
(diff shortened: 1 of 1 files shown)


r10s:
> Should I test such rather simple PRs before approving?

just having a look a the source code is often sufficient i'd say.

i tested it here :)

PR #1386, opened 2020-05-27
fix typos

r10s:


r10s:
> Are you sure you wanted to update the submodule though?

of course not, thanks!

PR #1391, opened 2020-06-03
Switching to "Default" Language behaves as it should be

### LINKED ISSUE #1209: "Default" language doesn't restore the OS-language
Hello,
I'm italian so I correctly got Delta Chat in italian when installed. Then I've switched the language from Default to English to do a bug report. 
Now, I've switched back to "Default" but Delta Chat remains in English. If I choose "Italian" it works correcly, but "Default" is always English. It seems that the App can't restore the Android native language after swhitching from Default to any other language.

/e/OS (Android Pie)
Delta Chat 1.0.3

angelo-fuchs:
Fixes #1209

To reproduce the bug this fixes the App needs to START with a Language that is not the System default. In doubt restart your phone after selecting a non-default language.

Code wise this change modifies the way the system default language is read. Currently `Locale.getDefault()` is used but this returns the locale the app was started in (and I suspect that it is the same it terminated the last time).
The new code is from here https://stackoverflow.com/a/28498119/881272

### DIFF
--- a/src/org/thoughtcrime/securesms/util/DynamicLanguage.java
+++ b/src/org/thoughtcrime/securesms/util/DynamicLanguage.java
@@ -70,14 +70,18 @@ public class DynamicLanguage {
   // Beware that Locale.getDefault() returns the locale the App was STARTED in, not the locale of the system.
   // It just happens to be the same for the majority of use cases.
   private static Locale getDefaultLocale() {
-    Locale locale;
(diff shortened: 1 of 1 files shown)


angelo-fuchs:
@r10s Nice, I prettified the exception logging and switched to used the android prebuilt config compat I found that does the same thing as I did manually before so we should be safe wrt future versions.

PR #1392, opened 2020-06-03
Allow enlarge of avatar

### LINKED ISSUE #1228: enlarge avatars
it would be great, if a click on a user avatar shows a larger view of the avatar.

see https://github.com/deltachat/deltachat-desktop/issues/1356

angelo-fuchs:
fixes #1228

Tap on the avatar in a profile to see a large version of it. Tap the large one to make it small again.

Should be discussed if the size of the overlay is proper.

I also noticed that I cannot modify a contacts avatar (as we can with their name). We might want to allow to at least intentionally delete / suppress it.

### DIFF
--- a/res/layout/profile_activity.xml
+++ b/res/layout/profile_activity.xml
@@ -44,4 +44,13 @@
             android:layout_height="match_parent"
             app:layout_behavior="@string/appbar_scrolling_view_behavior"/>
 
+    <org.thoughtcrime.securesms.components.AvatarImageView
+        android:id="@+id/profile_large"
+        android:visibility="invisible"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:layout_above="@id/pager"
+        android:layout_gravity="center"
+        android:adjustViewBounds="true"/>
+
 </androidx.coordinatorlayout.widget.CoordinatorLayout>
--- a/res/layout/profile_activity.xml
+++ b/res/layout/profile_activity.xml
@@ -51,6 +51,7 @@
         android:layout_height="wrap_content"
         android:layout_above="@id/pager"
         android:layout_gravity="center"
+        android:padding="8dp"
         android:adjustViewBounds="true"/>
 
 </androidx.coordinatorlayout.widget.CoordinatorLayout>
--- a/res/layout/profile_activity.xml
+++ b/res/layout/profile_activity.xml
@@ -44,14 +44,4 @@
             android:layout_height="match_parent"
             app:layout_behavior="@string/appbar_scrolling_view_behavior"/>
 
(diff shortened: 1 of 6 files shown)


Hocuri:
Zooming would be nicer, I'm pretty late with this idea but we could try using the media viewing activity to view avatars.

angelo-fuchs:
@r10s Now this is always converted to JPEG, which should decrease image size reliably. We might want to change the core code at some point to always convert to JPEG during rescaling no matter what is delivered (non-urgent, no issue).

angelo-fuchs:
@Hocuri What would be the advantage? We could put Avatars into the Gallery (treating them like attachments). Is that what you mean?

Hocuri:
No, I mean reusing the activity because we already have zooming and so on implemented there.

angelo-fuchs:
@Hocuri Like this?
cc @r10s Implemented reuse of MediaPreviewActivity

I personally liked the UX of the circle overlay better. But this is certainly the cleaner implementation.

Hocuri:
I couldnt test it:
```
2020-06-05 21:18:59.844 14575-14575/com.b44t.messenger.beta E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.b44t.messenger.beta, PID: 14575
    android.util.AndroidRuntimeException: Calling startActivity() from outside of an Activity  context requires the FLAG_ACTIVITY_NEW_TASK flag. Is this really what you want?
        at android.app.ContextImpl.startActivity(ContextImpl.java:912)
        at android.app.ContextImpl.startActivity(ContextImpl.java:888)
        at org.thoughtcrime.securesms.ProfileActivity.onEnlargeAvatar(ProfileActivity.java:409)
        at org.thoughtcrime.securesms.ProfileActivity.lambda$onCreate$1$ProfileActivity(ProfileActivity.java:101)
        at org.thoughtcrime.securesms.-$$Lambda$ProfileActivity$emAJaoTWhwoxEEMJOVr-rqhCFwo.onClick(Unknown Source:2)
        at android.view.View.performClick(View.java:6597)
        at android.view.View.performClickInternal(View.java:6574)
        at android.view.View.access$3100(View.java:778)
        at android.view.View$PerformClick.run(View.java:25889)
        at android.os.Handler.handleCallback(Handler.java:873)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6746)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)
```
But generally, yes.

PR #1395, opened 2020-06-05
Implement manual blurring

### LINKED ISSUE #1394: Update signal based image editing codebase
Signal just released a nice face blurring feature 🚀  (https://signal.org/blog/blur-tools/). Since we didn't diverge from that part of signals codebase (iirc), it should be easy to reuse it for DeltaChat.

cyBerta:
... adapting signal's new face detection / blurring feature

fixes #1394 

- all code related to face detection has been removed with commit 2caad455c001260a112459c1f4287e33a1c2aad1. If we ever want to have it we can revert that commit or use it at least as an inspiration.


angelo-fuchs:
fixes #1394 

Sounds like a useful feature to have.

r10s:
@cyBerta this does not compile for me. as also build.gradle is not modified, seems as if you forgot to push the latest commits :)

Issue #1390, opened 2020-06-03
check if maybeNetwork is called when the app comes to foreground


r10s:
check if maybeNetwork is called when the app comes to foreground, to get some general stability, see https://github.com/deltachat/deltachat-ios/pull/762/files

PR #1398, opened 2020-06-06
call maybeNetwork() when app comes to foreground to catch up the state

r10s:
closes https://github.com/deltachat/deltachat-android/issues/1390

### DIFF
--- a/src/org/thoughtcrime/securesms/connect/ForegroundDetector.java
+++ b/src/org/thoughtcrime/securesms/connect/ForegroundDetector.java
@@ -13,7 +13,7 @@ public class ForegroundDetector implements Application.ActivityLifecycleCallback
 
     private int refs = 0;
     private static ForegroundDetector Instance = null;
-    ApplicationContext application;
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/connect/ForegroundDetector.java
```diff
@@ -35,9 +35,16 @@ public boolean isBackground() {
 
     @Override
     public void onActivityStarted(Activity activity) {
+        if (refs == 0) {
+            application.dcContext.maybeStartIo();
+            if (application.dcContext.isNetworkConnected()) {
+                new Thread(() -> {
+                    application.dcContext.maybeNetwork();
+                });
```
Hocuri, line 43:
```suggestion
                new Thread(() -> application.dcContext.maybeNetwork()).start();
```
You also need to start the thread

r10s: good point!

PR #1401, opened 2020-06-07
Avatars again

Hocuri:
Change the enlarged avatar name to the chat name

Do not enlarge avatar if there is none (just a big letter in a circle)


Hocuri:
> the title has to be set to dcContact.getDisplayName() in case there is no chat_id present (the profile is used for both, chats and contacts without chat (the latter can be accessed eg. from new-chat-activity))

Done.

PR #1406, opened 2020-06-10
remove 'Delete emails from server' experiment

r10s:
this pr removes the 'Delete emails from server' experiment, the function is superseded by 'Auto-delete messages'.

for 2 months or so we had both functions, however, support-wise it would be cool if we can drop the experiment.

once merged/accepted, we should also drop the corresponding core/ios/~~desktop~~ parts. EDIT: desktop never used this experiment :)

cc @adbenitez


adbenitez:
:+1: IIRC this didn't worked if users disable INBOX/DeltaChat watching, so it was "buggy" and with the new auto-delete the old experimental option seems unnecessary

hpk42:
note that my question about autodelete-speed is not critical 
as it only affects accounts with lots of messages.  And once
the run is through autodelete-imap is less of a burden.  I do
wonder though why autodelete-imap does not appear to 
delete multiple messages in one imap command.

r10s:
yip, i also think, imap-deletion can be optimized.

i just had a closer look - one reason for the one-job-per-message-to-delete and why a bulk-deletion is maybe not super-simple, is the server-uid thing. as we do not save the uidvalidity per-record, there is some risk, that the server-uid points to a wrong message. to avoid deleting the wrong message, we check rfc-message-id before deletion.

we could either also check the rfc-message-id in a bulk-request - or, track the uidvalidity per message-record.

hpk42:
On Thu, Jun 11, 2020 at 02:49 -0700, bjoern wrote:
> yip, i also think, imap-deletion can be optimized.
> 
> i just had a closer look - one reason for the one-job-per-message-to-delete and why a bulk-deletion is maybe not super-simple, is the server-uid thing. as we do not save the uidvalidity per-record, there is some risk, that the server-uid points to a wrong message. to avoid deleting the wrong message, we check rfc-message-id before deletion.
> 
> we could either also check the rfc-message-id in a bulk-request - or, track the uidvalidity per message-record.

yes -- bulk receive is done for prefetch and receive-imf  so shouldn't
be too hard to do for autodelete-imap.

PR #1407, opened 2020-06-10
Nicer background

Hocuri:
https://support.delta.chat/t/new-chat-background/226
Why did we still have the old background?
For the dark background I used GIMP's "Invert value".


r10s:
> Why did we still have the old background?

well, i think, it is not _that_ clear, that it is an overall improvement - esp. i think the contrast to the white bubbles is a bit too low and i personally also like the tint of the old background more and more fitting to the title bar.

on the other hand, we have this background since some time on desktop and afaik, there were no complaints.  
other argument pro-new-background is that we want to be a bit consistent between the systems. also, maybe it is just time for sth. new :)

PR #1410, opened 2020-06-11
United colors of emoji

### LINKED ISSUE #1264: diversify emojis
Currently DC android supports only standard yellowish emojis whereas DC Desktop and iOS support emojis with different skin colors. 
We should adapt that on Android as well, in order to avoid the loss of content while messaging between the different platforms.
 I'm pretty sure, we can borrow some code from signal for an fast and easy solution.

cyBerta:
* updates the signal based sources for the keyboard anf input panel 

closes #1264


cyBerta:
>     * there is a new flickering when switching from keyboard to emoji-picker, probably because there is a redraw between keyboard-hiding and emoji-picker-showing, so that for a small moment, they are stacked over each other, the emoji-picker is visible in the upper half of the screen for 1/3rd of a second or so.

I see this is an issue, but the flickering is also visible on current master. Anyways, I'm digging into it

r10s:
> I see this is an issue, but the flickering is also visible on current master. Anyways, I'm digging into it

thanks for having a closer look.

for me, it is only there with this pr. maybe some timing issue, might be the new emoji-picker takes some ms more to get ready and an redraw is done in the state between show-emoji and hide-keyboard.

cyBerta:
I've fixed both issues. Signal has made an fix for the flickering issue that I included. 
The recent emoji issue is still there in Signal, but fixed now in Delta Chat.

r10s:
> The recent emoji issue is still there in Signal, but fixed now in Delta Chat

if this is the unchanged signal-code, maybe we should also fix that in upstream.

PR #1411, opened 2020-06-12
streamline rough logging on app-foreground etc.

r10s:
this should help on finding out what is going on on background network-reconnects etc. and improve reliability when app comes to foreground.

### DIFF
--- a/src/org/thoughtcrime/securesms/ApplicationContext.java
+++ b/src/org/thoughtcrime/securesms/ApplicationContext.java
@@ -13,6 +13,8 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
+import android.util.Log;
+
 import androidx.annotation.NonNull;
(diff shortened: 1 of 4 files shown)


Hocuri:
A problem with calling maybeNetwork more often I just came up with: Dc just tries to execute all jobs (see job::load_next()) and if we do not have network all will fail. This means that job::perform_job() will increase job.tries for all jobs and if we do this 17 times all jobs will be deleted. ("{} thread removes job {} as it exhausted {} retries") 

Am I wrong here? Otherwise this might be a problem independently of this pr (although this pr would make it worse because opening DC 17 times without network will delete all jobs).

cc @dignifiedquire

r10s:
oh, i was not aware that maybeNetwork() affects the retry cycle. in this case, indeed, we should not call it that often.

r10s:
@Hocuri @dignifiedquire i removed the unconditional maybeNetwork() call and tried to make the doc on that a bit clearer, https://github.com/deltachat/deltachat-core-rust/pull/1620/files

PR #1420, opened 2020-06-16
fix jumping background image on keyboard appearing / disappearing

### LINKED ISSUE #1417: wallpaper jumps up and down
this issue was introduced with the attempt to fix https://github.com/deltachat/deltachat-android/issues/466 in https://github.com/deltachat/deltachat-android/pull/1372

to reproduce: 
- open a chat,then the keyboard
- the wallpaper moves up, probably by the height of the keyboard, not sure though
- this is esp. ugly when a photo or so is selected as a wallpaper

the jumping also takes place when switching between keyboard- and emoji-picker.

~~i'd say, if in doubt, i would revert https://github.com/deltachat/deltachat-android/pull/1372~~ hm, by the new background image this would led to distortions that are very obvious and visible by default, i'd say we should fix that properly. the issue is there for years, i'd really like to get that done :)

cyBerta:
fixes #1417

For now, I just rotated the default background image for landscape, because the background pattern is so small that it makes almost no difference: https://github.com/deltachat/deltachat-android/blob/93ea91980c8e00c7ce4e3f3690baaa7af3b55b1a/res/drawable-land/background_hd.jpg
Nontheless we should add a proper default background image for landscape.

Custom images will be resetted after installing the PR. The reason for this kind of 'migration' is that the aspect ratio of old images is inadequate for the new approach. Also, we're now saving a specific image for landscape orientations when selecting a custom background.


r10s:
i do not like that the pref return different values depending on some external state (phone orientation). 

i was already playing around and i have prepared some additional commits, will push them tomorrow, too late now :)

r10s:
closing this for now, i did a quick-fix-alternative in #1423 

the ideas and findings of this pr can be recycled in a "final fix", this is very worthful, thanks @cyBerta 

my idea is to do a BackgroundImage view, that does the work:
- look at self how large it is
- crop and scale a passed image
- also images from the resource should be processed (this pr also shows them in a bad aspect ratio)
- all that should be done "on the fly", so that from preference point of view, it is a simple url
- no migration should be needed - ppl may have things as their background that are no longer there as  photo

all not that complicated but some work, and this should not block releases.

PR #1429, opened 2020-06-18
background image scaling good without jumping

### LINKED ISSUE #466: Fix aspect-ratio of custom background images

angelo-fuchs:
fixes #466 

When finished needs to:
1. Keep aspect ratio of background image in landscape and portrait orientation in the chat view
2. Fill screen with image (no empty borders)
3. Maintain position when the keyboard is opened.


angelo-fuchs:
I don't particularly like this solution as it is very forced.
I'm uncertain if it works as expected on large landscape screens (e.g. Pads) as I don't have any to test it on.

angelo-fuchs:
This introduced a bug in the Background picker.

r10s:
wondering, how whatsapp does that - but that is closed source. signal does not have a background-image, telegram does weird things all the time,  so wondering if there is some source we can peek at :) conversations?

angelo-fuchs:
@r10s I didn't find any workable code (I didn't search too hard, but spend about an hour searching for examples).
This solution at least works and doesn't do anything surprising.
Just take the image, rescale/crop it once per size and display it when that size is needed.

It could be better, but I don't see how without pouring even more effort into it.

src/org/thoughtcrime/securesms/components/ScaleStableImageView.java
```diff
@@ -0,0 +1,181 @@
...
+        if (orientation == ORIENTATION_PORTRAIT) {
+            portrait = true;
+        } else if (orientation == ORIENTATION_LANDSCAPE) {
+            portrait = false;
+        } else {
+            Log.i(TAG, "orientation was: " + orientation);
+            return; // something fishy happened.
+        }
+
+        measureViewSize(width, height, oldWidth, oldHeight, portrait);
```
cyBerta, line 89: This line can be moved below the early return block (`if (!(defaultDrawable instanceof BitmapDrawable)) {`...)

angelo-fuchs: Agreed, I changed the code.

angelo-fuchs:
I rebased the changes on master and tested again. Works good.

r10s:
i think, sth. went wrong with rebasing, for me, it shows >100 commits now and >1300 changed lines

angelo-fuchs:
Seems like the rebase went wrong, I will manually rebase from master again and open a new PR for it.

PR #1446, opened 2020-06-24
jni: use dc_array_get_id instead of dc_array_get_raw

link2xt:
dc_array_get_raw API prevents Rust code from storing IDs in other format
than an array of uint32_t.

iOS code and node bindings (see `dc_array_to_js_array`) already use `dc_array_get_id` loops, Android client is the only user of this API.


link2xt:
A PR that removes `dc_array_get_raw` from the core: https://github.com/deltachat/deltachat-core-rust/pull/1670

PR #1450, opened 2020-06-25
add link to blog-entry to device-message

r10s:
i think, it makes some sense to link from the release-message to the blog-entry - however, this should only be merged when the https://github.com/deltachat/deltachat-pages/pull/354 is merged (all these dependencies ...)

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -93,7 +93,7 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
     // it is not needed to keep all past update messages, however, when deleted, also the strings should be deleted.
     DcMsg msg = new DcMsg(dcContext, DcMsg.DC_MSG_TEXT);
     msg.setText(getString(R.string.update_1_10_android) + "\n\n" + "https://delta.chat/en/2020-06-24-releases");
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/ConversationListActivity.java
```diff
@@ -92,8 +92,8 @@ protected void onCreate(Bundle icicle, boolean ready) {
     // update messages - for new messages, do not reuse or modify strings but create new ones.
     // it is not needed to keep all past update messages, however, when deleted, also the strings should be deleted.
     DcMsg msg = new DcMsg(dcContext, DcMsg.DC_MSG_TEXT);
-    msg.setText(getString(R.string.update_1_10_android));
-    dcContext.addDeviceMsg("update_1_10b_android", msg); // addDeviceMessage() makes sure, messages with the same id are not added twice
+    msg.setText(getString(R.string.update_1_10_android) + "\n\n" + "https://delta.chat/en/2020-06-24-releases");
+    dcContext.addDeviceMsg("update_1_10d_android", msg); // addDeviceMessage() makes sure, messages with the same id are not added twice
```
link2xt, line 96: As a result of this, those who have installed testing versions will have two messages now? Not a problem for testers, but maybe just keep old message for them.

r10s: yip this change from 'b' to 'd' is not really needed.

PR #1341, opened 2020-05-06
Relay to multiple contacts, rework message relaying

Hocuri:
Fix #1340.

Also, I did extensive testing of message relaying:
- I. share and II. forward
- a. text
  b. text and one image
  c. 2 images
  d. 1 file (unknown type)
  e. 2 files (unknown type)
- to:
  1. 1 chat (just tap on it)
  2. If applicable: go back and forth a bit to test drafts
  3. 2 chats (multi-relaying)

and some of the bugs I found and fixed were probably already there before, although I did not test this.

After fixing things, I went through this list again and everything worked.

Seems like at some places in the UI the chatId is saved as a long, which seems not to make much sense because the core saves them as int. Anyway, I had to cast them.

PR #1475, opened 2020-07-06
Fix mailto links

### LINKED ISSUE #1472: mailto doesn't work
- Operating System (Linux/Mac/Windows/iOS/Android): Android 10
- Delta Chat Version: 1.10.4 (2020-07-01 Nightly)
- Expected behavior: When I click on a "mailto" link, Delta Chat opens with the parameters body ect.
- Actual behavior: It opens Delta Chat and the email to chat with, but the text in the body is not inserted.
- Steps to reproduce the problem: open a link with example: mailto:test@testrun.org?body=Hi

in stable version 1.10.4 it works.  The bug appears to have been introduced with this commit: 
a2dc69e27ea185566e42521c6b0290c0fdae5150
 Merge: b3389ff1 23decbe2
 Author: cyBerta
 Dates: Tue Jun 30 23:03:37 2020 +0200

Merge pull request # 1341 from deltachat / fix-sharing-multiple Relay to multiple contacts, rework message relaying.

This information was given to me by adb and I thank him for finding this bug.

cyBerta:
fixes #1472 
I found some more issues on current master, partly introduced by #1341 

* when forwarding messages the Activity stack order was broken, multiple Activity List views could appear ( fixed by
20189fd )
* when forwarding is finished, the user always returns to the original chat a it was intended before #1341
* when sharing images, they couldn't kept as drafts (now fixed by 69b587c ) 
* 69b587c also now disallows to go the following path: share menu -> conversation list -> chat a -> back to conversation list -> back b, it caused troubles in the pre #1341 implementation while the fix in #1341 introduced the above mentioned draft bug. Instead the user can only jump into one chat or select mulitple chats from the chat list by long press on the list conversation items, going back will result in returning to the share menu

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -194,6 +194,8 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
   private boolean    isSecurityInitialized    = false;
   private boolean    isShareDraftInitialized  = false;
 
+  private boolean successfulForwardingAttempt = false;
+
 
   private final DynamicTheme       dynamicTheme    = new DynamicTheme();
   private final DynamicLanguage    dynamicLanguage = new DynamicLanguage();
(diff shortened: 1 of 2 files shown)


r10s:
great, and a good description of the issues :)

PR #1479, opened 2020-07-08
Rename "Auto-delete messages" into "Delete old messages"

link2xt:
This change makes it clear which messages are deleted, so this setting
can't be confused with Ephemeral-Timer header, configured per chat.

2_scale
1_scale

### DIFF
--- a/res/values/arrays.xml
+++ b/res/values/arrays.xml
@@ -146,7 +146,7 @@
   </string-array>
 
   <string-array name="autodel_device_durations">
-      <item>@string/off</item>
+      <item>@string/never</item>
       <item>@string/autodel_after_1_hour</item>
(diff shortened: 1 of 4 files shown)


r10s:
i would be fine with that, however, current discussion is on irc

hpk42:
TBH i'd prefer to stick with auto-delete for now. I aim to write an FAQ section on ephemerality and/or a blog post with screen shots -- which can then be used for the release.

link2xt:
@hpk42 proposed "trim old messages" or "housekeeping" on IRC.

I think it is to make it more clear that deletion happens not once, but continuously? Maybe change the "off" (on the first screenshot, not on the second one where only on/off are present) to "never"?

PR #1483, opened 2020-07-13
background image scaling good without jumping (2)

### LINKED ISSUE #466: Fix aspect-ratio of custom background images

angelo-fuchs:
continues where #1429 left off due to rebase issues.

fixes #466

When finished needs to:

 -   Keep aspect ratio of background image in landscape and portrait orientation in the chat view
 -   Fill screen with image (no empty borders)
 -   Maintain position when the keyboard is opened.

@cyBerta Your suggestion to move the `measureViewSize` call after the early return is included.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -175,6 +175,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
   private   InputAwareLayout            container;
   private   View                        composePanel;
   protected Stub<ReminderView>          reminderView;
+  private   ImageView                   backgroundView;
 
   private   AttachmentTypeSelector attachmentTypeSelector;
   private   AttachmentManager      attachmentManager;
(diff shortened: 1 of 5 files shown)


src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -869,16 +874,17 @@ private void initializeViews() {
...
     String backgroundImagePath = Prefs.getBackgroundImagePath(this);
+    Drawable background;
     if(!backgroundImagePath.isEmpty()) {
-      Drawable image = Drawable.createFromPath(backgroundImagePath);
-      getWindow().setBackgroundDrawable(image);
+      background = Drawable.createFromPath(backgroundImagePath);
     }
-    else if(dynamicTheme.isDarkTheme(this)) {
-      getWindow().setBackgroundDrawableResource(R.drawable.background_hd_dark);
+    else if(DynamicTheme.isDarkTheme(this)) {
```
Hocuri, line 881: :+1: btw

src/org/thoughtcrime/securesms/components/ScaleStableImageView.java
```diff
@@ -0,0 +1,181 @@
...
+import static android.content.res.Configuration.ORIENTATION_PORTRAIT;
+import static android.content.res.Configuration.ORIENTATION_LANDSCAPE;
+
+/**
+ * An image view that maintains the size of the drawable provided.
+ * That means: when you set an image drawable it will be scaled to fit the screen once.
+ * If you rotate the screen it will be rescaled to fit.
+ * If you crop the screen (e.g. because the soft keyboard is displayed) the image is cropped instead.
+ *
+ * @author Angelo Fuchs
```
r10s, line 31: in general, we do not use the @author tag, this is hard to maintain. the git history should be sufficient.

r10s:
looking forward to try that out finally, thanks a lot @angelo-fuchs for your restless work in this issue!

angelo-fuchs:
@r10s I tried to top align the image but did not succeed. It was then squished when the keyboard was shown.

r10s:
>  It was then squished when the keyboard was shown.

k, i assumed sth. like that :) thanks for merging!

Issue #142, opened 2017-07-31
Emoticons do not work on Android 4.0.3


kollokollo:
![b](https://user-images.githubusercontent.com/15250834/28783913-1de3a498-7612-11e7-9930-f2fc2e2aa90c.jpg)

Emoticons are not displayed on my phone (Android 4.0.3). 
Not in the chats, nor in the emoticons menu (there is only one icon displayed, the black heart). All other menu fields are empty, or better: blank. Maybe the icons are missing.

On other/newer phones it works.

r10s:
Yes this may happen, esp. on older devices the system's unicode support is not that good (in fact, they're only glyphes as other characters; the same issue may happen if someone sends you some text using different alphabets as russian, arabic or greek)

However, as this issue is already fixed by more recent android versions, I do not think that it makes much sense to invest time and money into this issue (eg. by rendering all emoticons ourself) - esp. as this may open further other issues (eg. incompatible emoticons when using the clipboard on the same device, no aotumatic support of new emoticons, unfamiliar emoticons ...)

kollokollo:
Hm, yes, I see. Is there a possibility to install the missing glyphs e.g. by using a third-party app?

r10s:
As this issue is not related directly to Delta Chat, I think we can close it.

kollokollo:
Thanks for the link, but isnt this for iOS only?

r10s:
The linked forum posts are a about how to replace the Emoji-Font on Android devices. The font is replaced by the iOS-Emoji font IIRC.

kollokollo:
OK thanks; wer lesen kann...

kollokollo:
The method described above in the link didnt work out for me. I think this works only for Android 4.1 and later, I have Android 4.0.3. (Which does not support color fonts)

But:  I found a solution (with monochrome emotjis):

1. Download NotoEmoji-Regular.ttf from this project: https://github.com/googlei18n/noto-emoji
2. Using adb on your rooted phone: make /system writable (mount -o remount,rw /system)
3. put NotoEmoji-Regular.ttf in /system/fonts/
4. Add following lines to /system/etc/fallback_fonts.xml

`
    <family>
        <fileset>
            <file>NotoEmoji-Regular.ttf</file>
        </fileset>
    </family>


`
5. reboot the phone

r10s:
Okay - this seems to be a solution. If you like, you can create a blog entry for it a https://delta.chat/en/blog ...

[...]

PR #1494, opened 2020-07-22
Fix emojis on android 4

### LINKED ISSUE #1493: No more emoji's on Android 4.1.2
- Operating System (Linux/Mac/Windows/iOS/Android):Android 4.1.2
- Delta Chat Version:1.10.4; 1.10.5 (nightly 2020-07-21)
- Expected behavior:Emojis are displayed
- Actual behavior:Emojis are not displayed
- Steps to reproduce the problem:Enter a chat and click the emoji  #icon.
Screenshot_2020-07-21-18-12-54

- Screenshots:

Regards Mike

cyBerta:
fixes #1493

I was able to reproduce the bug and indeed webp and vector graphics haven't been displayed correctly on Android API < 18

I converted all webp files to pngs and enabled a compat option to display vector graphics correctly on old devices. Vector graphics are used in the lower tabbar for the different emoji catergories. 

I tested on a Samsung Galaxy S Android 4.4 cyanogenmod and a Samsung Galaxy S4 mini, Android 9 (Lineage OS). I couldn't visueally recognize any difference performance wise between loading webps and pngs on the S4 mini.


r10s:
good catch with the webp thingie!

unfortunately, this pr adds 12.4 mb to the apk (emojis 7.4 mb webp vs -19.8 mb png). wondering if that can be optimized somehow, i have the feeling a eg. 1024x1024 png with 1.6 mb (flags.png) is pretty big. maybe converting to 8bit optimized palette+dithering is worth a try.

wrt vector: so we can now savely use svg as https://github.com/deltachat/deltachat-android/pull/1480 - @cyBerta can you please double check that?

r10s:
just tried a random png-minifier (https://tinypng.com)

original (629 kb):

Activity

minified (100kb, 84% saved):

Activity

if you enlarge the images, you will notice differences, the minified version uses 8bit-palette dithering instead of 24bit-true-color.

it would be great, if we can use a 12bit- or 16-bit-palette, i think, then nothing would be noticable and the saving would still be huge.

cyBerta:
> wrt vector: so we can now savely use vectors as #1480 - @cyBerta can you please double check that?

Yes, the icon is visible using this PR on Android 4.4.

link2xt:
Use [oxipng](https://github.com/shssoichiro/oxipng) on top of it for lossless compression. `cargo install oxipng`, then `oxipng -Z` for best compression (zopfli)

Hocuri:
Why aren't we using jpg?
I could see a difference but not a lot, so I think that it's generally ok to show them this way.

cyBerta:
I've used https://tinypng.com/ for compression. The results were visually better than with https://compresspng.com. But there are still some artifacts, you see it on the tennis ball for example.

Activity

Simon-Laux:
And we can't use normal emoji fonts because older android has unicode bugs I assume?

r10s:
yip, old androids have partly pretty poor emoji supports. 

wrt jpg: jpg does not offer an alpha channel. also, to achieve a comparable quality, jpg is probably not smaller as it is optimized for photos mainly.

Simon-Laux:
> yip, old androids have partly pretty poor emoji supports.

Is the because the systems emoji font does not have the new characters or is that it also does not support new emoji fonts?

r10s:
just a random issue, including an image of the missing emoji: https://github.com/deltachat/deltachat-android/issues/142 , there are lots more, just browser through older, fixes issues.

the example is maybe extreme, the gaps are smaller in newer androids, however still present, so sending from android 10 to an android 9 may result in missing emojis. i would also prefer to just use native emoji support, however, it does not seems reasonable currently.

r10s:
@Simon-Laux back on-topic: do you see big/bad differences between the original imagses and the compressed on from @cyBerta? to me that looks pretty good! and we will save 18mb or so - so, a small regression seems to be reasonable :)

Hocuri:
For saving 18mb, it is ok to use these emojis IMO.

Simon-Laux:
@r10s I don't mean the system emoji font, we can ship an emoji font like we do in desktop. Question was if there is a fundamental problem with newer Unicode chars on older androids or if they are only missing on the systems font.

About the compression I don't care that much I'm having nothing against some retro looking dithering :smile:
:thinking: I think looking at this on a phone in a real chat makes more sense for comparing how it feels...

r10s:
i think, we had all kind of problems with unicode and emojis on older androids, also, there is not "the" system's font, unfortunately :) for sure, there are several ways to solve them, due to the signal-codebase, we followed the png/webp way.

AndreazLattmann:
You could use pngquant to reduce the size of the png files. 🤔

r10s:
@AndreazLattmann there seems to be lots of tools - feel free to play around with them, and, if there is an even better solution, a pr is very welcome - coding is not required in this case, the challenge is just to get the png from this pr smaller an nicer :) (as original files, i would always use the webp, however)

Simon-Laux:
We should try shipping the google noto color emoji font (https://github.com/googlefonts/noto-emoji) and see on android 4 if there are any bugs with it... do we have an android 4 device? who has one?

Simon-Laux:
But on the other hand its fine for now, but if we can get the font somehow to work it I would vote for using a font instead of image files.

Simon-Laux:
I think I found what we could use https://developer.android.com/guide/topics/ui/look-and-feel/emoji-compat

link2xt:
> we could try https://developer.android.com/guide/topics/ui/look-and-feel/emoji-compat but that maybe depends on the playstore.

It depends on Android 4.4, and we support Android 4.1, and @adbenitez supports a build running on Android 4.0.

Hocuri:
If we want to change the png pictures, then using the noto (https://github.com/googlefonts/noto-emoji) emojis might be an idea because they have fewer color gradients as far as I could see. And 1. it were the gradients that looked less good in the new emojis, 2. gradients make the images bigger.

Oblivion94:
I think that Google Noto Font would be an excellent option for fonts, since it has support for the SVG format. That way, the emojis would scale correctly on all devices.
However, Google Noto Fonts has been outdated for some time. One option that is faithful to each version of Unicode and is also Open source is Twitter's twi-emoji.
https://github.com/twitter/twemoji

Hocuri:
Yes, Twemoji seems to be a better choice, then, because they have few gradients, too.

Simon-Laux:
@Oblivion94 what is outdated at gogle noto? (besides the font they offer on the download page?)

Oblivion94:
@Simon-Laux Noto Color Emoji is a very cool font, but according to its official page it was only updated to Unicode version 10
https://www.google.com/get/noto/help/emoji/

For this reason many emojis are lost along the way.
However twi-emoji is constantly updated to new Unicode versions, currently 13
https://twemoji.twitter.com

From the Github page you can see the changelog
https://github.com/twitter/twemoji/releases

In addition, as well @Hocuri has less color gradients and is very simple.

Simon-Laux:
@Oblivion94 as I mentioned the website of noto is outdated, the github page has the emojis for version 13 (updated 5 days ago)
From what I saw noto reduced its gradients a bit.

I couldn't find the font file for Twemoji. though I have no opnion on whether we should or shouldn't use it.

Openmoji would also be an option but the style is rather unique which could be unfamiliar for users, they work on implementing version 13 too (https://github.com/hfg-gmuend/openmoji/issues/124)

Oblivion94:
@Simon-Laux I see, Noto also updated to Unicode 13 recently. Honestly it is a matter that deserves a survey. I opt for Twitter due to the support and the active community they have.
A while back I discussed just these three options in the support forum:
https://support.delta.chat/t/more-modern-and-vector-format-emojis-svg/851
It is the most convenient type of format for emojis for several reasons. Hopefully you decide to implement it in the next versions :)

icke68746:
Operating System (Linux/Mac/Windows/iOS/Android):Android 4.1.2
Since tonight's version (deltachat-gplay-debug-1.11.0.apk 26-Jul-2020 00:38 68830441), the emojis are back. Thanks!
Screenshot_2020-07-26-15-33-20
Regards Mike

r10s:
thanks for confirming, @icke68746 :)

PR #1498, opened 2020-07-25
Tell the user that they can turn off the background notification.

Hocuri:
Also, maybe we should make this translatable?

### DIFF
--- a/src/org/thoughtcrime/securesms/connect/KeepAliveService.java
+++ b/src/org/thoughtcrime/securesms/connect/KeepAliveService.java
@@ -123,8 +123,8 @@ public class KeepAliveService extends Service {
         if(!ch_created) {
             ch_created = true;
             NotificationChannel channel = new NotificationChannel(NotificationCenter.CH_PERMANENT,
-                "Receive messages in background.", NotificationManager.IMPORTANCE_MIN); // IMPORTANCE_DEFAULT will play a sound
-            channel.setDescription("Ensure reliable message receiving.");
+                "Backgrund connection. This can safely be turned off.", NotificationManager.IMPORTANCE_MIN); // IMPORTANCE_DEFAULT will play a sound
+            channel.setDescription("Delta Chat has to show this notification to reliably receive messages in background. But this will still work if you hide this notification.");
             NotificationManager notificationManager = context.getSystemService(NotificationManager.class);
             notificationManager.createNotificationChannel(channel);
         }


src/org/thoughtcrime/securesms/connect/KeepAliveService.java
```diff
@@ -123,8 +123,8 @@ static private void createFgNotificationChannel(Context context) {
         if(!ch_created) {
             ch_created = true;
             NotificationChannel channel = new NotificationChannel(NotificationCenter.CH_PERMANENT,
-                "Receive messages in background.", NotificationManager.IMPORTANCE_MIN); // IMPORTANCE_DEFAULT will play a sound
-            channel.setDescription("Ensure reliable message receiving.");
+                "Backgrund connection. This can safely be turned off.", NotificationManager.IMPORTANCE_MIN); // IMPORTANCE_DEFAULT will play a sound
```
angelo-fuchs, line 126: Telling the user that something can be "safely turned off" raises the question why its there in the first place.

So if we don't need it: Then we should remove it
If we only need it on special situations (old Androids?): Only use it there (and same wording as below).
We need it to display messages quickly: "Background connection. Needed to display new messages quickly."
We need it to display messages at all when DC is not in the foreground: "Background connection. Needed to inform you of new messages."

(We also should really put the strings to the strings.xml so they are translated)

src/org/thoughtcrime/securesms/connect/KeepAliveService.java
```diff
@@ -123,8 +123,8 @@ static private void createFgNotificationChannel(Context context) {
...
+            channel.setDescription("Delta Chat has to show this notification to reliably receive messages in background. But this will still work if you hide this notification.");
```
angelo-fuchs, line 127: Same as above. "This will work if you hide this notification" -> So why have it at all?

r10s: we need to add a so called "permantent notification" to get a reliable background service on many newer androids.

by default, this notification is visible to the user, which might be a annoying. there is nothing we can do on that, the system _wants_ the user to notice that there is sth. in the background.

however, depending on the system, the user may have the option to hide the notification, so that it is less annoying. this should not affect the reliability of the background-service (as the notification, is still there in a way).

two things:

- i am not sure, that hiding the notification does not affect the background-service (i also think so, however, i have not found sources for that) - also: is this true for all the various android versions? this kind of things also easily changes with newer android versions.

- if we actively point users to hide the permantent notification, this may result in increased support issues - eg. toggling the corresponding switch in the setting will not have an effect.

also, obviously, things are hard to explain and easily misleading.

Hocuri: @angelo-fuchs Björn explained it nicely here.

> i am not sure, that hiding the notification does not affect the background-service (i also think so, however, i have not found sources for that)

That's actually bad. We might test, but this might not be worth it. But I think I am still in favor of it.

> the whole thing is pretty invisible, somewhere deep in the system settings. so i do not think that this is a really important thing now.

That's actually why I am still in favor of it: users will mainly land there if they are looking for a way to turn off this notification. But "Ensure reliable message receiving" sounds as if the user must not turn this off. We guess otherwise, so we should tell this to the rather advanced users who get here.

angelo-fuchs: I agree that we can explain it better, but the right hand parts should be removed.

This: "Delta Chat has to show this notification to reliably receive messages in background."
is better then: "Ensure reliable message receiving."

But telling users to hide it without knowing for certain that its safe (by design, not just by chance on the devices we have at hand) is at least suspicious, most likely dangerous. I'm strongly opposed to that (until we have documentation from the Android Team that it is, in fact, safe.)

PR #1499, opened 2020-07-25
Explain the "Show all/normal emails" setting better

Hocuri:


res/values/strings.xml
```diff
@@ -461,10 +461,10 @@
     <string name="pref_auto_folder_moves">Automatic moves to DeltaChat folder</string>
     <string name="pref_auto_folder_moves_explain">Chat conversations are moved to avoid cluttering the Inbox</string>
     <string name="pref_email_interaction_title">Email interaction</string>
-    <string name="pref_show_emails">Show classic emails</string>
-    <string name="pref_show_emails_no">No, chats only</string>
+    <string name="pref_show_emails">Show all emails</string>
+    <string name="pref_show_emails_no">No, only chats via Delta Chat</string>
     <string name="pref_show_emails_accepted_contacts">For accepted contacts</string>
-    <string name="pref_show_emails_all">All</string>
+    <string name="pref_show_emails_all">Yes, show all emails from my mail account</string>
```
angelo-fuchs, line 467: I think this line is too long to be displayed nicely. How about

"Yes, all emails"

As we always display this together with "Show classic/all emails" it should be clear what is meant.

res/values/strings.xml
```diff
@@ -461,10 +461,10 @@
     <string name="pref_auto_folder_moves">Automatic moves to DeltaChat folder</string>
     <string name="pref_auto_folder_moves_explain">Chat conversations are moved to avoid cluttering the Inbox</string>
     <string name="pref_email_interaction_title">Email interaction</string>
-    <string name="pref_show_emails">Show classic emails</string>
-    <string name="pref_show_emails_no">No, chats only</string>
+    <string name="pref_show_emails">Show all emails</string>
+    <string name="pref_show_emails_no">No, only chats via Delta Chat</string>
```
angelo-fuchs, line 465: I think this line is too long to be displayed nicely. How about

"No, only Delta Chats"

Hocuri: I did not try this, but I do not think that non-techie people would understand the term "Delta Chats" here.

Hocuri: Of course your idea is way better than the original because that's a smart way to use the term "Delta Chat" without using "Chat" twice :+1:.

res/values/strings.xml
Hocuri: Ok, I asked my brother and my mother. My brother would have understood your formulations, my mother would have preferred my formulations. 

I am still in favor of my formulation because this is an important setting and it is not good if even a few people do not understand this. Also, these are the settings, people will not look at this too often.

OTOH, of course, most users will be rather as techie as my brother.

angelo-fuchs:
For comparsion two screenshots with the texts:
Longer:
longer texts

Shorter:
shorter texts

I think both are all right, even though I would prefer the shorter ones. But the longer ones are a definitive improvement over the current state.

@Hocuri I would leave the decision to you, if you want to keep the longer ones I will approve and merge. Else please update to the shorter ones and I'll merge that then.

Hocuri:
I asked @Simon-Laux and he was in favor of the longer ones, so if you leave the decision to me, I'll be in favor of the longer ones.

cyBerta:
I think "All chats" is quite clear in the context of Delta Chat. The settings menu itself is called "Chats and media" and I doubt people would think of other chat clients than the one they currently using. That's why I would leave "No, chats only" as it is currently.

I don't have a strong opinion about "Yes, show all emails from my email account" vs. "Yes, all mails" but I tend to the longer version. I just wonder if "Yes, show all emails _of_ my email account" would sound better to a native speaker. 

When comparing layouts: please keep in mind that these strings are also used on iOS and Desktop. So comparing Android's layout only doesn't make too much sense.

angelo-fuchs:
@cyBerta I have no means to check the layout of iOS or Desktop. Do you expect the strings to look bad there?

res/values/strings.xml
hpk42, line 467: Why suddenly "mail" instead of "emails"?

hpk42:
I don't see the evidence that the current language caused confusion that this change would remedy. 

That people don't know that they change the email interaction setting is a problem that you don't solve via this setting. 

Rather, during onboarding and after we scanned the history, we might ask the user something along the lines of 
"oh, it appears you are using this e-mail account also with another email program.  Do you want to a) see all your e-mails in Delta Chat or b) only see chat conversations from other Delta Chat users?" 

So i suggest to not merge this PR.

res/values/strings.xml
Hocuri: So that we do not write "mail" three times (it is also it the title)

Hocuri:
>  don't see the evidence that the current language caused confusion that this change would remedy.

I did not understand what this setting does before I got a member of the dev team myself. And we definitely need a "Yes" opiton somewhere if we have a "no" somewhere. Neither did the non-techie people I asked (my family).

> Rather, during onboarding and after we scanned the history, we might ask the user something along the lines of
> "oh, it appears you are using this e-mail account also with another email program. Do you want to a) see all your e-mails in Delta Chat or b) only see chat conversations from other Delta Chat users?"

Good idea (When you told me about this idea the first time, I thought that this was too difficult to explain for someone who does not open the settings. But I changed my mind after my family had no problems understanding a sentence like you proposed).

However, we will still need a setting in the settings for people who want to change this later. And it will still need to be understandable.

Why don't you like this change? Do you think that the old wording was more understandable? Or looked nicer because it fit in one line?

PR #1503, opened 2020-07-26
implement custom view for ephemeral messages dialog 

### LINKED ISSUE #1497: add a hint in disappearing-messages-dialog
in the disappearing-messages-dialog, below the radio-options, there should be a hint that the settings apply for all users and that no one should assume there is not another copy (ppl can take a photo, this is a very obvious example)

- the AlertDialog does not allow options _and_ a text, but it allows text and a view, so we have to move the options to a view ... we've done that already on several other places, eg. recently in https://github.com/deltachat/deltachat-android/commit/75955f70100b78a1d5f0fd8c9c61209eea14f449 - or before in the autocrypt-setup-input

- the whole layout should be put in an xml file

- i think, the hint should be _below_ the options, as message is shown above, we have to integrate the hint to the view

- the options should stay scrollable, so that everything is fine on smaller displays/landscape. a LinearLayout should handle that easily

- although using a view offers to use different, maybe nicer selectors, i would stay with the radio buttons as they are now - or maybe this scroll-selector(how is it called?) is also fine, but, if possible, no external libraries :)

once this is implemented here and the wording is settled, we should use the same text on ios/desktop.

cyBerta:
...showing a hint about the implications of this feature

closes #1497

ephemeralDialog

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -26,6 +26,7 @@
     <string name="unarchive">Unarchive</string>
     <string name="mute">Mute</string>
     <string name="ephemeral_messages">Disappearing messages</string>
+    <string name="ephemeral_messages_hint">These settings apply to all participants using Delta Chat. However, you must trust your conversation partners to use Delta Chat and not copy, save or forward any content.</string>
     <string name="save">Save</string>
     <string name="chat">Chat</string>
     <string name="media">Media</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -26,7 +26,7 @@
     <string name="unarchive">Unarchive</string>
     <string name="mute">Mute</string>
     <string name="ephemeral_messages">Disappearing messages</string>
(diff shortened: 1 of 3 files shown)


cyBerta:
screenshot was made on a s4 mini (rather small display)

res/values/strings.xml
```diff
@@ -26,6 +26,7 @@
     <string name="unarchive">Unarchive</string>
     <string name="mute">Mute</string>
     <string name="ephemeral_messages">Disappearing messages</string>
+    <string name="ephemeral_messages_hint">These settings apply to all participants using Delta Chat. However, you must trust your conversation partners to use Delta Chat and not copy, save or forward any content.</string>
```
link2xt, line 29: This does not sound good when translated. How about "However, they may copy, save and forward messages or use other chat clients."

cyBerta: sounds good to me!

Hocuri:
BTW, we should maybe also let selfchat-messges vanish if the user enables this setting.

link2xt:
@Hocuri This already works when enabled in Saved messages chat. If it doesn't, it's a bug.

Hocuri:
Sorry, it does work. I just did not understand how this feature is supposed to work.

PR #1504, opened 2020-07-27
generic foregroundservice

### LINKED ISSUE #1447: use an explicit foreground-service for long-ongoing processes
there are some few actions in Delta Chat that take some time; there is some risk, that this action is interrupted by the system and Delta Chat is killed before the process is finished. 

to avoid that, a so called "foreground service" is needed, these services always come with a notification that is shown as long the process takes place (in fact, our "reliable background notification" thing is such a "foreground service", see [`KeepAliveService`](https://github.com/deltachat/deltachat-android/blob/master/src/org/thoughtcrime/securesms/connect/KeepAliveService.java)).

actions:

- create a generic foreground-service that can be easily used with a simple call and closed when work is done by another call
- only one foreground-service is needed at the same time
- the existing `KeepAliveService` should not be touched, if it is active, there might be _two_ permanent notifications, which is fine
- i think we can use signal's [`GenericForegroundService`](https://github.com/signalapp/Signal-Android/blob/master/app/src/main/java/org/thoughtcrime/securesms/service/GenericForegroundService.java) - maybe some tweaking here and there :)
- the needed notifiation-id needs to be added to NotificationCenter as eg. ID_GENERIC_FOREGROUND_SERVICE=3
- call `GenericForegroundService.startForegroundTask()` before the long-ongoing process and `GenericForegroundService.stopForegroundTask()` or so afterwards - the  long-ongoing process as such should not be modified, the changes there should be small.

long-ongoing processes are currently import and export. maybe later we can also add recoding videos, however, this should be done in another step.

cyBerta:
used for import export 🚛 🚛 🚛 

closes #1447 

 changes in Signals NotificationController and GenericBackground service:
* removed AutoCloseable Interface from NotificationController because it is only supported from API 19+
* instead we assure that the service connection to the GenericBackgroundService gets explicitly closed  in close()
* added possibility to add a contentText of a notification. It is used to display and update the percentage in the notification.

backup

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -285,6 +285,8 @@
         </intent-filter>
     </service>
 
+    <service android:name=".service.GenericForegroundService" />
+
     <receiver android:name=".notifications.MarkReadReceiver"
               android:enabled="true"
               android:exported="false">
--- a/src/org/thoughtcrime/securesms/WelcomeActivity.java
+++ b/src/org/thoughtcrime/securesms/WelcomeActivity.java
@@ -5,10 +5,11 @@ import android.annotation.SuppressLint;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.os.Bundle;
-import androidx.annotation.NonNull;
-import androidx.appcompat.app.AlertDialog;
 import android.view.View;
 import android.widget.Button;
+
+import androidx.annotation.NonNull;
+import androidx.appcompat.app.AlertDialog;
 
 import com.b44t.messenger.DcContext;
 import com.b44t.messenger.DcEventCenter;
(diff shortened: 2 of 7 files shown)


r10s:
great! i'd use the strings `import_backup_title` and `pref_backup` or `export_backup_desktop` or so instead of "one moment" here.

cyBerta:
> great! i'd use the strings `import_backup_title` and `pref_backup` or `export_backup_desktop` or so instead of "one moment" here.

makes sense. I changed that.

r10s:
as a quick fix - maybe we can just do "nothing" on tapping the notification - that would move the whole thing to the state as before - with the advantage that eg. export/import is not aborted by the system.

cyBerta:
> as a quick fix - maybe we can just do "nothing" on tapping the notification - that would move the whole thing to the state as before - with the advantage that eg. export/import is not aborted by the system.

fixed

r10s (approved):
lgtm, for the remaining https://github.com/deltachat/deltachat-android/pull/1504#pullrequestreview-466004016 i've file a new issue at #1610

Issue #1477, opened 2020-07-07
Replace doze-reminder device message with a menu option and ask for it on first start


link2xt:
Users don't even open device chat sometimes, let alone click the message to enable battery optimizations. If there is no harm in asking for this permission, maybe ask for it immediately at the start? And if it is not granted, show an option to enable it somewhere in the settings menu.

AndreazLattmann:
I have already noticed that some friends did not go to see in the messages of the device. I agree with @link2xt .

Hocuri:
I'd personally leave the device message as is (maybe show a notification for it, the core has the possibility now, we just need to add the ffi), but thats not so important. 

And I agree that we could just ask on start.

r10s:
> If there is no harm in asking for this permission, maybe ask for it immediately at the start? 

well, thinking it over, there is some risk to lose the ability to do the request at all, i commented at https://github.com/deltachat/deltachat-android/pull/1505#pullrequestreview-479019318

> And if it is not granted, show an option to enable it somewhere in the settings menu.

this is probably no problem, i did a pr for that at https://github.com/deltachat/deltachat-android/pull/1608

r10s:
closed by #1608 and #1505

PR #1505, opened 2020-07-27
Already ask the user to disable doze on the first start.

Hocuri:
This fixes, well, one part of https://github.com/deltachat/deltachat-android/issues/1477, also because it was easier to implement only this for now.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationListFragment.java
@@ -160,6 +160,7 @@ public class ConversationListFragment extends Fragment
     super.onResume();
 
     updateReminders();
+    DozeReminder.maybeAskDirectly(getActivity());
     list.getAdapter().notifyDataSetChanged();
   }
 
--- a/src/org/thoughtcrime/securesms/ConversationListFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationListFragment.java
@@ -160,7 +160,6 @@ public class ConversationListFragment extends Fragment
     super.onResume();
 
     updateReminders();
-    DozeReminder.maybeAskDirectly(getActivity());
     list.getAdapter().notifyDataSetChanged();
   }
 
@@ -239,6 +238,7 @@ public class ConversationListFragment extends Fragment
 
       @Override
       protected void onPostExecute(Optional<? extends Reminder> reminder) {
+        DozeReminder.maybeAskDirectly(getActivity());
         if (reminder.isPresent() && getActivity() != null && !isRemoving()) {
           reminderView.showReminder(reminder.get());
         } else if (!reminder.isPresent()) {
--- a/src/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
+++ b/src/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
@@ -105,18 +105,22 @@ public class DozeReminder {
   }
 
   public static void maybeAskDirectly(Context context) {
-    if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M
-        && !Prefs.getBooleanPreference(context, Prefs.DOZE_ASKED_DIRECTLY, false)
-        && ContextCompat.checkSelfPermission(context, Manifest.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS)==PackageManager.PERMISSION_GRANTED
-        && !((PowerManager)context.getSystemService(Context.POWER_SERVICE)).isIgnoringBatteryOptimizations(context.getPackageName())) {
+    try {
+      if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M
+          && !Prefs.getBooleanPreference(context, Prefs.DOZE_ASKED_DIRECTLY, false)
+          && ContextCompat.checkSelfPermission(context, Manifest.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS) == PackageManager.PERMISSION_GRANTED
+          && !((PowerManager) context.getSystemService(Context.POWER_SERVICE)).isIgnoringBatteryOptimizations(context.getPackageName())) {
 
-      Intent intent = new Intent(Settings.ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS,
-              Uri.parse("package:" + context.getPackageName()));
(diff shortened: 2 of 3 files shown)


angelo-fuchs:
The commit is on the core, right. Shouldn't the PR then go to the core as well? (or am I just confused?)

Hocuri:
Oh, yes, I must somehow have gotten the git add wrong.

Anyway, fixed.

Hocuri:
Oh, yes, I must somehow have gotten the git add wrong.

Anyway, fixed.

angelo-fuchs:
I just tried deinstalling DC and reinstalling from android and was not asked until I clicked on the system message.
Do I need to do something else to see the question?

(Android Nougat 7.1.2, SDK 25)

Hocuri:
Strange... Could you tell me which of the conditions in `maybeAskDirectly()` returns false?

angelo-fuchs:
@Hocuri After trying it again it now works properly.
Strange, but nothing I would worry about. Its also not as annoying as I thought it would be, actually quite fast in the handling.

Hocuri:
@r10s could you remove your blocking review?

r10s:
is suggest to merge that in after the upcoming bugfix-release. we should also add a switch to the settings then (this is another pr, however)

Hocuri:
@r10s what about merging this now? (you still have to remove your review)

r10s:
we are doing another bugfix-release first, see https://github.com/deltachat/deltachat-android/pull/1597 - so this one has to wait a little longer :)

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -160,6 +160,7 @@ public void onResume() {
     super.onResume();
 
     updateReminders();
+    DozeReminder.maybeAskDirectly(getActivity());
```
r10s, line 163: i did not tested that, however, onResume is called whenever the activity is reused - i think it is not needed to check for the the DozeReminder in these cases. also, not sure why this goes to the fragment.

maybe ConversationListActivity.onStart() is a better place (ConversationListActivity.onCreate() where we add device-messages, however, is too soon as the activity is not visible)

src/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
```diff
@@ -101,4 +105,19 @@ public static void dozeReminderTapped(Context context) {
       context.startActivity(intent);
     }
   }
+
+  public static void maybeAskDirectly(Context context) {
+    if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M
```
r10s, line 110: i think, we should put the while function body to a try/catch block, eg. https://stackoverflow.com/search?page=3&tab=Relevance&q=ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS reports several situations where the things do not work and i can easily imagine that things may crash - esp. as this area is highly modified by the manufacturers.

r10s (commented):
while technically with the proposed changes, this pr will be fine, there is some, maybe low, risk that we will lose the right to fire ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS at all: there is some room for interpretation if Delta Chat is an "exemption-case", see the table at https://developer.android.com/training/monitoring-device-state/doze-standby.html#exemption-cases - we were classified as such years/sdk/policies ago, iirc there also was the exemption _email_ that time. however, when the query pops up directly in the face of a reviewer, they may have a closer look.

google seems to be quite strict about this specific right today, browsing through stackoverflow, i do not find a completely matching story, only some ["be careful"](https://stackoverflow.com/questions/33114063/how-do-i-properly-fire-action-request-ignore-battery-optimizations-intent#comment54538042_33114136) here and there. maybe these guys did not add the corresponding request to the manifest, however, i have the feeling, this could cause problems ... 

also, i am not sure if it is fine (from view of google) to ask the user without prior user-interaction.

link2xt:
I think Delta Chat falls under "No, can't use FCM because of technical dependency on another messaging service or Doze and App Standby break the core function of the app." It depends on another messaging service (email).

> also, i am not sure if it is fine (from view of google) to ask the user without prior user-interaction.

Yes, maybe it should happen after account configuration and with a dialog explaining why the permissions is needed. This page talks about "educational UI": https://developer.android.com/training/permissions/requesting

link2xt:
You can also install Briar and see how they do it. Can't take screenshot because it does not allow screenshots >_<

r10s:
> You can also install Briar and see how they do it. Can't take screenshot because it does not allow screenshots >_<

briar has a screen before the system-question, ftr, this is how it looks there:



however, their situation is even more worse as ours - you cannot use briar without battery-optimizations disabled at all. therefore, the dialog is designed so that you cannot continue until you have disabled battery-optimizations.

having in dontkillmyapp.com in mind - this whole disable-battery-optimizations is misleading at all - it only increases the _chance_ that things work in background. therefore, in case of delta, that at least _somehow_ works without, i would not tune this up too much. a simple Alert shown before the system question should do the job.

r10s:
another issue: the keyboard is open (or not closed)  while the dialog is shown.



this is not a very good first impression :)

src/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
```diff
@@ -21,11 +20,11 @@
 
 import org.thoughtcrime.securesms.R;
 import org.thoughtcrime.securesms.connect.DcHelper;
-import org.thoughtcrime.securesms.util.IntentUtils;
 import org.thoughtcrime.securesms.util.Prefs;
 
 @SuppressLint("BatteryLife")
 public class DozeReminder {
+  private static final String ALREADY_ASKED_DIRECTLY = "DOZE_ALREADY_ASKED_DIRECTLY";
```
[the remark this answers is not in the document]
r10s: i changed that.

src/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
r10s: i changed that.

r10s:
k, let's merge that in a add/discuss a permission rationale in a subsequent pr. probably a permission rationale is not really needed to keep ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS, however, some system only talk more about battery and not less about background, so it seems reasonable.

PR #1513, opened 2020-07-29
setting stock translations after context switch

### LINKED ISSUE #1502: English signature is default, even on non-english language settings
Hi there,

on my Samsung phone with German language environment Delta Chat uses English signature ("Sent with my...") as default for new profiles.

(GUI is German as expected.)
 
Found with 10.10.5 build. But I think I experienced the same bug already with 10.8.1 build.

angelo-fuchs:
The context contains the proper translations to be used by the core ("stock translations")
After a context switch they are unset.

This change sets them so they are of the appropriate language again.

fixes #1502


angelo-fuchs:
I'm open to moving the call to someplace else that might be more clean, but this change alone should do it.
We might want to check if other things need to be called on the new context as well.

PR #1515, opened 2020-07-29
handle videocall invitations

r10s:
this pr adds sending and receiving videocall invitations as provided by the new api from https://github.com/deltachat/deltachat-core-rust/pull/1735

the call itself is done in a browser or in apps as "jitsi meet".

Screen Shot 2020-07-29 at 16 52 17 Screen Shot 2020-07-29 at 16 54 15
Screen Shot 2020-07-29 at 16 54 37 Screen Shot 2020-07-29 at 16 55 10

to use jitsi for video calls, you can scan the following qr-code:

### DIFF
--- a/res/drawable/ic_map_white_24dp.xml
+++ b/res/drawable/ic_map_white_24dp.xml
@@ -1,0 +1,5 @@
+<vector android:height="24dp" android:tint="#FFFFFF"
+    android:viewportHeight="24.0" android:viewportWidth="24.0"
+    android:width="24dp" xmlns:android="http://schemas.android.com/apk/res/android">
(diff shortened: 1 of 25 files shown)


res/drawable/ic_map_white_24dp.xml
```diff
@@ -0,0 +1,5 @@
+<vector android:height="24dp" android:tint="#FFFFFF"
```
r10s, line 1: the map-icon is updated in this pr as at in the first iteration, there was a video-icon as well and the map-icon did not match standard-sizes. later, we decided to remove video-icon - but the map-icon is still nicer :)

cyBerta:
Currently the value of the entry 'Video chat instance' doesn't update when I add "https://meet.jit.si/$ROOM"

r10s:
@cyBerta looks to me as if not the correct core-rust-submodule is used or build, indeed.

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -731,20 +735,6 @@ public void onItemLongClick(DcMsg messageRecord) {
...
-            if (getContext() == null) return;
-
-//      ContactUtil.selectRecipientThroughDialog(getContext(), choices, locale, recipient -> {
-//        CommunicationActions.startConversation(getContext(), recipient, null);
-//      });
-        }
-
-        @Override
-        public void onInviteSharedContactClicked(@NonNull List<Recipient> choices) {
-            if (getContext() == null) return;
```
cyBerta, line 746: :)

Issue #1528, opened 2020-08-09
Add "go to last seen"


Hocuri:
- Operating System (Linux/Mac/Windows/iOS/Android): Android
- Delta Chat Version: Some versions or years ago (I don't know) until current master
- Expected behavior: When I open a chat with unread messages, I get to the position of the first unread message and a "3 unread messages" is shown
- Actual behavior: it is not shown
- Steps to reproduce the problem: Receive many messages (that do not fit on the screen) and open the chat

r10s:
well, it is just a missing feature, i would not consider that as a bug ;)

nevertheless, it is an important thing that we should add.

there is the rough idea to get that in on all platforms by the MARKER thing, https://github.com/deltachat/deltachat-core-rust/issues/1674

Hocuri:
Ah ok, the code looked as if this should be working to me. Do you know what "moveToLastSeen()" does?

(it is important for me because if this is a feature that is currently working, then I might break it with #1529)

r10s:
yip, there is some code, eg. when being inside a chat, switching display off+on and message arrive in between, the message marker is shown.

PR #1529, opened 2020-08-09
Fix sharing both an image and a text to DC

Hocuri:
This bug was re-introduced in https://github.com/deltachat/deltachat-android/pull/1475/.

Steps: Share media+text from another app to DC.

@cyBerta for the next time: could you please ping me when you revert one of my commits?
@r10s please wait until @cyBerta commented, maybe it is possible to prevent adding another bug again this time.

Concerning this SettableFuture: Turns out it is only used to determine whether to go to lastseen (or not) but this does not work anyway: #1528.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -725,26 +725,28 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
    */
   private ListenableFuture<Boolean> initializeDraft() {
 
-    final SettableFuture<Boolean> result = new SettableFuture<>();
+    SettableFuture<Boolean> result = new SettableFuture<>();
 
-    final String    draftText      = getIntent().getStringExtra(TEXT_EXTRA);
-    final Uri       draftMedia     = getIntent().getData();
-    final MediaType draftMediaType = MediaType.from(getIntent().getType());
+    final String draftText = RelayUtil.getSharedText(this);
+    final ArrayList<Uri> draftMediaList = RelayUtil.getSharedUris(this);
+    Uri draftMedia = null;
+    if (!draftMediaList.isEmpty()) draftMedia = draftMediaList.get(0);
+    final MediaType draftMediaType = MediaType.from(MediaUtil.getMimeType(this, draftMedia));
 
     if (draftText != null) {
       composeText.setText(draftText);
       result.set(true);
     }
     if (draftMedia != null && draftMediaType != null) {
-      return setMedia(draftMedia, draftMediaType);
+      result = setMedia(draftMedia, draftMediaType);
     }
 
-    if (draftText == null && draftMedia == null && draftMediaType == null) {
-      return initializeDraftFromDatabase();
+    if (draftText == null || draftMedia == null || draftMediaType == null) {
+      result = initializeDraftFromDatabase();
     } else {
-      updateToggleButtonState();
       result.set(false);
     }
+    updateToggleButtonState();
 
     return result;
   }
@@ -756,7 +758,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     attachButton.setEnabled(enabled);
   }
 
-  private ListenableFuture<Boolean> initializeDraftFromDatabase() {
+  private SettableFuture<Boolean> initializeDraftFromDatabase() {
     SettableFuture<Boolean> future = new SettableFuture<>();
 
     new AsyncTask<Void, Void, DcMsg>() {
@@ -969,7 +971,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     startActivityForResult(intent, PICK_CONTACT);
   }
 
-  private ListenableFuture<Boolean> setMedia(@Nullable Uri uri, @NonNull MediaType mediaType) {
+  private SettableFuture<Boolean> setMedia(@Nullable Uri uri, @NonNull MediaType mediaType) {
     if (uri == null) {
       return new SettableFuture<>(false);
     }
--- a/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
+++ b/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
@@ -212,7 +212,7 @@ public class AttachmentManager {
   */
 
   @SuppressLint("StaticFieldLeak")
-  public ListenableFuture<Boolean> setMedia(@NonNull final GlideRequests glideRequests,
+  public SettableFuture<Boolean> setMedia(@NonNull final GlideRequests glideRequests,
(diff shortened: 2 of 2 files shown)


Hocuri:
@r10s if `fragment.moveToLastSeen();` is doing something that currently works, then this pr might break it. Because I am not sure if initializeDraft() still always returns true or false correctly and in https://github.com/deltachat/deltachat-android/blob/46a14b954cb2e0a23282776dcc9abb0d0215b6dc/src/org/thoughtcrime/securesms/ConversationActivity.java#L238 it depends on the returned value whether moveToLastSeen() is invoked or not.

However, I still do not understand this feature and why we would only call moveToLastSeen() if the share draft was initialized.

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -725,26 +725,28 @@ public void onSuccess(Boolean result) {
    */
   private ListenableFuture<Boolean> initializeDraft() {
 
-    final SettableFuture<Boolean> result = new SettableFuture<>();
+    SettableFuture<Boolean> result = new SettableFuture<>();
 
-    final String    draftText      = getIntent().getStringExtra(TEXT_EXTRA);
```
cyBerta, line 730: This reintroduces #1472.

cyBerta: I'll add more comments later

cyBerta:
Thanks for your efforts @Hocuri 
since it's already late, I'll continue tomorrow. I didn't completely understand the fix yet, but tomorrow with fresh energy I will hopefully.

cyBerta:
in general I think this part is too complex and hard to follow. Several paths lead to the initialization part, initialization from an intent (e.g. through sendto intent) and drafts from the database are handled here. these cases can have different passed values and with regard to that the implications of changes like that https://github.com/deltachat/deltachat-android/pull/1529/files#diff-8b6badc7254f2d75634d87a85bc9cd31R744 are not easy to understand. That's why I think splitting it the method up into several for different cases might be an improvement.

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -725,26 +725,28 @@ public void onSuccess(Boolean result) {
...
 
-    if (draftText == null && draftMedia == null && draftMediaType == null) {
-      return initializeDraftFromDatabase();
+    if (draftText == null || draftMedia == null || draftMediaType == null) {
+      result = initializeDraftFromDatabase();
     } else {
-      updateToggleButtonState();
       result.set(false);
     }
+    updateToggleButtonState();
```
cyBerta, line 749: updateToggleButtonState() is called in initializeDraftFromDatabase(), no need to call it twice

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -725,26 +725,28 @@ public void onSuccess(Boolean result) {
...
-      return setMedia(draftMedia, draftMediaType);
+      result = setMedia(draftMedia, draftMediaType);
     }
 
-    if (draftText == null && draftMedia == null && draftMediaType == null) {
-      return initializeDraftFromDatabase();
+    if (draftText == null || draftMedia == null || draftMediaType == null) {
+      result = initializeDraftFromDatabase();
     } else {
-      updateToggleButtonState();
```
cyBerta, line 745: only in this case we need to make sure updateToggleButton gets called

cyBerta:
> Because I am not sure if initializeDraft() still always returns true or false correctly

There's another usage that depends on the correct return value: 
https://github.com/deltachat/deltachat-android/blob/d10891311fa706827f3ec50263c1298a7537332b/src/org/thoughtcrime/securesms/ConversationActivity.java#L711

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -725,26 +725,28 @@ public void onSuccess(Boolean result) {
    */
   private ListenableFuture<Boolean> initializeDraft() {
 
-    final SettableFuture<Boolean> result = new SettableFuture<>();
+    SettableFuture<Boolean> result = new SettableFuture<>();
 
-    final String    draftText      = getIntent().getStringExtra(TEXT_EXTRA);
-    final Uri       draftMedia     = getIntent().getData();
```
cyBerta, line 731: getIntent().getData() gets set here https://github.com/deltachat/deltachat-android/blob/d10891311fa706827f3ec50263c1298a7537332b/src/org/thoughtcrime/securesms/NewConversationActivity.java#L171, if a sendto link was clicked. 
Replacing it by `RelayUtil.getSharedUris(this);` doesn't work in that case, because the latter will return an empty list.

cyBerta:
@Hocuri which app did you use to share both text and images at the same time?

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -756,7 +758,7 @@ private void initializeEnabledCheck() {
     attachButton.setEnabled(enabled);
   }
 
-  private ListenableFuture<Boolean> initializeDraftFromDatabase() {
+  private SettableFuture<Boolean> initializeDraftFromDatabase() {
```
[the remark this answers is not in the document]
Hocuri: It does not compile otherwise, changing this was the quickest way I saw.

Hocuri:
> There's another usage that depends on the correct return value: 
https://github.com/deltachat/deltachat-android/blob/d10891311fa706827f3ec50263c1298a7537332b/src/org/thoughtcrime/securesms/ConversationActivity.java#L711

But here the returned boolean is not checked (at least if I understood correctly what SettableFuture does)

> @Hocuri which app did you use to share both text and images at the same time?

WhatsApp, should be possible with Telegram or other messengers too.

> in general I think this part is too complex and hard to follow.

Yes, maybe I should rearrange the code a little more, I just hope that I won't introduce a new bug again. Maybe I should make an integration test for this.

src/org/thoughtcrime/securesms/ConversationActivity.java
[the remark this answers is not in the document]
cyBerta: If you split initializeDraft into initializeDraftFromIntent() (`mailto:`) and initializeDraftFromDatabase() you shouldn't need that change anymore.

cyBerta:
> But here the returned boolean is not checked (at least if I understood correctly what SettableFuture does)

https://github.com/deltachat/deltachat-android/blob/d10891311fa706827f3ec50263c1298a7537332b/src/org/thoughtcrime/securesms/ConversationActivity.java#L1166

the returned boolean is saved in `isShareDraftInitialized` and used to handle the activity backstack correctly.

PR #1541, opened 2020-08-11
Update issue templates

r10s:


r10s:
just a test, using the issue generator, want to check that we did not specify the label in a wrong way as suggested at https://github.com/kubernetes/website/issues/13615

see https://github.com/deltachat/.github/pull/1

PR #1543, opened 2020-08-12
Fix sharing again 2

cyBerta:
proposal to simplify the code. Since we're saving drafts to the database while `handleSharing()` and do `immediatelyRelay` for all cases we do forwarding  (see ConversationActivity for the details) all we need to do is to distinguish between sendto: links or loading drafts from the database in `initializeDraft()`

@Hocuri Unfortunately I lost my phone so I don't have telegram or whatsapp. Signal/wire doesn't allow to share both text and images. Could you please try if sharing both works with this PR?

I also removed the AsyncTask which was only there to load a message object. I don't think this is needed and it seems to be a relict from old Signal code.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -125,6 +125,7 @@ import java.util.Date;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
 
+import static org.thoughtcrime.securesms.NewConversationActivity.MAILTO;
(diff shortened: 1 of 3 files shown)


Hocuri:
> Could you please try if sharing both works with this PR?

Works fine, didn't test anything else. I'll review tomorrow, my room is getting too hot to concentrate.

r10s:
great! it is also pretty hot here in the north :)

this pr superseeds #1529 then, or?

Hocuri:
> this pr replaces #1529 then, or?

Yes.

PR #1547, opened 2020-08-13
Make links in info messages clickable

### LINKED ISSUE #1546: Make links in info messages clickable
When email sending fails with a permanent error, a message with server error is added to the chat. Yandex puts a link to captcha there: https://github.com/deltachat/deltachat-core-rust/blob/526e76c59faaec5b0342d6de5a7c67b8668237f3/src/job.rs#L282

Hocuri:
Fix  #1546


link2xt:
How about the info message itself that is added to the chat, not the info dialog?

r10s:
ahh, it is about info-messages in-chat, not about message-info that is opened on a long-click on any message :) 

but this pr is probably also helpful as it makes links shown for errors in messages clickable.

but we should keep the issue #1546 open.

PR #1551, opened 2020-08-13
Update target sdk

### LINKED ISSUE #1522: target at least Android 10 (API level 29)
message from the playstore:
```
Warnings
Your app currently targets API level 28 and must target at least API level 29 to ensure it is built on the latest APIs optimized for security and performance.

From August 2020, new apps must target at least Android 10 (API level 29).
From November 2020, app updates must target Android 10 (API level 29).

Tip: Change your app's target API level. [Learn how.](https://developer.android.com/distribute/best-practices/develop/target-sdk.html)
```

cyBerta:
closes #1522


Hocuri:
I think that in #1551 @r10s and I agreed that we want to update all dependencies what updating the target sdk.

Issue #1437, opened 2020-06-22
save debug symbols before stripping


r10s:
for getting a grip on crashes that are reported from time to time, it would be useful to have the debug-symbols that are stripped from the release-version somewhere.

(having an extra "fat" version that includes the debug symbols seems to be of less use as this would require users to reproduce a crash - which is not always easily doable. also it may require users installing an extra version, maybe going the export/import way which may cause additional problems, esp. when having few disk space)

as @link2xt pointed out on irc, the native part of Delta Chat android, `libnative-utils.so`,  is an ELF:

```
> ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), dynamically linked, 
> BuildID[sha1]=ede5bfc42dc08bbf929672906d737f49371b86c9, stripped
> this is what "file" says
> I have unpacked it from the APK
> so all I need is debug symbols for this ELF stored separately until someone reports crash 
> or just an unstripped binary, that is equivalent to having a mapfile
> so before you strip the binary, save a copy somewhere 
```

the "stripping" is done implicitly in https://github.com/deltachat/deltachat-android/blob/master/jni/Android.mk#L20 (or at least, for me it is hidden :) - so, any hint/pr to achieve that would be very useful :)

link2xt:
Found this: https://stackoverflow.com/questions/37575043/not-strip-debug-symbols-ndk-build

If this does not increase the size of binary too much, maybe it's even possible to keep the symbols in the binary to get proper backtraces. Otherwise, save .so somewhere and strip the binary afterwards with `strip`.

link2xt:
Also this thread: https://groups.google.com/forum/#!topic/android-ndk/Q7PlBW2dmBk

r10s:
also came over that:
https://developer.android.com/ndk/guides/ndk-stack.html
(via https://stackoverflow.com/questions/40355200/how-to-avoid-stripping-for-native-code-symbols-for-android-app/40368127)

`ndk-stack` is a tool that proposes to replace the hex-numbers in an android-log by symbol-names and linenumbers - sounds nice. it needs the log and the raw .a object files from eg. obj/local/armeabi-v7a - so, these have to be saved, in my normal environment, they are just overwritten with the next build.

i am also checking  for the release 1.10.3 how much larger an .apk with debug symbols in would be:
- 55647703 bytes for gplay/release/deltachat-gplay-release-1.10.3-stripped-as-usual.apk. 
- ...

[...]

PR #1563, opened 2020-08-14
strip release libraries

elwerene:
Add strip commands to strip libraries to save about (25% unzipped) 12% zipped in release apk

### DIFF
--- a/ndk-make.sh
+++ b/ndk-make.sh
@@ -35,7 +35,7 @@ export CFLAGS=-D__ANDROID_API__=16
 RUSTFLAGS="-C lto=on -C embed-bitcode=yes" \
 TARGET_CC=armv7a-linux-androideabi16-clang \
 cargo +`cat rust-toolchain` build --release --target armv7-linux-androideabi -p deltachat_ffi
-armv7a-linux-androideabi-strip target/armv7-linux-androideabi/release/libdeltachat.a
+arm-linux-androideabi-strip target/armv7-linux-androideabi/release/libdeltachat.a
 
 echo "-- cross compiling to aarch64-linux-android (arm64) --"
 export CFLAGS=-D__ANDROID_API__=21
--- a/ndk-make.sh
+++ b/ndk-make.sh
@@ -35,24 +35,28 @@ export CFLAGS=-D__ANDROID_API__=16
 RUSTFLAGS="-C lto=on -C embed-bitcode=yes" \
 TARGET_CC=armv7a-linux-androideabi16-clang \
 cargo +`cat rust-toolchain` build --release --target armv7-linux-androideabi -p deltachat_ffi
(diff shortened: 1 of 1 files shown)


Hocuri:
lgtm, @r10s in my opinion we can merge this.

r10s:
it's maybe more complicated :) 

we decided in https://github.com/deltachat/deltachat-android/issues/1437 to  spend some extra mb to allow better debugging. the release script in tools/upload-release.sh uploads the symbols together with the apk to  https://download.delta.chat/android/

this way, when we get a crash,  ndk-stack should be able to replace the hex-numbers in an android-log by symbol-names and linenumbers. the thing is quite new, not sure if we ever made use of it.

so, i am wondering, if the armv7a-linux-androideabi-strip removed information we may need for making sense of logfiles.

r10s:
@Hocuri @elwerene does building with this pr work on your machines?

i just tried on mac, 

- first i got an error `./ndk-make.sh: line 38: armv7a-linux-androideabi-strip: command not found` - so at least on mac, the strip is called `arm-linux-androideabi-strip` (without `v7a`)

- fixing that, strip runs through, however, finally i am getting a lot of "undefined reference" errors [[*]](https://gist.github.com/r10s/59ad60188a73d348b15868079999a6f6)

elwerene:
I fixed the strip command. It was wrong on linux as well. I don't have any undefined reference errors.

r10s:
hm, for me, on mac, still all dc_* references are missing, log: https://gist.github.com/r10s/a17e6cca3348133011c210d9faa3405b

elwerene:
Oh nevermind, I just checked the output of ndk-build is stripped anyways. I used some rust on mobile before, but we didn't use ndk-build, we directly used the .so files produced by rust.
```
file libs/*/*.so
libs/arm64-v8a/libnative-utils.so:   ELF 64-bit LSB shared object, ARM aarch64, version 1 (SYSV), statically linked, BuildID[sha1]=4a07be890c0d578efbcec2d9a4b7b509b91f3fa0, stripped
libs/armeabi-v7a/libnative-utils.so: ELF 32-bit LSB shared object, ARM, EABI5 version 1 (SYSV), statically linked, BuildID[sha1]=cb181cc4f1e7ffc37cf3de1e9fd8bb5e4850fd2b, stripped
libs/x86_64/libnative-utils.so:      ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), statically linked, BuildID[sha1]=6a4c6f87a1c4a6907a7ca8454233865f674bdbf3, stripped
libs/x86/libnative-utils.so:         ELF 32-bit LSB shared object, Intel 80386, version 1 (SYSV), statically linked, BuildID[sha1]=95a3b83b2038db3c920f47c4a52cd79358b532b2, stripped
```

PR #1567, opened 2020-08-17
Linkify commands

link2xt:
This makes bot commands similar to `/echo` clickable.

Patch from @adbenitez

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -506,7 +506,6 @@ public class ConversationItem extends LinearLayout
     if (!shouldLinkifyAllLinks) {
       return messageBody;
     }
-    Linkify.addLinks(messageBody, Linkify.EMAIL_ADDRESSES|Linkify.WEB_URLS|Linkify.PHONE_NUMBERS);
     boolean hasLinks = false;
 
     Pattern cmdPattern = Pattern.compile("(?<=^|\\s)/[a-zA-Z][a-zA-Z@\\d_/.-]{0,254}");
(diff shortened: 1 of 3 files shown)


src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -500,15 +503,44 @@ private void setContactPhoto() {
...
+    boolean hasLinks = false;
+
+    Pattern cmdPattern = Pattern.compile("(?<=^|\\s)/[a-zA-Z][a-zA-Z@\\d_/.-]{0,254}");
+    hasLinks = Linkify.addLinks(messageBody, cmdPattern, "cmd:", null, null) || hasLinks;
+
+    Linkify.TransformFilter urlFilter = (match, url) -> {
+      if (url.startsWith("http://") || url.startsWith("https://")) {
+        return url;
+      } else {
+        return "http://" + url;
```
cyBerta, line 518: shouldn't we force https here?

Hocuri: http is the standard and if you type `github.com` then your browser will use http. (you will then be redirected to https://github.com)

So, technically http is correct but as very few sites still use http, I agree with you, we should actually use https, at the cost of breaking a very little percentage of links.

@link2xt @adbenitez what do you think?

hpk42:
On Wed, Aug 19, 2020 at 00:32 -0700, Hocuri wrote:

> @Hocuri commented on this pull request.
> 
> 
> 
> > @@ -500,15 +503,44 @@ private void setContactPhoto() {
>    }
>  
>    private SpannableString linkifyMessageBody(SpannableString messageBody, boolean shouldLinkifyAllLinks) {
> -    boolean hasLinks = Linkify.addLinks(messageBody,
> -        shouldLinkifyAllLinks ? Linkify.EMAIL_ADDRESSES|Linkify.WEB_URLS|Linkify.PHONE_NUMBERS : 0);
> +    if (!shouldLinkifyAllLinks) {
> +      return messageBody;
> +    }
> +    boolean hasLinks = false;
> +
> +    Pattern cmdPattern = Pattern.compile("(?<=^|\\s)/[a-zA-Z][a-zA-Z@\\d_/.-]{0,254}");
> +    hasLinks = Linkify.addLinks(messageBody, cmdPattern, "cmd:", null, null) || hasLinks;
> +
> +    Linkify.TransformFilter urlFilter = (match, url) -> {
> +      if (url.startsWith("http://") || url.startsWith("https://")) {
> +        return url;
> +      } else {
> +        return "http://" + url;
> 
> http is the standard and if you type `github.com` then your browser will use http. (you will then be redirected to https://github.com)
> 
> So, technically http is correct but as very few sites still use http, I agree with you, we should actually use https, at the cost of breaking a very little percentage of links.
> 
> @link2xt @adbenitez what do you think?

yes, default to https -- one can still explicitely enter http-links if needed.

PR #1570, opened 2020-08-17
Linkify cmds2

Hocuri:
@link2xt, @adbenitez Some suggestions for your pr #1567

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -506,7 +506,6 @@ public class ConversationItem extends LinearLayout
     if (!shouldLinkifyAllLinks) {
       return messageBody;
     }
-    Linkify.addLinks(messageBody, Linkify.EMAIL_ADDRESSES|Linkify.WEB_URLS|Linkify.PHONE_NUMBERS);
     boolean hasLinks = false;
 
     Pattern cmdPattern = Pattern.compile("(?<=^|\\s)/[a-zA-Z][a-zA-Z@\\d_/.-]{0,254}");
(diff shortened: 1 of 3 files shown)


src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -503,17 +503,13 @@ private void setContactPhoto() {
...
     }
 
-    boolean hasLinks = false;
-
     Pattern cmdPattern = Pattern.compile("(?<=^|\\s)/[a-zA-Z][a-zA-Z@\\d_/.-]{0,254}");
-    hasLinks = Linkify.addLinks(messageBody, cmdPattern, "cmd:", null, null) || hasLinks;
+    boolean hasLinks = Linkify.addLinks(messageBody, Linkify.EMAIL_ADDRESSES|Linkify.WEB_URLS|Linkify.PHONE_NUMBERS);
 
-    hasLinks = Linkify.addLinks(messageBody, Patterns.EMAIL_ADDRESS, "mailto:", null, null) || hasLinks;
-    hasLinks = Linkify.addLinks(messageBody, Patterns.WEB_URL, "http://", new String[]{"http://", "https://"},  null, null) || hasLinks;
```
Hocuri, line 515: This method can't be used in Android M (6) and older. (note the extra `new String[]{"http://", "https://"}` array argument; you can't do this on older devices).

I now just went back to the old approach of `Linkify.addLinks(messageBody, Linkify.EMAIL_ADDRESSES|Linkify.WEB_URLS|Linkify.PHONE_NUMBERS);`, does this have any drawback?

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -503,17 +503,13 @@ private void setContactPhoto() {
 
   private SpannableString linkifyMessageBody(SpannableString messageBody, boolean shouldLinkifyAllLinks) {
     if (!shouldLinkifyAllLinks) {
-	return messageBody;
+      return messageBody;
```
Hocuri, line 506: Spacing

src/org/thoughtcrime/securesms/util/LongClickCopySpan.java
```diff
@@ -29,10 +30,6 @@
   private int highlightColor;
   private int chatId;
 
-  public LongClickCopySpan(String url) {
-    this(url, 0);
-  }
-
```
Hocuri, line 35: Unused

src/org/thoughtcrime/securesms/util/LongClickCopySpan.java
```diff
@@ -52,22 +49,20 @@ private void openChat(Activity activity, DcContact contact) {
...
-	    DcContext dcContext = DcHelper.getContext(activity);
-	    dcContext.sendTextMsg(this.chatId, cmd);
-	}
-	catch(Exception e) {
-	    e.printStackTrace();
-	}
+      try {
+        String cmd = url.substring(PREFIX_CMD.length());
+        ConversationActivity activity = (ConversationActivity) widget.getContext();
+        activity.setDraftText(cmd + " ");
```
Hocuri, line 55: I am only setting the text as a draft to avoid accidental messages and to enable arguments (like /xkcd **30**).

link2xt: From @adbenitez 

> about setting as draft instead of sending directly, I think telegram sends them directly, for arguments there is dynamic commands like 
/xkcd_30
and anyway in long click the command is copied to clipboard

src/org/thoughtcrime/securesms/ConversationItem.java
link2xt: From @adbenitez 
> this has drawbacks, the linkify is lost if this come after the command linkify and if you put it before then numbers in commands are linked as phone numbers etc

src/org/thoughtcrime/securesms/util/LongClickCopySpan.java
link2xt: I still think copying them is ok, sending the message on behalf of the user might be unexpected.

Hocuri:
@adbenitez I applied your suggestion for links. But now if your tap on an email address, only the part until the @ will be highlighted. Likewise, for numbers in commands the numbers will not be highlighted. It somehow recognizes that this might be another link and then removes the highlighting (which is weird because the whole link can be clicked and is underlined). It is only an optical issue but still not nice. Not sure if this is an improvement. 

Also gonna write a pm.

r10s:
i suggest to merge that in after the upcoming bugfix release (1.12.3 then)

Hocuri:
@r10s ok (I'll merge this but not into master but into the other pr)

r10s:
seems i was commenting in the wrong pr :)

PR #1562, opened 2020-08-14
Almost fix #1544 image rotation not working sometimes

Hocuri:
Almost fix #1544 image rotation not working sometimes.

I did not find any better way to do this. Now it is not crashing anymore but it jumps again.

PR #1576, opened 2020-08-17
[WIP] Properly measuring view size to fix rotation.

### LINKED ISSUE #1544: image rotation not working sometimes
[deltachat-log-20200812-091030.txt](https://github.com/deltachat/deltachat-android/files/5062858/deltachat-log-20200812-091030.txt)
[deltachat-log-20200812-091030.txt]

Seen in Android 9 phone with the Google keyboard and Delta Chat 1.12.1 on redmi note 8 in MIUI 11 on the last patch, maybe it are the Android 9  

**How do the bug:** 
- open Delta Chat on landscape mode
- go to chat with a Image(in my case .png)
- put the keyboard in fullscreen likewise you are trying to do a message
- rotate to portrait again

angelo-fuchs:
This pr fixes #1544 by properly measuring the size of the image not only the first, but also the second time (or whenever a newer, larger size comes along).

@Hocuri please do review the change.
As I can only trigger it on an emulator I'm uncertain if this works on a regular device as well.

Maybe we should fix your PR as well to avoid future crashes.
Otoh we would not have know about this issue if we had caught it.

### DIFF
--- a/src/org/thoughtcrime/securesms/components/ScaleStableImageView.java
+++ b/src/org/thoughtcrime/securesms/components/ScaleStableImageView.java
@@ -13,6 +13,7 @@ import org.thoughtcrime.securesms.mms.GlideApp;
 import org.thoughtcrime.securesms.util.Util;
 
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
 
@@ -38,7 +39,8 @@ public class ScaleStableImageView
 
     private Drawable defaultDrawable;
     private Drawable currentDrawable;
-    private Map<String, Drawable> storedSizes = new HashMap<>();
+
+    private boolean keyboardShown = false;
 
     public ScaleStableImageView(Context context) {
         this(context, null);
@@ -54,27 +56,25 @@ public class ScaleStableImageView
 
     @Override
     public void setImageDrawable(@Nullable Drawable drawable) {
+        if(defaultDrawable != null && defaultDrawable.equals(drawable)) {
+            Log.i(TAG, "revoked new drawable, it was the same");
+            return;
+        }
+        Log.i(TAG, "new drawable");
         defaultDrawable = drawable;
-        storedSizes.clear();
         overrideDrawable(defaultDrawable);
     }
 
     private void overrideDrawable(Drawable newDrawable) {
+        Log.i(TAG, "override drawable");
         if (currentDrawable == newDrawable) return;
         currentDrawable = newDrawable;
         super.setImageDrawable(newDrawable);
     }
 
-    private int landscapeWidth = 0;
-    private int landscapeHeight = 0;
-    private int portraitWidth = 0;
-    private int portraitHeight = 0;
-    private boolean keyboardShown = false;
-
     @Override
     protected void onSizeChanged(int width, int height, int oldWidth, int oldHeight) {
         if (width == 0 || height == 0) return;
-        final String newKey = width + "x" + height;
         int orientation = getResources().getConfiguration().orientation;
         boolean portrait;
         if (orientation == ORIENTATION_PORTRAIT) {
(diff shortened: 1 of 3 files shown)


Hocuri:
Actually, with this pr, when I enter a chat normally (in portrait) and then open the keyboard, the image jumps again. So, that's a regression.

angelo-fuchs:
On newer code
an odd regression
ancient fiends
find new expression
Burma-Shave

So, now the jumping is gone when entered from the chat list in portrait.
But: Now when entering chat in landscape, opening keyboard, turning to portrait: The bg image is scaled and jumps to normal once the keyboard is down.

This keeps haunting me.

Hocuri:
>  But: Now when entering chat in landscape, opening keyboard, turning to portrait: The bg image is scaled and jumps to normal once the keyboard is down.

In this case You don't know how big the view is going to be without a keyboard when the chat is entered. So, fixing this would be very hard and this is a very rare case.

angelo-fuchs:
I was unhappy with the previous solution so I made a new one. This only works for custom images at the moment (as soon as we update the background images to be larger it works for them too).

This pins the image at the top center position and crops away from there. So, no more jumping around, no need for expensive rescaling (except when setting the image), etc.

Much simplified code, but if the element of interest on your picture is in the (vertical) middle it gets cut off when you rotate the phone horizontally. So this is not yet a sufficient solution (as most users would use a picture of their spouse or similar and that person usually is in the middle of the picture) and I will need to adapt for the landscape mode.

r10s:
is this pr still valid? #1544 is already closed by https://github.com/deltachat/deltachat-android/pull/1562.

angelo-fuchs:
@r10s #1562 is a workaround that fixes the crash (by accepting that the background jumps again) but not the underlying problem.

This PR fixes the underlying problem for good.

r10s:
thanks for information - and thanks for taking care.

src/org/thoughtcrime/securesms/components/ScaleStableImageView.java
```diff
@@ -54,27 +56,25 @@ public ScaleStableImageView(Context context, @Nullable AttributeSet attrs, int d
 
     @Override
     public void setImageDrawable(@Nullable Drawable drawable) {
+        if(defaultDrawable != null && defaultDrawable.equals(drawable)) {
+            Log.i(TAG, "revoked new drawable, it was the same");
+            return;
+        }
+        Log.i(TAG, "new drawable");
```
cyBerta, line 63: the logs can be removed, here and further down

src/org/thoughtcrime/securesms/components/ScaleStableImageView.java
```diff
@@ -85,88 +85,33 @@ protected void onSizeChanged(int width, int height, int oldWidth, int oldHeight)
             Log.i(TAG, "orientation was: " + orientation);
             return; // something fishy happened.
         }
-
         if (!(defaultDrawable instanceof BitmapDrawable)) {
             return; // need Bitmap for scaling and cropping.
         }
 
-        measureViewSize(width, height, oldWidth, oldHeight, portrait);
+        Log.i(TAG, "view size. w: " + width + " h: " + height + " p: " + portrait);
```
cyBerta, line 92: maybe comment these out for now (instead of removing them completely)?

angelo-fuchs:
@cyBerta @r10s I'm unsure who made our default background images (and how, they seemed tiled but I can't find the tiling edge), so can we have a 4.000 x 4.000 px bg image? Then it will work with the code I wrote here in all situations (where the larger side of the screen is smaller then 4k px).

I'm much happier with this code then the one we had before.
No more complicated measuring, failing attempts at caching the image, etc. and we only need one dark and one bright default image.

The way it works is like this:
During display the image is only ever cropped, not scaled.
In portrait we start at the top-center and crop the image from there.
In landscape we start at the middle-left and crop from there.

In case of custom images:
To make sure the original "middle" of the image is shown we take the original image as set in ChatBackgroundAcivity and scale it so that the shorter edge of the image fits the longer edge of the screen and then crop it squared.

I will need to rebase this branch on master before we can continue, though.

src/org/thoughtcrime/securesms/components/ScaleStableImageView.java
angelo-fuchs: I removed all the debugging comments in the cleanup commit.

Hocuri:
@angelo-fuchs the background is from here: https://support.delta.chat/t/new-chat-background/226. I made the dark background using GIMP's "Invert value" and added it in https://github.com/deltachat/deltachat-android/pull/1407. In https://github.com/deltachat/deltachat-android/pull/1424 landscape was added.

Hocuri:
Also, we have an svg in the Interfaces repo: https://github.com/deltachat/interface/blob/59b46a30a6e825a728f068a2c52d38e2750a2daa/background/background_hd.svg

angelo-fuchs:
Closing this, as there could be unforseen changes to the view size and so precalculating could lead to errors. The current master works good enough.

PR #1584, opened 2020-08-19
set notification group only for Android API 24+

### LINKED ISSUE #1525: Old Android not getting notified
Os: Android
Delta Chat version :1.11.0
Android version: 4.2.2 or 4.4.2 maybe more
Expected Behavior: on get message from server, the user receive a notification
Actual Behavior: the notification does not appear, with or without the reliable background connection setting

In the 1.8.1 it works ok but later version it are not working fine

cyBerta:
fixes #1525

### DIFF
--- a/src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
+++ b/src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
@@ -349,11 +349,15 @@ public class NotificationCenter {
                     .setColor(context.getResources().getColor(R.color.delta_primary))
                     .setPriority(Prefs.getNotificationPriority(context))
                     .setCategory(NotificationCompat.CATEGORY_MESSAGE)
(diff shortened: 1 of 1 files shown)


cyBerta:
info about grouping notifications and API version can be found here: https://developer.android.com/training/notify-user/group.html

r10s:
great finding!

however, i do not really get why setGroup() causes an issue on API<24 - i though using the NotifcationCompat class just does "nothing" for older APIs not supporting some aspect.

cyBerta:
> great finding!
> 
> however, i do not really get why setGroup() causes an issue on API<24 - i though using the NotifcationCompat class just does "nothing" for older APIs not supporting some aspect.

yeah, it NotifcationCompat _should_ handle these cases, but... notifications are a mess :D
I could both reproduce the bug without that change and see that it is fixed with that change.

r10s (approved):
great!

PR #1587, opened 2020-08-19
Improve keyboard layouts

### LINKED ISSUE #1556: Use slightly different keyboard layout for URL input field(s)
- Operating System (Linux/Mac/Windows/iOS/Android):
Android 10 on Fairphone FP2 (Lineage OS 17.1 build QQ3A.200705.002 from 2020-08-07).

- Delta Chat Version:
1.12.2 & 1.12.3.

- Expected behavior:
When an URL is to be entered in an input field, e.g. the one to enter a WebRTC server, the same keyboard layout provided in URL input fields in web browsers should be offered (with the slash in first place).

- Actual behavior:
The keyboard layout differs, it is not the one used in URL input fields in web browsers. To type the slash character in "https://", the user must switch the layout.

-- Focus an URL input field, e.g. the one an URL to a video chat (WebRTC) server is to be entered.
-- Focus an URL input field of a web browser and compare both keyboard layouts with each other.

- Screenshots:

-- Keyboard layout used in web browsers for URL input field(s):
grafik

-- Keyboard layout used in Delta Chat for URL input field(s):
grafik

- Logs:
N/A.

cyBerta:
closes #1556

### DIFF
--- a/src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
+++ b/src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
@@ -13,7 +13,6 @@ import androidx.preference.CheckBoxPreference;
 import androidx.preference.Preference;
 import android.util.Log;
 import android.view.View;
-import android.widget.CheckBox;
 import android.widget.EditText;
 import android.widget.Toast;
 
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
```diff
@@ -187,6 +186,7 @@ public boolean onPreferenceClick(Preference preference) {
       inputField.setHint(R.string.videochat_instance_placeholder);
       inputField.setText(dcContext.getConfig(DcHelper.CONFIG_WEBRTC_INSTANCE));
       inputField.setSelection(inputField.getText().length());
+      inputField.setInputType(TYPE_TEXT_VARIATION_URI);
```
r10s, line 189: this changes only the default keyboard, or?

i am not sure if sth. as `jitsi:https://example.org/?$ROOM` is actually a valid uri.

r10s (approved):
lgtm, i am pretty sure that the uri thing is correct, however.

src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
cyBerta: It doesn't matter if the entered string is really a valid URI. The InputType only defines the keyboard layout, independently of the actual content of the EditText view.

PR #1590, opened 2020-08-20
Fix #1585 User asked for camera permission twice

### LINKED ISSUE #1585: User asked for camera permission twice
- Operating System (Linux/Mac/Windows/iOS/Android):
Android 10 on Fairphone FP2 (Lineage OS 17.1 build QQ3A.200805.001 from 2020-08-14).

- Delta Chat Version:
1.12.3.

- Expected behavior:
When using the camera in a chat, the user is asked for permission once.

- Actual behavior:
When using the camera in a chat, the user is asked for permission twice. The dialogs look quite different to each other.

-- Start a new chat and tap on the camera icon. (The permission to use the camera must not be granted before.)

- Screenshots:

-- This is the first dialog:
grafik

-- This is the second dialog shown afterwards:
grafik

- Logs:
N/A.

Hocuri:
The R.string.perm_explain_need_for_camera_access string is also used elsewhere, so we can't delete it.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -1366,7 +1366,6 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
         Permissions.with(ConversationActivity.this)
                    .request(Manifest.permission.CAMERA)
                    .ifNecessary()
-                   .withRationaleDialog(getString(R.string.perm_explain_need_for_camera_access), R.drawable.ic_photo_camera_white_48dp)
                    .withPermanentDenialDialog(getString(R.string.perm_explain_access_to_camera_denied))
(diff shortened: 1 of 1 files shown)


r10s:
yip, looks good to me. we ask for camera-permission only directly on user action, so the additional question seems to be superfluous and maybe even annoying. it also does not really help - and by withPermanentDenialDialog() we already have some guidance what do do if permission was denied maybe accidentialy.

r10s (approved):

cyBerta:
The audio permission handling works similar and has 2 dialogs (on Android 10 at least). Maybe this can be added here as well then, to keep the permission handling consistent?

r10s:
yip, i agree, these double-questions are not of much use, esp. when resulting from a direct user action. but we should make sure, we have the withPermanentDenialDialog() active.

PR #1593, opened 2020-08-21
tweak and fix permissions

### LINKED ISSUE #1591: "permanent denial" hints are not shown on export from chat activity
when the user has, maybe by accident, permanently denied the permission to write-access disk, we show this to the user using withPermanentDenialDialog() that shows some information about how the user can change the settings in the system (we cannot do that from the app).

however, this dialog and its hints are shown only from within MediaPreviewActivity - but not from ConversationFragment (the chat activity) - maybe because some event is not forwarded to the fragment.

to preproduce:

- make sure, write-access is not yet granted (eg. new installation or remove the grant in the system settings)
- long-tap some image in the chat activity, select "export", then "continue"
- deny access
- long-tap again, select "export" again, deny access again - this time with "never ask again" enabled

if you now try to export - no further hint is shown. but there should be a hint about how to disable permantent denial. if you try to export from the MediaPreviewActivity (enlarge/gallery view) the hint is in place.

tested on lg with android6.

r10s:
this pr removes double questions for permissions, ensures, there is always a dialog also if the permission is permanently denied and removes unneeded toasts.

moreover, some bugs when trying to catch the onRequestPermissionsResult() are catched (one has to be careful on fragments here)

see the single commits for some more details about the single steps.

fixes #1591


cyBerta (approved):
lgtm

PR #1599, opened 2020-08-25
connect immediately after burner account scan

### LINKED ISSUE #1598: not connected after qr-account scan
on current master: directly after a qr-scan, the app is not connected, probably a startIo missing. once the app is short in background, things work as expected, also for all subsequent starts.

when entering credentials, things work as usual.

r10s:
the startIo() call after scanning a burner account was just missing,
putting the app for a second to background fixes the issue,
so it was maybe not that visible.
the issue was probably introduced on the switch to async,
before, startIo() was not existant.

fixes #1598

### DIFF
--- a/src/org/thoughtcrime/securesms/WelcomeActivity.java
+++ b/src/org/thoughtcrime/securesms/WelcomeActivity.java
@@ -224,6 +224,7 @@ public class WelcomeActivity extends BaseActionBarActivity implements DcEventCen
                 progressUpdate((int)progress);
             }
             else if (progress==1000/*done*/) {
+                dcContext.maybeStartIo();
                 progressSuccess(true);
             }
         }


link2xt:
Maybe put common code under `if (eventId==DcContext.DC_EVENT_CONFIGURE_PROGRESS)`, and check for `manualConfigure` only where the behaviour should be different? There is duplicate code right above, which is probably the source of this error.

r10s:
> Maybe put common code under if (eventId==DcContext.DC_EVENT_CONFIGURE_PROGRESS), and check for manualConfigure only where the behaviour should be different? There is duplicate code right above, which is probably the source of this error.

the duplicate code is in the import-branch, not in the manualConfigure-ones.

so combining manualConfigure and !manualConfigure and check for manualConfigure where needed won't help - also all three cases (ok/progress/error) are handled differently in manualConfigure and !manualConfigure.

and combining import-branch and !manualConfigure - well, not sure, if that would make things more readable. also, that would make adaptions harder and may introduce problems there.

also that would require shifting several things around and easily introduce logic-errors, so i'd vote for the easy one-liner.

link2xt (approved):

link2xt:
Yes, haven't noticed that the logic completely different in these branch.s

r10s:
thanks for approving!

PR #1601, opened 2020-08-25
use auto-linkify together with custom-linkify

### LINKED ISSUE #1600: linkify degradation
with linkify to bot commands such as `/echo` in https://github.com/deltachat/deltachat-android/pull/1567 , existant linkification is slightly degraded, eg.

- `Http://example.org` opens `http://Http://example.org` (instead of `http://example.org`)
- `1.2` is rendered as a link to a phone number (was not before, only some longer numbers became phone numbers)
- `space missing.after full stop` renders a link to `missing.after` (was not before on newer androids)
- maybe more.

the degradation happened because we stop using Linkify.addLinks(EMAIL_ADDRESSES|WEB_URLS|PHONE_NUMBERS) and use some manual regex patten - provided by android, however, it seems some logic is missing.

we can probably fix these issues, which, however, would be some effort. it would be great if we can find a way where we can continue using Linkify.addLinks(EMAIL_ADDRESSES|WEB_URLS|PHONE_NUMBERS) together with a manual pattern.

as described in https://github.com/deltachat/deltachat-android/pull/1570#pullrequestreview-468412214, there are some drawbacks when using both.

i just had a closer look at  Linkify.addLinks(EMAIL_ADDRESSES|WEB_URLS|PHONE_NUMBERS) - it  remove all URLSpan.class - maybe we can use a CmdSpan.class for our own links, i will have a look at that now.

(on master, not yet released, 1.12.3 does not have these issues)

r10s:
unfortunately, auto-linking such as Linkify.addLinks(WEB_URLS...)
replaces custom-links set before by Linkify.addLinks(custom-pattern).

calling custom-linking after auto-linking may be too inaccurate
as in case of bot-commands such as /xkcd_123456
are treated partly as a phone number.

skipping auto-linking is hard as there are lots of things to do to get a
similar experience, see #1600.

finally, we solve the problem by replacing URLSpan by some other span after.
in fact, we do that anyway to get a different long-click behaviour.

the only important thing is that the span-replacement
must not be derived from URLSpan.

fixes #1600
reverts the pattern-approach of #1567 and targets https://github.com/deltachat/deltachat-android/pull/1570#pullrequestreview-468412214

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -81,6 +81,7 @@ public class ConversationItem extends LinearLayout
 {
   private static final String TAG = ConversationItem.class.getSimpleName();
 
+  private static final Pattern CMD_PATTERN = Pattern.compile("(?<=^|\\s)/[a-zA-Z][a-zA-Z@\\d_/.-]{0,254}");
   private static final int MAX_MEASURE_CALLS = 3;
 
   private DcMsg         messageRecord;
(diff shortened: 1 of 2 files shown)


r10s:
@adbenitez i am really happy if these /commands can help on bot-usage :) thanks for a lot for pushing that forward!

PR #1608, opened 2020-08-31
add switch to enable/disable battery optimizations

r10s:
the switch fires the intent to request-ignore-battery-optimization the intent for opening battery-optimisation-settings (the latter on errors or if ignore-battery is already enabled - there is no direct way to revert ignoring).

Screen Shot 2020-09-01 at 01 21 43

targets #1477


r10s:
reworded to a suggestion from @link2xt on irc.

PR #1609, opened 2020-09-01
doc: clarify message in UI

monperrus:


r10s:
not sure.

esp. as we are telling users not much about encryption anyway. is this change the outcome of some users having problems, a forum entry or sth. like that? what do whatsapp/telegram/signal say in this case?

but apart from that, there is more to change than this string - there is the faq, translations, that has to be released on all platforms, the term may be referenced in the forum, in other strings etc.

so, this change will result in some amount of additional work.

in general, btw, we had the rough idea to rename _terms_ only when there is actually some work in the corresponding area.

monperrus:
Hi @r10s 

> as we are telling users not much about encryption anyway

We do, see the main [privacy statement](https://delta.chat/en/gdpr) listed from the home page says "Confidentiality: End-to-end encryption when possible, by implementing Autocrypt Level 1."

>  is this proposal the outcome of some users having problems, a forum entry or sth. like that?

Me spending some long time to understand the message I saw this morning in the UI while discussing with a friend.

r10s:
i'd say, the homepage is a different thing :)

however, let's discuss this in the forum and close this here for now. as mentioned above, it is not alone this single message anyway.

r10s:
doing a grep over "setup" in strings.xml, we are talking about "end-to-end setup" here and there - but often also only about "setup" alone.

monperrus:
yes, and "encrypt" appears 19 times also.

r10s:
yip, i am not saying this is perfectly streamlined :)

r10s:
leaving the issue open, maybe there are some more voices ...

Simon-Laux:
The key(`contact_setup_changed`) says nothing about encryption, so if we want to add 'encryption' to the text we should change the key too.

r10s:
yip, that is a technical detail.

however, to the gist of the pr - i am closing this for now here for reasons mentioned above. if needed we can discuss that on the forum.

PR #1611, opened 2020-09-03
[wip] fix launch if there is an ongoing process

### LINKED ISSUE #1610: do not open chatlist when database is not accessible
in some situations, the database is not accessible, eg. during export or import.

when re-launching Delta Chat during such a process (eg. start export -> go to home screen -> tap on Delta Chat icon), the chatlist is shown in an weird "error mode":

same happened when taping on the "generic foreground service" of https://github.com/deltachat/deltachat-android/pull/1504, however, this is "fixed" there by just doing nothing on tapping the permanent notification.

instead, the app should open the activity with the ongoing process in all these cases.

r10s:
this pr converts RoutingActivity to a real activity instead of being just an alias.

this way, we can forward to the ConversationListActivity only if there is _no_ ongoing process.

todo:
- [x] do not open activities that won't work correctly eg. during import/export
- [x] check for possible side-effects - eg. is there a "flash" or so? EDIT: seems to be fine, RoutingActivity is finish()'d immediately, that seems to have no noticeable impact
- [x] check if the other intent assigned to RoutingActivity alias are still working
- [ ] RoutingActivity does not start ConversationListActivity but just open the existing one 
- [ ] the routing does not work if the activity stack is thrown away (eg. by swiping from the tasklist) - what to do in this case?

fixes #1610 
successor of #1504

some reasoning about the explicit RoutingActivity:  
we could avoid activity-creation also in PassphraseRequiredActionBarActivity::onCreate() by calling finish() there, however, the ConversationListActivity has the singleTask flag, so onCreate() is not called when there is an activity somewhere (which is virtually always the case except for the WelcomeActivities)

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -100,9 +100,8 @@
               android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize"
               android:exported="true" />
 
-    <activity-alias android:name=".RoutingActivity"
-                    android:targetActivity=".ConversationListActivity"
-                    android:exported="true">
+    <activity android:name=".RoutingActivity"
+              android:exported="true">
 
         <intent-filter>
             <action android:name="android.intent.action.MAIN" />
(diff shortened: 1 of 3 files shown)


r10s:
closing, this comes with a behavior change when opening the app with the launcher - instead of starting with chatlist as in the past, the current, maybe deeply nested, activity is opened. might be okay, however, having a standardized way to open the chatlist realiably is probably better (eg. chrome always opens the main window this way, same signal) - either way, this is not a reasonable side effect.

PR #1612, opened 2020-09-04
make 'disappearing messages' a standard-feature

r10s:
this pr removes the requirement to enable the experimental feature 'disappearing messages' before disappearing messages can finally be used.

the experimental feature is available and is tested since about two months and i think, it can become a normal feature now.

we also hesitated to enable the feature by default to ensure, recipients using Delta Chat actually follow the deletion requests, this is now the case since july and since several updates.

counterpart of https://github.com/deltachat/deltachat-ios/pull/919


cyBerta (approved):
lgtm

PR #1614, opened 2020-09-04
close generic foreground-service when hitting 'cancel' button

### LINKED ISSUE #1613: generic-foregroundservice is not closed when "cancel" button is hit
this bug was introduced at #1504

r10s:
fixes #1613


r10s:
nb: the foreground-service would have been closed without this pr as well, however, partly much delayed, esp. as the core lacks some checks/events, see https://github.com/deltachat/deltachat-core-rust/issues/1892 - so ux-wise, it makes some sense to close  foreground-service together with progress-dialog, both when "cancel" is hit.

PR #1616, opened 2020-09-06
add rationale to ignore-battery-optimization request

r10s:
with #1505 we ask the user directly after configure to allow running Delta Chat in background.

while it is more straight-forward to let the system ask the user directly -
as we do for most other permissions -
this is probably not favorable for requesting ignore-battery-optimization:

- other permissions requests are preceded by a clear user intention -
  tap on the camera, record voice etc.
  so it is clear to the user why a permission is needed.
  this is not true at the moment after installation
  (maybe the moment can be improved, btw)

- different systems have completely different text,
  partly focusing on "battery drain" - it is not clear to the user
  what would be the advantage of that

- some guidelines say, user interaction is needed before querying permission.
  there is room for interpretation, of couse,
  however, adding a rational with a "continue" button
  seems to fit better to these guidelines.

the current rationale:

some permission requests wordings (from the system, we cannot influence that part):

  
(left to right: lg/android6, motog/android7, google/android9)

prior discussions at https://github.com/deltachat/deltachat-android/pull/1505
still targets #1477

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -718,6 +718,7 @@
     <!-- android specific strings, developers: please take care to remove strings that are no longer used! -->
     <string name="pref_background_notifications">Background notifications</string>
     <string name="pref_background_notifications_explain">Uses a background connection to your server and requires ignored battery optimizations.</string>
+    <string name="pref_background_notifications_ask">Do you want to receive messages in the background?\n\nThis requires ignored battery optimizations, however, Delta Chat uses few resources and takes care not to drain your battery.</string>
     <!-- disabling "Reliable service" will hide a the maybe annoying permanent-notification with the drawback that new-message-notifications get potentially unreliable -->
     <string name="pref_reliable_service">Reliable background connection</string>
     <string name="pref_reliable_service_explain">Requires a permanent notification</string>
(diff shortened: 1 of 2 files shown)


link2xt:
I think it is better to have only "Ok" button for the rationale, without "Not now" option. Just a dialog with "To maintain connection to your email server and receive messages in the background, Delta Chat needs to ignore battery optimizations. Delta Chat uses few resources and takes care not to drain your battery." and an Ok button, followed by the system dialog asking for permission. Otherwise user can click "Continue" and then "Cancel" anyway.

r10s:
@link2xt yip, maybe cancel is bearable as long we do not ask a quesiton :) added a commit, i think, we should point out in a way that the user has sth. to do in the next step (for the ones reading - for ppl just always clicking the rightmost button it does not matter anyway ;)

i also tried to keep the thing short.

this is how it looks like now:

Screen Shot 2020-09-07 at 13 54 34

btw, i'd like to use the same wording as a device-message, prefixed by the "tap here ..." string.

r10s:
this is how it looks after last commit:

r10s:
if the rationale-dialog on startup is cancelled by the user, we still add a device message as in the past.

the last commits uses the same explanation for the device message (and also makes it a bit nicer :)

Screen Shot 2020-09-08 at 12 38 54

the old text lacks the reasoning completely and just reads "Delta Chat uses few resources and takes care not to drain your battery"

cyBerta (approved):

PR #1617, opened 2020-09-08
fix launch if there is an ongoing process

r10s:
second attempt of https://github.com/deltachat/deltachat-android/pull/1611

this pr closes many cases of https://github.com/deltachat/deltachat-android/issues/1610, esp. all on onboarding.

~~leaving the issue on export open - this is currently hard to fix as i do not found a way to cancel the intent given to onNewIntent() - before the back stack is modified. all things i've tried raises more problems than it solves. the prober solution would be to decouple things completely from activities, but this is quite some work and maybe not worth the effort. also, the new tar-export has probably a much smaller window where there database is not accessible.~~ EDIT: the export-issue is fixed for now by aborting the export when the app is relauched, to make this clear, a toast is shown.

### DIFF
--- a/src/org/thoughtcrime/securesms/service/GenericForegroundService.java
+++ b/src/org/thoughtcrime/securesms/service/GenericForegroundService.java
@@ -47,6 +47,7 @@ public final class GenericForegroundService extends Service {
   private static final AtomicInteger NEXT_ID = new AtomicInteger();
   private static final AtomicBoolean CHANNEL_CREATED = new AtomicBoolean(false);
 
+  private static int startedCounter = 0;
 
   private final LinkedHashMap<Integer, Entry> allActiveMessages = new LinkedHashMap<>();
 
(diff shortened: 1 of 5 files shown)


src/org/thoughtcrime/securesms/service/GenericForegroundService.java
```diff
@@ -47,6 +49,7 @@
   private static final AtomicInteger NEXT_ID = new AtomicInteger();
   private static final AtomicBoolean CHANNEL_CREATED = new AtomicBoolean(false);
 
+  private static int startedCounter = 0;
```
cyBerta, line 52: why don't you use a bool here? It's never reset to 0 as far as I can see.

r10s: it is increased on adding a foreground-service and decreased on removing one (in theory, there could be several foreground-services) if it is 0 again, we know, there is nothing ongoing. a simple boolean would not allow using several foreground-services.

cyBerta (approved):

r10s:
thanks for review and for merging!

PR #1625, opened 2020-09-15
Add EditorConfig

link2xt:
It documents indentation style for source files.

See https://editorconfig.org/ for more.


r10s:
is suggest to format .c files the same way as .rs files, with 4 spaces indent, there is no open pr with dc_wrapper.c, so it seems to be a good timing :)

r10s (approved):
great!

link2xt:
In src/com/b44t/messenger indentation is 4 spaces, maybe it needs its own non-root .editorconfig, or style change.

PR #1627, opened 2020-09-17
tune down usage of screen-lock

r10s:
for various reasons,
screen-lock will be removed completely in one of the next versions.

- if screen-lock was enabled before, add a warning
  and keep the options (user can switch screen-lock on and off
  until it will be finally removed)

- if screen lock was disabled before,
  hide the corresponding options,
  (the user will not be able to enable screen-lock anymore)

reasons for removing are:

- hard maintainability on different android versions, see discussions at
  https://github.com/deltachat/deltachat-android/issues/1208 and
  https://github.com/deltachat/deltachat-android/pull/1319 and offline.
  ftr, also the used api createConfirmDeviceCredentialIntent() is deprectated,
  so another set of api calls would be needed in the soon future.

- compared to the effort, few security is added,
  the function is only available if the system is protected anyway,
  one just have to repeat the system secret.

- might even worsen overall security - if app login secret is peeked,
  and attacker will also have the system secret.

- the functions stands in the way of moving forward to a
  cross-platform solution (however, not sure if this will really come)

this is what the device-message looks like:
Screen Shot 2020-09-17 at 16 07 33
it will be shown only to users that have the function enabled before,
so i do not think it is worth making that localizable -
also as the reasoning might be hard to translate
for ppl not deep into the issue.

targets #1208

### DIFF
--- a/res/xml/preferences_app_protection.xml
+++ b/res/xml/preferences_app_protection.xml
@@ -13,7 +13,7 @@
         android:summary="@string/pref_incognito_keyboard_explain"
         android:title="@string/pref_incognito_keyboard" />
 
-    <PreferenceCategory android:title="@string/screenlock_title">
+    <PreferenceCategory android:key="pref_android_screen_lock_category" android:title="@string/screenlock_title">
 
         <org.thoughtcrime.securesms.components.SwitchPreferenceCompat
             android:defaultValue="false"
--- a/src/org/thoughtcrime/securesms/ApplicationPreferencesActivity.java
+++ b/src/org/thoughtcrime/securesms/ApplicationPreferencesActivity.java
@@ -176,8 +176,6 @@ public class ApplicationPreferencesActivity extends PassphraseRequiredActionBarA
 
       this.findPreference(PREFERENCE_CATEGORY_NOTIFICATIONS)
           .setSummary(NotificationsPreferenceFragment.getSummary(getActivity()));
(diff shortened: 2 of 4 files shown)


Hocuri (approved):

m1k3s0:
Any chance this could be revisited? Mobile devices have biometric unlock capabilities now so no secrets would be revealed. Many other private messaging apps take advantage of this functionality as an extra layer of security.

r10s:
> Any chance this could be revisited?

there are no plans currently.

most of the points from above are still true, eg. it will be hard to get it right on different android versions for all app features and click paths, there were  lots of issues with that in the past, and they're probably not solved. newer android versions might even have added new ones. if we spend much energy here, it will be missing somewhere else; there are tons of issues and ideas waiting, maybe ones with more impact, this is always a challenge :)

but be assured, we have that in mind and it is discussed again from time to time.

> Mobile devices have biometric unlock capabilities now so no secrets would be revealed. 

this is known - and Delta Chat makes use of it for adding extra protection to dedicated areas of the app - eg. the "password and account" or "add second device".

m1k3s0:
Thanks for the quick response. Do you know if the iOS app has those config locks? I don’t see a way to lock password and account.

r10s:
> Do you know if the iOS app has those config locks? I don’t see a way to lock password and account.

not yet. but there is a PR at "resurrection" and we'll still aim to get that in

m1k3s0:
Great! Thank you so much.

PR #1630, opened 2020-09-23
Adapt configuration error (was already changed in core)

Hocuri:


r10s:
took me a moment to get the context ;)

so, this pr is the successor of https://github.com/deltachat/deltachat-core-rust/pull/1917#discussion_r490462577 where the string `Configuration failed. Error: <details>` was changed to just `Error:  <details>` - plus the suggestion to just only say `<details>`.

i do not get the reasoning about this change in core. i think, the first string is better as it gives some context first - the detail-error string may be full of technical details. also the `Configuration failed. Error: ` will get translated, while the details string won't (which is also good as it is easier to find solutions in search engines this way.

so - what was the reason for the core change at all?

/me maybe missed some discussions about that :)

r10s (changes requested):
apart from the question above, technically this pr contains a probably unwanted submodule change and is we really want the change, we can drop `configuration_failed_with_error` completely an just use existing `error_x`

Hocuri:
If we want to give the user some hint what went wrong, we have to write `login failed` not `configuration failed` because no one knows what configuration is. But as the user just tried to login and this obviously did not work, they will easily guess this themselves.

> while the details string won't

in the most common cases (wrong pw and no internet) they will. And `Error: Please check your email address and password` is better than `Configuration failed. Error: Please check your email address and password`. Also note the new `\n\n` that will make sure there is some blank space so that it looks more clear.

r10s:
in general, i agree. however, the errors are not always that sophisticated. `configuration_failed_with_error` was meant to be used as a _fallback_ if there is only an english error message, maybe a rare one, whatever. this is meant to give some context.

> And `Error: Please check your email address and password` is better than `Configuration failed. Error: Please check your email address and password`

totally agree - but core can just _not_ use the configuration_failed_with_error-prefix in this case.

ui will just display the string received as data2 from DC_EVENT_CONFIGURE_PROGRESS on data1=0 - ui won't prefix the string or so.

Hocuri:
> totally agree - but core can just not use the configuration_failed_with_error-prefix in this case.

I'll do that, good idea.

I still don't think that users know what configuration is. Maybe `Logging in failed. Error:`?

Hocuri:
> > totally agree - but core can just not use the configuration_failed_with_error-prefix in this case.
> 
> I'll do that, good idea.

Actually thinking about it, this is going to be hard because the nice translated error message (or the ugly english one) is created in configure::configure() while the `Configuration failed...` is added in Context::inner_configure() (which calls the former one). And there are so many places where configure::configure() can return, I can't add the `Configuration failed…` everywhere.

r10s:
> I'll do that, good idea.

can you please also reset the string in core to the previous one.

> I still don't think that users know what configuration is. Maybe `Logging in failed. Error:?`

maybe that is better, however, that would be `log in failed` then, or?

however, let's close this pr now and i suggest that you can tweak the string in core first. we can then later adapt translations.

r10s:
> I can't add the Configuration failed… everywhere.

why not?

PR #1632, opened 2020-09-25
Lossless shrink image sizes

TPS:
I ran [ImgBot](httpS://ImgBot.Net) over these, & came up with some savings toward (@ least installed) filesize.


r10s:
nice, thanks for taking care.

i just had a look - in sum, this saves about 100k.

wondering if there are side effects or so - do you have experiences with that tool?

r10s:
ahh - i just noticed that this is a bot that comes along optimizing images?! that is nice, indeed!

TPS:
@r10s Yes, I've used ImgBot on quite a few repos of mine (public & private) & a few others'.  I've not run into nor heard of _any_ ill effects.  I certainly recommend signing the repo up for it if y'all are satified it's good for ongoing maintenance.

r10s:
@TPS thanks for coming back! i could also not see problems. 

also we have manually used some probably comparable tools in the past, oxipng iirc, see https://github.com/deltachat/deltachat-android/pull/1494 or https://github.com/deltachat/deltachat-core-rust/pull/1802

let's give it a try with this pr :)

i also added https://github.com/deltachat/deltachat-android/issues/1638, maybe we can subscribe to the bot.

r10s (approved):
lgtm

PR #1642, opened 2020-09-29
Prevent exception (would be caught but is annoying)

Hocuri:


cyBerta:
@Hocuri I think there is one commit too much in this PR?!

Hocuri:
@cyBerta oops... Fixed.

r10s:
do you know and can you explain what this code should do at all? and is it safe to just remove it? it seems to comes from signal code base.

Hocuri:
@r10s no idea.

Hocuri:
@r10s what about this PR? It would be really good to not have these many crashes in the log anymore.

No idea what KeyboardAwareLinearLayout is for, maybe I can try to remove it and then we see if there are problems after the release. But it would still be nice to merge this now to not have annoying backtraces in the log.

r10s:
yip, we should find a solution for that.

however, i think we should understand a bit better what we are doing here.

i just had a look: Signal seems to use the [same code](https://github.com/signalapp/Signal-Android/blob/5e58f0a21250aee4c4e1208c65be8198c0251b68/app/src/main/java/org/thoughtcrime/securesms/components/KeyboardAwareLinearLayout.java#L124)  - why is it working for them? or do they get the same crashes? or are they doing something slightly different? 

the only difference i noticed, they have removed the logging - so maybe that is the better approach here?

Hocuri:
> why is it working for them?

Pretty sure it is not because with Android 10 the non-SDK API "mAttachInfo" was removed in Android 10 (try searching for `mAttachInfo` on https://developer.android.com/about/versions/10/non-sdk-q). 

> the only difference i noticed, they have removed the logging - so maybe that is the better approach here? might be, that this is not a general Q problem but only happens on some systems?

Of course we could also just remove the logging but this will still unnecessarily throw and catch an exception here which is bad for performance.

> the link in the comment says, mAttachInfo is gone, yes, however, with no addition info, would not be the first time, that things are incomplete in the android-doc.

See android-doc link posted above.

Hocuri:
TL;DR: This PR now properly fixes the height of the emoji view.

> however, i think we should understand a bit better what we are doing here.

I finally found out what this `getViewInset` is needed for (that's the method using `mAttachInfo`). The result is used here:

    final int availableHeight = this.getRootView().getHeight() - viewInset - (!isFullscreen ? statusBarHeight : 0);

... and this is used here:

    final int keyboardHeight  = availableHeight - (rect.bottom - rect.top);

So, in the end, `viewInset` makes the resulting `keyboardHeight` lower. And it turns out that Delta Chat does always think that the keyboard was a little higher than it actually is: The emoji view and the in-chat camera view are both a little too high on Android 10, on Android 7 everything is fine. Signal has this problem, too.

...BUT I found a way to solve this with only some lines (so it was a good thing you insisted on investigating more).

r10s (approved):
great, i just want to ask why we do not need a replacement for Q ;)

Hocuri:
No more filled-up-with-garabage logs!!! :tada: :tada:

PR #1643, opened 2020-10-04
[WIP] Adapt to mailing lists

Hocuri:
Adapt to API changes at https://github.com/deltachat/deltachat-core-rust/pull/1964/

TODO: Bug: The sender names are not shown anymore over the messages


res/values/strings.xml
```diff
@@ -761,6 +761,12 @@
     <string name="a11y_disappearing_messages_activated">Disappearing messages activated</string>
 
 
+    <!-- mailing lists -->
+    <string name="mailing_list">Mailing list</string>
+    <string name="ask_show_mailing_list">Show mailing list with %1$s in Delta Chat?</string>
```
r10s, line 766: `%1$s` is the name of the list, right? maybe better say just `Show mailing list \"%1$s\" in Delta Chat?` then.

Hocuri: That's not possible with the current API design (which is probably a reason to change it):
- The currently only way to get the mailing list name is calling `chat.get_name()`. But as this is a deaddrop message, if we call `message.get_chat()`, we only get the deaddrop chat, not the mailing list chat. This already was a problem for `is_mailing_list()`, for this one I introduced `dc_msg_is_mailing_list()` additionally to `dc_chat_is_mailing_list()`.
- Maybe it would be worth it to introduce an ffi `dc_msg_get_original_chat()` that always returns the actual chat, even if the message is in the deaddrop. This would remove the need for extra functions on `Message` that get the title and whether it's a mailing list.

r10s: i see, get_chat_id() returns artificially DC_CHAT_ID_DEADDROP by an extra condition, iirc, there was some reason for that, not sure :)

src/com/b44t/messenger/DcChat.java
```diff
@@ -37,6 +37,7 @@ public DcChat(long chatCPtr) {
     public native boolean isProtected       ();
     public native boolean isSendingLocations();
     public native boolean isMuted           ();
+    public native boolean isMailingList  ();
```
r10s, line 40:
```suggestion
    public native boolean isMailingList     ();
```

res/values/strings.xml
Hocuri: @link2xt For asking the user about messages in the deaddrop it would be nice to have a possibility to get info about the original chat. What do you think about adding `dc_msg_get_original_chat()`? This would also remove the need for `dc_msg_get_override_sender_name()` and `dc_msg_get_sender_first_name()` without making performance any worse.

We could:
- make `get_chat_id()` return the original chat even if it's the deaddrop; disadvantage: probably breaking change
- introduce `dc_msg_get_original_chat()`, probably better

link2xt: > i see, get_chat_id() returns artificially DC_CHAT_ID_DEADDROP by an [explicit condition](https://github.com/deltachat/deltachat-core-rust/blob/master/src/message.rs#L510), iirc, there was some reason for that, not sure :)

Maybe try to drop this condition and see if any test breaks? Don't see much reason to call `dc_chat_id` on messages in the deaddrop if it does not return anything useful.

Hocuri: On Android we don't have any test, and `get_chat_id()` is only used in UIs. I don't know if there are tests on other UIs (dc_chat_id() is not used in core itself).

In the Android code it has 7 usages we'd have to check, so nothing undoable, and it would definitely be better for the API not to have two functions that get the chat id. OTOH, we'll not only have to do this on Android, and it's a "hidden" breaking change (i.e. compilation doesn't fail but a bug might be introduced). And it's not totally bad to have two functions here, there are legitimate use cases for this (e.g. it's useful to get the maybe-deaddrop chat if for some reason we want to open the chat of a specific message [to navigate to a quote or whatever]).

In essence, I think I'm rather in favor of another function.

r10s: i would also created an issue for that, https://github.com/deltachat/deltachat-core-rust/issues/2166, and i also have the feeling changing the behavior could be a call for troubles.

r10s:
leaving this for reference, successor is in https://github.com/deltachat/deltachat-android/pull/1781

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -300,7 +298,7 @@ protected void onPostExecute(Void result) {
       protected void executeAction(@Nullable Void parameter) {
         for (long chatId : selectedConversations) {
           if (chatId == DcChat.DC_CHAT_ID_DEADDROP) {
-            dcContext.marknoticedContact(getListAdapter().getDeaddropContactId());
+            dcContext.marknoticedChat(DcChat.DC_CHAT_ID_DEADDROP);
```
r10s, line 301: ftr, this is not implemented in core, not sure, however, if that is needed at all.

i think, the idea was to get the contact-request "out of sight", however, implicitly marking as noticed when one taps "archive" or "delete" (case below) might also be a bit weird. not sure.

Hocuri: I deprecated `marknoticed_contact()` in core and removed it from the C bindings as it's not needed for deciding on contact requests anymore; so, we should probably implement this functionality in `marknoticedChat()`

r10s: yip, either adapt or remove this functionality. i commented a bit more detailed at https://github.com/deltachat/deltachat-android/pull/1781 - let's continue the discussion there.

PR #1644, opened 2020-10-05
Add swipe-to-reply

Hocuri:


link2xt:
There is a quote button in the toolbar in device chat. Clicking it opens the keyboard which types into invisible input field.

If there is no input field, there should be no button.

Otherwise everything works perfectly.

Hocuri:
@link2xt fixed

src/org/thoughtcrime/securesms/glide/ContactPhotoFetcher.java
```diff
@@ -31,7 +31,7 @@ public void loadData(Priority priority, DataCallback<? super InputStream> callba
       inputStream = contactPhoto.openInputStream(context);
       callback.onDataReady(inputStream);
     } catch (IOException e) {
-      callback.onLoadFailed(e);
+      //callback.onLoadFailed(e);//TODO dbg
```
link2xt, line 34: What is this todo?

link2xt (approved):
Skimmed the code, tested on my device. Everything works.

src/org/thoughtcrime/securesms/glide/ContactPhotoFetcher.java
Hocuri: Ah yes, I had disabled this line because I suspected it to be noisy. But I'll revert it, maybe I'll redo this in another PR but this needs some thoughts about side effects first.

Hocuri:
I squashed all commits (and removed the submodule), for to see the individual commits see `swipe-to-reply-backup` branch.

PR #1650, opened 2020-10-07
remove unsued starred api

r10s:
successor of https://github.com/deltachat/deltachat-core-rust/pull/1965


adbenitez (approved):
one line less :)

PR #1651, opened 2020-10-07
direct forward to "saved messages"

r10s:
when forwarding to "saved messages",
the ui goes back to the chat and the position in the original chat
and shows a short "saved" hint.

this saves at least one tap (cc @hpk42 :)
and makes saving multiple messages easier
as you stay in context.
btw, this is also how telegram handles things
and makes the saving as easy to use as starring eg. in whatsapp.

forwarding messasges to other chats is not changed.


adbenitez (approved):
nice!

adbenitez:
it actually saves two taps :D

PR #1652, opened 2020-10-08
better multi-device notification removal

### LINKED ISSUE #1646: status-bar Notifications do not disappear when read on another device
this is happening for a long time: when i read messages on desktop, and enter Android Delta Chat list screen later, messages are correctly "not fresh", i.e. read.  However, the android-status-notification are not (all?) removed automatically.  I have to go to them and swipe them away or re-enter the chat.   The android-statusbar-notification even remain if i enter the respective chat manually --- so it seems that in general statusbar-notifications are, once there, not updated-removed other than going to them.

r10s:
this pr removes the system-notifications once we know that there are no more unread messages in a chat.

there is currently no event for that, however, it happens implicitly when one _answers_ in a chat on another device - then we get that message and the last message is no longer seen/fresh. the badge was always removed in this case, this pr now also removes the system notification and improves layout.

just _reading_ the chat on another device won't be sufficient, if we really want that, we need probably to do sth. in core. but even in the current state, it seems to be a good improvement :)

as a side-effect, this pr will save some calls to dc_get_fresh_msg_cnt() an may speed up things beyond https://github.com/deltachat/deltachat-core-rust/pull/1881

closes #1646

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationListItem.java
@@ -80,7 +80,6 @@ public class ConversationListItem extends RelativeLayout
   private DeliveryStatusView deliveryStatusIndicator;
   private ImageView          unreadIndicator;
 
-  private int             unreadCount;
   private AvatarImageView contactPhotoImage;
 
   public ConversationListItem(Context context) {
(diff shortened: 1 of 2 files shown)


adbenitez:
> it happens implicitly when one _answers_ in a chat on another device

I think something similar happens when we reply a message using the notification "reply" button, we write to the chat without entering the chat so the chat has no badge counter but the "last message preview" is in kept in bold

adbenitez (approved):
notification disappear so you can't keep replying to that chat, for example see Pix Art Messenger's way to do it which is nice.
That said, the old way also removed the notification and worse: added an empty "new messages" notification, then removed it, this is way better/faster also doesn't keep the chat bold once your reply
+1

r10s:
@adbenitez thanks for checking and approving this and the other prs! this is very much appreciated 👌

Issue #1653, opened 2020-10-12
add some padding around the remove-reply button


r10s:
the remove-reply-button (the "x") is hard to tap, there should be some padding around it (the visible size is okay)

r10s:
even better is the suggestion of @Hocuri in "Testing Delta Chat" group, just move the "x" right:

- easier to see
- make clear, the whole quote gets deleted, not only the attachment
- the "x" can have more padding and will be easier to tap

<img width=400 src=https://user-images.githubusercontent.com/9800740/95844364-246f9c00-0d49-11eb-8faa-2afd301e1ff9.jpg>

<img width=400 src=https://user-images.githubusercontent.com/9800740/95844370-25a0c900-0d49-11eb-9051-f8a77827e7e8.jpg>

Hocuri:
<img width=400 src=https://user-images.githubusercontent.com/9800740/95844364-246f9c00-0d49-11eb-8faa-2afd301e1ff9.jpg>
The two circles seem to interfere a little IMHO (before they were above each other).

r10s:
> The two circles seem to interfere a little IMHO (before they were above each other).

nah, i think that's fine. at least much better than before :)

maybe we can move the file-icon closer to the quote - also later, it is typically shown closer together:

![alt](https://user-images.githubusercontent.com/9800740/95875134-f94d7280-0d71-11eb-8552-3b0661fd6524.png)

maybe we can even remove it completely (for voice-messages we do not have them)

but as said, it is already an improvement with your fix, no need to target that directly.

PR #1655, opened 2020-10-13
Things that came up in testing group while testing quotes

### LINKED ISSUE #1654: move directly to first-unread message
with https://github.com/deltachat/deltachat-android/pull/1641, a chat is opened at the position of the first unread message.

however, there is a short flickering as the chat is opened at the end first and scrolls up afterwards. would be nice if we can get rid of that.

Hocuri:
See individual commits
fix https://github.com/deltachat/deltachat-android/issues/1653
fix #1654

### DIFF
--- a/src/com/b44t/messenger/DcMsg.java
+++ b/src/com/b44t/messenger/DcMsg.java
@@ -80,9 +80,9 @@ public class DcMsg {
         int msgs[] = dcContext.getChatMsgs(msg.getChatId(), 0, 0);
         int startingPosition = -1;
         int msgId = msg.getId();
-        for(int i=0; i< msgs.length; i++ ) {
-            if(msgs[i] == msgId) {
-                startingPosition = msgs.length-1-i;
+        for (int i = 0; i < msgs.length; i++) {
+            if (msgs[i] == msgId) {
+                startingPosition = msgs.length - 1 - i;
                 break;
             }
         }
--- a/src/com/b44t/messenger/DcMsg.java
+++ b/src/com/b44t/messenger/DcMsg.java
@@ -79,8 +79,9 @@ public class DcMsg {
     public static int getMessagePosition(DcMsg msg, ApplicationDcContext dcContext) {
         int msgs[] = dcContext.getChatMsgs(msg.getChatId(), 0, 0);
         int startingPosition = -1;
+        int msgId = msg.getId();
         for(int i=0; i< msgs.length; i++ ) {
-            if(msgs[i] == msg.getId()) {
+            if(msgs[i] == msgId) {
                 startingPosition = msgs.length-1-i;
                 break;
             }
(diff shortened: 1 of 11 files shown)


src/com/b44t/messenger/DcMsg.java
```diff
@@ -71,6 +73,21 @@ public boolean equals(Object other) {
         return this.getId()==that.getId() && this.getId()!=0;
     }
 
+    /**
+     * If given a message, calculates the position of the message in the chat
+     */
+    public static int getMessagePosition(DcMsg msg, ApplicationDcContext dcContext) {
+        int msgs[] = dcContext.getChatMsgs(msg.getChatId(), 0, 0);
+        int startingPosition = -1;
+        for(int i=0; i< msgs.length; i++ ) {
```
r10s, line 82:
```suggestion
        for (int i = 0; i < msgs.length; i++) {
```

src/com/b44t/messenger/DcMsg.java
```diff
@@ -71,6 +73,21 @@ public boolean equals(Object other) {
...
+            if(msgs[i] == msg.getId()) {
```
r10s, line 83: i would tend to get `msg.getId()` out of the loop and not only rely on the optimizer. a chat with some 1000 messages may result in lots of ffi calls, done before the chat is displayed, on slow devices, this may be some 100 ms.

src/org/thoughtcrime/securesms/MyLayoutManager.java
```diff
@@ -0,0 +1,49 @@
...
+
+import androidx.recyclerview.widget.LinearLayoutManager;
+import androidx.recyclerview.widget.RecyclerView;
+
+import org.thoughtcrime.securesms.util.ViewUtil;
+
+/**
+ * Like LinearLayoutManager but you can set a starting position
+ */
+class MyLayoutManager extends LinearLayoutManager {
```
r10s, line 14: can we have a more describing name for that?

also, overwriting a central function - why is that needed? this may cause troubles on different implementations.

i just had a look - signal uses SmoothScrollingLinearLayoutManager - is that comparable?

Hocuri: No, I guess that Signal uses SmoothScrollingLinearLayoutManager in order to be able to set the speed of smooth scrolling. We need a custom LayoutManager in order to be able to set the starting position ( 5c5e267 ). I don't know how Signal does this but their code has diverged a lot from ours in this area, so that I decided it's easier to implement this myself (would have been at least some 100 lines to pull in)

I thought about renaming, too, but didn't know a good name. SetStartingPositionLinearLayoutManager?

r10s: thanks for explaining. and yes, `SetStartingPositionLayoutManager` looks good to me.  maybe a bit long, however, it is not used that often and make things much clearer.

r10s (approved):
i suggest to rename from MyLayoutManager to SetStartingPositionLinearLayoutManager, however, i already approve, so you can merge directly after rename (in case you agree :)

r10s:
@Hocuri great, thanks for merging! quotes are so much better now!

PR #1659, opened 2020-10-14
Don't use quickContact in the ConversationListItem

Hocuri:
quickContact means that you get to the contact details screen by
clicking the avatar. In the conversations list this is usually disabled,
just for the search it stayed enabled, which was unexpected
(because it was usually disabled ;-) ).

Even worse, in the search clicking the avatar sometimes got me to the
wrong contact details. No idea why, but I think disabling is better
anyway.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationListItem.java
@@ -213,7 +213,7 @@ public class ConversationListItem extends RelativeLayout
     deliveryStatusIndicator.setNone();
 
     setBatchState(false);
-    contactPhotoImage.setAvatar(glideRequests, recipient, true);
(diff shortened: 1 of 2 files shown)


r10s (approved):
lgtm, i was not aware that a tap on the avatar in search does sth. different. i would also consider this as a bug.

@adb: to see the profile of ppl you do not have a chat with, we could add a long-tap option in the normal contact list (i would add that to the three-dot-menu as "View profile"), indeed, we have a similar option on ios.

adbenitez:
> i would add that to the three-dot-menu as "View profile"

that would be easier for users to discover +1

r10s:
> > i would add that to the three-dot-menu as "View profile"
>
> that would be easier for users to discover +1

@adbenitez i did sth. like that at https://github.com/deltachat/deltachat-android/pull/1668

PR #1660, opened 2020-10-14
Faster conversations list load

Hocuri:
it's not a big difference but these are small changes, too.
- Log.i instead of Log.w (seemed to make a difference, maybe only Log.i buffers?)
- Don't set a fallback avatar
  if there is no fallbackavatar the error avatar will be shown, which is the same in our case, but this way we don't have to call getFallbackAvatarDrawable twice


r10s (approved):
strange with Log.w and Log.i - if this is so much time consuming, we can also remove these logs.

however, for now, we do merge this pr :)

Hocuri:
> strange with Log.w and Log.i - if this is so much time consuming, we can also remove these logs.

I didn't double-check this so I may as well have overseen sth or whatever but Log.i fits better there anyway.

PR #1664, opened 2020-10-16
Use viewtype instead of mimetype for presenting messages

Hocuri:
See https://github.com/deltachat/deltachat-core-rust/pull/1973/


link2xt:
Looks like the code comes from here:
https://github.com/signalapp/Signal-Android/blob/f9a9ee6b0cdb694d1d3490e96271863a224a4794/app/src/main/java/org/thoughtcrime/securesms/util/MediaUtil.java#L68

Signal uses mimetypes, but DC has viewtypes.

link2xt (approved):

adbenitez (approved):

r10s (approved):
yeah, much better. there are still some places where the android-code is not a 100% match to the deltachat-objects, but these places become rare :)

PR #1665, opened 2020-10-16
If audio failed to play, open it in external app

Hocuri:
I don't think it is worth it to add support for e.g. m4a but we can let the user open it in another app so that they can still listen to it.


adbenitez (approved):
a friend was complaining about this recently, great!

r10s (approved):
lgtm

PR #1667, opened 2020-10-16
remove dead code

r10s:
a little cleanup ...


hpk42 (approved):
that sounds like an easy one to approve :)

r10s:
> that sounds like an easy one to approve :)

we will see :)

PR #1668, opened 2020-10-17
open profile from contact list

r10s:
this pr adds an "view profile" info-icon to the action-bar that is shown when a contact is long-tapped.

selecting the icon will open the contact's profile.

successor of #1659, getting the "hack" into a "real feature" :)

### DIFF
--- a/res/menu/contact_list.xml
+++ b/res/menu/contact_list.xml
@@ -2,7 +2,7 @@
 <menu xmlns:android="http://schemas.android.com/apk/res/android"
       xmlns:app="http://schemas.android.com/apk/res-auto">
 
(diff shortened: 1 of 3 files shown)


res/menu/contact_list.xml
```diff
@@ -2,7 +2,12 @@
 <menu xmlns:android="http://schemas.android.com/apk/res/android"
       xmlns:app="http://schemas.android.com/apk/res-auto">
 
-    <item android:title="@string/menu_delete_chat"
```
r10s, line 5: this is not related to the pr, but it turned out the tooltip delete_chat is just wrong for deleting contacts :)

adbenitez (approved):
nice!

PR #1669, opened 2020-10-17
reword protection message and make it translatable

r10s:
with this pr, the system messages will look like ...

 

**please regard the icons as placeholders,** they will probably be changed/refined at some point later, this pr is about the wording of the message, so that it can go to the translators.

the "... by ..." suffix cannot be changed for now, this would cause too much troubles. so wordings as "I enabled ..." are out of scope.

suggestions welcome :)

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -594,6 +594,12 @@
     <string name="systemmsg_ephemeral_timer_week">Disappearing messages timer set to 1 week.</string>
     <string name="systemmsg_ephemeral_timer_four_weeks">Disappearing messages timer set to 4 weeks.</string>
 
+    <!-- Translators: Protection disabled/enabled messages will be displayed together with a big "shield" and may be suffixed by systemmsg_action_by_user or systemmsg_action_by_me -->
+    <string name="systemmsg_chat_protection_disabled">Chat protection disabled.</string>
+    <string name="systemmsg_chat_protection_enabled">Chat protection enabled.</string>
+    <!-- Translators: The details are typically shown directly below systemmsg_chat_protection_enabled -->
+    <string name="systemmsg_chat_protection_enabled_explain">From now on, messages shown in this chat are protected even against compromised servers.</string>
+
 
     <!-- screen lock -->
     <string name="screenlock_title">Screen lock</string>
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -1337,6 +1337,12 @@ JNIEXPORT jboolean Java_com_b44t_messenger_DcMsg_isInfo(JNIEnv *env, jobject obj
 }
 
 
+JNIEXPORT jint Java_com_b44t_messenger_DcMsg_getInfoType(JNIEnv *env, jobject obj)
+{
+    return (jint)dc_msg_get_info_type(get_dc_msg(env, obj));
+}
+
+
 JNIEXPORT jboolean Java_com_b44t_messenger_DcMsg_isSetupMessage(JNIEnv *env, jobject obj)
 {
     return dc_msg_is_setupmessage(get_dc_msg(env, obj));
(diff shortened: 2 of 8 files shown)


res/values/strings.xml
```diff
@@ -594,6 +594,12 @@
     <string name="systemmsg_ephemeral_timer_week">Disappearing messages timer set to 1 week.</string>
     <string name="systemmsg_ephemeral_timer_four_weeks">Disappearing messages timer set to 4 weeks.</string>
 
+    <!-- Translators: Protection disabled/enabled messages will be displayed together with a big "shield" and may be suffixed by systemmsg_action_by_user or systemmsg_action_by_me -->
+    <string name="systemmsg_chat_protection_disabled">Chat protection disabled.</string>
```
comradekingu, line 598:
```suggestion
    <string name="systemmsg_chat_protection_disabled">Chat protection off.</string>
```

res/values/strings.xml
```diff
@@ -594,6 +594,12 @@
...
+    <string name="systemmsg_chat_protection_enabled">Chat protection enabled.</string>
```
comradekingu, line 599:
```suggestion
    <string name="systemmsg_chat_protection_enabled">Chat protection on.</string>
```

res/values/strings.xml
r10s: that would not fit when the suffix "by ..." is added. also imho, "enabled" implies some action taken before. this is not true for "on"

res/values/strings.xml
r10s: see comment  above.

res/values/strings.xml
comradekingu: Well, as anything other than a manifestation of itself, (which is weird). "Now" added.

r10s: what's wrong with "Chat protection disabled"?

comradekingu: The distinction between "off" and "disabled" is unimportant(?) so the primary result is more interesting.
Enabled/disabled is mechanical and off-putting. I thought the strings were consistently "on/off", but that got shut down https://github.com/deltachat/deltachat-android/pull/1122/files
Instead it is, both. That is however the type of consistency you will find.

res/values/strings.xml
comradekingu: In my head this could also mean some sort of +voice mode. "Chat now protected" sounds less like that to me.
Is there a short way to explain how?

res/values/strings.xml
hpk42: but the first message says "Chat protection enabled/disabled by XXXX" and so it's reasonable to re-use the same term. 
Bjoern: as noted somehwere else, i think it should be "chat protection enabled by last message" but maybe i am misinterpreting, only briefly in today.

r10s: @hpk42 yip, you're right. the "shown in this chat" is a bit misleading. messages that were in the chat before do not have the protection.

maybe say. 
"Chat protection enabled [by me|username]. Subsequent messages are protected against compromised servers."
or  
"Chat protection enabled [by me|username]. Subsequent messages are protected against compromised servers until protection is disabled again."

to the context: the initial post is about two system-messages, one for enabling, one for disabling. we can also leave out the explaining text completely, however, i think it is good to describe what "Chat protection" means here, similar as what whatsapp is doing - even if not everyone understands what that means.

i'd add a screenshot here, however, this requires the shield first, which is not ready yet (and also not required for the next release where we do not have the on/off switch)

r10s:
@hpk42 @all i updated the initial post and added some screenshots, maybe the context is clearer now :)

okdistribute:
"are protected against compromised servers" the even is unnecessary in my opinion. 

Also what is the word used for login and other parts of the app that deal with the email account? Provider or server or something else?

res/values/strings.xml
comradekingu: Doesn't stating the same thing using the same words take away an opportunity to understand what is going on?

flub:
Is it possible to add a "more info" link to the FAQ page?

r10s:
thanks all for helping - for the upcoming version, we will just do a minimal "Chat protection enabled/disabled [by user]", without icon or further explanation. 

this already ensures, clients will follow the enable/disable messages and the user will notice - while not confusing and mixing things with the still existent "verified groups".

when we finally ship protected-groups, including a switch, we will target the explanation and the icon again.

PR #1673, opened 2020-10-20
"forward" should not go to overflow menu, fix #1656

### LINKED ISSUE #1656: "forward" should not go to overflow menu
on a bit smaller screens, in group-chats, long-tapping a message may move forward to the overflow-menu, esp.

however, esp. as we use that also for "starring", i'd regard that as a major function that should be available with just one tap. "copy to clipboard", "info" and even "delete" are better candidates for the overflow-menu.

Hocuri:
cc @r10s

### DIFF
--- a/res/values/attrs.xml
+++ b/res/values/attrs.xml
@@ -249,7 +249,7 @@
         <attr name="quote_colorSecondary" format="color" />
     </declare-styleable>
 
-    <declare-styleable name="AdaptiveActionsToolbar">
+    <declare-styleable name="ConversationAdaptiveActionsToolbar">
         <attr name="aat_max_shown" format="integer" />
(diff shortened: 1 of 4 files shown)


r10s:
i commented on https://github.com/deltachat/deltachat-android/issues/1656#issuecomment-712722448 :)

r10s (approved):
lgtm, see reasoning in linked issue, thanks @Hocuri.

r10s:
k, let's move that in, it is an improvement.

i still like to have some more cleanup, also i do not like the shifting of the items when doing a multiple-select, an "always there" context menu would help there and moving COPY there looks reasonable to me (whatsapp does the same btw), however, this can be part of another discussion/pr.

PR #1677, opened 2020-10-20
Don't allow quoting system messages and voice message invites

Hocuri:
Swipe-to-reply already was disabled for them, now quoting by long-press
is disabled, too.


r10s (approved):
i would remove the BuildConfig.DEBUG, apart from that, things lgtm.

PR #1679, opened 2020-10-20
Add scripts/install-toolchains.sh

link2xt:


r10s (approved):
lgtm, thanks for cleaning up!

you put the script to `scripts` instead of `tools` because the name is more common and more descriptive?  
that is fine for me, however, in another pr, i think, it makes sense to move also the other scripts to `scripts` then, or?

link2xt:
Actually I just haven't noticed `tools`. `scripts` is more common indeed.

PR #1684, opened 2020-10-22
Copying does not contain the names or "Image - " for single messages

Hocuri:
We have replies for this now

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -406,6 +406,7 @@ public class ConversationFragment extends Fragment
     private void handleCopyMessage(final Set<DcMsg> dcMsgsSet) {
         List<DcMsg> dcMsgsList = new LinkedList<>(dcMsgsSet);
         Collections.sort(dcMsgsList, (lhs, rhs) -> Long.compare(lhs.getDateReceived(), rhs.getDateReceived()));
+        boolean singleMsg = dcMsgsList.size() == 1;
 
         StringBuilder result = new StringBuilder();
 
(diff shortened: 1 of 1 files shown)


adbenitez (approved):
tested, great!

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -415,11 +416,11 @@ private void handleCopyMessage(final Set<DcMsg> dcMsgsSet) {
...
-            if (msg.getFromId() != prevMsg.getFromId()) {
+            if (msg.getFromId() != prevMsg.getFromId() && !singleMsg) {
                 DcContact contact = dcContext.getContact(msg.getFromId());
                 result.append(contact.getDisplayName()).append(":\n");
             }
-            if(msg.getType() == DcMsg.DC_MSG_TEXT) {
+            if(msg.getType() == DcMsg.DC_MSG_TEXT || singleMsg) {
                 result.append(msg.getText());
             } else {
                 result.append(msg.getSummarytext(10000000));
```
r10s, line 426: that would result in copying an empty text if the user tries to copy eg. an image or a voice-message.

before, the image type was copied to the clipboard in the these cases (from summary then).

therefore, we should append the result from msg.getText() only if not empty - else use the summary.

r10s (changes requested):

src/org/thoughtcrime/securesms/ConversationFragment.java
Hocuri: Don't know what the usecase of copying the word "Image" would be but anyway, done (I don't think it's an improvement but it doesn't make anything worse, either)

r10s (approved):
lgtm now :)

Issue #1687, opened 2020-10-23
Don't show messages very thin if small or tall images are sent


Hocuri:
If I send a small (few px) or tall (e.g. screenshot) image with text, then the text area is so slim that rarely two words fit beside each other. Other messengers solve this by cutting tall images at the top and bottom and showing small images bigger.

r10s:
yip, i totally agree, this also annoys me since some time. for calculated cell-width should take higher priority on text, the image should be second priority only. i'd consider that as a bug even.

> Other messengers solve this by cutting tall images at the top and bottom

yip, this is also what i would suggest. tapping on the image will show it in full-size anyway.

>  and sending small images bigger.

not sure what you mean here. sending images should not be changed to solve this issue imho.

Hocuri:
> > and sending small images bigger.
> 
> not sure what you mean here. sending images should not be changed to solve this issue imho.

I meant "showing" :joy: if we have a 50x50 px image, we are showing it 50x50 px -> very thin, others enlarge it

r10s:
i seen :) yip, that makes sense +1

PR #1688, opened 2020-10-24
Wider messages for tall images

Hocuri:
Fix https://github.com/deltachat/deltachat-android/issues/1687

Don't know if such a change should wait until after the release because a bug could be introduced?

Screenshot_20201025-163623
Screenshot_20201025-163611
Screenshot_20201025-163550
Screenshot_20201025-163543


r10s:
> Don't know if such a change should wait until after the release because a bug could be introduced?

i'll have a look :)

thanks a lot for pushing that forward, this is a personally long-awaited fix :)

r10s (changes requested):
i did not look that much at the code but tested the pr more in practice with groups i have on my phone :)

- we should not cut _landscape_ images, they can easily be shown in complete width as before

- we should cut _portrait_ images a bit less, so that the result is a _square_. this is less surprisingly when the image is enlarged and the user has a clue that this is not a _landscape_ image

both cases are also handled by whatsapp/etc in this way.

wrt merging: i would do that after the 1.14 release to allow testing that a bit more.

Hocuri:
> * we should not cut _landscape_ images, they can easily be shown in complete width as before

Was a bug, I meant to make a minimum height but I wanted it to be 50dp, not 100dp (before there also was some limit on how low the height can get and I don' think we should change this)
> * we should cut _portrait_ images a bit less, so that the result is a _square_. this is less surprisingly when the image is enlarged and the user has a clue that this is not a _landscape_ image

That's a little harder to do, as: If the screen is in landscape and we show the image as a square then it will be easily twice as high as the screen and the image does not even fit on the screen. Maybe I'll just cut the image as I am currently doing but only while the device is in landscape mode (which should be the minority of cases by far, anyway)

r10s:
> If the screen is in landscape and we show the image as a square then it will be easily twice as high as the screen and the image does not even fit on the screen.

that's a point.

i can imagine two solutions for that:
1. show portrait images cutted to landscape (as now), use square only if the whole device is in portrait mode
2. do not use the full width for the bubbles (for the text) in device-landscape-mode - this way, also the squares will maybe be small enough. at a first glance this is what whatsapp is doing

but yip, this is not super-urgent to decide, we will do the next release without this pr anyway.

Hocuri:
> 1. show portrait images cutted to landscape (as now), use square only if the whole device is in portrait mode
> 
> 2. do not use the full width for the bubbles (for the text) in device-landscape-mode - this way, also the squares will maybe be small enough. at a first glance this is what whatsapp is doing

Actually, let's just do 1 if the text is longer than, like, 200 chars, 2 otherwise

Hocuri:
> Actually, let's just do 1 if the text is longer than, like, 200 chars, 2 otherwise

Actually actually, 2 would be quite some work to implement and I don't think it's worth it. Currently I only implemented 1.

Hocuri:
@r10s what about merging this now?

r10s (approved):
lgtm now - and yes, a good time for merging :)

PR #1696, opened 2020-10-28
Move docker/cargo-config into Dockerfile ENV directives

link2xt:
Having everything in a single file is easier to manage.


link2xt:
Maybe we should have `ndk-make.sh` configure all these environment variables based on `ANDROID_SDK_ROOT`.

r10s (approved):
i am not using these files directly on my system, but they lgtm :)

PR #1697, opened 2020-10-29
Refactor ndk-make.sh

link2xt:


ndk-make.sh
```diff
@@ -1,60 +1,49 @@
...
-export CFLAGS=-D__ANDROID_API__=21
-RUSTFLAGS="-C lto=on -C embed-bitcode=yes" \
-TARGET_CC=x86_64-linux-android21-clang \
-cargo +`cat rust-toolchain` build --release --target x86_64-linux-android -p deltachat_ffi
-
-echo -- copy generated .a files --
-cd ..
+: "${ANDROID_NDK_ROOT:=$ANDROID_NDK_HOME}"
+: "${ANDROID_NDK_ROOT:=$ANDROID_NDK}"
+if test ! -z "ANDROID_NDK_ROOT"; then
```
r10s, line 33:
```suggestion
if test ! -z "ANDROID_NDK_ROOT"; then
    echo "setting env."
```

ndk-make.sh
```diff
@@ -63,12 +52,73 @@ mkdir -p armeabi-v7a
...
-cp deltachat-core-rust/target/i686-linux-android/release/libdeltachat.a x86
-cp deltachat-core-rust/target/x86_64-linux-android/release/libdeltachat.a x86_64
+
+cd deltachat-core-rust
+
+# fix build on MacOS Catalina
+unset CPATH
+
+if test -z $1; then
+  # Full build
```
r10s, line 62:
```suggestion
  echo "Full build."
```

it is good to know what is going on :)

ndk-make.sh
```diff
@@ -63,12 +52,73 @@ mkdir -p armeabi-v7a
...
+
+  export RUSTFLAGS="-C lto=on -C embed-bitcode=yes"
+  RELEASE="release"
+else
+  # Fast build
```
r10s, line 67:
```suggestion
  echo "Fast, partial build. DO NOT UPLOAD THE APK SOMEWHERE."
```

ndk-make.sh
```diff
@@ -63,12 +52,73 @@ mkdir -p armeabi-v7a
...
+fi
 
 echo -- ndk-build --
+
+cd ..
+# Set the right arch in Application.mk:
+oldDotMk="$(cat Application.mk)"
+
+if test $1; then
+    sed -i "s/APP_ABI.*/APP_ABI := $1/g" Application.mk
```
r10s, line 111: this does not work on mac, results in error "invalid command code A", sed seems to be tricky cross-platform.

maybe we should just generate the whole Application.mk via an `echo` command here? there are only three lines in there.

r10s (changes requested):
nice, much more overview :)

however, this sed-thing does not work, iirc i tried some things already in  #1422 and came to the conclusion it is too hard (for me :) doing that cross-platform.

ndk-make.sh
link2xt: `-i` is really not portable: https://unix.stackexchange.com/questions/13711/differences-between-sed-on-mac-osx-and-other-standard-sed

Can use `ed` or just generate with a here-document.

r10s: whatever :) as said, also `echo "content" >  Application.mk` would be fine (assuming we can create the files on f-droid)

link2xt: Updated to use sed with temporary file, would be nice if you test on MacOs.

ndk-make.sh
```diff
@@ -1,60 +1,50 @@
...
-# [target.i686-linux-android]
-# linker = "PATH_TO_NDK/toolchains/llvm/prebuilt/HOST/bin/i686-linux-android16-clang"
-# [target.x86_64-linux-android]
-# linker = "PATH_TO_NDK/toolchains/llvm/prebuilt/HOST/bin/x86_64-linux-android21-clang"
-# ```
-# then, the following should work:
-# (If you want to speed up compilation, have look at the comment in ndk-make-fast.sh.)
+#
+# If you put this in your .bashrc, then you can directly build and deploy DeltaChat from the jni/deltachat-core-rust directory by typing nmake():
+# nmake() {sh -c 'cd ../..; ./ndk-make.sh arm64-v8a && ./gradlew installFatDebug; notify-send "install finished" }
```
Hocuri, line 26:
```suggestion
# If you put this in your .bashrc, then you can directly build and deploy DeltaChat from the jni/deltachat-core-rust directory by typing nmake():
# (you have to replace arm64-v8a by your architecture)
# nmake() {sh -c 'cd ../..; ./ndk-make.sh arm64-v8a && ./gradlew installFatDebug; notify-send "install finished"' }
```
esp. missing `'` (was probably me who forgot this when writing this text)

link2xt: Pushed a change that starts a subshell with `(...)`

r10s:
great, `sed` works now, full build passes (and this is the most important thing for me and for f-droid :)

trying a partial build, however, results in the following error for me:
```
~/projects/deltachat-android[refactor-ndk-make|✔]$ ./ndk-make.sh armeabi-v7a
starting time: Mon Nov  2 21:00:09 CET 2020
Setting CARGO_TARGET environment variables.
Fast, partial, slow debug build. DO NOT UPLOAD THE APK ANYWHERE.
-- cross compiling to armv7-linux-androideabi (arm) --
error: Found argument '--debug' which wasn't expected, or isn't valid in this context

USAGE:
    cargo build [OPTIONS]

For more information try --help
```

not sure, if this is a general issue or just my local rust installation :)

link2xt:
There is no `--debug` flag indeed, I updated the script.

r10s:
now, on mac, i am getting `error: linker `/Users/bpetersen/android-ndk-r20b/toolchains/llvm/prebuilt/linux-x86_64/bin/armv7a-linux-androideabi16-clang` not found` 

this is not unexpected, the file does just not exist. for mac, the following exports work:#

```
    export CARGO_TARGET_ARMV7_LINUX_ANDROIDEABI_LINKER="${ANDROID_NDK_ROOT}/toolchains/llvm/prebuilt/darwin-x86_64/bin/armv7a-linux-androideabi16-clang"
    export CARGO_TARGET_AARCH64_LINUX_ANDROID_LINKER="${ANDROID_NDK_ROOT}/toolchains/llvm/prebuilt/darwin-x86_64/bin/aarch64-linux-android21-clang"
    export CARGO_TARGET_I686_LINUX_ANDROID_LINKER="${ANDROID_NDK_ROOT}/toolchains/llvm/prebuilt/darwin-x86_64/bin/i686-linux-android16-clang"
    export CARGO_TARGET_X86_64_LINUX_ANDROID_LINKER="${ANDROID_NDK_ROOT}/toolchains/llvm/prebuilt/darwin-x86_64/bin/x86_64-linux-android21-clang"
```

so, the only difference is the path-component `prebuilt/darwin-x86_64/bin` instead of `prebuilt/linux-x86_64/bin`.

maybe we can move that to a variable and set it depending of the existance of the file `${ANDROID_NDK_ROOT}/toolchains/llvm/prebuilt/darwin-x86_64/bin/armv7a-linux-androideabi16-clang`

ndk-make.sh
```diff
@@ -1,60 +1,61 @@
...
-# [target.aarch64-linux-android]
-# linker = "PATH_TO_NDK/toolchains/llvm/prebuilt/HOST/bin/aarch64-linux-android21-clang"
-# [target.i686-linux-android]
-# linker = "PATH_TO_NDK/toolchains/llvm/prebuilt/HOST/bin/i686-linux-android16-clang"
-# [target.x86_64-linux-android]
-# linker = "PATH_TO_NDK/toolchains/llvm/prebuilt/HOST/bin/x86_64-linux-android21-clang"
-# ```
-# then, the following should work:
-# (If you want to speed up compilation, have look at the comment in ndk-make-fast.sh.)
+    if -z "$NDK_HOST_TAG"; then
```
r10s, line 39:
```suggestion
    if test -z "$NDK_HOST_TAG"; then
```

that works on mac, not sure, what `if -z` is for, maybe just a typo or so :)

r10s (approved):
now it works, thanks for including all the mac specials :)

PR #1699, opened 2020-10-30
Insert empty line correctly in res/values-de/strings.xml

gerryfrancis:
The title says it all.


r10s:
thanks for taking care.

as translations are not done via github but via [transifex](https://www.transifex.com/delta-chat/), however, i made the change there and close this pr.

(on the next translation-updated, res/values-de/strings.xml will finally be updated)

gerryfrancis:
Ah, sorry about that, I will do it in Transifex next time. Thanks for correcting this! :+1:

PR #1702, opened 2020-10-31
Handle sendto intents with ShareActivity, not NewConversationActivity

### LINKED ISSUE #1701: DC does not handle NewPipe email intents properly
- Operating System (Linux/Mac/Windows/iOS/Android): Android
- Delta Chat Version: 1.12.5
- Expected behavior: DC should open a chat with the email address that was passed in the intent and pre-fill the text field with the email body that was passed.
- Actual behavior: DC opens the chat list, like it does when it is opened using the app icon in the launcher.
- Steps to reproduce the problem: Use an app that can send email intents (i.e. a "report an error using email") to send an email intent to DC. (more info below, this is most likely **not** a problem caused by another app)
- Screenshots:
- Logs:

Let me tell you how I discovered this issue.

I tried to watch a YouTube video using the NewPipe app (version 0.20.1 from F-Droid). The app failed to load the video and showed me an error message. At the very bottom of the error screen there is a button labelled "report error via email". So I pressed that button and my Android asked me which email client I'd like to use to send the email. I picked DC from the list of installed email clients. Then DC opened and showed me the chat list, which was unexpected behaviour. Instead, I expected DC to open a chat to the email address that was passed to DC via the intent and fill the email text into the text field.

The story continues. Since I wanted to know what was going on, I repeated what I did previously. This time, when my Android asked me which email client to use, I chose FairEmail instead of DC. FairEmail opened the "write an email" screen and filled in the recipient, the subject and the email body using the corresponding data fields of the email intent.

Then I tried triggering an email intent from a web browser (Fennec version 82.1.1 from F-Droid). I created [this gist](https://gist.github.com/errhammr/afe737082c95850fb63a4c36d2e3ff33) to test different `mailto:` URIs. Both DC and FairEmail behaved as expected, filling out the recipient address, subject and body correctly.

To make sure this is **not** an issue **caused by NewPipe**, I then tried a different app to trigger an email intent. I used the "report a problem" feature of the App Unstoppable Wallet (version 0.16.2 from F-Droid). This email intent only contains the recipient email address, no subject and no body. As with NewPipe, DC did not behave as expected and only showed me the chat list instead of opening a chat to the email address that was passed to DC via the intent. I repeated the steps with FairEmail and FairEmail did open the "write an email" screen with the recipient filled out correctly.

As far as I can tell, this is most likely a problem caused by DC because FairEmail seems to handle the very same email intents correctly.

Hocuri:
Fix #1701 

I am not 100% sure if this is the way to go. On first sight, this seems totally legit: ShareActivity should handle `SENDTO` intents, I mean, why should NewConversationActivity do this? But then, there may be a reason why NewConversationActivity was chosen to do the job.

Of course NewConversationActivity could do the same as ShareActivity does or relay it to ShareActivity, but this seems like very bad code style (it currently does relay to ShareActivity sometimes in `handleIntent()` but then forgets to relay all the extras; that's why the recipient is lost).

I think it's best to ping @flub as he also did sharing-related stuff in the past and then wait until after the release.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -77,6 +77,13 @@
         </intent-filter>
 
         <intent-filter>
+            <data android:scheme="mailto"/>
+            <action android:name="android.intent.action.SENDTO"/>
+            <category android:name="android.intent.category.DEFAULT"/>
+            <category android:name="android.intent.category.BROWSABLE"/>
+        </intent-filter>
+
+        <intent-filter>
             <action android:name="android.intent.action.SEND_MULTIPLE" />
             <category android:name="android.intent.category.DEFAULT" />
             <data android:mimeType="audio/*" />
@@ -154,7 +161,6 @@
         <intent-filter>
             <data android:scheme="mailto"/>
             <action android:name="android.intent.action.VIEW"/>
-            <action android:name="android.intent.action.SENDTO"/>
             <category android:name="android.intent.category.DEFAULT"/>
             <category android:name="android.intent.category.BROWSABLE"/>
         </intent-filter>


r10s:
sounds reasonable to me, however, @cyBerta might have additional comments/thoughs.

AndroidManifest.xml
```diff
@@ -154,7 +161,6 @@
         <intent-filter>
             <data android:scheme="mailto"/>
```
cyBerta, line 162: mailto is probably not needed anymore here?

Hocuri:
I created a testing document at https://github.com/deltachat/interface/pull/40 (Jikstra just told me that the Desktop testing documents are also there).

It should test for most of the issues we had in the past with mailto links (unfortunately not this one because it only happens with SENDTO intents, and there seems to be no way to create such an intent using a link).

AndroidManifest.xml
Hocuri: If I remove only this line, then Android Studio is mad at me as there is a `data` element missing. If I remove the whole block then I can't open mailto links with DC at all anymore. And if I then add `mailto` and `action.VIEW` up to ShareActivity, then mailto links can be shared to DC but DC doesn't find the content.

So, I'd actually propose to merge this PR as is because everything is working fine (at least sharing from NewPipe, SuperFreezZ and all the mailto links in the testing document I just created). Maybe code style is not the best as we still have multiple activities caring about sharing but it's the minimal-change solution, hopefully with the lowest risk of breaking something.

r10s: hm, that's weird. i'll have look at that later.

r10s (commented):
i played around with that, and, indeed, `mailto:` links work all without regression. [`mailto-links.md`](https://github.com/deltachat/interface/blob/master/user-testing/mailto-links.md) is a great tool for testing, thanks again for that, btw.

- wrt https://github.com/deltachat/deltachat-android/pull/1702/files#r533377407 and the double `mailto`-intent: that is needed as - maybe unexpectedly - most things are still handled by NewConversationActivity and not ShareActivity

- first, i though there could be a conflict, therefore i was sceptical to merge that in (if two activities listen to the same intent, it might cause new issues), however, this is actually not the case: the two activities handle slightly different actions - Browser fires `android.intent.action.VIEW` and NewPipe seems to fire `android.intent.action.SENDTO` (did not test that, but otherwise that should have worked before)

it is a bit messed up, but not only our fault :) so i agree we can merge that in.

r10s (approved):

PR #1710, opened 2020-11-03
Fix crash in ImageEditor

Hocuri:
For some reason it is not possible to remove `setUri`, when I tried, I
got crashes each time I tried to edit an image.

### DIFF
--- a/src/org/thoughtcrime/securesms/scribbles/ImageEditorFragment.java
+++ b/src/org/thoughtcrime/securesms/scribbles/ImageEditorFragment.java
@@ -95,7 +95,7 @@ public final class ImageEditorFragment extends Fragment implements ImageEditorHu
     super.onCreate(savedInstanceState);
 
     if (imageUri == null) {
-      throw new AssertionError("No KEY_IMAGE_URI supplied");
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/scribbles/ImageEditorFragment.java
```diff
@@ -95,7 +95,7 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
     super.onCreate(savedInstanceState);
 
     if (imageUri == null) {
-      throw new AssertionError("No KEY_IMAGE_URI supplied");
+      imageUri = getArguments().getParcelable(KEY_IMAGE_URI);
```
r10s, line 98: maybe that error sneaked in when updating the editor last year.

in fact, this is roughly the same signal is doing, see https://github.com/signalapp/Signal-Android/blob/1730260343312ff22295ec3e145d54a09beef496/app/src/main/java/org/thoughtcrime/securesms/scribbles/ImageEditorFragment.java#L129 , they handle getArguments() a bit more carefully, however, indeed, it should be always set.

r10s (approved):

PR #1712, opened 2020-11-04
Do not display milliseconds when recording a voice message fix #1711

### LINKED ISSUE #1711: Do not display milliseconds when recording a voice message
It is just a random number displayed after the number of seconds. Refresh seems to happen 4 times per second on my phone.

Hocuri:
Also I have no idea why someone would set UPDATE_EVERY_MS to 173 but with 99 we get a totally smooth counter

### DIFF
--- a/res/layout/conversation_input_panel.xml
+++ b/res/layout/conversation_input_panel.xml
@@ -43,7 +43,7 @@
                 android:layout_height="wrap_content"
                 android:ellipsize="none"
                 style="@style/Signal.Text.Body"
-                android:text="00:00.00"
+                android:text="00:00.0"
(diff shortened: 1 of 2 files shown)


r10s (approved):
great, much smoother, indeed!

> Also I have no idea why someone would set UPDATE_EVERY_MS to 173 

iirc, basic idea was to have _any_ feedback that something is recorded; updating the time only once a second does not feel responsive here. a randomly appearing number, as produced by 173 ms update cycle, just did this job. but the new approach is better - and also uses fewer screenspace, that is also limted, esp. on small screens (i just tested on and 10yo nexus-s, very nice :)

ftr: we never displayed milliseconds, we displayed hundredths, now we display tenth of a second.

PR #1720, opened 2020-11-08
Quotes add on 3

Hocuri:
See the individual commits (hope it's ok that I did not open 4 PRs because this way I could test all 4 changes at once)

The third commit can be reproduced by setting "Font size" to "Small" and "Display size" to "Large" in "Display" settings (this problem was reported on the testing group)


r10s (approved):
lgtm, great fixes.

to fixing several bugs in on pr:

i'd say, in general, it is better to have multiple pr. 

however, if a combined-fix-pr is as good documented and explained as this one and one can easily compare things commit-by-commit as in this pr, an exception is totally reasonable when it helps developing.

PR #1721, opened 2020-11-09
do not use cached value to check for notifications

r10s:
the 'inboxes' value contains the last lines of notification,
however, if the app is restarted for whatever reasons,
but notifications are persisted,
they do not contain all lines.

for the notification display, this is no big deal,
however, this cached value is not suitable
to check if there is a notification or not for a chat.

even if there is a notification, 'inboxes' may be empty,
this 'speed up' check was introduced
for a workaround of notification removal that requires lots of calls.
meanwhile, however, we have DC_EVENT_MSGS_NOTICED,
so we also remove this workaround.

the bug of notifications getting stuck was reported by @adbenitez on irc/dc

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationListItem.java
@@ -139,18 +139,6 @@ public class ConversationListItem extends RelativeLayout
     int state       = dcSummary.getState();
     int unreadCount = (state==DcMsg.DC_STATE_IN_FRESH || state==DcMsg.DC_STATE_IN_NOTICED)? thread.getUnreadCount() : 0;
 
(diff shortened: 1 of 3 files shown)


adbenitez (approved):
already using it +1

adbenitez:
(with the app closed, I marked as read a message and all other notifications where also hidden, I guess this is expected?)

r10s:
> (with the app closed, I marked as read a message and all other notifications where also hidden, I guess this is expected?)

depending on android version, we have to create a so called summary-notification. when inboxes is empty due to restart or whatever, that gets removed. i think, this is the thing you described.

maybe that can be improved somehow, however, this is another issue an no regression since 1.12

PR #1732, opened 2020-11-13
add 'show in chat' function, fix enlarge-view-menu-issues

### LINKED ISSUE #1731: minor enlarge-media menu issues
the menu for enlarge-media-view-activity has some minor issues:

- "all media" leaves activity-stack dirty, eg.
  - when opened from already gallery, no new gallery should be opened but it should behave as "back"
  - when opened from chat, self is not removed

- "all media" does not open the correct tab

- "forward" is misleading/unexpected as not only the image is forwarded but also the text that is not shown in the gallery. it should be removed or, better, be replaced by sth. as "show in chat" that opens the correct message, that can be forwarded from there (also replied to, btw)

- the order of the menu items does not match the one used as other places

r10s:
in the enlarge-image-view, this pr replaces buggy and misleading 'forward' by more powerful 'show in chat' and fixes other issues mentioned at #1731

closes #1731

please merge that after 1.14 release cycle (which is expected to be done soon :)


Hocuri (approved):
definitely better now

PR #1736, opened 2020-11-17
fortify wording in delete-server dialog

r10s:
this pr changes the wording, not relativize things, underlining the worse impact.

i avoided adding "maybe" or "potentially" here, even if that would be more accurate as in reality fewer messages are deleted (we could add a "Info" link or so at some point, but that is more work and can be done later).

before/after screenshots:

### DIFF
--- a/res/layout/autodel_confirm.xml
+++ b/res/layout/autodel_confirm.xml
@@ -7,6 +7,8 @@
     <CheckBox
         android:id="@+id/i_understand"
         android:text="@string/autodel_confirm"
+        android:textSize="16sp"
+        android:textColor="@color/red_500"
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:layout_marginLeft="20dp"
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -540,7 +540,7 @@
     <!-- %1$d will be replaced by the number of messages, you can assume plural/lots here. %2$s will be replaced by a timespan option. -->
     <string name="autodel_device_ask">Do you want to delete %1$d messages now and all newly fetched messages \"%2$s\" in the future?\n\n• This includes all media\n\n• Messages will be deleted whether they were seen or not\n\n• \"Saved messages\" will be skipped from local deletion</string>
     <!-- %1$d will be replaced by the number of messages, you can assume plural/lots here. %2$s will be replaced by a timespan option. -->
-    <string name="autodel_server_ask">Do you want to delete %1$d messages now and all newly fetched messages \"%2$s\" in the future?\n\n• This includes all media\n\n• This will also delete emails affected by the \"Show classic emails\" option\n\n• \"Saved messages\" will be deleted from the server as well</string>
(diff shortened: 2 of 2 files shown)


gerryfrancis:
@r10s Very transparent! :+1:

Hocuri:
1. :+1: 
2. Maybe "Do not use this function if you are using other email clients **than**/**besides** Delta Chat
3. I liked the sentence "This will also delete emails affected by the 'show classic emails' option", OTOH it's very technical, maybe "This includes **~all~** emails in all folders, **not only chat messages**"? (bold means that I propose to change it, not to make it bold:wink:)
  I'd also be in favor of leaving out the word "all" because technically it's not correct that all emails are deleted - if `show_emails` is off, some emails are not deleted and people should not rely on that _all_ emails are deleted.
  :+1: for the wording "in all folders" as this is going to be the case as soon as my `scan_folders` is implemented
4. @elwerene what do you think?

elwerene:
@Hocuri I think it is a lot better now! I didn't know it would delete my normal emails when I activated this feature and with this dialog it's very clear!

Hocuri (approved):
Then let's approve so @r10s can merge it quickly after applying my suggestions (or reading through my suggestions and not liking them, whatever, it's probably most important to fortify wording at all)

r10s:
- i stroke the first "all", i think it is still strong by the "from all server folders".
- yip, you're probably right,  "beside" seems to be more correct
- mentioning the "show all emails option" - i tried to avoid that as this seems to be part of the problem: ppl may think "fine" at the moment the deletion is enabled and forget about this bit later. therefore, being quite negative seems to be better here. and yes, it is too complicated.
- i did not add "not only chat messages" as this seems to be pretty clear and would make the sentence a bit too complicated imho



thanks for approving, but i would like to leave the pr a little bit open to allow more comments (we do not have core50 yet anyway, so 1.14.5 won't be done before tomorrow anyway)

Hocuri:
- it's "beside***s***":joy:
- concerning "not only chat messages": My suggestion had been to remove "media and 'Saved Messages'" instead, so that the sentence becomes: "This includes emails in all folders, not only chat messages".
  However, of all my suggestions, this one was the least important to me so if you would like to keep "media and 'Saved Messages'", still :+1:

Hocuri:
> i was not aware there is a difference :) thanks for taking care about such details!

Ok I looked it up now, and sources disagree on this (I had thought you had just typed it wrongly) but if there is a difference then it's that ["beside" means "next to" while "besides" means "in addition to"](https://www.grammarly.com/blog/beside-vs-besides/). So, it was ok before but now it's perfect :)

Hocuri:
BTW I wanted to say how much I enjoy the way we discuss such things in general. I mean, we had totally different opinions in the beginning on a topic that feels so important (of course it does, it is important) and yet we got to a consensus without even starting to repeat arguments (feeling not understood).

r10s:
totally agree, @Hocuri 👍

PR #1737, opened 2020-11-18
show permanent hint about server-delete impact

r10s:
when server-deletion is enabled, always show a hint about the impact below the option.

the hint is a summary of the conformation-dialog so that the user is more aware of the impact if enabling was some time ago or the option was even enabled by the provider-db.

Screen Shot 2020-11-18 at 21 48 09

just came into my mind :) successor of https://github.com/deltachat/deltachat-android/pull/1736


Hocuri (approved):
Great idea, also helping for existing users!

Just wondering if we should add a ⚠️ somewhere or write "NOT" in capitals but doesn't have to be.

What's probably most important is to release now :) (IIRC link2xt also made a critical PR that is merged but not yet released)

r10s:
> Just wondering if we should add a ⚠️ somewhere or write "NOT" in capitals but doesn't have to be.

also thought about that, however, that might be a bit too annoying for a permanent reminder. it is not that sth. is "wrong" or so.

also, we have a. hint already in the settings overview, so, i think, all in all, it is already much better visible.

PR #1741, opened 2020-11-22
show hint if there are no blocked contacts

### LINKED ISSUE #1287: empty blocked-contacts list should show some hints
the empty blocked-contacts list should show a hint.

desktop already uses the string `none_blocked_desktop` - we should just also use this this.

(once done, i'd like to reword the hint to "If you block contacts, they are shown here" or so - but that is independent from android then)

corresponding ios issue: https://github.com/deltachat/deltachat-ios/issues/661

r10s:
add a hint to the "blocked contacts" activity if there are no blocked contacts.

before, the whole activity was just "white" and it was unclear what can be done here.

moreover, this pr cleans up with unused `SwipeRefreshLayout`.

closes #1287

once merged, we should use the string `blocked_empty_hint` instead of `none_blocked_desktop` also on ios and desktop. (the old string reads "no blocked contacts yet" which does not give a hint who blocks things and also the "yet" feels a bit weird to me - as if this is a goal or so :)

### DIFF
--- a/res/layout/contact_selection_list_fragment.xml
+++ b/res/layout/contact_selection_list_fragment.xml
@@ -16,7 +16,8 @@
             android:layout_width="match_parent"
             android:layout_height="match_parent"
             android:gravity="center|center_vertical"
-            android:layout_marginTop="15dp"
+            android:padding="16dp"
+            android:lineSpacingMultiplier="1.3"
             android:visibility="gone"
             android:text="@string/contacts_empty_hint"
             android:textSize="20sp" />
--- a/res/layout/contact_selection_list_fragment.xml
+++ b/res/layout/contact_selection_list_fragment.xml
@@ -19,8 +19,8 @@
             android:padding="16dp"
             android:lineSpacingMultiplier="1.3"
             android:visibility="gone"
-            android:text="@string/contacts_empty_hint"
-            android:textSize="20sp" />
+            android:textSize="20sp"
+            tools:text="No contacts." />
 
     <org.thoughtcrime.securesms.components.RecyclerViewFastScroller
         android:id="@+id/fast_scroller"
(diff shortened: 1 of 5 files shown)


res/layout/contact_selection_list_fragment.xml
```diff
@@ -21,12 +16,12 @@
             android:layout_width="match_parent"
             android:layout_height="match_parent"
             android:gravity="center|center_vertical"
-            android:layout_marginTop="15dp"
-            android:text="@string/one_moment"
+            android:padding="16dp"
+            android:lineSpacingMultiplier="1.3"
+            android:visibility="gone"
+            android:text="@string/contacts_empty_hint"
```
Hocuri, line 22: Will this text ever be shown to the user? (`No contacts.`) I didn't find a place where I can see it.

Also, aren't we directly changing the text in onCreate? Nothing will be shown to the user until onCreate is done (?)

Hocuri (approved):
+1 for the changed text

res/layout/contact_selection_list_fragment.xml
r10s: no it is not shown currently, however, the layout is shared for different purposes, so having a more generic "empty" text makes some sense. before there was "one moment" which i found pretty confusing :) but maybe we should just set no text at all in the layout.

r10s: i will move that to tools:text, i'll add a commit.

r10s:
> once merged, we should use the string blocked_empty_hint instead of none_blocked_desktop also on ios and desktop.

- ios changed that by the last translation update in https://github.com/deltachat/deltachat-ios/commit/ab86fa6c344afddec6ec766053ac0fa236c5e700

- for desktop i filed an issue at https://github.com/deltachat/deltachat-desktop/issues/1988

PR #1744, opened 2020-11-25
display forwarded messages also in quotes as such

### LINKED ISSUE #1743: quoting forwarded message  no longer look as beeing forwarded
when quoting a forwarded message, the information that the message is not from the sender, is lost.

this can easily lead to confusing situations, i just came over that in testing group, here is a more constructed case:

Screen Shot 2020-11-25 at 11 42 58

things would be even worse, if there are other messages between the first two shown in the example above.

solution:

if forwarded messages are quoted, this should be shown in the quote as usual by replacing the name by the string "Forwarded message" and show the title and the bar in "forward grey color".

no additional api is needed for that, just check [`dc_msg_is_forwarded()`](url) for the quoted message. if the quoted-message is unknown, things are fine as the author is not shown at all in this case. however, i suggest to use the "forwarded" bar also in this case (instread of the current purple fallback) - this is fine whether the message was forwarded or not.

r10s:
see #1743  for backgrounds and reasoning. this is how it looks like:

100225763-05a31e80-2f1f-11eb-92ef-d7d996ebb96f

i also introduced a "universal" forward color that is is also used for quotes where the sender is unknown for whatever reason (forwarded-messages and unnamed-quotes are quite similar in this regard and it makes sense to use the same color for both)

closes #1743

### DIFF
--- a/res/layout/quote_view.xml
+++ b/res/layout/quote_view.xml
@@ -20,8 +20,8 @@
 			android:id="@+id/quote_bar"
 			android:layout_width="@dimen/quote_corner_radius_bottom"
 			android:layout_height="match_parent"
-			android:background="@color/purple_400"
-			tools:tint="@color/purple_400" />
+			android:background="@color/unknown_sender"
+			tools:tint="@color/unknown_sender" />
 
 		<LinearLayout
 			android:layout_width="0dp"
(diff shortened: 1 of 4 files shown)


Hocuri (approved):
BTW is the "post-release-maybe-still-make-hotfixes" phase over now?

r10s:
> BTW is the "post-release-maybe-still-make-hotfixes" phase over now?

i also thought about that just today. yip, i'd say so, we can move forward. if really another bugfix realease is needed for core50/android1.14, we can always branch.

PR #1747, opened 2020-12-04
Put a black circle behind the edit options when editing an image

Hocuri:
It repeatedly annoyed me that I can't recognize the white edit tools on a bright surface. I now added a black half-opaque circle behind each one.

According to StackOverflow (if I understood correctly), selectableItemBackgroundBorderless is supposed to show some animation when pressing the icon but this seems not to have worked anyway, at least I could not recognize any animation.

### DIFF
--- a/res/layout/image_editor_hud.xml
+++ b/res/layout/image_editor_hud.xml
@@ -25,6 +25,7 @@
             android:orientation="horizontal"
             android:layout_marginEnd="20dp"
             android:layout_marginRight="20dp"
+            app:chipSpacing="8dp"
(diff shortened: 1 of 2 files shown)


r10s (approved):
great improvement!

the only thing i did not find so nice is that only some circles are visible (the ones outside the image are not visible).

i fixed that by tweaking the universal_overlay color, see https://github.com/deltachat/deltachat-android/pull/1747/commits/0980e9703fa8b8b5e3aeddf0ed7e99d2971978b6 for reasoning.

moreover, i tested the pr on android4, which is always a bit sensitive on shapes and attributes - but no problems this time :)

i rebased the pr and added a commit, feel free to merge (unless you do not like the new universal_overlay color)

PR #1752, opened 2020-12-12
Red and more visible error icon

Hocuri:
IIRC some time ago we talked about making the error icon red and make it
an exclamation mark to make it more visible.

Today someone complained to me that they thought the "x" looks like one
can close something and that it shouldn't be green, so I went ahead and
made it red :) (it's also a little bit bigger now)

- the image file is a vector (xml) file now, previously it was a png file.
Might this be a problem on newer devices?

Screenshot_20201212-164027


r10s (approved):
nice!

i added a tiny commit that aligns the height of the new icon to the height of the delivery line, (the icon should not add height)

in theory, the other delivery-icons are 12dp in height, however, that looks too small for a circle, 16dp is fine, and, due to remaining padding, no overall height is added.

wrt "problem on newer devices" - why do you think, that can cause problems? we are using xml graphics at several places, or? however, we had some problems with older devices and vectors, so i tested it with android4 - words as expected.

Hocuri:
oops, yes I meant "older devices" :) but if you tested it and it works fine, then great!

r10s:
k, then let's merge that in :)

PR #1751, opened 2020-12-12
Directly show error on msg tap

Hocuri:
fix #1749

Screenshot_20201212-151100
Screenshot_20201212-150401

PR #1754, opened 2020-12-14
add missing button to new error-dialog

r10s:
following common ui expectations,
there must always be at least one "ok" button
in a dialog or alert.
there might be good reasons for doing it differently,
however, i do not see any of them here :)

i've overseen that bit when reviewing https://github.com/deltachat/deltachat-android/pull/1751 :)

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -785,6 +785,7 @@ public class ConversationItem extends LinearLayout
         AlertDialog d = new AlertDialog.Builder(context)
                 .setMessage(messageRecord.getError())
                 .setTitle(R.string.error)
+                .setPositiveButton(R.string.ok, null)
                 .create();
         d.show();
         try {


Hocuri:
I don't know - is it a common ui expectation? I mean, an ok button does take away some space and does not provide any value in this case as the user can just tap beside the dialog or use the back button.

Also, it's not a dialog in the sense that the user confirms or acknowledges something. It's only showing the info the user requested.

Actually it's an interesting question in general (also for the "show message info" dialog), I'll ask my brothers whether they think it's needed.

Hocuri (approved):
Ok, according to my mini-poll we do need an OK button :)

r10s:
> I don't know - is it a common ui expectation?

yip, it is :) thanks for your mini-poll to prove that :)

adbenitez:
yes, I was using your new dialog since yesterday and it was quite unexpected for me, the error dialog was big so no much space left to tap and I was for some seconds like "how do I get out of here"

Issue #1748, opened 2020-12-07
Device run's in DarkMode - Background and font color is black


Meldario:
- Operating System (Android): 10 (V12.0.3.0.QFDEUXM, QKQ1.190825.002 test-keys)
- Delta Chat Version: Delta Chat 1.14.5-fat
- Expected behavior: Messages should be shown normally
- Actual behavior: Background ist black, also font color is black (Smylies are shown) after a fresh install
- Steps to reproduce the problem: Device runs in DarkMode. In DeltaChat Settings (Appearance) the Theme is "Light" = Font Color and Background is black, can't read Messages. Have to change Theme to "Dark"
- Screenshots:
![DarkMode](https://user-images.githubusercontent.com/75611145/101335092-03bf5080-3879-11eb-84de-989c7c6700af.jpg)

r10s:
thanks for reporting.

1. if that is mainly the bubble color, a quick fix might be to not use the "system-background" for bubbles but just white. then, the delta chat dark-mode can be selected independingly of the system dark mode.

2. alternatively, we could to tag our themes for light/dark in a way, the system understands it, so that the system can select the correct one see https://developer.android.com/guide/topics/ui/look-and-feel/darktheme?authuser=1. however, unfortunately, most androids out there do not support system-dark-mode at all, so that we would still need an internal switch, that would probably need three states then (default/light/dark). if the user has manually selected light/dark, we have to override the system setting, that might be tricky.

3. we could hide the in-app-option in android9, however, not sure if really all android9 systems have that switch. as this is not easy to find out, i would avoid that option for now.

so, all in all, it seems 1. seems to be the easiest thing to do. also, it is a precondition for 2., so the work would not be lost

Meldario:
Hello,
thanks for your answer.

It's not a Problem for me, I simply selected "Dark" in Themes (but it took a few Moments to understand, what's the Problem)
It may be a Problem for Users, that want to use DeltaChat "Out of the Box". Maybe they don't understand, why they can't read anything :-)

Greetings,
Alex

r10s:
> Maybe they don't understand, why they can't read anything :-)

yip, that needs to be fixed, no doubt :)

r10s:
this needs a retry, i tested darkmode on some common devices, looks good so far.

however, in the screenshot atop, also the background image looks weird - maybe the system tries to shift some colors around or so...

Meldario:
[...]

PR #1758, opened 2020-12-19
Dark theme / DayNight: Add option to follow system

Hocuri:
fix maybe #1748. Maybe not, because actually I don't understand why issue #1748 even happened because we don't use the system background for the bubble but an own one, `conversation_item_outgoing_bubble_color`, defined in `themes.xml`. They are retrived in `org.thoughtcrime.securesms.ConversationItem#initializeAttributes` (org/thoughtcrime/securesms/ConversationItem.java:263` and set on our own. The standard background color should be white (see conversation_item_sent.xml:50).

Probably the system is somehow trying to forcefully apply a dark mode, no idea how. Maybe it looks for a combination of "background" and "white" and then sets it to dark, similar to [ForceDark](https://developer.android.com/guide/topics/ui/look-and-feel/darktheme#disable-force-dark) (which is opt-in though, and also when I enabled force-dark in the developer options, this worked better than in #1748; esp. there was no black-on-black text). BTW, I also run Android 10 and never had this issue.

So, let's see whether this PR fixes the issue, if not, I'll try again.

Unfortunately, we now have `preferences_appearance.xml` twice. Signal also has it twice but it's way shorter there.
If we don't want this duplication, we have to use `<include>` or set the default programmatically (the problem is that before API29 the default should be `Light` and at API29+ it should be `SystemDefault`

### DIFF
--- a/res/values/arrays.xml
+++ b/res/values/arrays.xml
@@ -84,7 +84,8 @@
       <item>te</item>
   </string-array>
 
-    <string-array name="pref_theme_entries" tools:ignore="InconsistentArrays">
+  <!-- "InconsistentArrays" is needed as the array is duplicated in values-v29/arrays.xml with a different size -->
+  <string-array name="pref_theme_entries" tools:ignore="InconsistentArrays">
       <item>@string/pref_light_theme</item>
       <item>@string/pref_dark_theme</item>
   </string-array>
--- a/res/values/arrays.xml
+++ b/res/values/arrays.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<resources>
+<resources xmlns:tools="http://schemas.android.com/tools">
 
   <!-- take care, language_entries and languages_values are in sync ...
        ... some one has to clean up this mess some day ... -->
@@ -84,12 +84,12 @@
       <item>te</item>
   </string-array>
 
-    <string-array name="pref_theme_entries">
+    <string-array name="pref_theme_entries" tools:ignore="InconsistentArrays">
       <item>@string/pref_light_theme</item>
       <item>@string/pref_dark_theme</item>
   </string-array>
 
-  <string-array name="pref_theme_values" translatable="false">
+  <string-array name="pref_theme_values" translatable="false" tools:ignore="InconsistentArrays">
       <item>light</item>
       <item>dark</item>
   </string-array>
--- a/res/xml-v29/preferences_appearance.xml
+++ b/res/xml-v29/preferences_appearance.xml
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 
+<!-- when changing this file, consider adapting values/preferences_appearance.xml as well
+ (values/preferences_appearance.xml differs only wrt theme default value) -->
 <PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android">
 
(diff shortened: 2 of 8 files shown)


res/values/arrays.xml
```diff
@@ -84,12 +84,12 @@
       <item>te</item>
   </string-array>
 
-    <string-array name="pref_theme_entries">
+    <string-array name="pref_theme_entries" tools:ignore="InconsistentArrays">
```
r10s, line 87: what does `tools:ignore="InconsistentArrays"` do? and why is it needed for this array but not for the others?

res/xml-v29/preferences_appearance.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="utf-8"?>
```
r10s, line 1: i am not really happy about this duplication, we should at least add a bit WARNING atop of both files.

however, for now: the duplication is only needed because of the changed default value, or?
i suggest to just leave the default to "light" in all cases - we seem to have some issues with dark-mode on some systems, so we should not force using that until we really know it is fixed.

r10s (commented):
this is only a partly review so far, i did not test it.

one question: i assume, that the system may adapt the light/darkmode depending on the time/lightness - do we follow such changes immediately? at a first glance, i did not see that in the code, however, i also do not have the full picture yet :)

Hocuri:
> one question: i assume, that the system may adapt the light/darkmode depending on the time/lightness - do we follow such changes immediately? at a first glance, i did not see that in the code, however, i also do not have the full picture yet :)

Yes. The system automatically takes care of restarting all activities when night mode changes. I tested it and it works fine.

res/xml-v29/preferences_appearance.xml
Hocuri: Issue  #1748 is that the system is in dark mode and DC is in light mode. As soon as the user switched DC to dark mode, everything was fine. So, always following the system mode would definitely have prevented this special bug (that is, at least until the user changes the setting from "system default" to "light")

Or are you talking about other issues with the dark mode?

res/values/arrays.xml
Hocuri: In one config, we have only 2 entries, in another config we have 3 entries (because "follow system" was added as an option). That's what Android would complain about without the warning.

Hocuri:
BTW, I should probably have said that, for the most part of the code I just pulled from Signal.

res/xml-v29/preferences_appearance.xml
r10s: yip, you're right, so, changing default should not worsening things.

but i would still suggest to add a hint atop of the duplicated files. otherwise if will be forgotten on the next option added/changed.

> Or are you talking about other issues with the dark mode?

no, i do not know about other issues with dark mode.

res/values/arrays.xml
r10s: k, i would also add that bit as a short comment to the code.

r10s (commented):
thanks for the additional explanations, in general, lgtm. but i would really add the comments as suggested :)

r10s (approved):

Hocuri:
:tada: thanks for adding the comments!

r10s:
thanks for the whole fix, @Hocuri :)

PR #1764, opened 2021-01-02
Show controls on small screens

Hocuri:
Using both `setMessage()` and `setView()` on the same AlertDialog,
on small screens the "OK" and "Cancel" buttons were not be show. So, put the
message into our custom view.


r10s (approved):
great, thanks for fixing!

PR #1412, opened 2020-06-12
start ConversationListActivity with startActivityForResult()

Hocuri:
so that the conversations list is correctly updated before and after
sharing. Now it always correctly shows or does not show the devicetalk.

Second try after
https://github.com/deltachat/deltachat-android/pull/1402.

@r10s I tried again :wink:. 

I got to the idea because with forwarding everything worked fine and with forwarding startActivityForResult() is used:

    private void handleForwardMessage(final Set<DcMsg> messageRecords) {
        Intent composeIntent = new Intent(getActivity(), ConversationListActivity.class);
        int[] msgIds = DcMsg.msgSetToIds(messageRecords);
        setForwardingMessageIds(composeIntent, msgIds);
        startActivityForResult(composeIntent, REQUEST_RELAY);
        getActivity().overridePendingTransition(R.anim.slide_from_right, R.anim.fade_scale_out);
    }

In my tests everything worked fine.

PR #1770, opened 2021-01-15
Fix disappearing drafts

Hocuri:
Some people reported that drafts were disappearing, and I found a way to
reproduce it (my interpretation in brackets):
- Open Saved Messages chat, could be any other chat too
- Go to another app and share to DC
- In DC select Saved Messages. Now a draft with the share text is created. Edit the text if you like (but don't put too much effort in it as the draft will vanish :) ). _Don't_ hit the Send button.
  - (as sharing is done using startActivityForResult(), a second
  ConversationActivity is created without affecting the existing one)
- Leave DC. (The second ConversationActivity will correctly save its
draft to the db)
- Open DC again from the "Recent apps". (this is where my fix comes into play) (The first ConversationActivity
will be loaded. It will not know that there is a newer draft and still
show the old draft, or nothing if there was none. When it's closed, it
will overwrite the draft)

I came to this conclusion by logging all writes to the draft in the db.
Also, when I set `REQUEST_RELAY` to -1, `startActivityForResult(intent, -1)` 
will behave the same as `startActivity(intent)`) (but that's not a solution,
sometimes we need startActivityForResult(), see e.g.
https://github.com/deltachat/deltachat-android/pull/1412)

What's still a little weird: I tried replacing all calls to
`startActivityForResult(-intent for ConversationActivity-)` with
startActivity(), and the issue didn't go away. Apparently it's only
enough if all calls are replaced, which is the same as setting
`REQUEST_RELAY` to -1. Maybe it's enough if one activity in the call
chain is called using startActivityForResult().

The call to `initializeDraft()` makes the activity start & resume 2-3ms
slower, which is probably acceptable.

While going back and forth with activities & sharing, I also encountered
some more bugs (without trying to reproduce them), but they are not too
annoying (as opposed to losing your draft) and I'm hesitant to changing
the activity-switching logic too much as it is so easy to just break
things again.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -167,6 +167,17 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
   private static final int PICK_LOCATION       = 9;  // TODO: i think, this can be deleted
   private static final int SMS_DEFAULT         = 11; // TODO: i think, this can be deleted
 
+  /**
+   * If the user opens a chat, goes to another app, and shares some content to the same chat, we will have two ConversationActivity's
+   * with the same chat (because sharing uses startActivityForResult() and all activities started this way will be created a second time,
+   * without affecting the existing activity). So, when the sharing ConversationActivity closes, it sets this variable to `true` so that
+   * the still-running-currently-in-background ConversationActivity re-initializes the draft.
+   * There is always only one still-running-currently-in-background ConversationActivity, which is the one started normally (i.e. with
+   * `startActivity()`). Therefore, after reinitializing the draft in one activity, we can safely set this variable to `false`.
+   * See https://github.com/deltachat/deltachat-android/pull/1770
+   */
+  private static boolean doReinitializeDraft;
+
   private   GlideRequests               glideRequests;
   protected ComposeText                 composeText;
   private   AnimatingToggle             buttonToggle;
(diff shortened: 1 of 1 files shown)


r10s:
thanks for the detailed description, can you please add where you actually set drafts? in the first step? in the 3rd step - do you mean to send the shared data or save it again as a draft?

i also encountered weird things, but i am not totally sure if that is the same thing :)

at some point in the past, we decided that we need only one draft per chat, might be we're experienting the limits of this decision here.

Hocuri:
> at some point in the past, we decided that we need only one draft per chat, might be we're experienting the limits of this decision here.

I think that's fine as long as the draft doesn't suddenly go back to an older version (i.e. vanish if you didn't have a draft before) :)

I hope the description is clear now (draft is set in 3rd step, don't send). I can also do a screencast otherwise.

r10s:
> I hope the description is clear now (draft is set in 3rd step, don't send). I can also do a screencast otherwise.

thanks, i think it is clear enough now :)

r10s:
k, i could reproduce the issue now and yes, the pr would fix the issue.

however, i am a bit unhappy, that `initializeDraft()` is called twice now (one time in `onCreate()`/`onNewIntent()` and after that in `onResume()`. when the draft is an image or so, this might be significantly more than 2ms or so.

as the open-chat workflow is one of the most-used workflows, it makes sense to not waste time here.

we could fix that by:
1. skipping calls to  `initializeDraft()` in `onCreate()`/`onNewIntent()` - however, there is deeply nested code and futures depending on `initializeDraft()` to report success or whatever, so, going that way has a big risk of breaking more things

2. ~~doing the new call in `onRestart()`, that is called only if the activity was paused before~~ seems not to work as most time activity we open the activity using `onNewIntent()` - [and that seems to be followed by `onRestart()`](https://www.nickagas.com/using-onnewintent-and-activity-lifecycle/), so we would still have double initialisation

3. adding a `boolean doReinitializeDraft` that is set to `true` in `onPause()`. in `onResume()` we would do `if(doReinitializeDraft) { initializeDraft(); }`

did not test that out :)

the only reasonable options seems to be 3. - but that is easy to do.

Hocuri:
Can we just appreciate how much text we have already written about this one line? :astonished: :astonished: :smile: :smile: 

Anyway,

> when the draft is an image or so, this might be significantly more than 2ms or so.

This is not a problem, the draft media is only loaded once.

When I benchmarked this time, it was + around 5ms if there is a draft (no matter if it's text or image) and + around 2ms if there is no draft.

OTOH doing 3. or even 4. will make things a little faster, maybe it's worth it. 4 will need a global (static) variable but it should be fine as only the UI thread accesses it.

r10s:
> Can we just appreciate how much text we have already written about this one line? 😲 😲 😄 😄

yeah, it is the most-discussed line currently :D

tbh, i would feel better if we can avoid double calls to initializeDraft() - one thing to add, in addition to wasted time: we should also keep in mind that the stuff executed from initializeDraft() is async, so the stuff runs more or less at the same time, i can also imagine that that can cause problems. so, i would avoid that, and maybe do that with 4. - also, as that is probably not that hard.

Hocuri:
Ok, both the comment-to-code-changes and the github-discussion-to-code-changes ratios decreased a bit, but now it should be faster, no possible race condition anymore and still work.

r10s (approved):
lgtm, good job.

the use of `getCallingActivity()` is elegant as it avoid using a global and tearing up the logic across several files.

otoh, we rely on `getCallingActivity()` never to return non-null when called outside sharing, this seems to be true, but the condition might be a bit weak, also when regarding the different android versions and the troubles we have with that activity-resuts ...  but - one of the ideas was to et `doReinitialzeDraft` even unconditionally, so, even when the check fails under some circumstances, it will not be that worse :)

so, no blocker, things are very probably fine :)

PR #1772, opened 2021-01-20
minor mailing lists tweaks

r10s:
this pr shall be merged to #1643, not to `master`:

this pr are just some minor tweaks i came over while first reviewing-and-play-around round:
- change name of tab and subtitle in profile
- hide "Edit group name and image" for mailing lists

### DIFF
--- a/src/org/thoughtcrime/securesms/ProfileActivity.java
+++ b/src/org/thoughtcrime/securesms/ProfileActivity.java
@@ -67,6 +67,7 @@ public class ProfileActivity extends PassphraseRequiredActionBarActivity
   private int                  chatId;
   private boolean              chatIsGroup;
   private boolean              chatIsDeviceTalk;
+  private boolean              chatIsMailingList;
   private int                  contactId;
   private boolean              fromChat;
 
@@ -130,7 +131,9 @@ public class ProfileActivity extends PassphraseRequiredActionBarActivity
 
       if (chatId != 0) {
         inflater.inflate(R.menu.profile_chat, menu);
-        if (chatIsGroup) {
+        if (chatIsMailingList) {
+          menu.findItem(R.id.edit_name).setVisible(false);
+        } else if (chatIsGroup) {
           menu.findItem(R.id.edit_name).setTitle(R.string.menu_group_name_and_image);
         }
       }
@@ -198,6 +201,7 @@ public class ProfileActivity extends PassphraseRequiredActionBarActivity
     contactId        = getIntent().getIntExtra(CONTACT_ID_EXTRA, 0);
     chatIsGroup      = false;
     chatIsDeviceTalk = false;
+    chatIsMailingList= false;
     fromChat         = getIntent().getBooleanExtra(FROM_CHAT, false);
 
     if (contactId!=0) {
@@ -207,6 +211,7 @@ public class ProfileActivity extends PassphraseRequiredActionBarActivity
       DcChat dcChat = dcContext.getChat(chatId);
       chatIsGroup = dcChat.isGroup();
       chatIsDeviceTalk = dcChat.isDeviceTalk();
+      chatIsMailingList = dcChat.isMailingList();
       if(!chatIsGroup) {
         final int[] members = dcContext.getChatContacts(chatId);
         contactId = members.length>=1? members[0] : 0;
@@ -304,7 +309,9 @@ public class ProfileActivity extends PassphraseRequiredActionBarActivity
           if(isContactProfile()) {
             return getString(contactId==DcContact.DC_CONTACT_ID_SELF? R.string.self : R.string.tab_contact);
           }
-          else {
+          else if (chatIsMailingList) {
+            return getString(R.string.mailing_list);
+          } else {
             return getString(R.string.tab_group);
           }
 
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/ProfileActivity.java
```diff
@@ -130,7 +131,9 @@ public boolean onCreateOptionsMenu(Menu menu) {
 
       if (chatId != 0) {
         inflater.inflate(R.menu.profile_chat, menu);
-        if (chatIsGroup) {
+        if (chatIsMailingList) {
+          menu.findItem(R.id.edit_name).setVisible(false);
```
Hocuri, line 135: I am using mailing lists on my phone a little (very little, and only GitHub notifications) and I actually liked the possibility to change the group name and image (e.g. renaming `deltachat/deltachat-core-rust` to `deltachat-core-rust`).

The core will not send any mail out if the name is changed.

OTOH, this might break with the expectation that the name is changed for all users (as this is the case for all other groups).

src/org/thoughtcrime/securesms/ProfileSettingsAdapter.java
```diff
@@ -261,10 +266,13 @@ public void changeData(@Nullable int[] memberList, @Nullable DcContact dcContact
     itemDataMemberCount = 0;
     itemDataSharedChats = null;
     itemDataContact = null;
+    itemDataChat    = null;
 
     if (memberList!=null) {
       itemDataMemberCount = memberList.length;
-      if (!dcChat.isMailingList()) {
+      if (dcChat.isMailingList()) {
+        itemDataChat = dcChat;
```
Hocuri, line 274: What do we need this variable for?

src/org/thoughtcrime/securesms/ProfileActivity.java
r10s: let's keep things simple for now. the changed name won't be synced in a multi-device-setup and it is also not helpful when user have different names for the same group.

also, this breaks the expectation, that the name change is for all users, as in normal groups.

src/org/thoughtcrime/securesms/ProfileSettingsAdapter.java
r10s: it is used at https://github.com/deltachat/deltachat-android/pull/1772/files#diff-d483d329b19c7c242afce2368bf6fcf6acbe10a21c80996b74831eaf99cba30bR219 to set a different subtitle - "X members" is wrong for mailing lists, in fact, we do not know the number of members.

`itemDataChat` is a bit weird, however, i am just following the style already used in the file :)

Hocuri: It somehow confuses me that `isMailingList()` is checked twice (and generally, that we only have an `itemDataChat` if we have a mailing list) :) Maybe just unconditionally set it here - `isMailingList()` is [checked above](https://github.com/deltachat/deltachat-android/pull/1772/files#diff-d483d329b19c7c242afce2368bf6fcf6acbe10a21c80996b74831eaf99cba30bR219), anyway - or make a `bool` `isMailingList`?

Hocuri (commented):
Just wondering - if we make the tab title `Mailing list`, maybe we should add everyone that writes to a mailing list to the contacts list (instead of just self and the mailing-list-pseudo-contact)? (to be done in the core) Because, if the tab title is "Mailing list", I'd expect to get some info about the mailing list.

And, if we do this - do we somehow express the fact that this list is incomplete? I didn't have any good idea for this so far. Or do we just expect users to know that DC has no way of knowing who is subscribed to a mailing list?

src/org/thoughtcrime/securesms/ProfileSettingsAdapter.java
r10s: yip, we can do a `isMailingList` list, that is probably more straight-forward. i'll add a commit.

r10s:
> Just wondering - if we make the tab title Mailing list, maybe we should add everyone that writes to a mailing list to the contacts list (instead of just self and the mailing-list-pseudo-contact)? (to be done in the core) Because, if the tab title is "Mailing list", I'd expect to get some info about the mailing list.

i do not think, ppl will expect that this is the complete mailing list.

and even when: then they would expect that even more when the subtitle says "2 members". so, i think, things are mitigated even in this case.

> And, if we do this - do we somehow express the fact that this list is incomplete? I didn't have any good idea for this so far. Or do we just expect users to know that DC has no way of knowing who is subscribed to a mailing list?

i would not go for that now, i think, it is fine as it is for now. i just want to mitigate strongly misleading things with this pr, as showing the concrete member amount which is just wrong.

i am not saying, things are perfect with this pr, we can improve things later anyway :)

/me not even finished with checking the other parts affecting mailing lists :)

Hocuri:
> i would not go for that now, in the scope of this pr

this would be sth. to do in core anyway, sorry for bringing it up here off-topicly :) I just started to wonder whether I should implement this in core.

Just hiding the tab would also be a possibility of course.

This PR here lgtm (and yes, it definitely improves things, thanks), feel free to merge it if you are done adding things (and could you please remove the updated jni/deltachat-core-rust submodule?)

r10s:
k, let's merge that in :)

PR #1773, opened 2021-01-22
add preview-apk workflow

adbenitez:
generate APK previews for pull request testing (for faster generation, only architecture armeabi-v7a is compiled, change it before merging, if not convenient)

## How to download the generated debug APK preview?

once the action ends to run, click in details:
action-details

then click the artifacts drop-down menu and select the APK file:
artifacts

### DIFF
--- a/.github/workflows/preview-apk.yml
+++ b/.github/workflows/preview-apk.yml
@@ -1,0 +1,48 @@
+name: Upload Preview APK
+
+on: pull_request
+
+jobs:
+  build:
+    name: Upload Preview APK
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: textbook/git-checkout-submodule-action@master
+      - run: cp jni/deltachat-core-rust/rust-toolchain .
+      - uses: actions-rs/toolchain@v1
+      - uses: Swatinem/rust-cache@v1
+        with:
+          working-directory: jni/deltachat-core-rust
+      - uses: actions/setup-java@v1.4.3
+        with:
+          java-version: 11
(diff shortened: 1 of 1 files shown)


r10s (approved):
wow - that's HUGE if that works :) thanks for pushing that forward!

feel free to play around and merge, we are not using the github workflows at all currently :)

.github/workflows/preview-apk.yml
```diff
@@ -0,0 +1,48 @@
...
+        id: setup-ndk
+        with:
+          ndk-version: r21d
+
+      - name: Compile core
+        env:
+          ANDROID_NDK_ROOT: ${{ steps.setup-ndk.outputs.ndk-path }}
+        run: |
+          export PATH="${PATH}:${ANDROID_NDK_ROOT}/toolchains/llvm/prebuilt/linux-x86_64/bin/:${ANDROID_NDK_ROOT}"
+          ./scripts/install-toolchains.sh && ./ndk-make.sh armeabi-v7a
```
link2xt, line 39: Why only one architecture? I think we should switch to simply `ndk-make.sh` before merging, so it generates universal APK.

link2xt: I'm one minute late, but anyway.

adbenitez: yikes! :( well I was using `ndk-make.sh` without args, but since in the comments inside `ndk-make.sh` seemed that for testing it is common to only use one arch to speed up things used that, what we do now? do I open another PR or just commit directly to master?

link2xt: Just commiting to master is ok

PR #1778, opened 2021-01-28
Don't use AsyncTask

Hocuri:
AsyncTasks [are deprecated](https://developer.android.com/reference/android/os/AsyncTask), and actually by not using it, the code gets shorter and IMHO more readable. Also, AsyncTask has "inconsistent behavior on different versions of the platform".


r10s (approved):
thanks, i missed completely that `AsyncTask` is deprecated, well, well, well.

but i agree, the code is more readable. currently `Util.runOnBackground()` has a reference to `AsyncTask`,  however, that can be replaced much more easily than these things derived from AsyncTask (at the day where AsyncTask will be gone completely then :)

PR #1779, opened 2021-01-28
message-html-view: harden settings, add pinch-to-zoom

r10s:
this pr adds pinch-to-zoom to the message-html-view and hardens the settings.

i've read through the settings at https://developer.android.com/reference/android/webkit/WebView and https://developer.android.com/reference/android/webkit/WebSettings and https://wiki.sei.cmu.edu/confluence/display/android/DRD02-J.+Do+not+allow+WebView+to+access+sensitive+local+resource+through+file+scheme

successor of #1763

### DIFF
--- a/src/org/thoughtcrime/securesms/FullMsgActivity.java
+++ b/src/org/thoughtcrime/securesms/FullMsgActivity.java
@@ -37,6 +37,11 @@ public class FullMsgActivity extends WebViewActivity
     loadRemoteContent = Prefs.getAlwaysLoadRemoteContent(this);
     webView.getSettings().setBlockNetworkLoads(!loadRemoteContent);
 
+    // setBuiltInZoomControls() adds pinch-to-zoom as well as two ugly buttons;
+    // the latter are hidden with setDisplayZoomControls() again.
+    webView.getSettings().setBuiltInZoomControls(true);
+    webView.getSettings().setDisplayZoomControls(false);
+
     getSupportActionBar().setTitle(getString(R.string.chat_input_placeholder));
 
     dcContext = DcHelper.getContext(this);
--- a/src/org/thoughtcrime/securesms/FullMsgActivity.java
+++ b/src/org/thoughtcrime/securesms/FullMsgActivity.java
@@ -3,6 +3,7 @@ package org.thoughtcrime.securesms;
 import android.os.Bundle;
 import android.view.Menu;
 import android.view.MenuItem;
+import android.webkit.WebSettings;
 
 import androidx.appcompat.app.AlertDialog;
 
@@ -37,10 +38,22 @@ public class FullMsgActivity extends WebViewActivity
     loadRemoteContent = Prefs.getAlwaysLoadRemoteContent(this);
     webView.getSettings().setBlockNetworkLoads(!loadRemoteContent);
 
-    // setBuiltInZoomControls() adds pinch-to-zoom as well as two ugly buttons;
-    // the latter are hidden with setDisplayZoomControls() again.
+    // setBuiltInZoomControls() adds pinch-to-zoom as well as two on-screen zoom control buttons.
+    // The latter are a bit annoying, however, they are deprecated anyway,
+    // and also Android docs recommend to disable them with setDisplayZoomControls().
     webView.getSettings().setBuiltInZoomControls(true);
     webView.getSettings().setDisplayZoomControls(false);
+
+    // disable useless and unwanted features:
+    // - JavaScript and Plugins are disabled by default, however,
+    //   doing it explicitly here protects against changed base classes or bugs
+    // - Content- and File-access is enabled by default and disabled here
+    // - the other setAllow*() functions are related to enabled JavaScript only
+    // - "safe browsing" comes with privacy issues and already disabled in the base class
+    webView.getSettings().setJavaScriptEnabled(false);
+    webView.getSettings().setPluginState(WebSettings.PluginState.OFF);
+    webView.getSettings().setAllowContentAccess(false);
+    webView.getSettings().setAllowFileAccess(false);
 
     getSupportActionBar().setTitle(getString(R.string.chat_input_placeholder));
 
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/FullMsgActivity.java
```diff
@@ -37,6 +38,21 @@ protected void onCreate(Bundle state, boolean ready) {
...
 
+    // setBuiltInZoomControls() adds pinch-to-zoom as well as two ugly buttons;
+    // the latter are hidden with setDisplayZoomControls() again.
+    webView.getSettings().setBuiltInZoomControls(true);
+    webView.getSettings().setDisplayZoomControls(false);
+
+    // disable useless and unwanted features:
+    // - JavaScript and Plugins are disabled by default, however,
+    //   doing it explicitly here protects against changed base classes or bugs
+    // - Content- and File-access is enabled by default an disabled here
```
link2xt, line 49: "and disabled"

src/org/thoughtcrime/securesms/FullMsgActivity.java
```diff
@@ -37,6 +38,21 @@ protected void onCreate(Bundle state, boolean ready) {
     loadRemoteContent = Prefs.getAlwaysLoadRemoteContent(this);
     webView.getSettings().setBlockNetworkLoads(!loadRemoteContent);
 
+    // setBuiltInZoomControls() adds pinch-to-zoom as well as two ugly buttons;
+    // the latter are hidden with setDisplayZoomControls() again.
```
link2xt, line 42: maybe note that on-screen zoom controls are also deprecated and recommended to be disabled by API docs

link2xt (approved):
Is it possible to follow the links from the email in the embedded browser? Are we affected by `setSafeBrowsingEnabled`? Is it possible to disable safe browsing globally on non-ungoogled android (this setting is controlled by the user) or should we take care of it? On my phone FAQ opens external browser when I click links, but I'm not sure whether it checks the link *before* opening it or assumes the browser will check it.

src/org/thoughtcrime/securesms/FullMsgActivity.java
Hocuri: :joy:

Hocuri (approved):

src/org/thoughtcrime/securesms/WebViewActivity.java
```diff
@@ -41,15 +41,27 @@ protected void onCreate(Bundle state, boolean ready) {
 
     webView = findViewById(R.id.webview);
     webView.setWebViewClient(new WebViewClient() {
+      // `shouldOverrideUrlLoading()` is called when the user clicks a URL,
+      // returning `true` means, the URL is passed to `loadUrl()`, `false` aborts loading.
+      // the method is not called for POST request nor for on-page-links.
+      //
+      // nb: from API 24, `shouldOverrideUrlLoading(String)` deprecated and
```
Hocuri, line 48: I've seen it multiple times now, and - what does nb mean :) ?

link2xt: https://en.wikipedia.org/wiki/Nota_bene I guess

link2xt, line 48: "*is* deprecated"

r10s:
> Is it possible to follow the links from the email in the embedded browser?

that would be possible, however, as most pages require scripts and loading external content, cookies and more, also login etc. therefore, we pass these links explicitly to the system in `WebViewActivity.shouldOverrideUrlLoading()`. 

however, when checking that code, **i also aborted to load any unknown protocol** in the WebView, see comments in corresponding commit.

> On my phone FAQ opens external browser when I click links, but I'm not sure whether it checks the link before opening it or assumes the browser will check it.

`WebViewActivity.shouldOverrideUrlLoading()` is called before the url is opened in WebView. if we do not call `openOnlineUrl()` and always return `true`, nothing would happen for any link.

src/org/thoughtcrime/securesms/WebViewActivity.java
r10s: thanks a lot <3

r10s: "nota bene", "by the way", draw attention to the reader to sth. else, https://en.wikipedia.org/wiki/Nota_bene

r10s:
> Are we affected by setSafeBrowsingEnabled?

imu, "safe browsing" is enabled by the WebView by default on system that support it. not sure if google-free androids do support that at all, https://developer.android.com/reference/android/webkit/WebSettings#setSafeBrowsingEnabled(boolean)

calling the api explictily would have a higher precedence, however, if it was disabled by the user somehow, maybe we should just respect that. might be that safe-browsing has privacy-issues (or at least some user think that is true)

link2xt:
> calling the api explictily would have a higher precedence, however, if it was disabled by the user somehow, maybe we should just respect that. might be that safe-browsing has privacy-issues (or at least some user think that is true)

What I propose is to disable it to avoid external HTTP requests. We have no use for it as we only show local content and open the browser for all URLs, so if it decides to check any URL, it does nothing good.

src/org/thoughtcrime/securesms/WebViewActivity.java
Hocuri: Thanks :)

src/org/thoughtcrime/securesms/WebViewActivity.java
```diff
@@ -65,6 +66,14 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
       }
     });
     webView.setFindListener(this);
+
+    // disable "save browsing" as this has privacy issues,
```
link2xt, line 70: save->safe

src/org/thoughtcrime/securesms/WebViewActivity.java
```diff
@@ -65,6 +66,14 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
...
+    // eg. at least false positives are sent to the "Safe Browsing Lookup API".
+    // as all URLs opened in the WebView are local anyway,
+    // "save browsing" will never be able to report issues anyway, so it can be disabled.
```
link2xt, line 73: "save" -> "safe"

link2xt (approved):

PR #1780, opened 2021-01-30
Change list of ephemeral message options

link2xt:
Also match unknown options to some known option, not "disabled".

### DIFF
--- a/res/values/arrays.xml
+++ b/res/values/arrays.xml
@@ -131,14 +131,14 @@
   </string-array>
 
     <string-array name="ephemeral_message_durations">
-        <item>@string/off</item>
-        <item>@string/after_1_minute</item>
-        <item>@string/after_5_minutes</item>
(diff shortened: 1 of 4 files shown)


r10s (approved):
this is the outcome of user tests with @kermoshina, right?

it should work, it is a bit hard to check the values, i needed some comments to really review that, if you like, you can merge them in, see https://github.com/deltachat/deltachat-android/pull/1782

an idea might be to put the values as seconds to array.xml (as done for mute and the other autodel-stuff), that would at least make the java-part more readable (for the closes-value one could use a loop then). however, maybe for the next time, otoh these values are not changed that often and now it seems right :)

core probably also needs more strings for the system messages? maybe also add them in this pr, so that the things, once merged, can be translated together then.

link2xt:
Related core PR: https://github.com/deltachat/deltachat-core-rust/pull/2183

r10s (approved):

PR #1781, opened 2021-01-30
mailinglists3

r10s:
needs https://github.com/deltachat/deltachat-core-rust/pull/1964 to be merged first.

rebased and submodule-free successor of #1643

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationListFragment.java
@@ -437,20 +437,22 @@ public class ConversationListFragment extends Fragment
         DcMsg msg = dcContext.getMsg(msgId);
         int contactId = item.getContactId();
         DcContact contact = dcContext.getContact(contactId);
+        int textNo = R.string.menu_block_contact;
+        int textQuestion = R.string.ask_start_chat_with;
+        if (msg.isMailingList()) {
+          textNo = R.string.never;
+          textQuestion = R.string.ask_show_mailing_list;
+        }
         new AlertDialog.Builder(getActivity())
-                .setMessage(getActivity().getString(R.string.ask_start_chat_with, contact.getNameNAddr()))
                 .setPositiveButton(android.R.string.ok, (dialog, which) -> {
                   int belongingChatId = msg.decideOnContactRequest(DcMsg.DC_DEADDROP_DECISION_YES);
                   if (belongingChatId != 0) {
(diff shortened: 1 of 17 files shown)


src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -300,7 +298,7 @@ protected void onPostExecute(Void result) {
       protected void executeAction(@Nullable Void parameter) {
         for (long chatId : selectedConversations) {
           if (chatId == DcChat.DC_CHAT_ID_DEADDROP) {
-            dcContext.marknoticedContact(getListAdapter().getDeaddropContactId());
+            dcContext.marknoticedChat(DcChat.DC_CHAT_ID_DEADDROP);
```
r10s, line 301: `dc_marknoticed_chat(DC_CHAT_ID_DEADDROP)` won't work, it would just be a no-op as core the [corresponding sql-statement](https://github.com/deltachat/deltachat-core-rust/blob/master/src/chat.rs#L1944) works with real-chat-ids.

i think, the idea of this handling was to "move the deaddrop away" when the user taps "archive" for it, not sure, however, if that is nice ux (there is no "archive" for the deaddrop)

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -348,7 +346,7 @@ protected void onPreExecute() {
           protected Void doInBackground(Void... params) {
             for (long chatId : selectedConversations) {
               if (chatId == DcChat.DC_CHAT_ID_DEADDROP) {
-                dcContext.marknoticedContact(getListAdapter().getDeaddropContactId());
+                dcContext.marknoticedChat(DcChat.DC_CHAT_ID_DEADDROP);
```
r10s, line 349: same here: `dc_marknoticed_chat(DC_CHAT_ID_DEADDROP)` won't work, it would just be a no-op as core the [corresponding sql-statement](https://github.com/deltachat/deltachat-core-rust/blob/master/src/chat.rs#L1944) works with real-chat-ids.

i think, the idea of this handling was to "move the deaddrop away" when the user taps "delete" for it. while archiving above is already a bit weird, moving the deaddrop only "out of sight" when the user selects "delete" seems to be wrong to me.

r10s (commented):
so, from my point of view, this is okay and can be merged once core is ready.

PR #1786, opened 2021-02-02
Replace 4 weeks option with 5 weeks option

link2xt:
While 4 weeks always fits in a month,
5 weeks (35 days) always covers at least a month,
so it can be used in a situation where messages are required to
be stored for a month.


link2xt:
This was requested on the forum: https://support.delta.chat/t/ephemeral-messages/1415

Related recent PR: #1780

adbenitez (approved):

r10s:
i pushed the new string to transifex :)

PR #1792, opened 2021-02-04
show emojis on map

### LINKED ISSUE #1783: Emoji markers are not displayed on the map
One-letter markers are displayed directly (no need to click on the marker), but emojis don't.

Probably something is wrong with this length check and emojis length is not "1"? I tried replacing it with `neq` and `0`, but then it stopped displaying any markers at all, even 1-letter english markers:
https://github.com/deltachat/deltachat-android/blob/370335bf6e4f28a79a3f0a654035f4dafecb85ad/src/org/thoughtcrime/securesms/map/MapDataManager.java#L459

It's fixed in the core (https://github.com/deltachat/deltachat-android/blob/370335bf6e4f28a79a3f0a654035f4dafecb85ad/src/org/thoughtcrime/securesms/map/MapDataManager.java#L459), and I have tested with python that markers are returned by the FFI just as well as letters.

cyBerta:
closes #1783


r10s:
is this still WIP?

cyBerta:
I would like to double check that next monday

r10s:
> I would like to double check that next monday

great!

cyBerta:
@link2xt maybe you can have a look at this PR and try it out?

src/org/thoughtcrime/securesms/components/emoji/EmojiProvider.java
```diff
@@ -158,6 +172,26 @@ public int getIntrinsicHeight() {
...
+      final int rowIndex = info.getIndex() % EMOJI_PER_ROW;
+
+      Rect desRect = new Rect(0, 0, (int) intrinsicWidth, (int) intrinsicWidth);
+      Rect srcRect = new Rect((int)(rowIndex * intrinsicWidth),
+        (int)(row * intrinsicHeight + row * verticalPad)+1,
+        (int)(((rowIndex + 1) * intrinsicWidth)-1),
+        (int)((row + 1) * intrinsicHeight + row * verticalPad)-1);
+
+      Canvas canvas = new Canvas(singleEmoji);
+      canvas.scale(0.5f, 0.5f);
```
link2xt, line 188: Where does the 0.5 scale comes from? Why the emoji returned is half the size of intrinsicWidth and intrinsicHeight?

src/org/thoughtcrime/securesms/map/DataCollectionTask.java
```diff
@@ -64,29 +70,32 @@ public static void cancelRunningTasks() {
...
+          emojiCodePoints,
+          emojiProvider,
+          boundingBuilder);
         for (int contactId : contactIds) {
            dataCollector.updateSource(chatId,
-                    contactId,
-                    System.currentTimeMillis() - TIME_FRAME,
-                    TIMESTAMP_NOW);
+             contactId,
+            System.currentTimeMillis() - TIME_FRAME,
```
link2xt, line 86: Something is wrong with the indentation

src/org/thoughtcrime/securesms/map/EmojiBitmapGenerationTask.java
```diff
@@ -0,0 +1,56 @@
...
+    super.onProgressUpdate(values);
+    EmojiBitmapGenerationCallback callback =  callbackWeakReference.get();
+    if (callback != null) {
+      callback.onEmojiBitmapCreated(values[0].first(), values[0].second());
+    }
+  }
+
+  @Override
+  protected void onPostExecute(Void aVoid) {
+    super.onPostExecute(aVoid);
```
link2xt, line 49: Is it necessary to override these methods?

src/org/thoughtcrime/securesms/map/EmojiBitmapGenerationTask.java
```diff
@@ -0,0 +1,56 @@
...
+
+import android.graphics.Bitmap;
+import android.os.AsyncTask;
+
+import org.thoughtcrime.securesms.components.emoji.EmojiProvider;
+import org.thoughtcrime.securesms.util.Pair;
+
+import java.lang.ref.WeakReference;
+
+public class EmojiBitmapGenerationTask extends AsyncTask<String, Pair<String, Bitmap>, Void> {
```
link2xt, line 11: https://developer.android.com/reference/android/os/AsyncTask says AsyncTask is deprecated.

Given that progress update and other features are not used, maybe use a simple thread like suggested in https://stackoverflow.com/a/63500657 ?

src/org/thoughtcrime/securesms/map/MapDataManager.java
```diff
@@ -279,6 +303,16 @@ public void onDataCollectionFinished() {
         }
     }
 
+    public void handleEmojiCodepoints(Set<String> emojiCodePoints) {
+      // generate only new emoji bitmaps, so check the diff
```
link2xt, line 307: I don't understand this comment :/

link2xt (approved):
Seems to work, I commented on the code as much as I can.

I noticed that "Use system emoji" setting is ignored, but this is not expected, rendering system fonts into images is a lot of unnecessary additional work.

src/org/thoughtcrime/securesms/components/emoji/EmojiProvider.java
cyBerta: the original size of each emoji is 64x64 which is far too big, if it's rendered on the map. So we need to scale the cut out emoji down.

link2xt: But EmojiProvider itself and getEmojiBitmap has nothing to do with the map. Maybe add an argument to set bitmap scale and pass 0.5 from the map code here?

src/org/thoughtcrime/securesms/map/EmojiBitmapGenerationTask.java
cyBerta: Dang! Thanks for the hint. It makes to not implement new stuff using the pattern.

src/org/thoughtcrime/securesms/map/MapDataManager.java
cyBerta: I'll try to explain it more in detail

src/org/thoughtcrime/securesms/map/EmojiBitmapGenerationTask.java
cyBerta: nope. thanks, that's generated boilerplate

src/org/thoughtcrime/securesms/components/emoji/EmojiProvider.java
cyBerta: good idea +1

src/org/thoughtcrime/securesms/map/EmojiBitmapGenerationTask.java
link2xt: It's used, but only to return the result once.

cyBerta: no, look at line 32, it's in a loop and it's the method to invoke onProgressUpdate()

cyBerta: Anyways the whole class vanished in the meanwhile

src/org/thoughtcrime/securesms/components/emoji/EmojiProvider.java
```diff
@@ -158,6 +172,26 @@ public int getIntrinsicHeight() {
       this.intrinsicHeight = EMOJI_RAW_HEIGHT * decodeScale;
     }
 
+    private Bitmap getEmojiBitmap(float scale) {
+      Bitmap singleEmoji = Bitmap.createBitmap((int) intrinsicWidth/2, (int) intrinsicHeight/2, Bitmap.Config.ARGB_8888);
```
link2xt, line 176: Shouldn't this be `intrinsicWidth * scale`? Probably makes sense to try that setting different scale (`2.5f`, `0.7f`) actually works.

cyBerta: +1 nice catch

cyBerta: fixed and tested with different scales

PR #1795, opened 2021-02-07
show name of forwarder in groups

r10s:
recent discussions with @adbenitez @flub @Simon-Laux and other on DC, this was also one of the suggestions of @Simon-Laux at https://github.com/deltachat/deltachat-desktop/issues/1977 :

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -113,6 +113,7 @@
     <string name="audio">Audio</string>
     <string name="voice_message">Voice message</string>
     <string name="forwarded_message">Forwarded message</string>
+    <string name="forwarded_by">Forwarded by %1$s</string>
     <string name="video">Video</string>
     <string name="documents">Documents</string>
     <string name="contact">Contact</string>
--- a/src/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationItem.java
@@ -654,12 +654,15 @@ public class ConversationItem extends LinearLayout
 
   private void setGroupMessageStatus() {
     if (messageRecord.isForwarded()) {
-      this.groupSender.setText(context.getString(R.string.forwarded_message));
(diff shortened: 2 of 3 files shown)


flub (approved):
lgtm, even though i don't know much about this code.  but the change in isolation seems reasonable.

On a more general note however, I'm a little surprised you chose to do this in the UIs rather than in the core as now all UIs have to implement this (this is obviously fine for this PR, it was already done like this with "forwarded message").  I was sort of expecting this to be handled by the core with the UIs checking a message attribute for the any leading text to display.

r10s:
there is no existing api that gets the "forwarded message text", so tbh, i do not really came to the idea, that this can be solved in core.

and adding an api (this was your expectation?) would also add some work for all uis, so, not sure if that is worth the thing. also, eg. desktop shows the headlines a bit different, might be they stay with that (cmp, https://github.com/deltachat/deltachat-desktop/issues/1977).

or have i overseen a much simpler solution?

flub:
Sorry, my comment was not meant for this PR, this is how it currently works and that's fine.  I was wondering if it would have made more sense to move this kind of work in the core instead and was kind of expecting to find it there.  If that's the case then it might be something to work on in the future.  However I don't work on any of the apps so I could be entirely wrong on which side this should land :)
And yes, if this is moved to the core it depends on all UIs to adopt it (in this case probably wouldn't hurt if they didn't, but that doesn't generalise), so I realise that's also part of the trade-off.

r10s:
yes, it is not always simple to decide where to draw the line between core and ui, in general, compared with other libs i've seen, our core, however, is pretty close to ui whereever this makes sense. but we've also stepped backward here and there, eg. we removed the chat-subtitle-generation from core as it turns out, the plurals was nothing we want to deal with (and callbacks were too much hazzle). i assume this "forward title" would be in the same area - much work in core with only few benefit in ui plus pushing the ui in one direction (you cannot easily eg. on desktop, where there is more screenspace, split that into two lines if we decide so)

PR #1805, opened 2021-02-11
Show status

### LINKED ISSUE #1804: Display contact status in the profile
See related core PR: https://github.com/deltachat/deltachat-core-rust/pull/2218

New API is `dc_contact_get_status`

r10s:
this is how it looks like:

pretty neutral, with title (same as in the profile), i think, that should not raise many questions.

on protocol-level, the text comes in the email footer/signature, it can be defined in Delta Chat as "Signature text" in the profile and it can also be defined in nearly all other mua.

some words to the layout changes: the e-mail address went to the header (where it is also displayed in similar layouts); the 'send message' button is atop for simplicity if the footer/status becomes longer (when opening contact-profiles from group-members, this button should always be at the same position and should be easily reachable)

to the wording: as we are not using the wording "status" in the shipped apps and the predefined text is indeed more a footer/signature, i would not say "status" for now.

Todo:
- [x] some cleanup in profile code
- [x] add ffi
- [x] show basic signature
- [x] make text copyable
- [x] linkify
- [x] figure out how to make the subview clickable

closes #1804 

<details><summary>previous layouts</summary>

_while that is a bit more shiny, the new yellow color introduces a new thing to explain. also, a title is missing. so, i'd say, using a more unagitated layout is better for now, also as this thing is currently a mix of status/signature/classic-footer ..._
</details>

### DIFF
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -1547,6 +1547,15 @@ JNIEXPORT jint Java_com_b44t_messenger_DcContact_getColor(JNIEnv *env, jobject o
 }
 
 
+JNIEXPORT jstring Java_com_b44t_messenger_DcContact_getStatus(JNIEnv *env, jobject obj)
+{
+    char* temp = dc_contact_get_status(get_dc_contact(env, obj));
+        jstring ret = JSTRING_NEW(temp);
+    dc_str_unref(temp);
+    return ret;
+}
+
+
 JNIEXPORT jboolean Java_com_b44t_messenger_DcContact_isBlocked(JNIEnv *env, jobject obj)
 {
     return (jboolean)(dc_contact_is_blocked(get_dc_contact(env, obj))!=0);
--- a/res/layout/profile_status_item.xml
+++ b/res/layout/profile_status_item.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<org.thoughtcrime.securesms.ProfileStatusItem
+             xmlns:android="http://schemas.android.com/apk/res/android"
+             android:layout_width="match_parent"
+             android:layout_height="wrap_content"
+             android:background="?attr/conversation_list_item_background"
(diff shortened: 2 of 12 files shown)


r10s:
btw, another layout idea that came up after the current layout was just finished: maybe just use a normal bubble, pointing to the left, similar to the ones used in the normal chat. that would make it pretty clear "who" is "saying" that.

link2xt:
The text is not copyable and the status is highlighted like a menu item when clicked. Making it look and behave like a message could help indeed, maybe with a word "Signature" in bold at the top, like "Forwarded message". There are sometimes links to home pages I want to click or copy.

Hocuri:
+1 for some way of indicating that this is a signature/status and not a warning from DC

r10s:
@Hocuri the text is linkified now and things should work, however, the EmojiTextView and its links does not really get clicks, i played around with adding onClick handlers to the view., however, no direct success. the web is full of things that could be done, and i am pretty sure i did sth. like that before, but i forgot :) maybe it is just too late :) do you have an idea?

link2xt (approved):
Works great. Links are clickable. Would be perfect if I could select any part of the text (a single word, for example) and copy it rather than the whole status.

r10s:
thanks for review!

> Would be perfect if I could select any part of the text (a single word, for example) and copy it rather than the whole status.

i think, that would be useful in general, also for the bubbles. but that's for another pr, for now i'd like to go for sth. else :)

Master0ne:
I'm not sure if re-purposing the email signature for use as "Status" in Delta Chat is the best idea, because it makes sense to put some info into the signature relevant for receives who are not using Delta Chat (yet).

I have posted my thoughts on this in the following forum thread:

[Catchy signature for convincing people to have a look at Delta Chat?](https://support.delta.chat/t/catchy-signature-for-convincing-people-to-have-a-look-at-delta-chat/1605)

PR #1808, opened 2021-02-14
mark overridden names by a "~" in the bubbles

r10s:
this is just a very simple approach to move forward.

to mark overridden names, a "~" is prefixed.

the idea with using a dedicated color for these names:
as we do not really know
if the address is a generic sender address or belongs to the user,
it makes sense to keep the color used for the name and do not use a
dedicated grey or so.

i think, it is not even worth to detect that better:
it is pretty nice to have senders in mailinglist marked nicely,
all the time.

currently, the "~" is only added to the bubbles - it is not there
in the chatlist-summary (that would needed to be done in core)
and also not in forwarded by, notifications, clipboard etc. i'd say currently,
this is not needed, however, with the current approach it would be doable
(if we mark possible impersonation by colors or icons that may be
much more complicated)

adaption on other uis would also be pretty simple with this approach,
and imho, this `~Max Mustermann` looks pretty nicely
and even intuitive in a way.

drawback of the approach is that user may user `~`
as the first character of their names, we could target that by
adapting dc_msg_get_displayname() in the core, if we want to keep this approach,
however, maybe it is also not worth, the "~" is more a flaw than a feature,
also that would result to "~~Name" in the bubbles.
i do not see how that can be used to trick users when beeing added.
names as "foo 🔒" might be more dangerous here, when it comes to trick users.

### DIFF
--- a/src/com/b44t/messenger/DcMsg.java
+++ b/src/com/b44t/messenger/DcMsg.java
@@ -131,10 +131,10 @@ public class DcMsg {
     public native String  getError           ();
     private native @Nullable String getOverrideSenderName();
 
-    public @NonNull String getSenderName(@NonNull DcContact dcContact) {
+    public @NonNull String getSenderName(@NonNull DcContact dcContact, boolean markOverride) {
         String overrideName = getOverrideSenderName();
         if (overrideName != null) {
-            return overrideName;
+            return (markOverride ? "~" : "") + overrideName;
         } else {
             return dcContact.getDisplayName();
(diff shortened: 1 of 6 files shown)


adbenitez (approved):

Simon-Laux:
>  i do not see how that can be used to trick users when beeing 

I think an ui element that isn't part of the name text, like a [BOT] label box , is even safer, but for mailinglists `~` could be sufficient.

PR #1822, opened 2021-03-01
Call stop_io()/start_io() before/after backup export

Hocuri:
should ideally be merged before https://github.com/deltachat/deltachat-core-rust/pull/2255

I tested key import/export and backup import/export, all work fine.


r10s (approved):
lgtm, thanks for taking care!

PR #1825, opened 2021-03-04
remove android dependencies from bindings

r10s:
targets https://github.com/deltachat-bot/echo/issues/9

now, it should be possible to check out the android project and use only the `jni`  and the `src/com/b44t/messenger` directory in combination with plain java, without android.

at some point, we could think over to move the bindings to a separate project, however, i am not in favor of doing that now as this would comes at costs wrt android development. but if it is actually in use independently of android, we can think over it.

### DIFF
--- a/src/com/b44t/messenger/DcChatlist.java
+++ b/src/com/b44t/messenger/DcChatlist.java
@@ -1,8 +1,5 @@
 package com.b44t.messenger;
 
-import androidx.annotation.NonNull;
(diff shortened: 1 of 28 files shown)


r10s (commented):
just noticed: DcContact still references android.graphics.Color, i'll add a commit.

r10s:
also removed android.graphics.Color from bindings, now that should be usable standalone.

Hocuri (approved):

PR #1826, opened 2021-03-05
mark overridden senders in quotes

r10s:
if the original author of a quote is shown
and known to be overridden,
that should be marked with the character `~`
as for the normal names,
esp. as they are look very similar otherwise,
eg. also bold, also colored.
the missing `~` looks like a bug here.

successor of #1808

this also aligns display of overridden names with desktop, where the `~` is also added to quotes at https://github.com/deltachat/deltachat-desktop/pull/2132


r10s:
tested that, that feels much better. esp. as you can tap on the quote with the name and see the same name then.

Hocuri (approved):

PR #1829, opened 2021-03-07
Use the message subject as the full-msg-view title

Hocuri:
For long subject it's ellipsized (`…`) now, but I think that there is no need to show the full subject somewhere, it usually doesn't contain _that_ important information and can be seen in full in the normal chat view


r10s (approved):
lgtm, and yes, for now an ellipse for too long subjects is totally fine.

i also just updated core on master to get https://github.com/deltachat/deltachat-core-rust/pull/2274 in, so that this can be merged directly.

PR #1831, opened 2021-03-08
Make the upper left back button return to conversation list even if the keyboard is open

Hocuri:
As discussed in the Testing group


r10s (approved):
lgtm

PR #1832, opened 2021-03-10
Allow to share media from Delta Chat

Hocuri:
The IMHO main open point is to discuss where exactly to put the share button.

I put it before the "export attachment" button now, because I think that it will be used more often.

Also, many people will not know what this "export attachment" diskette symbol means, so that it's probably better if it goes to the overflow menu, where the user doesn't see the diskette symbol but "export attachment". 
Although I'm not sure whether "export attachment" is clear, either. Maybe we should rename it to "Save to gallery"(image)/"Save to downloads"(others)?


r10s:
the diskette is indeed a bit antiquated, just putting that to the overflow menu and having it as a named entry is better. for the naming, i would stuck with  "export attachement" for now, although other wordings may work as well, we did not have many issues with that. also, it is used for documents as well, being too specific here, opens other issues (also, we do not really know if things really land in the gallery on every system)

Simon-Laux (approved):
Nice!

Hocuri:
> lgtm, the icon with the diskette is no longer used and can be removed then?

Currently not, it's just shown less often currently. But it probably makes sense to completely remove it.

Hocuri:
I pushed two more commits:
- Don't use the diskette icon for export attachment at all
- completely delete diskette icon, don't use it for "save log", either

Hocuri:
I also ran auto-indent lines, so, use "ignore whitespace" :)

Hocuri:
> also, it is used for documents as well, being too specific here, opens other issues

I meant "Save to gallery" for photos, "Save to downloads" for all others (documents etc.)

> (also, we do not really know if things really land in the gallery on every system)

Isn't "the gallery" just the name of the app that shows all images that are in the public storage? I hope that the images land there, if not then most users will never find the image.

r10s:
>I meant "Save to gallery" for photos, "Save to downloads" for all others (documents etc.) [...]
> Isn't "the gallery" just the name of the app that shows all images that are in the public storage? I hope that the images land there, if not then most users will never find the image.

this is how it usually works. but already when the gallery app is not called "Gallery" things may not match perfectly ...

therefore, i think, the current, a bit broader wording is "good enough" and simpler than more fine-tuned wordings that may not fit.

r10s (approved):
still lgtm, nice to get rid of the diskette :)

PR #1833, opened 2021-03-10
add lock recording feature

adbenitez:
allow to lock recording so you don't have to keep your finger pushing
(adapted from Signal code)

deltachat-2021-03-10-183120

### DIFF
--- a/res/layout/microphone_recorder_view.xml
+++ b/res/layout/microphone_recorder_view.xml
@@ -1,0 +1,64 @@
+<?xml version="1.0" encoding="utf-8"?>
+<merge xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    tools:showIn="@layout/conversation_input_panel">
+
+    <ImageButton
+        android:id="@+id/quick_audio_toggle"
+        android:layout_width="24dp"
+        android:layout_height="match_parent"
+        android:layout_gravity="center_vertical"
+        android:layout_marginEnd="2dp"
+        android:background="@null"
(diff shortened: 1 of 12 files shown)


Hocuri:
For some reason, Signal made the touch area for recording a lot narrower:

Current DC:
Screenshot_20210311-095546
This PR:
Screenshot_20210311-095539_1
Signal:
Screenshot_20210311-095529

Personally, I liked the wider touch area, making it easier to hit the button, so I vote for making it wider again.

Hocuri:
Apart from that, lgtm

gerryfrancis:
> For some reason, Signal made the touch area for recording a lot narrower

I think they did this because a wider touch area was necessary before. Now the user is not forced to tap and hold the finger that precisely anymore. ;)

Hocuri:
It was never necessary to hold the finger in that position; only if you move your finger 2cm to the left, the recording is cancelled. BTW WhatsApp and Telegram both have wide touch areas, too (also, they are higher, but anyway)

res/layout/microphone_recorder_view.xml
```diff
@@ -0,0 +1,67 @@
...
+        android:contentDescription="@string/audio"
+        android:src="?quick_mic_icon" />
+
+    <LinearLayout
+        android:id="@+id/lock_drop_target"
+        android:layout_width="40dp"
+        android:layout_height="64dp"
+        android:layout_gravity="center"
+        android:layout_marginLeft="-9dp"
+        android:layout_marginStart="-9dp"
```
Hocuri, line 24: This corrects a small shift to the right of the lock drop target.

I have no idea why the shift was there in the first place, though.

Hocuri (approved):
@r10s said that my additions don't need another approval :upside_down_face: So, lets merge it!

PR #1834, opened 2021-03-11
Don't open media preview for attachments with file viewtype

### LINKED ISSUE #1816: Clicking on SVG with File viewtype opens media preview
Here ThumbnailClickListener is set for "documents" (files):
https://github.com/deltachat/deltachat-android/blob/a19e739d89f76a7126a984102e57d908215f0f5d/src/org/thoughtcrime/securesms/ConversationItem.java#L436

When clicked, the check is `MediaPreviewActivity.isContentTypeSupported`, which supports anything `image/*` and `video/*`:
https://github.com/deltachat/deltachat-android/blob/a19e739d89f76a7126a984102e57d908215f0f5d/src/org/thoughtcrime/securesms/ConversationItem.java#L758

So media preview opens, but instead of SVG some random image is opened.

Maybe the check `isContentTypeSupported` should be removed and instead only viewtype checked? Or check that we can actually load the image? Why random image is opened at all?

cc @adbenitez

Hocuri:
fix #1816

The problem was:
- as @link2xt already wrote, the onclick listener only checks if the mime type is `image/` or `video/`.
- then `PagingMediaLoader` loads all images and videos of the current chat with `context.getChatMedia(msg.getChatId(), DcMsg.DC_MSG_IMAGE, DcMsg.DC_MSG_GIF, DcMsg.DC_MSG_VIDEO)`. We need _all_ images so that the user can swipe back and forth.
- The correct image to be shown can't be found, the MediaPreview is opened anyway. As the correct image can't be shown, the first image in the chat is shown.

### DIFF
--- a/src/org/thoughtcrime/securesms/MediaPreviewActivity.java
+++ b/src/org/thoughtcrime/securesms/MediaPreviewActivity.java
@@ -26,14 +26,6 @@ import android.os.AsyncTask;
 import android.os.Build.VERSION;
 import android.os.Build.VERSION_CODES;
 import android.os.Bundle;
-import androidx.annotation.NonNull;
-import androidx.annotation.Nullable;
-import androidx.loader.app.LoaderManager;
-import androidx.loader.content.Loader;
-import androidx.viewpager.widget.PagerAdapter;
-import androidx.viewpager.widget.ViewPager;
-import androidx.appcompat.app.AlertDialog;
-
 import android.util.Log;
 import android.view.LayoutInflater;
 import android.view.Menu;
@@ -44,7 +36,14 @@ import android.view.ViewGroup;
 import android.view.Window;
 import android.view.WindowManager;
 import android.widget.FrameLayout;
-import android.widget.Toast;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import androidx.appcompat.app.AlertDialog;
+import androidx.loader.app.LoaderManager;
+import androidx.loader.content.Loader;
+import androidx.viewpager.widget.PagerAdapter;
+import androidx.viewpager.widget.ViewPager;
 
 import com.b44t.messenger.DcMediaGalleryElement;
 import com.b44t.messenger.DcMsg;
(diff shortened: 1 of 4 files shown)


src/org/thoughtcrime/securesms/MediaPreviewActivity.java
```diff
@@ -230,13 +230,6 @@ private void initializeMedia() {
 
     // if you search for the place where the media are loaded, go to 'onCreateLoader'.
 
-    if (!isContentTypeSupported(initialMedia.type)) {
-      Log.w(TAG, "Unsupported media type sent to MediaPreviewActivity, finishing.");
-      Toast.makeText(getApplicationContext(), R.string.error, Toast.LENGTH_LONG).show();
-      finish();
-      return;
-    }
```
Hocuri, line 238: As `isContentTypeSupported` takes a `Slide` now, and we don't have one here, I removed this check. It turned out not to work, anyway.

r10s (approved):
lgtm code wise and makes sense. did not test, though.

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
```diff
@@ -407,8 +400,8 @@ private boolean isMediaInDb() {
     }
   }
 
-  public static boolean isContentTypeSupported(final String contentType) {
-    return contentType != null && (contentType.startsWith("image/") || contentType.startsWith("video/"));
+  public static boolean isContentTypeSupported(final Slide slide) {
```
link2xt, line 403: The name of this function does not make sense anymore, as it does not check MIME type. Maybe remove this function completely, as it is only used in one place anyway?

link2xt (approved):

src/org/thoughtcrime/securesms/MediaPreviewActivity.java
Hocuri: It has two usages unfortunately:/ But I can rename it to `isTypeSupported`

PR #1837, opened 2021-03-13
Better profile and group picture selection by using attachment selector

### LINKED ISSUE #1836: Icons and labels not shown when capturing image for avatar or deleting it
- Operating System (Linux/Mac/Windows/iOS/Android): Android 10
- Delta Chat Version: nightly/master
When changing the profile image, then instead of "Capture image" and "Remove" it says "Android S...". Also, the icons just show a gear in a green circle.
- Screenshots:
Screenshot_20210313-130536

**Possible fixes:**
- Pull code from Signal. They implemented  this in quite a complex way, but OK.
- Just show an AlertDialog with 3 options "Take photo", "Choose from gallery" and "Remove photo". The easier solution, but won't look as nice.
- Imitate Telegram and re-use the attachment-chooser dialog. This will have the nice effect that users already know it and they are already offered some options.
- _Probably doesn't work:_ "Directly" fix this, just set the correct label and icon: I tried to use `LabeledIntent`s , but it didn't work. I tried some other things, but they didn't work, either.

Hocuri:
fix #1836

AvatarSelector is mostly copied from AttachmentTypeSelector. I first started to make it a subclass, but there were so many functions that had to be changed that I decided against it (esp. also the constructor, so I would have had to create another constructor to allow AvatarSelector to bypass the "normal" constructor). Could still be done though, of course.

I think (hope :joy:) that the commits are clear enough that this can be reviewed commit-by-commit. In the first commit, commenting out code means "Moved somewhere else"

I found it easier to look at the first commit using gitk than GitHub as GitHub doesn't show it as copied from AttachmentTypeSelector but as a new file. (of course you can also just review normally, going throug it file-by-file :)


Hocuri:
Lol, the documentation at https://developer.android.com/training/permissions/evaluating#take-photo is misleading, you do have to ask for camera permission before letting the user choose a photograph with another app

Hocuri:
But you do not need storage permission to let the user choose a file

adbenitez (approved):
wow! looks much better now :ok_hand:

Hocuri:
Oops, but I just noticed that the question when the user wants to remove an image is wrong ("Remove profile image?" instead of "Remove group image?"). But actually, we can probably just remove this confirmation, the user has to confirm again, anyway.

PR #1844, opened 2021-03-25
Explain Podman setup better for new users

Hocuri:
A version where the in-document links work (they don't work in GitHubs `Files changed` view): https://github.com/deltachat/deltachat-android/blob/podman-setup/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,26 +51,41 @@ or
 docker run -it --name deltachat -v $(pwd):/home/app -w /home/app localhost/deltachat-android
 ```
 
+You can leave the container with Ctrl+D or by typing `exit` and re-enter it with `podman start -ia deltachat`.
+
 Within the container, install toolchains and build the native library:
 ```
 deltachat@6012dcb974fe:/home/app$ scripts/install-toolchains.sh
 deltachat@6012dcb974fe:/home/app$ ./ndk-make.sh
 ```
 
-Leave the container with Ctrl+D (you can return later using `podman start -ia deltachat`).
-
 Then, [build an APK](https://developer.android.com/studio/build/building-cmdline):
 ```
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -58,9 +65,42 @@ Then, [build an APK](https://developer.android.com/studio/build/building-cmdline
...
+
+Then in /etc/containers/storage.conf, replace the line
+```
+driver = ""
+```
+with:
+```
+driver = "overlay"
+```
+You can also set the `driver` option to something else, you just need to set it to _something_.
```
link2xt, line 90: I don't have the file `/etc/containers/storage.conf` at all on my Debian Sid. Can't check my Arch setup right now, but I don't remember setting this either.

README.md
```diff
@@ -58,9 +65,42 @@ Then, [build an APK](https://developer.android.com/studio/build/building-cmdline
...
 ```
 
-If you don't want to use Docker or Podman, proceed to the next section.
+## Troubleshooting
+
+- `./gradlew assembleDebug` fails with `The SDK directory '/home/user/Android/Sdk' does not exist.`:
+  
+  You have to
+  - either: remove the line starting with `sdk.dir=` from your `local.properties` file
+  - or: execute `./gradlew assembleDebug` from outside the container
```
link2xt, line 74: Not sure this is a good idea. Core is linked using a toolchain from android NDK, so if you have a different version installed outside the container, this may cause incompatibility issues.

Probably better just say that this can be caused by `sdk.dir` in `local.properties` file created by Android Studio and let the user decide what to do about it.

link2xt (approved):

README.md
Hocuri: I'll move it to the "Troubleshooting" section then

link2xt (approved):

Issue #1849, opened 2021-03-27
target api level 30, until  November 2021


r10s:
currently, we are targeting api level 29; message from 23 Mar 2021 in gplay console inbox:

> **New Android App Bundle and target API level requirements in 2021**
>
> Starting August 2021, new apps will be required to target API level 30 (Android 11) and use the Android App Bundle publishing format. Starting November 2021, all app updates will be required to target API level 30 (Android 11). Apps with a download size of more than 150 MB are now supported by Play Asset Delivery and Play Feature Delivery.

seems as if we can stay with the .apk format at least:

<img width=500 src=https://user-images.githubusercontent.com/9800740/112733239-fa98e300-8f3e-11eb-8b84-84351e202cf1.png>
 
source: https://android-developers.googleblog.com/2020/11/new-android-app-bundle-and-target-api.html  
related: https://github.com/deltachat/deltachat-android/issues/1777

r10s:
cmp https://github.com/deltachat/deltachat-android/issues/1561

PR #1852, opened 2021-03-31
 Adapt to new direct share api

Hocuri:
Actually, I wouldn't be surprised if the new API is changed again soon, as there are lots of rough edges. Like, when are you supposed to update your shortcuts? Why does `addDynamicShortcuts()` throw an exception if there are too many existing shortcuts, instead of just pushing old shortcuts out of the list? (that's why I call `removeAllDynamicShortcuts()` if they might get too many shortcuts)

Anyway, enough grumbling :joy: If you have questions about the new Android API, it may be best to make an a/v.


Hocuri:
BTW just wondering if this should even get in before the next release. It's difficult to test as lots of Android versions behave differently here, and there are lots of pitfalls in the Android API, so that there may well be more bugs. 

OTOH, direct sharing isn't actually a core feature, and currently (without this PR) direct-sharing often as-good-as doesn't work on very recent Android versions as we use the old sharing API and these are ranked very low, often pushed out of by other apps. @r10s ?

Hocuri:
@r10s @adbenitez we had postponed this PR to after the release, so I think it can be reviewed and merged now

Hocuri:
So, maybe review now? Then we can merge it as soon as the 1.22 "merge stop" is over and have a long testing period before the next release (and IIRC this has to get in before we can target API 30 (https://github.com/deltachat/deltachat-android/issues/1849)).

The downside is, I won't have much time until the 24th of September and may not be able to fix appearing bugs. Until the 11th of September I may not even answer questions, or only with some delay.

r10s:
@Hocuri hm, after all experience with "share on android", it is reasonable to expect issues on merging this pr. so, maybe better wait until you are really back and can take care. we have time [until end of October to finally target API 30](https://github.com/deltachat/deltachat-android/issues/1849), so it is not (yet) super-urgent.

Hocuri:
@adb @r10s review now, then?

adb:
> @adb @r10s review now, then?

you probably want @adbenitez

Hocuri:
Oops, yes :joy:

src/org/thoughtcrime/securesms/connect/DirectShareUtil.java
```diff
@@ -0,0 +1,160 @@
...
+      Util.runOnAnyBackgroundThread(() -> ShortcutManagerCompat.removeDynamicShortcuts(context, Collections.singletonList(Integer.toString(chatId))));
+    }
+  }
+
+  public static void resetAllShortcuts(@NonNull Context context) {
+    Util.runOnBackground(() -> {
+      ShortcutManagerCompat.removeAllDynamicShortcuts(context);
+      triggerRefreshDirectShare(context);
+    });
+  }
```
Hocuri, line 61: TODO: Wrap with try-catch

PR #1859, opened 2021-04-07
Add APNG and animated webp support

adbenitez:
besides support on normal image messages, this means Delta Chat Android will also have "animated stickers" which, btw, already work in Delta Chat Desktop.

Images for testing:
* animated web:
https://isparta.github.io/compare-webp/image/gif_webp/webp/2.webp
* animated png:
https://misc.aotu.io/ONE-SUNDAY/SteamEngine.png


Hocuri (approved):

r10s (approved):
displaying seems to be okay, when it comes to generating, we should have a closer look what is actually supported on other os (at a first glance, ios supports apng but not webp out-of-the-box, https://github.com/SDWebImage/SDWebImage, but even for apng, support may be buggy). but, as said, just displaying makes sense - esp. if it is only one line of code change :)

PR #1860, opened 2021-04-08
Connectivity view

Hocuri:
See https://github.com/deltachat/deltachat-core-rust/pull/2319 (you need to be on core branch `connectivity-view` to use this)


adbenitez:
@Hocuri CI is failing :(

Hocuri:
Yes, because the core PR is not merged yet

r10s:
nb: i'd say, we should merge that after the final email-compat releases, otherwise, tweaking that would delay releases even further.

src/com/b44t/messenger/DcContext.java
```diff
@@ -184,6 +195,23 @@ public void unref() {
...
     public native static boolean data2IsString(int event);
 
+    /**
+     * Gets a string you can show to the user with basic information about connectivity.
+     * @param context
+     * @param connectedString Usually "Connected", but when using this as the title in
+     *                        ConversationListActivity, we want to write "Delta Chat" there instead.
+     * @return
+     */
+    public static String getConnectivitySummary(Context context, int connectedString) {
```
adbenitez, line 205: this is a regression, needs android specific stuff and parts of the delta chat application, reverting previous efforts to keep pure Java bindings modular and separated from the rest of the application, 

the bindings should be used as if they were an external dependency, and actually would be great if at some point they were just a line in the dependencies section in build.gradle to avoid future temptations ;)

I guess this should be added to ApplicationDcContext maybe?

adbenitez (changes requested):

src/com/b44t/messenger/DcContext.java
Hocuri: Will change that, thanks for taking care

r10s:
to move forward here - can we split the "remove screen lock features" off this pr? that would make things much easier reviewable and allow to focus on things better.

Hocuri:
Sure

Hocuri:
@adbenitez could you remove your blocking "changes requested" review, or even review again?

res/values/strings.xml
```diff
@@ -63,7 +63,6 @@
     <string name="next">Next</string>
     <string name="error">Error</string>
     <string name="error_x">Error: %1$s</string>
-    <string name="error_no_network">Network unavailable.</string>
```
Hocuri, line 66: Is it OK to just remove this string, or will there be problems as other UIs depend on it? (same below)

adbenitez (approved):
LGTM 👌

res/values/strings.xml
cyBerta: `error_no_network` is not used on iOS

res/values/strings.xml
```diff
@@ -447,8 +455,6 @@
     <string name="import_backup_no_backup_found">No backups found.\n\nCopy the backup to \"%1$s\" and try again. Alternatively, press \"Start messaging\" to continue with the normal setup process.</string>
     <!-- Translators: %1$s will be replaced by the e-mail address -->
     <string name="login_error_cannot_login">Cannot login as \"%1$s\". Please check if the e-mail address and the password are correct.</string>
-    <!-- Translators: %1$s will be replaced by the server name (eg. imap.somewhere.org) and %2$s will be replaced by the human-readable response from the server. this response may be a single word or some sentences and may or may not be localized. -->
-    <string name="login_error_server_response">Response from %1$s: %2$s\n\nSome providers may e-mail you additional information about this problem; you should be able to check your inbox using your regular e-mail app or web site. Consult your provider or friends if you run into problems.</string>
```
cyBerta, line 451: this one seems to be still used in the core on iOS (DC_STR_SERVER_RESPONSE). I didn't follow the core development, so maybe it's already deprecated.

Hocuri: Well it's deprecated with the introduction of the connectivity-view, as error_network's are removed. But I can leave it in for some time if you say it's better

res/values/strings.xml
Hocuri: Not used on Desktop, either as far as I can see (https://github.com/deltachat/deltachat-desktop/search?p=1&q=error_no_network)

res/values/strings.xml
Hocuri: Let's leave it in for now, then

res/values/strings.xml
r10s: i once wrote a script to answer these kind of questions :)

```
./scripts/grep-string.sh error_no_network
```

shows usages only in node's CHANGELOG.md, so, yes, the string can be removed :)

PR #1862, opened 2021-04-09
use mute API directly

adbenitez:


r10s (approved):
lgtm, thanks for cleaning up! 🧹

PR #1864, opened 2021-04-10
Let stickers handle clicks

Hocuri:
Esp:
- Long-click on a sticker to enter multi-select mode
- click on the sticker again to deselect it

Expected: The sticker is deselected
Actual: Nothing happens


adbenitez:
this also happens for audio messages, would be nice to fix all at once

adbenitez (approved):

PR #1868, opened 2021-04-13
add more options to Gallery and Documents profile tabs

### LINKED ISSUE #1865: Gallery and Documents tabs lacks more menu options
In the chat's Documents tab you have only two options: clicking/opening the document or deleting it, you can't export the document or open in context, so you have to manually search for the document in chat.

In the Gallery tab you can open an element to then have access to more options but would be nice to have some of them directly in the Gallery menu

adbenitez:
close #1865

### DIFF
--- a/src/org/thoughtcrime/securesms/MessageSelectorFragment.java
+++ b/src/org/thoughtcrime/securesms/MessageSelectorFragment.java
@@ -18,7 +18,7 @@ import org.thoughtcrime.securesms.util.SaveAttachmentTask;
 
 import java.util.Set;
 
-public class MessageSelectorFragment extends Fragment
+public abstract class MessageSelectorFragment extends Fragment
 {
   protected ActionMode actionMode;
   protected ApplicationDcContext dcContext;
--- a/src/org/thoughtcrime/securesms/MessageSelectorFragment.java
+++ b/src/org/thoughtcrime/securesms/MessageSelectorFragment.java
@@ -5,6 +5,7 @@ import android.content.Intent;
 import android.net.Uri;
 import android.os.AsyncTask;
 import android.text.util.Linkify;
+import android.view.Menu;
 import android.widget.TextView;
 import androidx.appcompat.app.AlertDialog;
 import androidx.appcompat.view.ActionMode;
@@ -22,6 +23,13 @@ public abstract class MessageSelectorFragment extends Fragment
 {
   protected ActionMode actionMode;
   protected ApplicationDcContext dcContext;
+
+  protected abstract void setCorrectMenuVisibility(Menu menu);
+
+  protected DcMsg getSelectedMessageRecord(Set<DcMsg> messageRecords) {
+    if (messageRecords.size() == 1) return messageRecords.iterator().next();
+    else                            throw new AssertionError();
+  }
 
   protected void handleDisplayDetails(DcMsg dcMsg) {
     String infoStr = dcContext.getMsgInfo(dcMsg.getId());
--- a/src/org/thoughtcrime/securesms/MessageSelectorFragment.java
+++ b/src/org/thoughtcrime/securesms/MessageSelectorFragment.java
@@ -14,12 +14,15 @@ import androidx.fragment.app.Fragment;
 import com.b44t.messenger.DcMsg;
 
 import org.thoughtcrime.securesms.connect.ApplicationDcContext;
+import org.thoughtcrime.securesms.connect.DcEventCenter;
 import org.thoughtcrime.securesms.permissions.Permissions;
 import org.thoughtcrime.securesms.util.SaveAttachmentTask;
 
 import java.util.Set;
 
-public abstract class MessageSelectorFragment extends Fragment
+public abstract class MessageSelectorFragment
+    extends Fragment
+    implements DcEventCenter.DcEventDelegate
 {
   protected ActionMode actionMode;
   protected ApplicationDcContext dcContext;
--- a/src/org/thoughtcrime/securesms/MessageSelectorFragment.java
+++ b/src/org/thoughtcrime/securesms/MessageSelectorFragment.java
@@ -1,0 +1,79 @@
+package org.thoughtcrime.securesms;
+
+import android.Manifest;
+import android.content.Intent;
+import android.net.Uri;
+import android.os.AsyncTask;
+import android.text.util.Linkify;
+import android.widget.TextView;
+import androidx.appcompat.app.AlertDialog;
+import androidx.appcompat.view.ActionMode;
+import androidx.fragment.app.Fragment;
+
+import com.b44t.messenger.DcMsg;
+
+import org.thoughtcrime.securesms.connect.ApplicationDcContext;
+import org.thoughtcrime.securesms.permissions.Permissions;
+import org.thoughtcrime.securesms.util.SaveAttachmentTask;
+
+import java.util.Set;
+
+public class MessageSelectorFragment extends Fragment
+{
+  protected ActionMode actionMode;
+  protected ApplicationDcContext dcContext;
+
+  protected void handleDisplayDetails(DcMsg dcMsg) {
+    String infoStr = dcContext.getMsgInfo(dcMsg.getId());
+    AlertDialog d = new AlertDialog.Builder(getActivity())
+            .setMessage(infoStr)
+            .setPositiveButton(android.R.string.ok, null)
+            .create();
+    d.show();
+    try {
+      //noinspection ConstantConditions
+      Linkify.addLinks((TextView) d.findViewById(android.R.id.message), Linkify.WEB_URLS);
+    } catch(NullPointerException e) {
+      e.printStackTrace();
+    }
+  }
+
+  protected void handleDeleteMessages(final Set<DcMsg> messageRecords) {
+    int messagesCount = messageRecords.size();
+
+    new AlertDialog.Builder(getActivity())
+            .setMessage(getActivity().getResources().getQuantityString(R.plurals.ask_delete_messages, messagesCount, messagesCount))
+            .setCancelable(true)
+            .setPositiveButton(R.string.delete, (dialog, which) -> {
+                int[] ids = DcMsg.msgSetToIds(messageRecords);
+                dcContext.deleteMsgs(ids);
+                actionMode.finish();
+            })
+            .setNegativeButton(android.R.string.cancel, null)
+            .show();
+  }
+
+  protected void handleSaveAttachment(final DcMsg message) {
+    SaveAttachmentTask.showWarningDialog(getContext(), (dialogInterface, i) -> {
+        Permissions.with(getActivity())
+                .request(Manifest.permission.WRITE_EXTERNAL_STORAGE, Manifest.permission.READ_EXTERNAL_STORAGE)
+                .ifNecessary()
+                .withPermanentDenialDialog(getString(R.string.perm_explain_access_to_storage_denied))
+                .onAllGranted(() -> {
+                    SaveAttachmentTask saveTask = new SaveAttachmentTask(getContext());
+                    SaveAttachmentTask.Attachment attachment = new SaveAttachmentTask.Attachment(
+                            Uri.fromFile(message.getFileAsFile()), message.getFilemime(), message.getDateReceived(), message.getFilename());
+                    saveTask.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR, attachment);
+                    actionMode.finish();
+                })
+                .execute();
+    });
+  }
+
+  protected void handleShowInChat(final DcMsg dcMsg) {
+    Intent intent = new Intent(getContext(), ConversationActivity.class);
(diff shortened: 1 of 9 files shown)


adbenitez:
I had to manually format some part of the code so it might be wrong formatted, Emacs has a format style for Java different than the used in this project, which btw is different in some parts, I should find a way to tweak that at some point.

Feel free to tweak/improve this PR

src/org/thoughtcrime/securesms/ProfileGalleryFragment.java
```diff
@@ -184,39 +188,84 @@ public void onMediaLongClicked(DcMsg mediaRecord) {
...
+
+  private void setCorrectMenuVisibility(Menu menu) {
+    Set<DcMsg> messageRecords = getListAdapter().getSelectedMedia();
+
+    if (actionMode != null && messageRecords.size() == 0) {
+      actionMode.finish();
+      return;
+    }
+
+    if (messageRecords.size() > 1) {
```
cyBerta, line 260: you can refactor that: define a boolean variable (e.g. hasMultipleMessages) with the condition `messageRecords.size() > 1` and use the value for the visibility

src/org/thoughtcrime/securesms/ProfileGalleryFragment.java
```diff
@@ -184,39 +188,84 @@ public void onMediaLongClicked(DcMsg mediaRecord) {
...
-
-          for (DcMsg record : records) {
-            dcContext.deleteMsgs(new int[]{record.getId()});
-          }
-          return null;
-        }
-
-      }.execute(mediaRecords.toArray(new DcMsg[mediaRecords.size()]));
+  private void handleDisplayDetails(DcMsg dcMsg) {
+    String info_str = dcContext.getMsgInfo(dcMsg.getId());
```
cyBerta, line 192: minor style thing: I think we're mostly using camelCase in DC Android

src/org/thoughtcrime/securesms/ProfileGalleryFragment.java
```diff
@@ -184,39 +188,84 @@ public void onMediaLongClicked(DcMsg mediaRecord) {
...
+
+  private void handleDeleteMedia(final Set<DcMsg> messageRecords) {
+    int messagesCount = messageRecords.size();
+
+    new AlertDialog.Builder(getActivity())
+            .setMessage(getActivity().getResources().getQuantityString(R.plurals.ask_delete_messages, messagesCount, messagesCount))
+            .setCancelable(true)
+            .setPositiveButton(R.string.delete, (dialog, which) -> {
+                int[] ids = DcMsg.msgSetToIds(messageRecords);
+                dcContext.deleteMsgs(ids);
```
cyBerta, line 214: You moved the deletion here to the main thread. Do you know if the deletion happens async in the core?

src/org/thoughtcrime/securesms/ProfileGalleryFragment.java
adbenitez: that was copy+pasted from somewhere, makes sense to change also in ConversationFragment.java

src/org/thoughtcrime/securesms/ProfileGalleryFragment.java
adbenitez: I basically tried to be consistent because it was done in one approach here and another in ConversationFragment.java since I actually don't fully understand what I am doing :) I decided to take the simple one, if this is wrong please consider fixing it yourself here and in ConversationFragment.java, help is appreciated

src/org/thoughtcrime/securesms/ProfileDocumentsFragment.java
```diff
@@ -172,39 +176,84 @@ public void onMediaLongClicked(DcMsg mediaRecord) {
...
     });
-    builder.setNegativeButton(android.R.string.cancel, null);
-    builder.show();
+  }
+
+  private DcMsg getSelectedMessageRecord() {
+    Set<DcMsg> messageRecords = getListAdapter().getSelectedMedia();
+
+    if (messageRecords.size() == 1) return messageRecords.iterator().next();
+    else                            throw new AssertionError();
```
cyBerta, line 237: Does multi-selection work with this PR? (I didn't build it, but here only single-selection seems to be allowed)

src/org/thoughtcrime/securesms/ProfileGalleryFragment.java
```diff
@@ -184,39 +188,84 @@ public void onMediaLongClicked(DcMsg mediaRecord) {
...
+                    saveTask.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR, attachment);
+                    actionMode.finish();
+                })
+                .execute();
     });
-    builder.setNegativeButton(android.R.string.cancel, null);
-    builder.show();
+  }
+
+  private DcMsg getSelectedMessageRecord() {
```
cyBerta, line 245: same here: https://github.com/deltachat/deltachat-android/pull/1868/files#r613536093

src/org/thoughtcrime/securesms/ProfileDocumentsFragment.java
```diff
@@ -172,39 +176,84 @@ public void onMediaLongClicked(DcMsg mediaRecord) {
...
+
+  private void setCorrectMenuVisibility(Menu menu) {
+    Set<DcMsg> messageRecords = getListAdapter().getSelectedMedia();
+
+    if (actionMode != null && messageRecords.size() == 0) {
+      actionMode.finish();
+      return;
+    }
+
+    if (messageRecords.size() > 1) {
```
cyBerta, line 248: see also https://github.com/deltachat/deltachat-android/pull/1868/files#r613525466

src/org/thoughtcrime/securesms/ProfileDocumentsFragment.java
adbenitez: yes multiselection works, that only means that multiselection is not allowed for "share" and "export attachments" that exception is never thrown because menu items are hidden when multiple are selected, it is just an extra precaution that the items are not left visible by accident

adbenitez: it means "give me one element, I expect only one element to be selected", btw, I just copy+pasted `getSelectedMessageRecord()` from `ConversationFragment.java` :D

adbenitez: it makes sense to group all that utilities somehow in one common class or so

cyBerta:
Hey @adbenitez, nice to see your contributions upstream :).  I've added some comments, but had rather only a brief look at the PR  (I'm in vacation mode right now). Besides these, it looks like some refactoring might be something to consider to reduce the overall code duplication (if it doesn't add too much complexity).

src/org/thoughtcrime/securesms/ProfileGalleryFragment.java
adbenitez: good idea! thanks!

adbenitez:
thanks for your review, already solved some of your points, and have nice vacations ;)

src/org/thoughtcrime/securesms/MessageSelectorFragment.java
```diff
@@ -0,0 +1,79 @@
...
+  protected void handleDeleteMessages(final Set<DcMsg> messageRecords) {
+    int messagesCount = messageRecords.size();
+
+    new AlertDialog.Builder(getActivity())
+            .setMessage(getActivity().getResources().getQuantityString(R.plurals.ask_delete_messages, messagesCount, messagesCount))
+            .setCancelable(true)
+            .setPositiveButton(R.string.delete, (dialog, which) -> {
+                int[] ids = DcMsg.msgSetToIds(messageRecords);
+                dcContext.deleteMsgs(ids);
+                actionMode.finish();
```
Hocuri, line 50:
```suggestion
                if (actionMode != null) actionMode.finish();
```
On my phone, this crashed without this line

src/org/thoughtcrime/securesms/MessageSelectorFragment.java
```diff
@@ -0,0 +1,79 @@
...
+        Permissions.with(getActivity())
+                .request(Manifest.permission.WRITE_EXTERNAL_STORAGE, Manifest.permission.READ_EXTERNAL_STORAGE)
+                .ifNecessary()
+                .withPermanentDenialDialog(getString(R.string.perm_explain_access_to_storage_denied))
+                .onAllGranted(() -> {
+                    SaveAttachmentTask saveTask = new SaveAttachmentTask(getContext());
+                    SaveAttachmentTask.Attachment attachment = new SaveAttachmentTask.Attachment(
+                            Uri.fromFile(message.getFileAsFile()), message.getFilemime(), message.getDateReceived(), message.getFilename());
+                    saveTask.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR, attachment);
+                    actionMode.finish();
```
Hocuri, line 67:
```suggestion
                    if (actionMode != null) actionMode.finish();
```
Same here

src/org/thoughtcrime/securesms/ProfileDocumentsFragment.java
Hocuri: This method should probably go into MessageSelectorFragment, then?

src/org/thoughtcrime/securesms/ProfileDocumentsFragment.java
```diff
@@ -172,39 +165,25 @@ public void onMediaLongClicked(DcMsg mediaRecord) {
...
-    builder.setNegativeButton(android.R.string.cancel, null);
-    builder.show();
+  private DcMsg getSelectedMessageRecord() {
+    Set<DcMsg> messageRecords = getListAdapter().getSelectedMedia();
+
+    if (messageRecords.size() == 1) return messageRecords.iterator().next();
+    else                            throw new AssertionError();
+  }
+
+  private void setCorrectMenuVisibility(Menu menu) {
```
Hocuri, line 175: This could go into MessageSelectorFragment too I guess, and then either be overridden in ConversationFragment? It's probably not possible to use the setCorrectMenuVisibility() from ConversationFragment for all three classes?

src/org/thoughtcrime/securesms/ProfileGalleryFragment.java
```diff
@@ -248,10 +228,20 @@ public boolean onPrepareActionMode(ActionMode mode, Menu menu) {
     @Override
     public boolean onActionItemClicked(ActionMode mode, MenuItem menuItem) {
```
Hocuri, line 229: Just wondering if this can be put into MessageSelectorFragment too, but you probably can judge this better than me, maybe it doesn't make sense

Hocuri (approved):
Thanks a huge lot for pushing this forward:heart:! I commented at some places, mainly code duplication and one NullPointerException I encountered, but then I think that can be merged

src/org/thoughtcrime/securesms/ProfileDocumentsFragment.java
adbenitez: yes, it could be added as an abstract method, didn't added it because it can't be reused by the three

src/org/thoughtcrime/securesms/ProfileDocumentsFragment.java
adbenitez: @Hocuri to move `getSelectedMessageRecord()` to MessageSelectorFragment would then need to refactor the three adapters under a common class/interface that has `getSelectedMessage()`, that is if I add it with implementation, I can add it as abstract but that will not help in the code duplication

adbenitez: solved

adbenitez:
fixed all recommendations, only point left to clarify is https://github.com/deltachat/deltachat-android/pull/1868#discussion_r613535926 

@Hocuri you might be able to reply my question?

src/org/thoughtcrime/securesms/ProfileGalleryFragment.java
Hocuri: The deletion does not happen async in the core. For every deleted message, there is one read access (loading the message) and two write accesses (trashing the message locally and saving a job to delete it on the server) to the SQL db.

I just deleted 170 messages, and it went fine; according to the log, it took 1.04s.

On the other hand - I'll just make a commit that copy-pastes uses the async approach, just because it's done so easily.

Hocuri: Actually actually, let's leave things as they are I'd say. The async approach is waaaaay slower because it calls deleteMsgs() for each message on its own, instead of all at once. And no one complained until now (we always used the synchronous approach in the chat activity), so let's focus on things people complain about :)

Hocuri (approved):

adbenitez:
so can this be merged? btw, for some reason the WIP bot is preventing this to be merged, but there is no WIP in name

Hocuri:
Yup, can be merged. I just edited the title back and forth, WIP problem is gone now

PR #1877, opened 2021-04-17
allow to add POI with text of any length

### LINKED ISSUE #1821: POI text entry field should not be limited
Currently only short texts can be entered, even shorter than the text field itself. Core and maps display are capable of transmitting and displaying long texts, so the limit can be simply removed.

adbenitez:
* unlimited message length
* text editor can be up to 5 lines tall (same as the chat's input text editor)
* POI text display is truncated to 16 characters (previous text editor limit was 16 characters)

close #1821


adbenitez:
## Screenshot:

Screenshot_20210417-011116_DeltaLab

Hocuri (approved):
Didn't test, but code lgtm

PR #1879, opened 2021-04-17
make repo's local issue template work

### LINKED ISSUE #1878: ISSUE_TEMPLATE.md is ignored
instead templates from https://github.com/deltachat/.github/blob/master/.github/ISSUE_TEMPLATE/ are used

if this is intended then the local template should be deleted?

would be nice to have a dedicated template for android, I tried to move the template to a ISSUE_TEMPLATE folder and see if it overwrites the global template, but it also hides the other links to community etc so also config.yml needs to be copied so it defeats the point of having the templates at the organization level. Still would be good to have the android-specific template, 

opinions?

adbenitez:
close #1878

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,3 +1,11 @@
+---
+name: Bug report
+about: Report something that isn't working.
+title: ''
+assignees: ''
+
+---
+
 <!--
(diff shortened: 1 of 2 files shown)


r10s (approved):
it is "Settings menu -> Advanced -> View log", but apart from that, this lgtm :)

the similar thing is also already working on ios, so, things seem to fit nicely now :)

adbenitez:
@r10s done

r10s:
greet! thanks for taking care!

Issue #1797, opened 2021-02-09
(Preview) Images of photos and videos distorted after toggling portrait->landscape->portrait view


gerryfrancis:
- Operating System (Linux/Mac/Windows/iOS/Android):
Android 10 on Fairphone FP2 (Lineage OS 17.1 build QQ3A.200805.001 from 2021-02-05).

- Delta Chat Version:
1.15.0-beta.

- Expected behavior:
Images of photos and videos are shown with correct ratio after toggling from portrait to landscape view and back again.

- Actual behavior:
Images of photos and videos are distorted after toggling from portrait to landscape view and back again.

- Steps to reproduce the problem:
-- Enlarge/Play a photo/video while in portrait view.
-- Rotate the view from portrait to landscape.
-- Select "Show in chat" from the three-dot menu (on right top).
-- Rotate back to portrait view. (Result: The image of the selected photo/video is distorted.)

- Screenshots:
-- Expected:
![grafik](https://user-images.githubusercontent.com/51095533/107416307-9f3c9e00-6b14-11eb-809c-2194b8ee690a.png)
-- Actual:
![grafik](https://user-images.githubusercontent.com/51095533/107416346-a95e9c80-6b14-11eb-9a28-cf50aac045d4.png)

- Logs:
N/A.

gerryfrancis:
Yeah, this is solved! Thank you so much! :+1:

PR #1883, opened 2021-04-18
fix sticker scaling

adbenitez:
avoid cropping stickers, scale them as appropriate


adbenitez:
Screenshot:
Screenshot

Hocuri (approved):
Nice to see how the bugs are squashed :tada:

Could you test whether a bug similar to https://github.com/deltachat/deltachat-android/issues/1797 occurs? To me, the code looks like it will probably not occur, though. (I can't test this as I don't have any non-square stickers)

adbenitez:
@Hocuri since `applySizing()` was removed the issue was solved also for this case, I tested it just in case and it works fine :)

PR #1884, opened 2021-04-18
speedup global search

r10s:
this pr aims to do the global search improvements from https://github.com/deltachat/deltachat-ios/pull/1140 on android.

the speedup is done by the following changes:

- if the query-string is updated while actually performing a search,
  the new search is not directly scheduled as before;
  instead the flag needsAnotherBgSearch is set;
  in the search-thread, the flag is checked
  and another search will be performed

  this way, searches as skipped, but the most recent is always executed

- if there are already chat- or contact-results and
  the query-search has changed meanwhile, search is aborted
  and the partial result is used (next search will be executed immediately)

  this way, esp. the expensive searchMsgs() search is executed less often
  on fast typing

- for a one-letter-search, search for messages is always skipped

moreover, this pr cleans up some things,
putting logic to the same java class.

the search code is still a bit weird, however, larger refactorings are out of scope of this pr :)

### DIFF
--- a/src/org/thoughtcrime/securesms/search/SearchFragment.java
+++ b/src/org/thoughtcrime/securesms/search/SearchFragment.java
@@ -66,9 +66,7 @@ public class SearchFragment extends Fragment implements SearchListAdapter.EventL
 
     this.locale = (Locale) getArguments().getSerializable(EXTRA_LOCALE);
 
-    SearchRepository searchRepository = new SearchRepository(getContext(),
-                                                             Executors.newSingleThreadExecutor());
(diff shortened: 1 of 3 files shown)


adbenitez (approved):
WOW! this is huge! searching for "hello" in my account's db was taking like ~5 seconds, now it is taking ~700 millisecond (I checked the exact number in the log ;)

r10s:
thanks for checking!

and with https://github.com/deltachat/deltachat-core-rust/pull/2365 & Co. things will even get faster :)

PR #1885, opened 2021-04-18
fix clicks on system messages

### LINKED ISSUE #1882: can't accept contact request if it is system message
if you open the contact request view, you will be unable to accept all contact requests that are system messages, for example adding you to a group.

adbenitez:
close #1882 

:cloud_with_rain: rain of fixes related to system messages: 
* clicks on system messages in the contact requests chat now work, displaying the dialog to accept the contact request (#1882)
* clicks on system messages that failed (`msg.isFailed() == true`) now display a dialog with the error message just like normal message bubbles do
* additionally, now system messages that have `Linkify`-ed text can be selected with long tap, this didn't worked in the past and users had to long-press empty areas close to the message record.

### DIFF
--- a/src/org/thoughtcrime/securesms/BaseConversationItem.java
+++ b/src/org/thoughtcrime/securesms/BaseConversationItem.java
@@ -42,9 +42,9 @@ public abstract class BaseConversationItem extends LinearLayout
     this.dcContext = DcHelper.getContext(context);
   }
 
-  public void bind(@NonNull DcMsg            messageRecord,
-                   @NonNull DcChat           dcChat,
-                   @NonNull Set<DcMsg>       batchSelected)
+  protected void bind(@NonNull DcMsg            messageRecord,
+                      @NonNull DcChat           dcChat,
+                      @NonNull Set<DcMsg>       batchSelected)
   {
     this.messageRecord  = messageRecord;
     this.dcChat         = dcChat;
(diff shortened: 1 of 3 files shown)


Hocuri (approved):
:tada:

code lgtm (didn't test)

src/org/thoughtcrime/securesms/BaseConversationItem.java
```diff
@@ -0,0 +1,125 @@
...
+import com.b44t.messenger.DcContext;
+import com.b44t.messenger.DcMsg;
+
+import org.thoughtcrime.securesms.connect.ApplicationDcContext;
+import org.thoughtcrime.securesms.connect.DcHelper;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public abstract class BaseConversationItem extends LinearLayout
```
r10s, line 25: is that BaseConversationItem from the Signal upstream? not saying that it needs to be, just curious :) if not, maybe worth checking what they are doing, the must have similar problems.

adbenitez: @r10s no, I could have just added the needed stuff to ConversationUpdateItem, but then things would be duplicated so I moved things to this base class, stuff from Signal usually have copyright and author notices everywhere :D maybe I should have add something like that so it is clear who to blame :D

adbenitez: wrt Signal, just checked, while trying to remove someone from a group while offline, they just don't add the system message until the person is actually removed, and they don't have a contact request "catch all" chat like us

adbenitez:
do you have any request or can this be merged?
(I still don't feel confident merging :sweat_smile: )

r10s (commented):
yeah, great! also tested that now a bit in practise!

PR #1886, opened 2021-04-19
Improve profile menus

adbenitez:
with #1868 new items were added to `profile_context.xml` menu, which are used by the gallery and documents tabs, it turns out that the settings tab also uses this menu, so unneeded items should be hidden.

Also, in the Gallery menu, would be nice to show the share button directly without having to open the image to share

### DIFF
--- a/src/org/thoughtcrime/securesms/ProfileDocumentsFragment.java
+++ b/src/org/thoughtcrime/securesms/ProfileDocumentsFragment.java
@@ -177,6 +177,7 @@ public class ProfileDocumentsFragment
     menu.findItem(R.id.details).setVisible(singleSelection);
     menu.findItem(R.id.show_in_chat).setVisible(singleSelection);
     menu.findItem(R.id.save).setVisible(singleSelection);
+    menu.findItem(R.id.share).setVisible(false);
   }
 
   private ProfileDocumentsAdapter getListAdapter() {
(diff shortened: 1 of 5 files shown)


src/org/thoughtcrime/securesms/ProfileDocumentsFragment.java
```diff
@@ -177,6 +177,7 @@ protected void setCorrectMenuVisibility(Menu menu) {
     menu.findItem(R.id.details).setVisible(singleSelection);
     menu.findItem(R.id.show_in_chat).setVisible(singleSelection);
     menu.findItem(R.id.save).setVisible(singleSelection);
+    menu.findItem(R.id.share).setVisible(false);
```
r10s, line 180: documents can never be shared - is that intended?

r10s (approved):
lgtm, apart from the open question.

src/org/thoughtcrime/securesms/ProfileDocumentsFragment.java
adbenitez: @r10s ups! for some reason I had the wrong memory that it was not possible to share a document when selecting it in the conversation, I just checked and indeed it is possible! fixing now

r10s: thanks for re-checking!

r10s (approved):

PR #1891, opened 2021-04-21
disable "reply privately" in contact requests chat

### LINKED ISSUE #1889: "reply privately" in contact requests at least partially superfluous
this is about the contact request chat that is available at "menu / contact requests":

when long-tapping a contact request message, the option "reply privately" is offered in the context menu.

the option is also offered for messages that will starting a one-to-one-chat, where it does not make sense at all, so i assume, that it was just forgotten to be disabled.

but also for contact request messages that will starting a group-chat, the option is of limited use, and also ux-wise a bit hard to explain, why some messages in the contact request have that option and other do not - so i would just disable "reply privately" in contact requests unconditionally.

adbenitez:
close #1889


r10s (approved):

r10s:
that was a fast fix!

adbenitez:
the hard part was to wait for the apk to compile and test it works :D

PR #1893, opened 2021-04-23
improve quote style when replying with a sticker

adbenitez:
replying  a message with a sticker is not possible with official DC clients, so this is not super important, nor does it tries to be the definitive final way to do things, but just to make stickers with quotes luck nice for people with older DC versions in the future when the app actually allows to reply with stickers. Currently the quote is shown without background, with this, it will look like:

deltachat-2021-04-23-180239


Hocuri (approved):
LGTM, however, not sure if we should merge that in that short before the release, maybe it's better to wait until afterwards

adbenitez:
the point was to have it in before the apk with support for viewing stickers is released :)

gerryfrancis:
I would apply this background to the timestamp as well to maintain readability.

adbenitez:
> I would apply this background to the timestamp as well to maintain readability.

agree, but that is also true for videochat invitations (the issue was there before stickers)

adbenitez:
@gerryfrancis I improved the footer also:

deltachat-2021-04-26-200655

adbenitez:
solved the merge conflicts, if everyone is happy with how this look, would be nice to merge this to avoid more conflicts, the changes are simple just some small UI tweaks and I am using them for some days and everything is fine.

Hocuri (approved):

PR #1897, opened 2021-04-26
tune down "Show full message" button

### LINKED ISSUE #1895: tune down "Show full message" button
currently, the "Show full message" is pretty bold.

as the button is shown for many messages, and does not always offer really additional content (and we cannot really say if that is the case or not), it would better to tune the button down a bit, let them appear as a usual text link.

i suggest, to do roughly the same as we're doing on ios:

so:

- use roughly the same font size and style as for the normal text
- add roughly one line height of padding above button
- color the text with `delta_accent` color, that should be nice enough on light- and dark-mode

this was discussed with some ppl on the last meeting, and the was some consent on that.

as a side effect, when we start using dedicated buttons "Block" and "Start chat" in the contact requests as already discussed here and there, these buttons cannot be mixed with the "Show full message" button (we can re-use the existing design for "Block" and "Start chat")

adbenitez:
close #1895

I am not good with layout so any feedback/request is welcome


adbenitez:
Screenshot:
deltachat-2021-04-26-181059

adbenitez:
mmm... I think font size is smaller than normal text...

adbenitez:
should the button font size adapt to message font size? (adjustable in settings)

r10s (approved):
lgtm :)

nitpicking: i would maybe remove the padding left of the button, so that it has the same horizontal starting position as the text above.

r10s:
> should the button font size adapt to message font size? (adjustable in settings)

either that or use the same style+size as the name/forwarded title (this is how it looks like currently)

but it is already much better than before, so, at the end, i do not have a strong opinion on that currently :)

adbenitez:
I think the forwarded size is about the same as the current size of the button, and the button was always the same size in the past

adbenitez:
I fixed the horizontal margin

r10s:
i tweaked fontsize and padding a bit (that was easy, now i know the correct lines to adapt :)

r10s (approved):

PR #1899, opened 2021-04-28
prepare 1.19.2 preview release

r10s:
a device message is still missing.


r10s:
just another preview release, mainly to do an gplay-signed without sqlx, merging that in a final release with an updated core will follow soon™ ;)

PR #1903, opened 2021-04-29
don't collapse search menu on group changes

### LINKED ISSUE #1902: Search is canceled when group state changes
if a member is added/removed the search is canceled and the input panel remains hidden

- Delta Chat version: current master
- Steps to reproduce the problem: select the "search in chat" option in a group, with another member account, remove/add someone, search is canceled, input panel is missing.

adbenitez:
close #1902

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -1538,7 +1538,6 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
       dcChat = dcContext.getChat(chatId);
       titleView.setTitle(glideRequests, dcChat);
       initializeSecurity(isSecureText, isDefaultSms);
-      invalidateOptionsMenu();
     }
   }
 
@@ -1579,11 +1578,11 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
   }
 
   private void searchCollapse(final Menu menu, final MenuItem searchItem) {
+    searchMenu = null;
     composePanel.setVisibility(beforeSearchComposeVisibility);
     attachmentManager.setVisibility(beforeSearchAttachVisibility);
 
     ConversationActivity.this.makeSearchMenuVisible(menu, searchItem, false);
-    invalidateOptionsMenu();
   }
 
   private void handleMenuSearchNext(boolean searchNext) {
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -404,7 +404,6 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     case GROUP_EDIT:
       dcChat = dcContext.getChat(chatId);
       titleView.setTitle(glideRequests, dcChat);
-      supportInvalidateOptionsMenu();
       break;
     case TAKE_PHOTO:
       if (attachmentManager.getImageCaptureUri() != null) {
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1538,7 +1538,9 @@ public void handleEvent(DcEvent event) {
       dcChat = dcContext.getChat(chatId);
       titleView.setTitle(glideRequests, dcChat);
       initializeSecurity(isSecureText, isDefaultSms);
-      invalidateOptionsMenu();
+      if (searchMenu == null) {
+        invalidateOptionsMenu();
```
link2xt, line 1542: Couldn't this check be moved into `onPrepareOptionsMenu`, so every time `invalidateOptionsMenu` is called, it actually updates the menu to the correct state? Then it will not be necessary to avoid calling `invalidateOptionsMenu` on the caller side.

This check looks rather like a workaround for incorrect behavior of `onPrepareOptionsMenu`.

adbenitez: in private conversations we found out that `onPrepareOptionsMenu` is not the reason the menu is collapsed

adbenitez: maybe @Hocuri can tell

Hocuri: `onPrepareOptionsMenu()` is called everytime the user opens the options menu, anyway. What `invalidateOptionsMenu()` does, is trigger a new call to `onCreateOptionsMenu()`. Source: https://developer.android.com/reference/android/app/Activity.html#invalidateOptionsMenu()

Just wondering why we need to call `invalidateOptionsMenu()` at all, if we do everything in `onPrepareOptionsMenu()` anyway, so that the menu is updated everytime the user opens it.

Hocuri: Apart, I wonder why we call `invalidateOptionsMenu()` in `handleEvent()` specifically, as the menu can't change at `DC_EVENT_CHAT_MODIFIED`, `DC_EVENT_CHAT_EPHEMERAL_TIMER_MODIFIED` or `DC_EVENT_CONTACTS_CHANGED`, except if I missed something.

The only reason that the options menu can change inside a chat I can think of is, the chat is muted ("Mute notifications"->"Unmute") and "Unarchive"->"Archive". And `invalidateOptionsMenu()` isn't even called at either of these, which is no problem, because we update the complete menu everytime it's opened, anyway.

All in all, I think you could just remove all `invalidateOptionsMenu()` calls from `ConversationActivity`.

link2xt: One problem is if you are removed from the group, "Leave group" menu should be changed.

Hocuri: > I think it just needs to be tested after the change, that when you are removed from the group and open menu immediately after that, there is no "Leave group" option.

Actually, currently, the "Leave group" option never disappears, but that's another issue. Also, `can_send()` should probably return false if you are not in the group, so that the input panel disappears.

adbenitez: my first impression was also that all this was unnecessary

Hocuri (approved):
I just pushed a commit that also removes supportInvalidateOptionsMenu() calls

PR #1904, opened 2021-04-29
improve app accesibility

### LINKED ISSUE #1901: Poor screenreader support
there are buttons unlabeled, for example:
- send button
- global search button
- go back to chat list button inside chat
- when you press add attachment, the button to go back keeps the label of "add attachment" button

Also would be nice to add labels to the message status button (sending, sent, read, error)

Voice message recording button also lacks hints to tell the user to swipe up to lock and swipe left to cancel

adbenitez:
close #1901 

- [x] add label to "global search", "go back to chatlist", "cancel contact search", and a lot more of other unlabeled buttons
- [x] add labels to emojis in emoji selector
- [x] fix: when you press add attachment, the button to go back keeps the label of "add attachment" button
- [x] fix send button label
- [x] add labels to quick camera buttons
- [x] add labels to staged image and image editor
- [x] add labels for message status (sending, sent, read, failed)

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -121,6 +121,9 @@
     <string name="contact">Contact</string>
     <string name="verified_contact">Verified contact</string>
     <string name="camera">Camera</string>
+    <string name="camera_capture_description">Capture</string>
+    <string name="camera_change_description">Change camera</string>
+    <string name="camera_fullscreen_toogle_description">Alternate full screen mode</string>
     <string name="location">Location</string>
     <string name="gallery">Gallery</string>
     <string name="images_and_videos">Images and videos</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -703,6 +703,13 @@
 
     <!-- ImageEditorHud -->
     <string name="ImageEditorHud_draw_anywhere_to_blur">Draw anywhere to blur</string>
+    <string name="ImageEditorHud_add_text">Add text</string>
+    <string name="ImageEditorHud_blur">Blur</string>
+    <string name="ImageEditorHud_brush_marker">Marker brush</string>
+    <string name="ImageEditorHud_brush_highlight">Highlight brush</string>
+    <string name="ImageEditorHud_crop">Crop</string>
+    <string name="ImageEditorHud_flip">Flip</string>
+    <string name="ImageEditorHud_rotate">Rotate</string>
 
     <!-- dc_str_* resources -->
     <string name="encrypted_message">Encrypted message</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -122,7 +122,7 @@
     <string name="verified_contact">Verified contact</string>
     <string name="camera">Camera</string>
     <string name="camera_capture_description">Capture</string>
-    <string name="camera_change_description">Change camera</string>
+    <string name="camera_change_description">Switch camera</string>
     <string name="camera_fullscreen_toogle_description">Toggle full screen mode</string>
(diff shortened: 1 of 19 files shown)


adbenitez:
wow, it is nice that now blind people can know if message was sent, failed or was read by the receiver, and this is even accessible from the chat list

res/values/strings.xml
```diff
@@ -120,6 +121,9 @@
     <string name="contact">Contact</string>
     <string name="verified_contact">Verified contact</string>
     <string name="camera">Camera</string>
+    <string name="camera_capture_description">Capture</string>
+    <string name="camera_change_description">Change camera</string>
+    <string name="camera_fullscreen_toogle_description">Alternate full screen mode</string>
```
Hocuri, line 126:
```suggestion
    <string name="camera_fullscreen_toogle_description">Toggle full screen mode</string>
```

res/values/strings.xml
```diff
@@ -120,6 +121,9 @@
     <string name="contact">Contact</string>
     <string name="verified_contact">Verified contact</string>
     <string name="camera">Camera</string>
+    <string name="camera_capture_description">Capture</string>
+    <string name="camera_change_description">Change camera</string>
```
Hocuri, line 125: Maybe "switch" instead of "change"? Sounds nicer to me

res/values/strings.xml
```diff
@@ -699,6 +703,13 @@
 
     <!-- ImageEditorHud -->
     <string name="ImageEditorHud_draw_anywhere_to_blur">Draw anywhere to blur</string>
+    <string name="ImageEditorHud_add_text">Add text</string>
+    <string name="ImageEditorHud_blur">Blur</string>
+    <string name="ImageEditorHud_brush_marker">Marker brush</string>
+    <string name="ImageEditorHud_brush_highlight">Highlight brush</string>
+    <string name="ImageEditorHud_crop">Crop</string>
+    <string name="ImageEditorHud_flip">Flip</string>
+    <string name="ImageEditorHud_rotate">Rotate</string>
```
Hocuri, line 712: BTW is it possible for blind people to use the image editor? If not, do we need these?

res/values/strings.xml
adbenitez: I just take that from Signal, but agree that switch sounds better +1

res/values/strings.xml
adbenitez: there are people that are not blind but can't see well, they might use the image editor and not see the details of the icons or wonder what the icons mean, for example the markers icons are almost identical, etc.

res/values/strings.xml
Hocuri, line 125:
```suggestion
    <string name="camera_change_description">Switch camera</string>
```

Hocuri (approved):

res/values/strings.xml
Hocuri: Makes sense

res/values/strings.xml
r10s: and can we just name the keys `capture` and `switch_camera`? although there are strings following other approaches, this is what we are doing usually as default in the last time, only if that turns out not to be sufficient, we add more information to the key.

adbenitez:
@r10s I applied your recommendations, something else? otherwise feel free to merge

r10s:
@adbenitez thanks for the adaptions! i also copied the spanish strings to transifex so that they do not get overwritten on the next pull.

Gh0stz0x:
Thanks @adbenitez, you did a great job!

PR #1913, opened 2021-05-03
highlight system messages in search results

### LINKED ISSUE #1912: system messages not highlighted in search
- Steps to reproduce the problem: search for some text that match also system messages, the system messages are included in results but when you use the arrows to navigate through the results, normal messages are highlighted but system messages aren't and you don't have visual feedback.

adbenitez:
close #1912

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationUpdateItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationUpdateItem.java
@@ -55,9 +55,8 @@ public class ConversationUpdateItem extends BaseConversationItem
                    @NonNull Recipient               conversationRecipient,
                             boolean                 pulseUpdate)
   {
-    bind(messageRecord, dcChat, batchSelected);
+    bind(messageRecord, dcChat, batchSelected, pulseUpdate);
     setGenericInfoRecord(messageRecord);
-    setSelected(batchSelected.contains(messageRecord));
   }
 
   private void initializeAttributes() {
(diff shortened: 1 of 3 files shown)


src/org/thoughtcrime/securesms/ConversationUpdateItem.java
```diff
@@ -55,9 +55,8 @@ public void bind(@NonNull DcMsg                   messageRecord,
                    @NonNull Recipient               conversationRecipient,
                             boolean                 pulseUpdate)
   {
-    bind(messageRecord, dcChat, batchSelected);
+    bind(messageRecord, dcChat, batchSelected, pulseUpdate);
     setGenericInfoRecord(messageRecord);
-    setSelected(batchSelected.contains(messageRecord));
```
Hocuri, line 60: What does this change do?

Hocuri (approved):
Thanks!

Tested and looked through code

src/org/thoughtcrime/securesms/ConversationUpdateItem.java
adbenitez: that is what adds support for the "pulse" effect in both normal messages and system messages, previously it was implemented only for normal messages, I moved that to the parent base class

Hocuri: Do you know why the `setSelected(batchSelected.contains(messageRecord));` was there before?

adbenitez: @Hocuri yes, it was to mark message as selected when you are in "selection mode" (to put the light blue background) that logic is inside `BaseConversationItem.setInteractionState()` which takes care of the interaction state (pulse effect and selected row)

PR #1914, opened 2021-05-03
handle URLs starting with OPENPGP4FPR and other case variants

adbenitez:
for some reason, URLs starting with "OPENPGP4FPR:" are not handled if the custom schema is not specified also in upper case.

in DeltaLab there is markdown support so extracting QR data and creating a nice invitation link is a possibility, also bots might offer this kind of link in html messages taking the QR data provided by core which is in upper case, and it would be nice if it works in upper case

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -122,6 +122,7 @@
             <category android:name="android.intent.category.DEFAULT"/>
             <category android:name="android.intent.category.BROWSABLE"/>
             <data android:scheme="openpgp4fpr" />
+            <data android:scheme="OPENPGP4FPR" />
         </intent-filter>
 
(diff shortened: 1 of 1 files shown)


adbenitez:
actually it turns out this is also the same behavior for urls like `HTTP://...` using upper case seems to be wrong

adbenitez:
it turns out that in the wild this URIs may not be always in lowercase, apps like **Open Keychain** handle several variations:

https://github.com/open-keychain/open-keychain/blob/fdd7c00a80e35b4b252f5f814341271385c2bf87/OpenKeychain/src/main/AndroidManifest.xml#L739

r10s (approved):
weird, but somehow reasonable :)

thanks for taking care and for the link to openkeychain.

Issue #1851, opened 2021-03-30
location streaming permission issues


cyBerta:
as a result of https://github.com/deltachat/deltachat-android/issues/1790

- [x] Add link to privacy policy in FAQ
- [x] add alert before location permission is not yet granted and needs to be requested

r10s:
this is a blocker as without we cannot go to gplay.

r10s:
> Add link to privacy policy in FAQ

@cyBerta not totally sure what this link/FAQ means here, what is missing is a statement in the policy imu

r10s:
fixes are in https://github.com/deltachat/deltachat-android/pull/1919 and https://github.com/deltachat/deltachat-pages/pull/441

PR #1919, opened 2021-05-06
background location disclosure

r10s:
part two of https://github.com/deltachat/deltachat-android/issues/1851

### DIFF
--- a/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
+++ b/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
@@ -424,8 +424,11 @@ public class AttachmentManager {
       return;
     }
 
+    // see https://support.google.com/googleplay/android-developer/answer/9799150#zippy=%2Cstep-provide-prominent-in-app-disclosure
+    // for rationale dialog requirements
     Permissions.PermissionsBuilder permissionsBuilder = Permissions.with(activity)
             .ifNecessary()
+            .withRationaleDialog("This app will collect location data to enable sharing your locations with chat members for a given time frame.\n\nLocation data is collected even when the app is closed or not in use.", R.drawable.ic_location_on_white_24dp)
             .withPermanentDenialDialog(activity.getString(R.string.perm_explain_access_to_location_denied))
             .onAllGranted(() -> {
               ShareLocationDialog.show(activity, durationInSeconds -> {
--- a/res/layout/permissions_rationale_dialog.xml
+++ b/res/layout/permissions_rationale_dialog.xml
@@ -19,8 +19,8 @@
               android:layout_width="match_parent"
               android:layout_height="wrap_content"
               android:padding="20dp"
-              android:textSize="15sp"
-              android:lineSpacingMultiplier="1.3"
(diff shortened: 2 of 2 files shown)


src/org/thoughtcrime/securesms/mms/AttachmentManager.java
```diff
@@ -424,8 +424,11 @@ public static void selectLocation(Activity activity, int chatId) {
       return;
     }
 
+    // see https://support.google.com/googleplay/android-developer/answer/9799150#zippy=%2Cstep-provide-prominent-in-app-disclosure
+    // for rationale dialog requirements
     Permissions.PermissionsBuilder permissionsBuilder = Permissions.with(activity)
             .ifNecessary()
+            .withRationaleDialog("This app will collect location data to enable sharing your locations with chat members for a given time frame.\n\nLocation data is collected even when the app is closed or not in use.", R.drawable.ic_location_on_white_24dp)
```
Hocuri, line 431: I don't know if this is the time to discuss the wording, but I'll just go ahead, if not just merge it without changing it:
- I'd drop out "for a given time frame" as it will be clear to users that the location streaming can be stopped again
- I'd use "location streaming" or "live location"
- Why "collect"?
- At https://support.google.com/googleplay/android-developer/answer/9799150?hl=en#zippy=%2Cstep-provide-prominent-in-app-disclosure%2Csee-an-example-of-prominent-in-app-disclosure it says we shall explain _why_ we want location access while DC is in the background.

All in all, this would give sth like:

```
To share your live location with your friends, allow Delta Chat to access your location.

If you want to continue location streaming while the app is in the background, allow access all of the time.
```

Hocuri (approved):

src/org/thoughtcrime/securesms/mms/AttachmentManager.java
r10s: the fist part is better, i will change that. however, "friends" is a bit unspecific, we know it more concrete at that moment, things are shared with the "chat members", so maybe better stay with that.

the second part, "If you want to continue location streaming while the app is in the background, ..." does not work if the user does not have a choice.

r10s: @Hocuri maybe that one:

```
To share your live location with chat members,
allow Delta Chat to use your location data.

Background location is used
to make live location work gaplessly
even when the app is closed or not in use.
```

Hocuri: > does not work if the user does not have a choice.

I thought they do have a choice? They can choose "Allow all the time", "Allow only while using the app" and "Deny" when the system asks.

Anyway, your version is nice too.

Hocuri (approved):

r10s:
> I thought they do have a choice?

on most systems out there, there is no choice. the choice is added in newer androids, i added two screenshots to the initial post above.

Hocuri:
Ah, ok. Then let's use your wording.

r10s:
k, this is what it looks like now:

PR #1920, opened 2021-05-06
Dockerfile: switch to Debian and use free SDK

link2xt:
NDK is still built by Google.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,11 @@
-FROM docker.io/ubuntu:18.04
+FROM docker.io/debian:bullseye
 
-# Install Android Studio requirements
-# See https://developer.android.com/studio/install#linux
-RUN apt-get update -y \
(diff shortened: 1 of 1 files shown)


link2xt:
Got an error during `gradlew assembleDebug`:
```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'app'.
> Failed to install the following Android SDK packages as some licences have not been accepted.
     build-tools;28.0.3 Android SDK Build-Tools 28.0.3
     platforms;android-29 Android SDK Platform 29
  To build this project, accept the SDK license agreements and install the missing components using the Android Studio SDK Manager.
  Alternatively, to transfer the license agreements from one workstation to another, see http://d.android.com/r/studio-ui/export-licenses.html
  
  Using Android SDK: /usr/lib/android-sdk
```

PR #1926, opened 2021-05-11
Clarify disappearing message hint

Jikstra:
I got the user feedback that it's not clear that this setting only applies for this chat/group and is not a global setting like device/server deletion

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -33,7 +33,7 @@
     <string name="unarchive">Unarchive</string>
     <string name="mute">Mute</string>
     <string name="ephemeral_messages">Disappearing messages</string>
-    <string name="ephemeral_messages_hint">These settings apply to all participants using Delta Chat. However, they may copy, save and forward messages or use other e-mail clients.</string>
+    <string name="ephemeral_messages_hint">These settings apply to all participants in this group using Delta Chat. However, they may copy, save and forward messages or use other e-mail clients.</string>
     <string name="save">Save</string>
     <string name="chat">Chat</string>
     <string name="media">Media</string>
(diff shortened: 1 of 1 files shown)


r10s:
i think, "in this group" is not correct, as the setting is also used for one-to-one chat.

maybe better use "These settings apply to all chat members using Delta Chat."

Jikstra:
@r10s I think combining both is fine. At least on Desktop it won't be bad if the text is a little bit longer. Just think about how long the hint for device/server deletion is :P

`When enabled, messages sent and received in this chat will disappear after they have been seen. However, they may copy, save and forward messages or use other e-mail clients.

r10s:
"they" is ambiguous now, first it refers to messages, second to users.
also, compared to the original, the information about to whom that applies also to others, are gone.

well, barriers are high :) iirc, the current string was also the outcome of a looooooooong discussion with @link2xt and others, so it is not that amazing that it is not easy to improve and to top :)

so, not sure if a complete rewording makes sense. what about the initial idea, but replacing "participants" by "members"? that already makes things a bit better, as we are re-using an existing wording.

nb: is it just one user that has problems? or did we get several complains about the existing string?

link2xt:
Replacing "participants" with "chat members" looks good.

r10s:
some thumbs up to the post of @link2xt  :) - so, wording would be "These settings apply to all chat members using Delta Chat. However, they may copy, save and forward messages or use other e-mail clients." - @Jikstra if you agree on that, can you modify the pr? or ping me, i can do that as well together with some tx actions. i just do not want to capture this pr without consent :)

r10s:
k, i updated the string, also on transifex.

PR #1928, opened 2021-05-14
request composer's focus after recording is done

### LINKED ISSUE #1785: Some defects related to the user interface
- Operating System: Android 8.1
- Delta Chat Version: 1.14.5
- Expected behavior: n/a
- Actual behavior: n/a
a) First make the input method keyboard pop up, and then send the voice message. At this time, the keyboard will not disappear, and the input box has no focus. If you try to use the keyboard in this state, no characters will be entered.
b) The read receipt mark on the right side of the contact list is displayed in white. This problem was found on MIUI 9 based on Android 4.4.2, and Android 8.1 does not have this problem.
- Screenshots: n/a
- Logs: n/a

adbenitez:
close #1785


r10s (approved):
lgtm, thanks a lot for taking care about details!

PR #1929, opened 2021-05-15
In-Chat-Search: Start searching at the newest (=last) message, not the oldest

Hocuri:


r10s:
makes proably, sense, i just had a look, whatsapp/ios also starts at the end.

two things, that should be considered:

- is it okay that the 1st result is shown as eg.  _12_/12? shouldn't we swap that as well? what are whatsapp/signal/telegram doing here?

- whatsapp/ios disables the up/down buttons when at start/end of the list - we should consider doing that as well to make it clearer that you have to tap "up" and you're at the end of the chat; this was less important before as search up-down is more expectable.

adbenitez:
>     * whatsapp/ios disables the up/down buttons when at start/end of the list

at first it makes sense, but that is less useful IMHO, if you have 100 matches, you are at the last and want to jump to the older match, you will need to press the up button 100 times, with current functionality you can cycle through the matches so you can press next to quickly go to the older match, also there is this trick of pressing the arrows without any matches to go to the top of the chat, if we get too smart about it the arrows will be disabled and the "temporal workaround" will be gone and we could get more requests for the "go to beginning" feature

adbenitez (approved):

r10s (approved):
i also considered things while having a walk - having an inversed counter is also not-so-great wrt internal consistency (we have a similar search in the help and in the full-message-view), and i agree that non-cycling would be a feature-removal.

therefore, yip, just merge the pr as it is now.

PR #1932, opened 2021-05-15
set navigation bar color in dark mode

### LINKED ISSUE #1931: navigation bar keeps in light theme when app is in dark mode
- Android version: 10
- Delta Chat version: current master
- Steps to reproduce the problem: enable dark mode or use system settings to set phone in "night mode"

adbenitez:
close #1931


r10s (approved):
lgtm, however i did not test it.

adbenitez:
> lgtm, however i did not test it - @adbenitez did you test on an android10?

Yes I tested it in my phone which has Android 10

PR #1934, opened 2021-05-16
notify quote-mentions in muted groups

adbenitez:
if group is muted, it would still be nice to get notifications when our messages are quoted, I added no extra setting for this as I think in general it would be a desired behavior and you can set notification sound of the chat to "silence" or temporally leave the group if you for don't want any interaction.

Without this feature, in a muted noise chat, you would miss mentions if you get like +100 messages and you don't skim over all of them to see if you got mentioned, etc. with this you will get a notification for the mention and then you can jump to the message directly.

related feature request: https://support.delta.chat/t/get-notified-when-your-message-is-quoted/1241

cc @link2xt

### DIFF
--- a/src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
+++ b/src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
@@ -311,8 +311,11 @@ public class NotificationCenter {
         Util.runOnAnyBackgroundThread(() -> {
 
             DcChat dcChat = dcContext.getChat(chatId);
+            DcMsg dcMsg = dcContext.getMsg(msgId);
+            DcMsg quotedMsg = dcMsg.getQuotedMsg();
+            boolean isGroupMention = quotedMsg != null && dcChat.isGroup() && quotedMsg.isOutgoing();
 
-            if (!Prefs.isNotificationsEnabled(context) || dcChat.isMuted()) {
+            if (!Prefs.isNotificationsEnabled(context) || (!isGroupMention &&  dcChat.isMuted())) {
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


link2xt:
Looks good. Maybe this whole logic of muting chats and disabling notifications should be moved to the core so it doesn't have to implemented it in all UIs. Like, `DC_EVENT_NOTIFY` right after `DC_EVENT_INCOMING_MSG`.

adbenitez:
indeed, having all this notifications logic in core would be good, also for @ **mentions**, but if that is not planned anytime soon would be good to at least have this in the meanwhile, it can be removed together with the other notifications logic later

r10s:
in general, i agree that it may be nice to have some more options wrt notifications.

> if group is muted, it would still be nice to get notifications when our messages are quoted, I added no extra setting for this as I think in general it would be a desired behavior

hm, not sure about that. if i say "shut up" to a group and my phone rings for whatever reason, that i cannot even switch off, i might be annoyed :)

so, a global option is probably needed. at some point, it can be used for mentions as well, as a compromise wrt core/ui, we can maybe let `dc_set_config()` handle such a setting, that way, it would also go to the backup.

link2xt:
In Skype there is an option to enable "Notifications" in a chat. If you enable it, you can then enable "Smart notifications" which reduces notifications to quotes and mentions only.

adbenitez:
hmm... if this turns out to be too controversial as is, feel free to close the PR, we can wait for a better solution in core :D

r10s:
yip, closing this for now and putting it to https://github.com/orgs/deltachat/projects/19

PR #1935, opened 2021-05-21
push untranslated strings from desktop

Simon-Laux:


res/values/strings.xml
```diff
@@ -41,6 +41,10 @@
     <string name="main_menu">Main menu</string>
     <string name="start_chat">Start chat</string>
     <string name="show_full_message">Show full message…</string>
+    <string name="show_full_message_in_browser">Show full message in browser…</string>
+    <string name="loading">Loading</string>
```
r10s, line 45: is that a message as "One moment…"? then maybe also add an ellipsis. in any way, i think, this strings deserves a little translator hint (just a `<!-- comment -->` atop of the string)

res/values/strings.xml
```diff
@@ -215,6 +221,12 @@
     <string name="menu_zoom_in">Zoom in</string>
     <string name="menu_zoom_out">Zoom out</string>
     <string name="menu_save_log">Save log</string>
+    <string name="menu_more_options">More Options</string>
```
r10s, line 224:
```suggestion
    <string name="menu_more_options">More options</string>
```

res/values/strings.xml
```diff
@@ -215,6 +221,12 @@
...
+    <string name="menu_learn_spelling">Learn Spelling</string>
```
r10s, line 225:
```suggestion
    <string name="menu_learn_spelling">Learn spelling</string>
```

res/values/strings.xml
```diff
@@ -215,6 +221,12 @@
...
+    <string name="menu_chat_audit_log">Chat Audit Log</string>
+    <string name="jump_to_message">Jump to Message<string>
```
r10s, line 227:
```suggestion
    <string name="jump_to_message">Jump to message<string>
```

res/values/strings.xml
```diff
@@ -215,6 +221,12 @@
...
+    <string name="copy_json">Copy JSON</string>
+    <string name="replace_draft">replace draft</string>
```
r10s, line 229: should that really start lower case? if so, add a hint for the translators.

res/values/strings.xml
```diff
@@ -339,6 +351,7 @@
     <string name="messaging_disabled_not_in_group">You cannot write because you are not in this group. To join, ask another member.</string>
     <string name="messaging_disabled_device_chat">This chat contains locally generated messages. Therefore you cannot write any messages.</string>
     <string name="messaging_disabled_deaddrop">Click on a message to accept or deny the contact request</string>
+    <string name="messaging_disabled_mailing_list">Sending Messages in Mailinglists is not supported yet<string>
```
r10s, line 354:
```suggestion
    <string name="messaging_disabled_mailing_list">Sending messages in mailing lists is not supported yet<string>
```

res/values/strings.xml
```diff
@@ -347,13 +360,23 @@
     <!-- mailing lists -->
     <string name="mailing_list">Mailing list</string>
     <string name="ask_show_mailing_list">Show mailing list \"%1$s\"?</string>
-
+    <string name="mailing_list_profile_info">These settings (chat name and image) apply to this device only.</string>
```
r10s, line 363: "chat name" is a term that is used nowhere else yet, so maybe not introduce that here.

as already discussed, i do not like the brackets here much. my suggestion would be 

```suggestion
    <string name="mailing_list_profile_info">Mailing list name and image changes apply to this device only.</string>
```

res/values/strings.xml
```diff
@@ -347,13 +360,23 @@
...
 
     <!-- map -->
     <string name="filter_map_on_time">Show locations in time frame</string>
     <string name="show_location_traces">Show traces</string>
     <string name="add_poi">Send point of interest</string>
 
+    <!-- chat audit log -->
+    <string name="chat_audit_log_empty_message">System messages of this chat would appear here</string>
+    <string name="chat_audit_log_title">Audit Log for %1$s</string>
+    <string name="chat_audit_log_description">This view shows only System- and Info-Messages. Useful for finding the last chat actions without scrolling through many normal messages.</string>
```
r10s, line 373: the chat audit log is still experimental, right?

usually, we do not pass strings in that stage to the translators, as they are very subject to change which is annoying for translators.

i suggest to remove these three strings for now until that is more settled, this would also remove the need to discuss these strings here currently.

```suggestion
```

res/values/strings.xml
```diff
@@ -347,13 +360,23 @@
...
+
+    <!-- punycode warning / labeled links -->
+    <string name="open_external_url_trust_domain">Don't ask again for</string>
+    <string name="puny_code_warning_header">Suspicious Link detected</string>
+    <string name="puny_code_warning_question">Are you sure you want to visit $$asciiHostname$$?</string>
```
r10s, line 378: please use the normal placeholders as `%1$s`

res/values/strings.xml
```diff
@@ -347,13 +360,23 @@
...
     <string name="show_location_traces">Show traces</string>
     <string name="add_poi">Send point of interest</string>
 
+    <!-- chat audit log -->
+    <string name="chat_audit_log_empty_message">System messages of this chat would appear here</string>
+    <string name="chat_audit_log_title">Audit Log for %1$s</string>
+    <string name="chat_audit_log_description">This view shows only System- and Info-Messages. Useful for finding the last chat actions without scrolling through many normal messages.</string>
+
+    <!-- punycode warning / labeled links -->
+    <string name="open_external_url_trust_domain">Don't ask again for</string>
```
r10s, line 376: is a placeholder missing here? at the end of the string?

otherwise, that would need some explanations and a translator hint.

res/values/strings.xml
```diff
@@ -347,13 +360,23 @@
...
+    <string name="puny_code_warning_header">Suspicious Link detected</string>
+    <string name="puny_code_warning_question">Are you sure you want to visit $$asciiHostname$$?</string>
+    <string name="puny_code_warning_description">You followed a link which could be misrepresenting characters by using similarly looking ones from different alphabets. Following the link labelled $$originalHostname$$ will lead to $$asciiHostname$$ which is normal for non-Latin characters. If you did not expect such characters this link could be harmful</string>
```
r10s, line 379: a full stop is missing at the end of the string.

also, use the normal placeholders here, `%1$s` and `%2$s` and document the usage in the translator's hint.

res/values/strings.xml
```diff
@@ -347,13 +360,23 @@
...
     <string name="add_poi">Send point of interest</string>
 
+    <!-- chat audit log -->
+    <string name="chat_audit_log_empty_message">System messages of this chat would appear here</string>
+    <string name="chat_audit_log_title">Audit Log for %1$s</string>
+    <string name="chat_audit_log_description">This view shows only System- and Info-Messages. Useful for finding the last chat actions without scrolling through many normal messages.</string>
+
+    <!-- punycode warning / labeled links -->
+    <string name="open_external_url_trust_domain">Don't ask again for</string>
+    <string name="puny_code_warning_header">Suspicious Link detected</string>
```
r10s, line 377:
```suggestion
    <string name="puny_code_warning_header">Suspicious link detected</string>
```

res/values/strings.xml
```diff
@@ -543,7 +566,10 @@
     <string name="profile_image_select">Select profile image</string>
     <string name="select_your_new_profile_image">Select your new profile image</string>
     <string name="profile_image_delete">Delete profile image</string>
-
+    <string name="pref_show_tray_icon">Show tray icon</string>
+    <string name="pref_edit_profile">Edit profile</string>
+    <string name="pref_set_profile_photo">Set profile photo</string>
+    <string name="pref_remove_profile_photo">Remove profile photo</string>
```
r10s, line 572: the "profile photo" is called "profile image" everywhere else; we should keep this term.

so, please use the already existing string `profile_image_delete` that reads "Delete profile image".

```suggestion
```

res/values/strings.xml
```diff
@@ -543,7 +566,10 @@
     <string name="profile_image_select">Select profile image</string>
     <string name="select_your_new_profile_image">Select your new profile image</string>
     <string name="profile_image_delete">Delete profile image</string>
-
+    <string name="pref_show_tray_icon">Show tray icon</string>
+    <string name="pref_edit_profile">Edit profile</string>
+    <string name="pref_set_profile_photo">Set profile photo</string>
```
r10s, line 571: the "profile photo" is called "profile image" everywhere else; we should keep this term.

so, please use the already existing string `profile_image_select` that reads "Select profile image" - i assume this is for the button opening the file selector to select an image? in this case, "Select" also fits much better than "Set".

```suggestion
```

Select profile image

res/values/strings.xml
```diff
@@ -681,7 +707,9 @@
     <string name="contact_setup_changed">Changed setup for %1$s.</string>
     <string name="verified_group_explain">Verified groups (experimental) provide safety against active attacks. Members are verified with a second factor by other members and messages are always end-to-end-encrypted.</string>
     <string name="copy_qr_data_success">Copied QR url to clipboard</string>
-
+    <string name="mailto_dialog_header_select_chat">Select chat to send the message to</string>
+    <string name="mailto_dialog_confirm_replace_draft">%1$s already has a draft message, do you want to replace it?</string>
```
r10s, line 711: please add a translator comment about what `%1$s` will be. (i assume the name of the chat?)

res/values/strings.xml
```diff
@@ -740,6 +768,8 @@
     <string name="global_menu_help_contribute_desktop">Contribute on GitHub</string>
     <string name="global_menu_help_report_desktop">Report an issue</string>
     <string name="global_menu_help_about_desktop">About Delta Chat</string>
+    <string name="global_menu_file_open_desktop">Open DeltaChat</string>
```
r10s, line 771:
```suggestion
    <string name="global_menu_file_open_desktop">Open Delta Chat</string>
```

res/values/strings.xml
```diff
@@ -740,6 +768,8 @@
...
+    <string name="global_menu_minimize_to_tray">Minimize</string>
```
r10s, line 772: if the wording is really just "minimize", i would also name the key as such.

res/values/strings.xml
```diff
@@ -785,6 +815,10 @@
     <string name="menu.view.developer.open.log.folder">Open the log folder</string>
     <string name="menu.view.developer.open.current.log.file">Open current logfile</string>
     <string name="user_location_permission_explanation">Delta Chat needs the location permission in order to show and share your location.</string>
+    <string name="ask_delete_message_desktop">Are you sure you want to delete this message?</string>
```
r10s, line 818: while this string is introduced on desktop, it is not desktop specific any may be used somewhere else as well. therefore, remove the `_desktop` suffix and maybe move it up below `ask_delete_messages` (maybe even that string can be used)

res/values/strings.xml
```diff
@@ -785,6 +815,10 @@
...
+    <string name="explain_desktop_minimized_disabled_tray_pref">Tray icon cannot be disabled as Delta Chat was started with the --minimized option</string>
```
r10s, line 819:
```suggestion
    <string name="explain_desktop_minimized_disabled_tray_pref">Tray icon cannot be disabled as Delta Chat was started with the --minimized option.</string>
```

res/values/strings.xml
```diff
@@ -785,6 +815,10 @@
...
+    <string name="no_spellcheck_suggestions_found">No Spelling Suggestions found</string>
```
r10s, line 820:
```suggestion
    <string name="no_spellcheck_suggestions_found">No spelling suggestions found.</string>
```

res/values/strings.xml
```diff
@@ -785,6 +815,10 @@
...
+    <string name="mac_show_windows">Show Window</string>
```
r10s, line 821: "Show windows" or "Show window"?

as the string is pretty neutral, i'd also remove the "mac_" prefix.

r10s (changes requested):
a general hint wrt upper and lower case: both variants, eg. "Open File" and "Open file" are reasonable, maybe the one is better than the other, i do not really know.

however, we should not mix the variants, therefore, we keep things lowercase to be consistent with the other strings.

res/values/strings.xml
Simon-Laux, line 45:
```suggestion
    <string name="loading">Loading…</string>
```

res/values/strings.xml
Simon-Laux: I use it for a button that does the all uppercase transformation so I did not care.
```suggestion
    <string name="replace_draft">Replace draft</string>
```

res/values/strings.xml
Simon-Laux: Users might not think of those changeable properties as settings so it could be confusing. That's why I insist on the clarification.

res/values/strings.xml
Simon-Laux: we also translated the voice message strings so I don't see why this is the case.

res/values/strings.xml
Simon-Laux, line 711:
```suggestion
    <!-- first placeholder is the name of the chat -->
    <string name="mailto_dialog_confirm_replace_draft">%1$s already has a draft message, do you want to replace it?</string>
```

res/values/strings.xml
Simon-Laux: considered using `ask_delete_messages` but it says something too different. We'll probably use it when we have multi select messages in desktop.

res/values/strings.xml
Simon-Laux:
```suggestion
    <!-- this is used as placeholder when something is loading -->
    <string name="loading">Loading</string>
```

res/values/strings.xml
Simon-Laux: IDK I think this could also change in the future :shrug: 
CC @Jikstra

res/values/strings.xml
Simon-Laux: lets keep the case here as this is shown in the mac titlebar and the style there is each word uppercase

res/values/strings.xml
r10s: k, then maybe leave that as is.

res/values/strings.xml
```diff
@@ -347,13 +362,26 @@
...
+    <string name="mailing_list_profile_info">Changes on mailing list name and image apply to this device only.</string>
 
     <!-- map -->
     <string name="filter_map_on_time">Show locations in time frame</string>
     <string name="show_location_traces">Show traces</string>
     <string name="add_poi">Send point of interest</string>
 
+    <!-- chat audit log -->
+    <string name="chat_audit_log_empty_message">System messages of this chat would appear here</string>
+    <string name="chat_audit_log_title">Audit Log for %1$s</string>
```
r10s, line 374:
```suggestion
    <string name="chat_audit_log_title">Chat Audit Log for %1$s</string>
```

just "Audit Log for Bjoern" looks scary to me :) also that thing is called "Chat Audit Log" in the menu, so, let's use the same in the title.

res/values/strings.xml
```diff
@@ -347,13 +362,26 @@
...
+    <string name="chat_audit_log_description">This view shows only System- and Info-Messages. Useful for finding the last chat actions without scrolling through many normal messages.</string>
```
r10s, line 375:
```suggestion
    <string name="chat_audit_log_description">This view shows only system and info messages. Useful for finding the last chat actions without scrolling through many normal messages.</string>
```

res/values/strings.xml
```diff
@@ -347,13 +362,26 @@
...
+
+    <!-- punycode warning / labeled links -->
+    <!-- placeholder is domain/hostname that should be trusted -->
+    <string name="open_external_url_trust_domain">Don't ask again for %1$s</string>
+    <string name="puny_code_warning_header">Suspicious link detected</string>
+    <!-- placeholder contains the hostname converted to ascii -->
+    <string name="puny_code_warning_question">Are you sure you want to visit %1$s?</string>
+    <!-- first placeholder is original hostname with special chars, second placeholder is hostname encoded in ascii -->
+    <string name="puny_code_warning_description">You followed a link which could be misrepresenting characters by using similarly looking ones from different alphabets. Following the link labelled %1$s will lead to %2$s which is normal for non-Latin characters. If you did not expect such characters this link could be harmful</string>
```
r10s, line 384:
```suggestion
    <!-- this message is shown whenever a link with non-latin characters is clicked. first placeholder is original hostname with special chars, second placeholder is hostname encoded in ascii -->
    <string name="puny_code_warning_description">You followed a link which could be misrepresenting characters by using similarly looking ones from different alphabets. Following the link labelled %1$s will lead to %2$s which is normal for non-Latin characters. If you did not expect such characters this link could be harmful.</string>
```

r10s (approved):
k, some minor suggestions, but all in all this lgtm now.

res/values/strings.xml
Simon-Laux: I miss the "You followed a link which could be misrepresenting characters by using similarly looking ones from different alphabets.".
It helps the more confusing technical explanation bellow, because it is easier to understand I believe.

res/values/strings.xml
```diff
@@ -347,13 +362,26 @@
...
+    <string name="chat_audit_log_title">Audit Log for %1$s</string>
+    <string name="chat_audit_log_description">This view shows only System- and Info-Messages. Useful for finding the last chat actions without scrolling through many normal messages.</string>
+
+    <!-- punycode warning / labeled links -->
+    <!-- placeholder is domain/hostname that should be trusted -->
+    <string name="open_external_url_trust_domain">Don't ask again for %1$s</string>
+    <string name="puny_code_warning_header">Suspicious link detected</string>
+    <!-- placeholder contains the hostname converted to ascii -->
+    <string name="puny_code_warning_question">Are you sure you want to visit %1$s?</string>
+    <!-- first placeholder is original hostname with special chars, second placeholder is hostname encoded in ascii -->
```
Simon-Laux, line 383:
```suggestion
    <!-- this message is shown whenever a link with non-latin characters is clicked. first placeholder is original hostname with special chars, second placeholder is hostname encoded in ascii -->
```

r10s (approved):

PR #1942, opened 2021-05-31
Remove screenlock

Hocuri:
@r10s there you go, a PR that removes the screenlock.

However, I just pulled out the three commits, I unfortunately didn't have the time to look at the files yet, so there may be things that should go in here that currently are in https://github.com/deltachat/deltachat-android/pull/1860 (or vice versa, though that seems less likely to me).


r10s (approved):
lgtm, also tested a bit.

@Hocuri i added three commits, please check if that makes sense to you :) then, this pr can be merged :)

Hocuri:
Makes sense. 75fa444 were things I forgot to pop out of https://github.com/deltachat/deltachat-android/pull/1860/.

PR #1947, opened 2021-06-07
send images from 'Image keyboard' as stickers

### LINKED ISSUE #1945: send keyboard-stickers as such
many system keyboards these days have the option to select/handle stickers.

Delta Chat supports these stickers, however, they are sent as images and not as stickers.

we should change them to be sent out as DC_MSG_STICKER, esp. as all [UI display stickers nicely meanwhile](https://support.delta.chat/t/no-chat-bubble-with-stickers/412/9).

this is eg. how the sticker-selector looks on an motog with android7:

(original feature request in the forum at https://support.delta.chat/t/stickers-simplified/1708/4)

r10s:
see https://developer.android.com/guide/topics/text/image-keyboard

closes #1945

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -68,6 +68,7 @@ import com.b44t.messenger.DcEvent;
 import com.b44t.messenger.DcMsg;
 
 import org.thoughtcrime.securesms.attachments.Attachment;
+import org.thoughtcrime.securesms.attachments.UriAttachment;
 import org.thoughtcrime.securesms.audio.AudioRecorder;
 import org.thoughtcrime.securesms.audio.AudioSlidePlayer;
 import org.thoughtcrime.securesms.components.AnimatingToggle;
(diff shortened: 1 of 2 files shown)


adbenitez:
maybe for the future:

- if a quote-reply is stagged the sticker should reply the quoted message,
- draft should not be reused, just set the sticker to quote that message and remove the quote from the draft, preserving text or maybe it is also ok to discard the draft completely?

adbenitez:
btw I am already using this, it works like a charm :D

r10s:
> btw I am already using this, it works like a charm :D

great to hear that!

do you get these weird toasts as of https://github.com/deltachat/deltachat-android/issues/1948 ?

r10s:
@adbenitez i added a commit that handles quotes as you suggested; see commit-message for details. i think, this or is now ready for review.

adbenitez:
> do you get these weird toasts as of #1948 ?

no, with the stock Samsung keyboard and with AnySoftKeyboard it works fine

adbenitez (approved):
code lgtm, also tested it already with two keyboards for a while

PR #1951, opened 2021-06-14
Hide unnecessary controls if you can't send

Hocuri:
Companion for https://github.com/deltachat/deltachat-core-rust/pull/2479, should be merged before to make sure that the compose panel is hidden immediately when the user leaves a group.

- Imediately hide the compose panel when you left the group and
EVENT_CHAT_CHANGED was sent
- Hide the "Leave" button if you are not a member of a group. I solved
this by checking `canSend()`; I could also have checked
`dc_is_contact_in_chat(CONTACT_ID_SELF)`, but this would be more
complexity for now.

  When we add the possibility to leave mailing lists, the core will have
  to explicitly tell the UI whether it's possible to leave a chat, but
  that's not for now.


r10s (approved):
lgtm

PR #1952, opened 2021-06-15
Update README.md

trollkotze:
Additional steps for setting up build environment without Docker or Podman, taken from a message by mbrubeck on the rust-lang forum (see https://users.rust-lang.org/t/build-fails-with-is-armv7a-linux-androideabi18-clang-installed/36099/2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,23 @@ done with care.
 Then, install Rust using [rustup](https://rustup.rs/). Install Rust
 toolchains for cross-compilation by executing `scripts/install-toolchains.sh`.
 
+After that, add `$ANDROID_NDK_ROOT/toolchains/llvm/prebuilt/HOST/bin` to your $PATH,
+where HOST is a descriptor for your host cpu architecture (e.g. linux-x86_64) and likely
+the name of the only subdirectory within `$ANDROID_NDK_ROOT/toolchains/llvm/prebuilt`.
+
+Then also add the correct clang-linkers to `~/.cargo/config.toml` (create it if it doesn't exist):
+
+```
+[target.armv7-linux-androideabi]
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -111,6 +111,23 @@ done with care.
...
 
+After that, add `$ANDROID_NDK_ROOT/toolchains/llvm/prebuilt/HOST/bin` to your $PATH,
+where HOST is a descriptor for your host cpu architecture (e.g. linux-x86_64) and likely
+the name of the only subdirectory within `$ANDROID_NDK_ROOT/toolchains/llvm/prebuilt`.
+
+Then also add the correct clang-linkers to `~/.cargo/config.toml` (create it if it doesn't exist):
+
+```
+[target.armv7-linux-androideabi]
+linker = "PATH_TO_NDK/toolchains/llvm/prebuilt/HOST/bin/armv7a-linux-androideabi18-clang"
```
adbenitez, line 122: hi, thanks for contributing but I think this is automatically done in `ndk-make.sh` via `CARGO_TARGET_ARMV7_LINUX_ANDROIDEABI_LINKER`, etc environment variables instead of writing it in `~/.cargo/config.toml`

see `ndk-make.sh` script line 48.

r10s: i also had a closer look, indeed, these steps are no longer needed since this refactoring: https://github.com/deltachat/deltachat-android/commit/18a19f98b6acd9180edc7a50aef8ca197634d292

before, more or less the suggestion of this pr was present in the comment / readme.

r10s:
closing this for reasons mentioned above.

@trollkotze if it still does not work for you for whatever reason, we can add something to the readme, however, we should tune this down then as it seems not to be an always necessary step.

r10s:
and also here - sorry for the delay, @trollkotze

PR #1955, opened 2021-06-20
don't open chat directly if user clicks in blocked addresses

### LINKED ISSUE #1937: contacts directly unblocked if you tap mailto link with their address and you had a chat with them before blocking
if you tap a mailto link or an address in a message, if that contact was blocked and you had an accepted chat with them in the past, the contact is unblocked and chat is open, instead, users should be asked for confirmation as if there is no accepted chat with that contacts

adbenitez:
close #1937


Hocuri (approved):
lgtm, not tested

PR #1958, opened 2021-06-21
handle openpgp4fpr schema in WebView

### LINKED ISSUE #1957: make openpgp4fpr links work in webview
with AEAP we aim to provide howtos to use QR codes on webpages that should be both, scannable and tappable.

tappable links are currently now implemented for our internal webview.

adbenitez:
close #1957


r10s (commented):
great, thanks for taking care!

r10s (approved):

PR #1960, opened 2021-06-22
App strings reworked

comradekingu:
One "eg." leaked into the strings. Noticed a lot of inconsistencies, complex conventions and errors have made it in.
An e-mail app should be ideally not get the spelling of it wrong…
Errors introduced in translation from "email" are not just strictly less correct, but wrong.
If anyone wants to use "email", I suggest setting up a glossary (and forbidden entries) on Weblate.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -59,7 +59,7 @@
     <string name="done">Done</string>
     <string name="undo">Undo</string>
     <string name="offline">Offline</string>
-    <!-- Translators: used eg. for the next view, could also be "continue" or so. as used in ios headers, the string should be as short as possible, though. -->
+    <!-- Translators: used e.g. for the next view, could also be "continue" or so. as used in iOS headers, the string should be as short as possible, though. -->
     <string name="next">Next</string>
     <string name="error">Error</string>
     <string name="error_x">Error: %1$s</string>
@@ -69,8 +69,8 @@
     <string name="file_not_found">Could not find %1$s.</string>
     <string name="copied_to_clipboard">Copied to clipboard.</string>
     <string name="contacts_headline">Contacts</string>
-    <string name="email_address">Email address</string>
-    <string name="bad_email_address">Bad email address.</string>
+    <string name="email_address">E-mail address</string>
+    <string name="bad_email_address">Erroneous e-mail address.</string>
     <string name="password">Password</string>
     <string name="existing_password">Existing password</string>
     <string name="now">Now</string>
@@ -120,14 +120,14 @@
     <string name="voice_message">Voice message</string>
     <string name="forwarded">Forwarded</string>
     <string name="forwarded_message">Forwarded message</string>
-    <!-- %1$s will be replaced by the name or the email-address of the person who has forwarded the message -->
+    <!-- %1$s will be replaced by the name or the e-mail address of the person who has forwarded the message -->
     <string name="forwarded_by">Forwarded by %1$s</string>
     <string name="video">Video</string>
     <string name="documents">Documents</string>
     <string name="contact">Contact</string>
     <string name="verified_contact">Verified contact</string>
     <string name="camera">Camera</string>
-    <!-- "capture" here means "start a video recording" or "take a photo"; eg. the description of the "shutter button" in the camera controls -->
+    <!-- "capture" here means "start a video recording" or "take a photo"; e.g. the description of the "shutter button" in the camera controls -->
     <string name="capture">Capture</string>
     <string name="switch_camera">Switch camera</string>
     <string name="toggle_fullscreen">Toggle full screen mode</string>
@@ -197,7 +197,7 @@
     <string name="menu_unmute">Unmute</string>
     <string name="menu_export_attachment">Export attachment</string>
     <string name="menu_all_media">All media</string>
-    <!-- menu entry that opens eg. a gallery image or a document in the chat at the correct position -->
+    <!-- menu entry that opens e.g. a gallery image or a document in the chat at the correct position -->
     <string name="show_in_chat">Show in chat</string>
     <string name="menu_share">Share</string>
     <!-- this is the action "to block sth." usually a mailing list or a contact. this is NOT "a large solid piece of hard material" :) -->
@@ -228,11 +228,11 @@
     <string name="jump_to_message">Jump to message</string>
     <string name="copy_json">Copy JSON</string>
     <string name="replace_draft">Replace draft</string>
-    <string name="title_share_location">Share location with all group members</string>
+    <string name="title_share_location">Share your location with the group</string>
     <string name="device_talk">Device messages</string>
     <string name="device_talk_subtitle">Locally generated messages</string>
-    <string name="device_talk_explain">Messages in this chat are generated locally by your Delta Chat app. Its makers use it to inform about app updates and problems during usage.</string>
-    <string name="device_talk_welcome_message">Welcome to Delta Chat! – Delta Chat looks and feels like other popular messenger apps, but does not involve centralized control, tracking or selling you, friends, colleagues or family out to large organizations.\n\nTechnically, Delta Chat is an email application with a modern chat interface. Email in a new dress if you will 👻\n\nUse Delta Chat with anyone out of billions of people: just use their e-mail address. Recipients don\'t need to install Delta Chat, visit websites or sign up anywhere - however, of course, if they like, you may point them to 👉 https://get.delta.chat</string>
+    <string name="device_talk_explain">Messages in this chat are generated locally by your Delta Chat app. It is used to notify you of app updates and problems during usage.</string>
+    <string name="device_talk_welcome_message">Welcome to Delta Chat. Looks and feels like other popular messenger apps, but no centralized control, tracking or selling you, friends, colleagues or family.\n\nAn e-mail app with a modern chat interface.\n\nUse it to chat over e-mail, (encrypted if both parties use https://get.delta.chat or any other Autocrypt program). Recipients don\'t need to install Delta Chat, visit websites or sign up anywhere.</string>
     <string name="edit_contact">Edit contact</string>
     <!-- Translators: "Pin" here is the verb for pinning, making sth. sticky. this is NOT the appreviation for "pin number". -->
     <string name="pin_chat">Pin chat</string>
@@ -245,18 +245,18 @@
     <string name="ConversationFragment_quoted_message_not_found">Original message not found</string>
     <string name="reply_privately">Reply privately</string>
 
-    <string name="mute_for_one_hour">Mute for 1 hour</string>
-    <string name="mute_for_two_hours">Mute for 2 hours</string>
-    <string name="mute_for_one_day">Mute for 1 day</string>
-    <string name="mute_for_seven_days">Mute for 7 days</string>
+    <string name="mute_for_one_hour">Mute one hour</string>
+    <string name="mute_for_two_hours">Mute two hours</string>
+    <string name="mute_for_one_day">Mute one day</string>
+    <string name="mute_for_seven_days">Mute seven days</string>
     <string name="mute_forever">Mute forever</string>
 
     <string name="share_location_once">once</string>
-    <string name="share_location_for_5_minutes">for 5 minutes</string>
+    <string name="share_location_for_5_minutes">for five minutes</string>
     <string name="share_location_for_30_minutes">for 30 minutes</string>
-    <string name="share_location_for_one_hour">for 1 hour</string>
-    <string name="share_location_for_two_hours">for 2 hours</string>
-    <string name="share_location_for_six_hours">for 6 hours</string>
+    <string name="share_location_for_one_hour">for one hour</string>
+    <string name="share_location_for_two_hours">for two hours</string>
+    <string name="share_location_for_six_hours">for six hours</string>
 
     <plurals name="ask_send_following_n_files_to">
         <item quantity="one">Send the following file to %s?</item>
@@ -307,8 +307,8 @@
 
     <!-- contact list -->
     <string name="contacts_title">Contacts</string>
-    <string name="contacts_enter_name_or_email">Enter name or email address</string>
-    <string name="contacts_type_email_above">Type email address above</string>
+    <string name="contacts_enter_name_or_email">Enter name or e-mail address</string>
+    <string name="contacts_type_email_above">Type e-mail address above</string>
     <string name="contacts_empty_hint">No contacts.</string>
 
 
@@ -334,19 +334,19 @@
     <string name="chat_record_slide_to_cancel">Slide to cancel</string>
     <string name="chat_record_explain">Tap and hold to record a voice message, release to send</string>
     <string name="chat_no_chats_yet_title">Empty inbox.\nPress \"+\" to start a new chat.</string>
-    <string name="chat_no_chats_yet_hint">You can chat with other Delta Chat users and with any email address.</string>
+    <string name="chat_no_chats_yet_hint">You can chat with other Delta Chat users and with any e-mail address.</string>
     <string name="chat_all_archived">All chats archived.\nPress \"+\" to start a new chat.</string>
     <string name="chat_share_with_title">Share with</string>
     <string name="chat_input_placeholder">Message</string>
     <string name="chat_archived_label">Archived</string>
     <string name="chat_no_messages">No messages.</string>
-    <string name="chat_self_talk_subtitle">Messages I sent to myself</string>
+    <string name="chat_self_talk_subtitle">Messages sent to self</string>
     <string name="saved_messages">Saved messages</string>
     <string name="saved_messages_explain">• Forward messages here for easy access\n\n• Take notes or voice memos\n\n• Attach media to save them</string>
     <!-- this "Saved" should match the "Saved" from "Saved messages" -->
     <string name="saved">Saved</string>
     <string name="chat_contact_request">Contact request</string>
-    <string name="chat_no_contact_requests">No contact requests.\n\nIf you want classic emails to appear here as contact requests, you can change the corresponding setting in the app settings.</string>
+    <string name="chat_no_contact_requests">No contact requests.\n\nIf you want classic e-mails to appear here as contact requests, you can change the corresponding setting in the app settings.</string>
     <string name="retry_send">Retry to send message</string>
     <string name="send_failed">Could not send message</string>
     <!-- reasons for a disabled message composer -->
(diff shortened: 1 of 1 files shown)


res/values/strings.xml
```diff
@@ -516,9 +515,9 @@
     <string name="pref_language">Language</string>
     <string name="pref_incognito_keyboard">Incognito keyboard</string>
     <!-- Translators: Keep in mind that this is a Request - it must be clear in the wording that this cannot be enforced. -->
-    <string name="pref_incognito_keyboard_explain">Request keyboard to disable personalized learning</string>
+    <string name="pref_incognito_keyboard_explain">Request keyboard to turn off personalized learning</string>
```
comradekingu, line 518: I have no idea what this means.

res/values/strings.xml
```diff
@@ -593,40 +592,40 @@
     <string name="autodel_device_title">Delete messages from device</string>
     <string name="autodel_server_title">Delete messages from server</string>
     <!-- %1$d will be replaced by the number of messages, you can assume plural/lots here. %2$s will be replaced by a timespan option. -->
-    <string name="autodel_device_ask">Do you want to delete %1$d messages now and all newly fetched messages \"%2$s\" in the future?\n\n• This includes all media\n\n• Messages will be deleted whether they were seen or not\n\n• \"Saved messages\" will be skipped from local deletion</string>
+    <string name="autodel_device_ask">Do you want to delete %1$d messages now and all newly fetched messages \"%2$s\" in the future?\n\n• This includes all media\n\n• Messages will be deleted whether they were seen or not\n\n• \"Saved messages\" will be skipped from local deletion.</string>
```
comradekingu, line 595: Is this "all chat content"?
Some places that say "messages" make it seem like it is all the history.

res/values/strings.xml
```diff
@@ -827,22 +826,22 @@
...
-    <string name="InfoPlist_NSMicrophoneUsageDescription">Delta Chat uses your microphone to record and send voice messages and videos with sound.</string>
-    <string name="InfoPlist_NSPhotoLibraryUsageDescription">Delta Chat will let you choose which photos from your library to send.</string>
-    <string name="InfoPlist_NSPhotoLibraryAddUsageDescription">Delta Chat wants to save images to your photo library.</string>
+    <string name="InfoPlist_NSCameraUsageDescription">Your camera scans QR codes for logins, and captures photos and videos to send directly.</string>
+    <string name="InfoPlist_NSContactsUsageDescription">A list of e-mail addresses in your contacts is assembled. It is not sent anywhere.</string>
+    <string name="InfoPlist_NSMicrophoneUsageDescription">Your mic is used to record messages and videos with sound to send directly.</string>
+    <string name="InfoPlist_NSPhotoLibraryUsageDescription">Access to your library lets you pick existing files, images and videos to send.</string>
+    <string name="InfoPlist_NSPhotoLibraryAddUsageDescription">Access to your gallery is needed to save images and video you capture.</string>
```
comradekingu, line 833:
```suggestion
    <string name="InfoPlist_NSPhotoLibraryAddUsageDescription">Gallery access is needed to save images and video you capture.</string>
```

res/values/strings.xml
```diff
@@ -771,7 +770,7 @@
     <string name="contact_request_title_desktop">Contact request</string>
     <string name="delete_message_desktop">Delete message</string>
     <string name="more_info_desktop">More info</string>
-    <string name="logout_desktop">Logout</string>
+    <string name="logout_desktop">Log out</string>
     <string name="timestamp_format_m_desktop">MMM D</string>
```
comradekingu, line 774: Is this correct?

r10s:
as there are at least some changes that may result in some discussion as they are more a matter of taste or some wording has some background that should not be altered (eg. "setup code"), i suggest to split of the simple typo changes from this pr so that they can be merged more easily.

r10s (changes requested):
i do not think, a pr with that many wording changes is easily mergable.

- there are at least some changes resulting in some discussion as they are more a matter of taste
- other rewording change the meaning "gallery" vs. "your gallery"
- some wording has some background that should not be altered (eg. "setup code")

in general, i am skeptical on larger rewordings without concrete reasons eg. user feedback.

to move forward, i suggest to split off the typo changes to a separate pr that can be merged more easily. esp. as we can push them to transifex without forcing retranslation of all languages (./scripts/tx-update-changed-sources.sh); we should not mix typo and semantic changes therefore.

res/values/strings.xml
Gh0stz0x: From wikipedia seem correct "Logout", because: Logout is the English term to indicate the exit procedure from a computer system or computer application. It comes from the contraction of *log out*, exit from the log, the chronological log of events typical of any information system.

res/values/strings.xml
```diff
@@ -827,22 +826,22 @@
...
+    <string name="InfoPlist_NSPhotoLibraryUsageDescription">Access to your library lets you pick existing files, images and videos to send.</string>
+    <string name="InfoPlist_NSPhotoLibraryAddUsageDescription">Access to your gallery is needed to save images and video you capture.</string>
 
 
     <!-- android specific strings, developers: please take care to remove strings that are no longer used! -->
     <string name="pref_background_notifications">Background notifications</string>
-    <string name="pref_background_notifications_explain">Uses a background connection to your server and requires ignored battery optimizations.</string>
-    <string name="pref_background_notifications_rationale">To maintain connection to your email server and receive messages in the background, ignore battery optimizations in the next step.\n\nDelta Chat uses few resources and takes care not to drain your battery.</string>
+    <string name="pref_background_notifications_explain">Uses a background connection to your mail server and requires ignored battery optimizations.</string>
```
Gh0stz0x, line 838: It seems to me that there is a "consistency" problem here.  Until now we have used e-mail and here mail.

res/values/strings.xml
```diff
@@ -827,22 +826,22 @@
...
+    <string name="pref_background_notifications_rationale">To maintain connection to your mail server and receive messages in the background, ignore battery optimizations in the next step.\n\nDelta Chat uses few resources and takes care not to drain your battery.</string>
```
Gh0stz0x, line 839: It seems to me that there is a "consistency" problem here.  Until now we have used e-mail and here mail.

res/values/strings.xml
comradekingu: There was a problem, because no one convention was used, I think that may be what leads to so many erroneous variants being used. It should always be "mail server", unfortunately.

Gh0stz0x: Conventions should probably be followed, if this does not create confusion for the end user. 🤔

comradekingu:
@r10s It is as easily merged as it is reverted. :)
I did explain my rationale on IRC, which I think qualifies as user feedback. (That is the intention here anyway.)
I remember that including why "setup code" doesn't work. If it doesn't explain anything, to anyone needing it explained, and can be omitted fully, does it add value?
I think that was a compromise, but I think it is time to take a hard look at the more technical explanations, because they all add to much complexity.

I don't do extra work because of Transifex, because I don't use Transifex.
If you want to, sure. Nobody wants to track or work on strings from there, because it is a pain.
Rarely if ever are the translations any good, and without source strings that make sense, chances are slim.
I am just trying to solve my issues with the source strings, since I have to see so many of them.

If "Gallery access" got a thumbs up, it stands to reason photos aren't saved in a "Photo library".
It is always "Gallery" on Android anyway. Is the string used between clients?
"Photo library" to me sounds like it is some app internal thing.
I didn't write "your" gallery? Do you mean that I should?

res/values/strings.xml
comradekingu: Here it is a direct action rather than a noun, is what I gathered.
Getting to the logout, vs. logging out. You don't "logout", you log out.

Gh0stz0x: Ok

comradekingu: Oh, my question was actually about the "MMM D" Why isn't it "MM DD"?

Hocuri:
> It is as easily merged as it is reverted.

Unfortunately not, as soon as it's merged, people will start translating it, so we shoudn't quickly go back and forth, to not annoy translators.

Also, as @r10s said, our current process involves ./scripts/tx-update-changed-sources.sh for typos, also to not annoy translators, we can't just skip this step.

>  Also offering to set it up on Weblate.

Yes, Weblate would definitely be nicer than Transifex, but migrating would be a big effort too; @r10s would have to do most of it.

I agree with about half of the semantic changes, but with such a big PR that's difficult to merge only partly, that's hard to discuss. Also, we should probably be pretty sure that we want to change a wording and not go back later, both to not annoy users and translators with our back-and-forth.

res/values/strings.xml
```diff
@@ -228,11 +228,11 @@
     <string name="jump_to_message">Jump to message</string>
     <string name="copy_json">Copy JSON</string>
     <string name="replace_draft">Replace draft</string>
-    <string name="title_share_location">Share location with all group members</string>
+    <string name="title_share_location">Share your location with the group</string>
```
r10s, line 231: "the group" seems to be less specific. but i am wondering if there are issues with the existing string at all.

res/values/strings.xml
```diff
@@ -245,18 +245,18 @@
...
     <string name="reply_privately">Reply privately</string>
 
-    <string name="mute_for_one_hour">Mute for 1 hour</string>
-    <string name="mute_for_two_hours">Mute for 2 hours</string>
-    <string name="mute_for_one_day">Mute for 1 day</string>
-    <string name="mute_for_seven_days">Mute for 7 days</string>
+    <string name="mute_for_one_hour">Mute one hour</string>
+    <string name="mute_for_two_hours">Mute two hours</string>
+    <string name="mute_for_one_day">Mute one day</string>
+    <string name="mute_for_seven_days">Mute seven days</string>
```
r10s, line 251: not sure, if these written numbers here and at similar places an improvement, although maybe more correct.

res/values/strings.xml
```diff
@@ -120,14 +120,14 @@
...
-    <!-- %1$s will be replaced by the name or the email-address of the person who has forwarded the message -->
+    <!-- %1$s will be replaced by the name or the e-mail address of the person who has forwarded the message -->
     <string name="forwarded_by">Forwarded by %1$s</string>
     <string name="video">Video</string>
     <string name="documents">Documents</string>
     <string name="contact">Contact</string>
     <string name="verified_contact">Verified contact</string>
     <string name="camera">Camera</string>
-    <!-- "capture" here means "start a video recording" or "take a photo"; eg. the description of the "shutter button" in the camera controls -->
+    <!-- "capture" here means "start a video recording" or "take a photo"; e.g. the description of the "shutter button" in the camera controls -->
```
r10s, line 130: not sure which one is better, there are lots of discussions around it :) iirc, in core we decided for "eg." (i personally also did use "e.g." before, but i do not care at the end :)

res/values/strings.xml
```diff
@@ -465,11 +465,10 @@
     <!-- share and forward messages -->
     <!-- Translators: Title shown above a chat/contact list; the user selects the recipient of the messages he wants to forward to -->
     <string name="forward_to">Forward to…</string>
-    <string name="share_multiple_attachments">Send %1$d files to the selected chat?\n\nThe files are sent unmodified in their original size, eg. images and videos are not recoded.</string>
-    <string name="share_multiple_attachments_multiple_chats">Send %1$d file(s) to %2$d chats?\n\nThe files are sent unmodified in their original size, eg. images and videos are not recoded.</string>
+    <string name="share_multiple_attachments">Send %1$d file(s) unmodified to the selected chat?</string>
+    <string name="share_multiple_attachments_multiple_chats">Send %1$d unmodified file(s) to %2$d chats?</string>
```
r10s, line 469: not sure if it is useful to leave out details. wondering if there was some feedback before.

res/values/strings.xml
```diff
@@ -465,11 +465,10 @@
...
     <string name="share_text_multiple_chats">Send this text to %1$d chats?\n\n\"%2$s\"</string>
-    <string name="share_abort">Sharing aborted due to missing permissions.</string>
-
+    <string name="share_abort">Grant all required permissions to share this.</string>
```
r10s, line 471: not sure if that is better error message, however, also not 100% sure where the string appears at all.

res/values/strings.xml
```diff
@@ -59,7 +59,7 @@
     <string name="done">Done</string>
     <string name="undo">Undo</string>
     <string name="offline">Offline</string>
-    <!-- Translators: used eg. for the next view, could also be "continue" or so. as used in ios headers, the string should be as short as possible, though. -->
+    <!-- Translators: used e.g. for the next view, could also be "continue" or so. as used in iOS headers, the string should be as short as possible, though. -->
```
r10s, line 62: i will fix the iOS and other typos in another pr.

res/values/strings.xml
```diff
@@ -120,14 +120,14 @@
     <string name="voice_message">Voice message</string>
     <string name="forwarded">Forwarded</string>
     <string name="forwarded_message">Forwarded message</string>
-    <!-- %1$s will be replaced by the name or the email-address of the person who has forwarded the message -->
+    <!-- %1$s will be replaced by the name or the e-mail address of the person who has forwarded the message -->
```
r10s, line 123: yip, that mix of email and e-mail is not good. i will fix that in another pr.

res/values/strings.xml
```diff
@@ -528,15 +527,15 @@
...
-    <string name="pref_saved_log">Saved the log to \"Downloads\" folder</string>
-    <string name="pref_save_log_failed">Failed to save the log</string>
+    <string name="pref_saved_log">Log saved in the \"Downloads\" folder</string>
+    <string name="pref_save_log_failed">Could not save log</string>
     <string name="pref_log_header">Log</string>
     <string name="pref_other">Other</string>
     <string name="pref_backup">Backup</string>
-    <string name="pref_backup_explain">Backup chats to external storage</string>
-    <string name="pref_backup_export_explain">A backup helps you to set up a new installation on this or on another device.\n\nThe backup will contain all messages, contacts and chats and your end-to-end Autocrypt setup. Keep the backup file in a safe place or delete it as soon as possible.</string>
+    <string name="pref_backup_explain">Back up chats to external storage</string>
```
r10s, line 535: +1

res/values/strings.xml
```diff
@@ -576,11 +575,11 @@
     <string name="emoji_search_results">Search results</string>
     <string name="emoji_not_found">No emoji found</string>
     <string name="emoji_recent">Recently used</string>
-    <string name="emoji_people">People &amp; Body</string>
-    <string name="emoji_nature">Animals &amp; Nature</string>
-    <string name="emoji_foods">Food &amp; Drink</string>
+    <string name="emoji_people">People and Body</string>
+    <string name="emoji_nature">Animals and Nature</string>
+    <string name="emoji_foods">Food and Drink</string>
```
r10s, line 580: i think, the reason for the ampersand was to save some space on the tabs and to have less text overall there.

res/values/strings.xml
```diff
@@ -593,40 +592,40 @@
...
     <string name="autocrypt_continue_transfer_title">Autocrypt Setup Message</string>
-    <string name="autocrypt_continue_transfer_please_enter_code">Please enter the setup code displayed on the other device.</string>
-    <string name="autocrypt_continue_transfer_succeeded">End-to-end setup transferred. This device is now ready to use Autocrypt with the same setup as the other device.</string>
+    <string name="autocrypt_continue_transfer_please_enter_code">Please enter the confirmation code displayed on the other device.</string>
+    <string name="autocrypt_continue_transfer_succeeded">Autocrypt End-to-end encryption transferred to yourself. This device is now ready to use like the other device.</string>
     <string name="autocrypt_continue_transfer_retry">Retry</string>
-    <string name="autocrypt_bad_setup_code">Bad setup code. Please try again.\n\nIf you do not remember the setup code, send a new Autocrypt Setup Message from the other device.</string>
+    <string name="autocrypt_bad_setup_code">Wrong confirmation code. Please try again.\n\nIf you do not remember it, send a new Autocrypt Setup Message from the other device.</string>
```
r10s, line 628: one may like or dislike the term "setup code", however, changing it is not only a thing of this file, it is used at several other places in other apps as well.

also, not sure if "confirmation code" is better at all.

res/values/strings.xml
```diff
@@ -638,30 +637,30 @@
...
+    <string name="systemmsg_ephemeral_timer_minutes">Messages now dissapear in %1$s minutes.</string>
+    <string name="systemmsg_ephemeral_timer_hours">Messages now dissapear in %1$s hours.</string>
+    <string name="systemmsg_ephemeral_timer_days">Messages now dissapear in %1$s days.</string>
+    <string name="systemmsg_ephemeral_timer_weeks">Messages now dissapear in %1$s weeks.</string>
 
     <!-- Translators: Protection disabled/enabled messages may be displayed together with a "big shield" and may be suffixed by systemmsg_action_by_user or systemmsg_action_by_me -->
-    <string name="systemmsg_chat_protection_disabled">Chat protection disabled.</string>
-    <string name="systemmsg_chat_protection_enabled">Chat protection enabled.</string>
+    <string name="systemmsg_chat_protection_disabled">Chat protection off.</string>
+    <string name="systemmsg_chat_protection_enabled">Chat protection on.</string>
```
r10s, line 663: i do not think "off" is the same as "disabled", the one is the state, the other, an action just taken by someone.

but anyway, the string might be expected by the user taking the action (see comment), so "off" would not work anyway here (we should get rid of this expansion, however, currently it is like that)

r10s:
i picked up the typos and merged them to master via #1966, thanks a lot for that, @comradekingu , i also commented a bit, but not exhaustively.

it is often better to change strings only if there is some incidence from users, that the existing string does not work. also, as a rule of thumb, it is a good idea to do that only when an area is reworked anyway.

if one still wants semantic changes or rewordings, they should be more focused, otherwise things are hard to merge. please take #1926, #1944 or the maybe starting discussion at #1965 as examples.

res/values/strings.xml
comradekingu: yes, because it isn't "all group members excluding group owners".

res/values/strings.xml
comradekingu: Just looks more natural. Technically it is probably a list and so on, but in translation you avoid problems with languages that write out all numbers 1-10.

res/values/strings.xml
comradekingu: The abbreviation is "exempli gratia", and "e.g." is correct.
In those discussions, like DeltaChat core, people saying "eg." are wrong.
Glad we cleared that up :)

res/values/strings.xml
comradekingu: Is there an option to send them compressed or without metadata?
If not, sending files is sending files and "unmodified" can be omitted.

res/values/strings.xml
comradekingu: I like active hints better.

res/values/strings.xml
comradekingu: Be that as it may, the code only belongs to the confirmation-part, not to the setup itself.

res/values/strings.xml
comradekingu: Not necessarily, and depends on the translation.
I think there is no distinction to be made for something that can be on and disabled at that.
This is just "off" or "on" (?).

res/values/strings.xml
r10s: indeed, there is a translator comment missing. i assume it is some sort of date formatting, where `MMM D` expands to `Jul 15` where `MM DD` expands to `07 15`.

PR #1965, opened 2021-06-29
allow withdrawing and reviving qr codes, add corresponding strings

### LINKED ISSUE #1962: withdraw/revive own qr-codes
as https://github.com/deltachat/deltachat-core-rust/pull/2512 is merged now, ui can implement  DC_QR_{WITHDRAW|REVIVE}_VERIFY{CONTACT|GROUP}:

- if one of the codes are returned by [`dc_check_qr()`](https://c.delta.chat/classdc__context__t.html#a34a865a52127ed2cc8c2f016f085086c), show an appropriate alert (sth. as "This QR code was generated by you for ... do you want to withdraw it?" resp. "This QR code was withdrawn. Do you want to revive it for ...") (lot.text1 contains the groupname in case of withdrawing/reviving group join QR codes)

- if the users chooses to continue, pass the scanned qr code to [`dc_set_config_from_qr()`](https://c.delta.chat/classdc__context__t.html#ac95d7275f01862fba70169b8cbb937c0)

~~once done and the strings are added to strings.xml, corresponding issues for ios/desktop should be filed.~~ EDIT corresponding issues are https://github.com/deltachat/deltachat-ios/issues/1291 and https://github.com/deltachat/deltachat-desktop/issues/2297

r10s:
the code-part of this pr is pretty simple, most work was already done in core at https://github.com/deltachat/deltachat-core-rust/pull/2512

what should be reviewed mainly is the wording of the 4 dialogs that appear when a user scans ones own qr codes:

**widthdraw/revive secure join qr codes:**

 

**widthdraw/revive group joins:**

 

closes #1962

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -682,8 +682,8 @@
     <string name="withdraw_verifycontact_explain">This QR code can be scanned by others to setup a contact with you.\n\nIf you do not want that any longer, you can withdraw it. You can revive it by scanning it again.</string>
     <string name="withdraw_verifygroup_explain">This QR code can be scanned by others to join the group \"%1$s\".\n\nIf you do not want that any longer, you can withdraw it. You can revive it by scanning it again.</string>
     <string name="withdraw_qr_code">Withdraw QR code</string>
-    <string name="revive_verifycontact_explain">This QR code was withdrawn and was used to setup a contact with you. Currently the QR code is not functional.</string>
-    <string name="revive_verifygroup_explain">This QR code was withdrawn and was used to join the group \"%1$s\". Currently the QR code is not functional.</string>
+    <string name="revive_verifycontact_explain">This QR code could be scanned to setup a contact with you, but was withdrawn. Currently the QR code is not functional.</string>
+    <string name="revive_verifygroup_explain">This QR code could be scanned by others to join the group \"%1$s\", but was withdrawn. Currently the QR code is not functional.</string>
     <string name="revive_qr_code">Revive QR code</string>
     <string name="qrshow_title">QR invite code</string>
     <string name="qrshow_x_joining">%1$s joins.</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -682,7 +682,7 @@
     <string name="withdraw_verifycontact_explain">This QR code can be scanned by others to setup a contact with you.\n\nIf you do not want that any longer, you can withdraw it. You can revive it by scanning it again.</string>
     <string name="withdraw_verifygroup_explain">This QR code can be scanned by others to join the group \"%1$s\".\n\nIf you do not want that any longer, you can withdraw it. You can revive it by scanning it again.</string>
     <string name="withdraw_qr_code">Withdraw QR code</string>
-    <string name="revive_verifycontact_explain">This QR code could be scanned to setup a contact with you, but was withdrawn. Currently the QR code is not functional.</string>
+    <string name="revive_verifycontact_explain">This QR code could be scanned by others to setup a contact with you, but was withdrawn. Currently the QR code is not functional.</string>
     <string name="revive_verifygroup_explain">This QR code could be scanned by others to join the group \"%1$s\", but was withdrawn. Currently the QR code is not functional.</string>
     <string name="revive_qr_code">Revive QR code</string>
     <string name="qrshow_title">QR invite code</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -679,6 +679,12 @@
     <string name="qrscan_contains_url">Scanned QR code URL:\n\n%1$s</string>
     <string name="qrscan_fingerprint_label">Fingerprint</string>
     <string name="qrscan_x_verified_introduce_myself">%1$s verified, introduce myself…</string>
+    <string name="withdraw_verifycontact_explain">This QR code can be scanned by others to setup a contact with you.\n\nIf you do not want that any longer, you can withdraw it. Withdrawn QR codes can be revived by scanning again.</string>
+    <string name="withdraw_verifygroup_explain">This QR code can be scanned by others to join the group \"%1$s\".\n\nIf you do not want that any longer, you can withdraw it. Withdrawn QR codes can be revived by scanning again.</string>
+    <string name="withdraw_qr_code">Withdraw QR code</string>
+    <string name="revive_verifycontact_explain">This QR code was withdrawn and was used to setup a contact with you. Currently the QR code it is not functional.</string>
+    <string name="revive_verifygroup_explain">This QR code was withdrawn and was used to join the group \"%1$s\". Currently the QR code it is not functional.</string>
+    <string name="revive_qr_code">Revive QR code</string>
     <string name="qrshow_title">QR invite code</string>
     <string name="qrshow_x_joining">%1$s joins.</string>
     <string name="qrshow_x_verified">%1$s verified.</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -679,8 +679,8 @@
     <string name="qrscan_contains_url">Scanned QR code URL:\n\n%1$s</string>
     <string name="qrscan_fingerprint_label">Fingerprint</string>
     <string name="qrscan_x_verified_introduce_myself">%1$s verified, introduce myself…</string>
-    <string name="withdraw_verifycontact_explain">This QR code can be scanned by others to setup a contact with you.\n\nIf you do not want that any longer, you can withdraw it. You can revive it by scanning it again.</string>
-    <string name="withdraw_verifygroup_explain">This QR code can be scanned by others to join the group \"%1$s\".\n\nIf you do not want that any longer, you can withdraw it. You can revive it by scanning it again.</string>
+    <string name="withdraw_verifycontact_explain">Prevent others from contacting you when scanning this QR code?\n\n(Can be undone by scanning it yourself afterwards.)</string>
+    <string name="withdraw_verifygroup_explain">Prevent others from joining your \"%1$s\" group when scanning this QR code?\n\n(Can be undone by scanning it yourself afterwards.)</string>
     <string name="withdraw_qr_code">Withdraw QR code</string>
     <string name="revive_verifycontact_explain">This QR code could be scanned by others to setup a contact with you, but was withdrawn. Currently the QR code is not functional.</string>
     <string name="revive_verifygroup_explain">This QR code could be scanned by others to join the group \"%1$s\", but was withdrawn. Currently the QR code is not functional.</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -679,11 +679,11 @@
     <string name="qrscan_contains_url">Scanned QR code URL:\n\n%1$s</string>
     <string name="qrscan_fingerprint_label">Fingerprint</string>
     <string name="qrscan_x_verified_introduce_myself">%1$s verified, introduce myself…</string>
-    <string name="withdraw_verifycontact_explain">This QR code can be scanned by others to setup a contact with you.\n\nIf you do not want that any longer, you can withdraw it. Withdrawn QR codes can be revived by scanning again.</string>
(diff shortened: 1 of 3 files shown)


Hocuri:
1. In the both withdrawing QR codes: "This QR code was withdrawn and was used" -> "This QR code was withdrawn and could be used"?
2. In the both setupcontact QR codes: Actually, that's not only about setting up a contact, it's also about verifying them. So, maybe "to setup a _verified_ contact with you" or "to add you as a contact and verify you"? OTOH, in this case we should probably also change the other places where this wording is used.

r10s:
> In the both withdrawing QR codes: "This QR code was withdrawn and was used" -> "This QR code was withdrawn and could be used"?

well, my english is not good enough to decide that :) 
if you are sure, of course, we can change it to "could" :)

> In the both setupcontact QR codes: Actually, that's not only about setting up a contact, it's also about verifying them. So, maybe "to setup a verified contact with you" or "to add you as a contact and verify you"? OTOH, in this case we should probably also change the other places where this wording is used.

yip, we are not talking about verification elsewhere, so i would not start with that just here.

Hocuri:
> well, my english is not good enough to decide that :)
> if you are sure, of course, we can change it to "could" :)

OK, I had a lengthy conversation with my mother who studied Anglistics (English language) and is non-technical :upside_down_face: 

- There is one "it" too much: "Currently the QR code *it* is not functional"
- "Withdrawn QR codes can be revived by scanning again" -> It was not clear to her that _you_ have to scan it again. "You can revive it by scanning it again" would be nicer (and a little shorter).
- For the revival questions, these would be nice formulations:
  {This was a QR invite code for the group "XY"|This QR code could be scanned by others to join the group "XY"}{, but was withdrawn.|. Currently the QR code is not functional.}

  '*This was a QR invite code for the group "XY"*' is probably not a good choice though, as I can't come up with a good wording for the verifycontact case.

  What do you think?

r10s:
> What do you think?

i think: thanks a lot to your mother!

i added two commit and incorporated the suggestions.

res/values/strings.xml
```diff
@@ -679,6 +679,12 @@
     <string name="qrscan_contains_url">Scanned QR code URL:\n\n%1$s</string>
     <string name="qrscan_fingerprint_label">Fingerprint</string>
     <string name="qrscan_x_verified_introduce_myself">%1$s verified, introduce myself…</string>
+    <string name="withdraw_verifycontact_explain">This QR code can be scanned by others to setup a contact with you.\n\nIf you do not want that any longer, you can withdraw it. You can revive it by scanning it again.</string>
+    <string name="withdraw_verifygroup_explain">This QR code can be scanned by others to join the group \"%1$s\".\n\nIf you do not want that any longer, you can withdraw it. You can revive it by scanning it again.</string>
+    <string name="withdraw_qr_code">Withdraw QR code</string>
+    <string name="revive_verifycontact_explain">This QR code could be scanned to setup a contact with you, but was withdrawn. Currently the QR code is not functional.</string>
```
Hocuri, line 685:
```suggestion
    <string name="revive_verifycontact_explain">This QR code could be scanned by others to setup a contact with you, but was withdrawn. Currently the QR code is not functional.</string>
```
the other 3 strings all include "by others"

res/values/strings.xml
```diff
@@ -679,6 +679,12 @@
...
+    <string name="revive_verifygroup_explain">This QR code could be scanned by others to join the group \"%1$s\", but was withdrawn. Currently the QR code is not functional.</string>
```
Hocuri, line 686: Actually, I think that it's enough to either write "but was withdrawn" or "currently the QR code is not functional", because that's saying the same thing, and users tend to like shorter dialogs

Hocuri (approved):

Hocuri:
BTW what happens when I try to join your group by scanning a withdrawn QR code?

res/values/strings.xml
```diff
@@ -679,6 +679,12 @@
     <string name="qrscan_contains_url">Scanned QR code URL:\n\n%1$s</string>
     <string name="qrscan_fingerprint_label">Fingerprint</string>
     <string name="qrscan_x_verified_introduce_myself">%1$s verified, introduce myself…</string>
+    <string name="withdraw_verifycontact_explain">This QR code can be scanned by others to setup a contact with you.\n\nIf you do not want that any longer, you can withdraw it. You can revive it by scanning it again.</string>
+    <string name="withdraw_verifygroup_explain">This QR code can be scanned by others to join the group \"%1$s\".\n\nIf you do not want that any longer, you can withdraw it. You can revive it by scanning it again.</string>
```
cyBerta, line 683: why is it necessary to add 'by others' at all? You could also join a group with a second address.

res/values/strings.xml
cyBerta: I would rather remove 'by others' in the strings above

res/values/strings.xml
cyBerta: 'but was withdrawn' is the explanation for 'currently the QR code is not functional'. From my point of view, that's not exactly the same. It would be probably nicer to show this causality within one sentence.

Hocuri: Not sure if I understood correctly: You propose to leave it as is?

res/values/strings.xml
Hocuri: Because if we say "This QR code can be scanned to join the group XY", some people will wonder "But I just scanned it, why didn't I join the group?"

There are two types of "scan" here:
- scan your own QR code
- Let others scan your QR code

and we have to make it clear which one we mean every time we write "scan". This was the result of an actual user test with n=1 non-technical participants I made :)

> You could also join a group with a second address.

I think that it will be clear that "others" includes "other addresses of yourself"; in case you don't agree it's probably best to go ask non-technical users again (of course they will have to be technical enough to understand why you would want to join your own group at all, and to understand that you could scan your QR code with another account for this)

res/values/strings.xml
```diff
@@ -679,6 +679,12 @@
     <string name="qrscan_contains_url">Scanned QR code URL:\n\n%1$s</string>
     <string name="qrscan_fingerprint_label">Fingerprint</string>
     <string name="qrscan_x_verified_introduce_myself">%1$s verified, introduce myself…</string>
+    <string name="withdraw_verifycontact_explain">This QR code can be scanned by others to setup a contact with you.\n\nIf you do not want that any longer, you can withdraw it. You can revive it by scanning it again.</string>
```
comradekingu, line 682:
```suggestion
    <string name="withdraw_verifycontact_explain">This QR code can be scanned to get in contact with you.\n\nIt can be withdrawn at any time, and revived by scanning it again.</string>
```

res/values/strings.xml
comradekingu:
```suggestion
    <string name="withdraw_verifygroup_explain">This QR code can be scanned to join the \"%1$s\" group.\n\nIt can be withdrawn at any time, and revived by scanning it again.</string>
```

res/values/strings.xml
comradekingu:
```suggestion
    <string name="revive_verifycontact_explain">This withdrawn QR code will no longer set up contact with you.</string>
```

res/values/strings.xml
comradekingu:
```suggestion
    <string name="revive_verifycontact_explain">Scanning this withdrawn QR code will no longer set up contact with you.</string>
    <string name="revive_verifygroup_explain">Scanning this withdrawn QR code will no longer allow joining the \"%1$s\" group.</string>
```

res/values/strings.xml
comradekingu: I don't understand what "reviving" a code does, nor who is able to do it?
In my head it is withdrawn so that scanning it does nothing, but all of a sudden it can be revived that way…?
How is it ever withdrawn then?

Hocuri: Hm, maybe this needs some more thought then, if it wasn't clear even to you as a technical person.

It's:
- First, _You_ create a group invite QR code.
- then, _others_ can scan it to join the group
- Now, _you_ scan it to withdraw it, after confirming that you want to withdraw it. `withdraw_verifygroup_explain` is the confirmation & explanation that's shown at this point.
- then, when _others_ scan it now, trying to join your group, it does nothing.
- Now, _you_ scan it again to revive it. `revive_verifygroup_explain` is the confirmation & explanation that's shown it this point.
- then, _others_ can scan it again to join your group.

Withdrawing contact invite codes works the same way.

Is it clear now? I guess you want to change your suggested changes again? Esp. can you say what other formulation would have made it clear?

comradekingu: Ah, so _creators_ can scan it to rescind it. Understood.

comradekingu:
@Hocuri 
Isn't it a bit strange to say "scanning this QR code" or "This code can be scanned" if the user has already scanned the code?
Shouldn't it be something like

> Since you have scanned your own code for establishing contact with you, you can cancel it. (Do the same again to revive it.)

res/values/strings.xml
```diff
@@ -679,6 +679,12 @@
...
+    <string name="withdraw_verifygroup_explain">This QR code can be scanned by others to join the group \"%1$s\".\n\nIf you do not want that any longer, you can withdraw it. You can revive it by scanning it again.</string>
+    <string name="withdraw_qr_code">Withdraw QR code</string>
```
comradekingu, line 684:
```suggestion
    <string name="withdraw_qr_code">Yes</string>
```

res/values/strings.xml
cyBerta: I like this proposal

Hocuri: In an a/v with @cyBerta we said that we would like to use the longer version that also explains the context to the user.

cyBerta (approved):

res/values/strings.xml
comradekingu: Why?

res/values/strings.xml
r10s: ui-wise, "yes" and "no" buttons are not always the best choice, cmp. https://developer.apple.com/design/human-interface-guidelines/ios/views/alerts/

just saying, i know, we are not following these guidelines everywhere :)

comradekingu: They went full circle with that. Curiously the image above that text has fallen out of line with its advice.
I use it for the single option notification.
Some of that advice is good reason for going with my suggestions.
The alternative is leaving it up there and going with GNOME style imperative buttons.
The difference between "Yes" and "OK" isn't the problem. Exacerbating the problem of overloading info is.
It doesn't matter what the buttons say if the user is gone by then.

res/values/strings.xml
```diff
@@ -679,6 +679,12 @@
     <string name="qrscan_contains_url">Scanned QR code URL:\n\n%1$s</string>
     <string name="qrscan_fingerprint_label">Fingerprint</string>
     <string name="qrscan_x_verified_introduce_myself">%1$s verified, introduce myself…</string>
+    <string name="withdraw_verifycontact_explain">Prevent others from contacting you when scanning this QR code?\n\n(Can be undone by scanning it yourself afterwards.)</string>
+    <string name="withdraw_verifygroup_explain">Prevent others from joining your \"%1$s\" group when scanning this QR code?\n\n(Can be undone by scanning it yourself afterwards.)</string>
+    <string name="withdraw_qr_code">Withdraw QR code</string>
+    <string name="revive_verifycontact_explain">This QR code could be scanned by others to setup a contact with you, but was withdrawn. Currently the QR code is not functional.</string>
```
comradekingu, line 685: This should not only be "set up", it is conjecture.
Nobody can tell anyone why this is a good approach, because it was never attempted.

res/values/strings.xml
```diff
@@ -679,6 +679,12 @@
...
+    <string name="revive_verifygroup_explain">This QR code could be scanned by others to join the group \"%1$s\", but was withdrawn. Currently the QR code is not functional.</string>
```
comradekingu, line 686: Self-referencing that foregoes replication, for no reason.
"Could" is left up to interpretation. Both meanings thereof are completely reversed after the comma.
The last sentence parades what should already be clear from what the possible actions are.
This explains nothing and instead operates in the realm of riddles.
2/10

PR #1966, opened 2021-06-30
fix some typos

r10s:
this pr shall only contain typo fixes so that they can be pushed without forcing retranslations.

thanks a lot @comradekingu for the suggestions at https://github.com/deltachat/deltachat-android/pull/1960

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -59,7 +59,7 @@
     <string name="done">Done</string>
     <string name="undo">Undo</string>
     <string name="offline">Offline</string>
-    <!-- Translators: used eg. for the next view, could also be "continue" or so. as used in ios headers, the string should be as short as possible, though. -->
+    <!-- Translators: used eg. for the next view, could also be "continue" or so. as used in iOS headers, the string should be as short as possible, though. -->
     <string name="next">Next</string>
     <string name="error">Error</string>
     <string name="error_x">Error: %1$s</string>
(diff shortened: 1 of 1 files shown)


res/values/strings.xml
```diff
@@ -837,7 +837,7 @@
     <!-- android specific strings, developers: please take care to remove strings that are no longer used! -->
     <string name="pref_background_notifications">Background notifications</string>
     <string name="pref_background_notifications_explain">Uses a background connection to your server and requires ignored battery optimizations.</string>
-    <string name="pref_background_notifications_rationale">To maintain connection to your email server and receive messages in the background, ignore battery optimizations in the next step.\n\nDelta Chat uses few resources and takes care not to drain your battery.</string>
+    <string name="pref_background_notifications_rationale">To maintain connection to your e-mail server and receive messages in the background, ignore battery optimizations in the next step.\n\nDelta Chat uses few resources and takes care not to drain your battery.</string>
```
comradekingu, line 840: @r10s It is never an "e-mail server", always "mail server".
"The server that holds your e-mails" avoids this problem.

r10s: interesting, indeed, there are few results for "e-mail server" - but still some, eg. https://www.ionos.com/office-solutions/email-server

so, i am wondering if people really will take care, esp. as it looks more consistent the way we're doing it currently, but at the end, i would not really care.

("The server that holds your e-mails" seems to be too long here, things are already complicated enough in this sentence)

comradekingu: I would argue use by people opting for "About IONOS Mail Server" just below it solidifies my point.
What I would do is
"Turn off battery optimizations to receive messages as they arrive." Y/N
Fluff is bad UI, and harder yet to translate than it is to read. Delta Chat seems to target users wanting technicalities semi-explained at length, when there are no such users.

PR #1968, opened 2021-07-05
avoid NullPointerException in DcMsg.msgSetToIds()

### LINKED ISSUE #1967: Possible null-pointer-exception.
- Android version:
- Device:
- Delta Chat version:
- Expected behavior:
- Actual behavior:
- Screenshots:
- Logs:

deltachat-android/src/com/b44t/messenger/DcMsg.java 

 public static int[] msgSetToIds(final Set<DcMsg> dcMsgs) {
        int   cnt = dcMsgs==null? 0 : dcMsgs.size(); // If dcMsgs is null, cnt is set to zero. See below.
        int[] ids = new int[cnt];
        int   i = 0;
        for (DcMsg dcMsg : dcMsgs) { // dcMsgs == null?
            ids[i++] = dcMsg.getId();
        }
        return ids;
    }

adbenitez:
close #1967

### DIFF
--- a/src/com/b44t/messenger/DcMsg.java
+++ b/src/com/b44t/messenger/DcMsg.java
@@ -145,8 +145,9 @@ public class DcMsg {
 
     // aliases and higher-level tools
     public static int[] msgSetToIds(final Set<DcMsg> dcMsgs) {
-        if(dcMsgs==null)
+        if (dcMsgs == null) {
             return new int[0];
+        }
         int[] ids = new int[dcMsgs.size()];
         int   i = 0;
         for (DcMsg dcMsg : dcMsgs) {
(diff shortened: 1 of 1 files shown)


src/com/b44t/messenger/DcMsg.java
```diff
@@ -145,8 +145,9 @@ public File getFileAsFile() {
 
     // aliases and higher-level tools
     public static int[] msgSetToIds(final Set<DcMsg> dcMsgs) {
-        int   cnt = dcMsgs==null? 0 : dcMsgs.size();
-        int[] ids = new int[cnt];
+        if(dcMsgs==null)
+            return new int[0];
```
r10s, line 149: in general, it is a good idea to always surround code blocks by curly brackets:

```suggestion
        if (dcMsgs == null) {
            return new int[0];
        }
```

r10s (approved):
lgtm, apart from the minor code style comments.

thanks for taking care!

PR #1969, opened 2021-07-08
[WIP] allow to save images in webview

### LINKED ISSUE #1828: add option to get an image out of full-message-view
in (html-)full-message-view, you can select text with a long-tap and copy it to the clipboard.

but you currently cannot get out an image.

the issue is not super-urgent as most images are shown also in the normal chat, however, for the few that are not shown there, it would be nice to have an option beside screenshot".

i suggest to show a simple option-alert when long tapping an image with at least the option "export image" and "cancel".  might be that there is already something built-in in `WebView`, that does this job, maybe a bit differently - if so, i would prefer that. whatever is simple :)

adbenitez:
close #1828


r10s:
is this still wip?

adbenitez:
yes, I jumped to other stuff and left this unfinished, it doesn't implement the download part, I am not sure how to do it, ex. I could download it directly in background or use the Download Manager (maybe would need some extra permissions) with notification etc.

r10s:
closing stale pr, we can always come back  this attemt if needed

PR #1970, opened 2021-07-09
avoid code duplication, use Utils.writeTextToClipboard() everywhere

adbenitez:


src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -424,13 +423,8 @@ private void handleCopyMessage(final Set<DcMsg> dcMsgsSet) {
...
         if (result.length() > 0) {
-            try {
-                ClipboardManager clipboard = (ClipboardManager) getActivity().getSystemService(Context.CLIPBOARD_SERVICE);
-                clipboard.setText(result.toString());
-                Toast.makeText(getActivity(), getActivity().getResources().getString(R.string.copied_to_clipboard), Toast.LENGTH_LONG).show();
-            } catch (Exception e) {
-                e.printStackTrace();
-            }
+            Util.writeTextToClipboard(getActivity(), result.toString());
+            Toast.makeText(getActivity(), getActivity().getResources().getString(R.string.copied_to_clipboard), Toast.LENGTH_LONG).show();
```
Hocuri, line 427: This has the disadvantage that now we will show the "Copied to clipboard" toast unconditionally, even if copying failed. Everytime we call `writeTextToClipboard()`, we either show "Copied to clipboard" or "Done", so maybe we should let `writeTextToClipboard()` show the toast? (I think that "Copied to clipboard" would always fit)

src/org/thoughtcrime/securesms/util/Util.java
```diff
@@ -272,15 +272,24 @@ public static float clamp(float value, float min, float max) {
   }
 
   public static void writeTextToClipboard(@NonNull Context context, @NonNull String text) {
-    try {
-      ClipboardManager clipboardManager = (ClipboardManager)context.getSystemService(Context.CLIPBOARD_SERVICE);
-      clipboardManager.setPrimaryClip(ClipData.newPlainText(context.getString(R.string.app_name), text));
```
Hocuri, line 277: Didn't this code work on older devices?

src/org/thoughtcrime/securesms/ConversationFragment.java
adbenitez: > even if copying failed

in theory now the copy doesn't fail, if I get it right it could fail because incompatible API, but now `Util.writeTextToClipboard()` uses the right API based on android version

src/org/thoughtcrime/securesms/util/Util.java
adbenitez: > Didn't this code work on older devices?

seems it didn't, tbh I am not sure, it just was done in two different ways, one cared about android versions and didn't checked for errors and another used one way and checked for errors, so I just used one of them

Hocuri (approved):

Issue #1890, opened 2021-04-21
allow viewing eg. pdf from contact request


r10s:
this is about the contact request chat that is available at "menu / contact requests":

when a contact request message contains eg. a pdf attachment, it may be useful to view/handle this attachment as usual - esp. if we push using the app more as an email-client, this becomes more important, as there will usually be many contact requests, parts of them with attachments - and you may want to see them before finally starting a chat. maybe just reading them from the contact requests is sufficient.

also, as links and "show full messages" work in contacts requests as usual, we should not make a difference here. maybe also images could be enlargeable, but that is even less important.

but: we should take care not to lose the option to finally _accept_ a contact request - if the file/image/whatever covers the whole bubble and there is no tap possible to accept the contact request (currently, taps outside the bubbles work as well, maybe that is sufficient - most messages are not only file/image/whatever)

i just rested 1.14.5 - also there, this does not work, so, this is no new issue and no degradation and not super-important.

adbenitez:
not sure about  this, because would be hard to accept contact requests with images/files, and clicking in empty spaces to accept it is kinda of a hack (not easy to discover)

I would like at some point this contact requests could be reworked to be just another chat list, where you can see all contact requests as a single entry in the list, open the chat, see all messages from that contact and interact with documents and images like in a normal chat and have buttons at the end of the chat to accept, block etc. just like Signal and other apps do

adbenitez:
something like this: 

![deltachat-2021-04-21-171630](https://user-images.githubusercontent.com/24558636/115622221-8e9f6580-a2c5-11eb-8cb6-3118c3965587.jpg)

r10s:
yes, maybe. thanks for the image.

or maybe a button a "start chat" to the bubbles.

but, as this such changes should be done on all uis and as there is already some backlog, maybe do not add too much new ui for now.

not sure. but as said, i do not think that this is important just now, it just came into my mind when i noticed that [links do not work in ios contact requests at all](https://github.com/deltachat/deltachat-ios/issues/1169) :) 

however, as said, that is not super-important now, there are lower-hanging fruits and also more general issues for contact requests, https://github.com/deltachat/deltachat-core-rust/issues/2371 , trying to solve these "little things" first may also harm the broader view, so, i think we should close this issue for now and put it to [resurrection project](https://github.com/orgs/deltachat/projects/19)

r10s:
[...]

PR #1975, opened 2021-07-12
Remove Deaddrop chat (Android UI)

### LINKED ISSUE #1963: adapt to upcoming contact-request api
with https://github.com/deltachat/deltachat-core-rust/pull/2514 the contact requests get a major update with some api changes involved:

- there is no longer a virtual contact request chat that contains all contact request messages, apis and constants around that that "deaddrop" chat are removed

- instead, contact requests appear mostly as normal chats in the chatlist, there is no api action needed to let these chats appear. however, there should be a hint similar to "Archived", we can use "New" or "Request" for that (use `dc_chat_is_contact_request()` to check for that state) (the special background color we were using for the deaddrop can be removed)

- when a contact request chat is opened, it should be browsable as usual, no need to do anything special on tapping messages (this also removes some issues).  
however, instead of the input bar and the buttons "Block" and "Accept" must be shown. taps in them result in the new apis `dc_block_chat()` and `dc_accept_chat()` to be called.

core makes sure, no read-receipts are sent out and that the contact-requests bubble up and down in the chatlist as usual chats. 

see `deltachat.h` at https://github.com/deltachat/deltachat-core-rust/pull/2514/files for details and documentation about the api changes.

Hocuri:
See https://github.com/deltachat/deltachat-core-rust/pull/2514/

Fixes #1963
Fixes https://github.com/deltachat/deltachat-android/issues/1890

Screenshots:
<details>

image_2021-07-22_15-09-34
image_2021-07-22_15-09-07
image_2021-07-22_15-08-47

</details>

Before deleting a chat the user is asked for confirmation. Before blocking not, because blocking is not "destructive" (i.e. can be completely reverted by unblocking the contact). I'm wondering whether we should show a snackbar with an Undo button though, but we can probably wait for user feedback on this.


src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -614,7 +559,7 @@ public void handleEvent(DcEvent event) {
     if (event.getId() == DcContext.DC_EVENT_CONNECTIVITY_CHANGED) {
       ((ConversationListActivity) getActivity()).refreshTitle();
     } else {
-      getLoaderManager().restartLoader(0, null, this);
+      LoaderManager.getInstance(this).restartLoader(0, null, this);
```
Hocuri, line 562: Just to get rid of the deprecation warning

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -161,13 +158,15 @@ public void onResume() {
...
+    LoaderManager.getInstance(this).restartLoader(0, null, this);
 
     reloadTimer = new Timer();
     reloadTimer.scheduleAtFixedRate(new TimerTask() {
       @Override
       public void run() {
-        Util.runOnMain(() -> { list.getAdapter().notifyDataSetChanged(); });
+        Util.runOnMain(() ->
+                LoaderManager.getInstance(ConversationListFragment.this)
+                        .restartLoader(0, null, ConversationListFragment.this));
```
Hocuri, line 169: Without this change, after blocking a contact, the chatlist was not properly refreshed, i.e the chat didn't vanish from the chatlist. I didn't investigate why notifyDataSetChanged didn't do the job, I guess that it only reloads the elements of the list but doesn't check if items were added/removed.

I didn't measure this, but I guess this will make performance a little worse. Although this reloading is only needed when a contact was blocked, so we could think about remembering that we blocked a contact and only restartLoader() if yes.

jni/Application.mk
```diff
@@ -1,3 +1,3 @@
 APP_PLATFORM := android-16
-APP_ABI := armeabi-v7a arm64-v8a x86 x86_64
+APP_ABI := arm64-v8a
```
adbenitez, line 2: it seems this sneaked in while you was testing

adbenitez:
I don't like the "contact request" badge, it is too long and in case of a group "contact request" doesn't sounds right to me, IMHO the short options proposed in #1963 look better:

> we can use "New" or "Request"

I like "Request"

jni/Application.mk
Hocuri: Ooops, thanks for watching out!

src/org/thoughtcrime/securesms/ConversationListFragment.java
Hocuri: It doesn't get significantly slower (still around 300ms), but after loading the chatlist, but `onBindViewHolder()` is called again for all 10 or so view holders, which takes 250ms. However, this happens in background and there seemed not to be any negative consequences because of this.

So, all in all, should be fine.

link2xt:
I think the original plan was to have only two buttons: Accept and Decline/Block (it normally blocks chats, but groups are currently deleted so probably name it "delete" for now). No need to have a third "Delete" button for 1:1 chats, there is a delete option in the menu if you really want it.

res/drawable-night/message_request_button_background.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
```
adbenitez, line 1: this is forcing dark mode for the buttons if the system is in night mode, but app could be in "light" theme which should force light theme ignoring system theme

src/org/thoughtcrime/securesms/ConversationListFragment.java
Hocuri: It's not fine, @adbenitez reported:

"Now when I go back to the chat list I see briefly the chat with the new message badge and an old message as latest message, then it is updated"

Hocuri:
In 10ec25c7f I hope to have resolved https://github.com/deltachat/deltachat-android/pull/1975#discussion_r675972875: The chatlist is reloaded _synchronously_ before onResume() exits and the list of chats is shown to the user. We only do this after a contact request was declined, because otherwise returning to the conversations list would take 60ms longer then currently.

We don't use a Loader anymore as it started the loading at wrong times and was very hard to understand.

This also means that on first startup, we don't load the chatlist asynchronously on a background thread anymore now, but I think that this is fine as it takes only 60ms and we have to wait with showing the list until it loaded, anyway. If we want to load it asynchronously again, we can do so in a later PR, by simply putting it on a background thread ourselves, which is way easier than using a Loader by the way. And also, we can then start loading the chatlist earlier so that it is done loading at the time it's needed.

We don't call `notifyDataSetChanged()` in onResume() anymore, as the Android system calls this after onResume(), anyway.

res/drawable-night/message_request_button_background.xml
Hocuri: I'll figure it out if not, but do you know how to fix this?

adbenitez: adding an attribute in `attrs.xml` then in `themes.xml` add an item to dark and light themes each one using the appropriate drawable, the property could be called something like "message_request_button_bg" and then name the drawables `message_request_button_bg_light.xml` and `message_request_button_bg_dark.xml`

Hocuri: Nevermind, I couldn't wait and figured it out myself :joy: (did it like Signal)

Hocuri: This added 6ms to startup time, but I think that's OK, as we will also be able to use more `-night` views in the future

Hocuri: Wait, you had already answered, but I didn't see it. Anyway, I it's solved now, but in another way :shrug:

src/org/thoughtcrime/securesms/ConversationListFragment.java
adbenitez: seems fast now

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1653,4 +1656,41 @@ public boolean onQueryTextChange(String query) {
...
+      extras.putInt(ConversationListFragment.RELOAD_LIST, 1);
+      handleReturnToConversationList(extras);
+    });
+
+    String question;
+    if (dcChat.isMailingList()) {
+      question = getString(R.string.ask_show_mailing_list, dcChat.getName());
+    } else {
+      final int[] members = dcContext.getChatContacts(chatId);
+      if (members.length == 1) {
```
adbenitez, line 1684: doesn't this means that if someone creates a group with two members, send a message and then leaves the group, group will have only 1 member and the contact request will be treated as a 1:1 chat with block option, didn't try it yet but seems it is what will happen

adbenitez:
this closes #1890

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -153,21 +149,28 @@ public void onActivityCreated(Bundle bundle) {
...
     super.onResume();
 
     updateReminders();
-    list.getAdapter().notifyDataSetChanged();
+
+    if (getActivity().getIntent().getIntExtra(RELOAD_LIST, 0) == 1
+        && !chatlistJustLoaded) {
+      loadChatlist();
+    }
+    chatlistJustLoaded = false;
```
adbenitez, line 167: shouldn't `chatlistJustLoaded = true;` if the previous `loadChatlist();` got executed?

adbenitez (approved):

src/org/thoughtcrime/securesms/ApplicationPreferencesActivity.java
```diff
@@ -126,6 +126,7 @@ public boolean onSupportNavigateUp() {
   @Override
   public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
     if (key.equals(Prefs.THEME_PREF)) {
+      DynamicTheme.setDefaultDayNightMode(this);
```
r10s, line 129: was that always missing/wrong?

i am wondering what that has to do with contact requests :)

r10s (approved):
there are proably some minor tweaks needed, but they can go to a subsequent pr.

so, i suggest to merge that in to move forward (also to get translations to other os) and to have android-master work with core-master again.

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1676,19 +1676,22 @@ public void initializeContactRequest() {
...
     if (dcChat.isMailingList()) {
       question = getString(R.string.ask_show_mailing_list, dcChat.getName());
     } else {
-      final int[] members = dcContext.getChatContacts(chatId);
-      if (members.length == 1) {
-        DcContact dcContact = dcContext.getContact(members[0]);
-        question = getString(R.string.ask_start_chat_with, dcContact.getNameNAddr());
-      } else {
-        question = "";
+      if (dcChat.getType() == DcChat.DC_CHAT_TYPE_GROUP) {
```
adbenitez, line 1683: maybe more readable to use `dcChat.isGroup()`

src/org/thoughtcrime/securesms/ApplicationPreferencesActivity.java
adbenitez: a modern approach was used for the dark theme version of the buttons using `drawable-night` it is nicer than the current approach in other parts, but I hope that is compatible with old Android versions?

src/org/thoughtcrime/securesms/ConversationActivity.java
Hocuri: Although `isGroup()` actually does `dcChat.getType() == DcChat.DC_CHAT_TYPE_GROUP || dcChat.getType() == DcChat.DC_CHAT_TYPE_MAILINGLIST`, which is rather a hack and I didn't want to add more usages. We should probably make a PR that renames `isGroup()` to `isMultiuser()`, and then later introduce a method `isGroup()` that checks what the name says.

src/org/thoughtcrime/securesms/ApplicationPreferencesActivity.java
r10s: i did not tested it on android 4, however, i assume it is fine as these drawable-_someVersionOrConfig_ are usually just ignored by older androids.

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -153,21 +149,28 @@ public void onActivityCreated(Bundle bundle) {
...
 
     reloadTimer = new Timer();
     reloadTimer.scheduleAtFixedRate(new TimerTask() {
       @Override
       public void run() {
-        Util.runOnMain(() -> { list.getAdapter().notifyDataSetChanged(); });
+        Util.runOnMain(() -> loadChatlist());
```
r10s, line 173: this change is not correct, i think.

before, just the cells were redrawn, to update relative timestamps - with the change, the data are actually reloaded, which is not needed.

(coming over that when targeting https://github.com/deltachat/deltachat-android/issues/1994)

PR #1979, opened 2021-07-19
capitalize some strings

r10s:
as discussed several times on different channels,
nowadays, it is more usual to use capitalized titles;
fitting better to existing environments.

this pr capitalizes menu entries, buttons and titles;
"titles" include all titles of all settings.

whats left first-character-lowercase (just to have a clear focus):
- states
- values
- whole sentences
- titles mentioned in whole sentences

(nb: there is the rough idea of having something as "Group or Subject"
instead of just "Group" -
this change is also easier with the new capitalisation rules, cc @hpk42  ;)

as this all does not affect other languages
(they have their own rules),
we should be careful not to force re-translations.

cc @Simon-Laux, @Jikstra as we recently discussed that.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -4,7 +4,7 @@
     <string name="app_name">Delta Chat</string>
     <string name="ok">OK</string>
     <string name="cancel">Cancel</string>
-    <string name="clear_search">Clear search</string>
+    <string name="clear_search">Clear Search</string>
     <string name="yes">Yes</string>
     <string name="no">No</string>
     <string name="select">Select</string>
@@ -18,7 +18,7 @@
     <string name="automatic">Automatic</string>
     <string name="strict">Strict</string>
     <string name="open">Open</string>
-    <string name="open_attachment">Open attachment</string>
+    <string name="open_attachment">Open Attachment</string>
     <string name="join">Join</string>
     <string name="rejoin">Rejoin</string>
     <string name="delete">Delete</string>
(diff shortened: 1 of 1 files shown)


res/values/strings.xml
```diff
@@ -161,119 +161,119 @@
...
     <string name="mute_for_one_day">Mute for 1 day</string>
     <string name="mute_for_seven_days">Mute for 7 days</string>
     <string name="mute_forever">Mute forever</string>
 
-    <string name="share_location_once">once</string>
-    <string name="share_location_for_5_minutes">for 5 minutes</string>
-    <string name="share_location_for_30_minutes">for 30 minutes</string>
-    <string name="share_location_for_one_hour">for 1 hour</string>
-    <string name="share_location_for_two_hours">for 2 hours</string>
-    <string name="share_location_for_six_hours">for 6 hours</string>
```
r10s, line 258: that was also wrong by the old conventions.

Simon-Laux (approved):
lgtm

adbenitez (approved):
looks better that way

Hocuri:
I must have missed the previous discussion on different channels, but:

In WA I didn't find a single example of mid-string word-capitalisation.

Telegram does mid-string word-capitalisation, but not for menu entries, only settings and buttons.

In both WA and TG, all menu entries, like "Mute notifications", "Leave/Exit group", "Clear chat", "Add to contacts", were lowercase except for the first word.

So, it seems not to be _that_ common, and we may want to leave menu entries as-is.

r10s:
> In WA I didn't find a single example of mid-string word-capitalisation.

i just had a rough look at WA ios, in settings, it reads "Starred Messages", "Linked Devices", "Storage and Data" and so on.

so, seems they're doing differently on android.

telegram, signal writes "Notifications and Sounds", "Chat Settings" and so on on both, android, ios, desktop.

macos usues all-caps, windows seems to be mixed (everything is mixed in windows :)

libreoffice also uses all-caps, at least on ios, but i think, they also share localisations.

so, all in all, i have the feeling all-caps is a bit more spreaded, first-word-only-caps seems to be a bit more old school.

personally i also like first-word-only-caps a bit more (it was me starting with that :), however, i do not want to stand in the way of adaption to maybe wider usage, desktop brought that up several times, that was the reason for this pr (cc @Jikstra @Simon-Laux).

so, all in all i am +/-0 - however, it seems to help on some platforms (where menus are mixed out of system and app) - and on other it seems at least not to be unusual.

res/values/strings.xml
```diff
@@ -161,119 +161,119 @@
...
-    <string name="videochat_tap_to_join">Tap to join</string>
-    <string name="videochat_tap_to_open">Tap to open</string>
-    <string name="videochat_instance">Video chat instance</string>
-    <string name="videochat_instance_placeholder">Your video chat instance</string>
-    <string name="videochat_instance_explain">If a video chat instance is defined, you can start a video chat from each chat. Video chats require a compatible app or a compatible browser on both ends.\n\nExamples: https://meet.jit.si/$ROOM or basicwebrtc:https://your-server</string>
+    <string name="videochat_tap_to_join">Tap to Join</string>
+    <string name="videochat_tap_to_open">Tap to Open</string>
+    <string name="videochat_instance">Video Chat Instance</string>
+    <string name="videochat_instance_placeholder">Your Video Chat Instance</string>
+    <string name="videochat_instance_explain">If a video chat Instance is defined, you can start a video chat from each chat. video chats require a compatible app or a compatible browser on both ends.\n\nExamples: https://meet.jit.si/$ROOM or basicwebrtc:https://your-server</string>
```
Hocuri, line 276: "video" should probably stay uppercased as it's start-of-sentence?

And "Instance" should probably stay lowercased as it's mid-sentence?

Hocuri (approved):
>  telegram also seems to be mixed, they write "Invite Friends" in the main menu but "Clear history" in the chat menu.

yes, with "menu" I meant "three-dots menu" (which is the only menu Delta Chat has). In the hamburger menu = "main menu" Telegram has each-word-capitalized.

Anyway, seems like capitalizing all words in titles is the correct spelling: https://english.stackexchange.com/questions/14/which-words-in-a-title-should-be-capitalized. And e.g. Telegram writes "Invite Friends" and "Last Seen" but "Freunde einladen" and "Zuletzt gesehen" (German), so this should probably be done only in English.

r10s:
> so this should probably be done only in English

yip, all that is about English only. translations are not affected directly, they should do what is appropriate/convention in their language.

res/values/strings.xml
r10s: that's a typo that sneaked in, everywhere else we say "video chat instance" when inside a sentence.

PR #1980, opened 2021-07-21
add yggmail support: allow dotless email address and localhost server

adbenitez:
related core PR: https://github.com/deltachat/deltachat-core-rust/pull/2112
related issue: https://github.com/deltachat/deltachat-core-rust/issues/2540

### DIFF
--- a/src/org/thoughtcrime/securesms/RegistrationActivity.java
+++ b/src/org/thoughtcrime/securesms/RegistrationActivity.java
@@ -430,14 +430,10 @@ public class RegistrationActivity extends BaseActionBarActivity implements DcEve
         }
     }
 
-    private boolean matchesEmailPattern(String email) {
-        return !TextUtils.isEmpty(email) && Patterns.EMAIL_ADDRESS.matcher(email).matches();
-    }
-
     private void verifyEmail(TextInputEditText view) {
         String error = getString(R.string.login_error_mail);
         String email = view.getText().toString();
-        if (!matchesEmailPattern(email)) {
+        if (!DcHelper.getContext(this).mayBeValidAddr(email)) {
             view.setError(error);
         }
     }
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -500,7 +497,7 @@ private void onLogin() {
 
     private boolean verifyRequiredFields() {
         String email = emailInput.getText().toString();
-        return !email.isEmpty() && matchesEmailPattern(email)
```
adbenitez, line 503: `!email.isEmpty()` was always unnecessary here

src/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -447,7 +443,8 @@ private void verifyServer(TextInputEditText view) {
         String server = view.getText().toString();
         if (!TextUtils.isEmpty(server) && !Patterns.DOMAIN_NAME.matcher(server).matches()
                 && !Patterns.IP_ADDRESS.matcher(server).matches()
-                && !Patterns.WEB_URL.matcher(server).matches()) {
+                && !Patterns.WEB_URL.matcher(server).matches()
+                && !"localhost".equals(server)) {
```
r10s, line 447: that still does not allow dotless servernames beside `localhost`, right?

probably not needed for yggmail, however, maybe we should be more relaxed here in general.

r10s (approved):
lgtm

src/org/thoughtcrime/securesms/RegistrationActivity.java
adbenitez: > that still does not allow dotless servernames beside `localhost`, right?

that is only used to show a warning in the server fields, it doesn't prevent configuration, I just added "localhost" because it is a common name so the app don't warn in that case, but if you think we should not warn in any dotless domain let me know

PR #1985, opened 2021-07-22
use dc_accounts_t

### LINKED ISSUE #1964: adapt to dc_account_t api
to prepare account-migration, it is required that all UI use [`dc_accounts_t`](https://c.delta.chat/classdc__accounts__t.html) for handling multiple accounts (currently android handles multiple accounts somehow on its own).

as an improvement, that will let all accounts run together in the background, so that switching accounts will be much smoother and shared qr-codes will also work when an account is in the background.

in some future, we can think over adding notifications for background accounts, disabling accounts etc. however, doing that would require various adaptions (badges pointing to bg accounts with new messages etc) and discussions.

so, any ui change is out of scope of this issue and also not needed for AEAP.

cmp https://github.com/deltachat/deltachat-ios/pull/1286, https://github.com/deltachat/deltachat-desktop/issues/2296

r10s:
closes #1964, which also shows details and reasoning.

**CAVE:** for testing, esp. from inside a pr, keep in mind that there is no easy way to go back to a previous version without export/import!  

~moreover, [there is a known bug](https://github.com/deltachat/deltachat-core-rust/issues/2522) that makes account creation pretty weird (after adding an account, you have to restart the app) - before that bug is fixed, this pr should **not be merged**, however, of course, can be tested and reviewed before :)~

- [x] add `dc_accounts_t` api to jni (`DcAccounts` and `DcAccountsEventEmitter` in java land)
- [x] cleanup: get rid of `ApplicationDcContext` that is derived from `DcContext`; as there are no real data attached (lots of functions are put more or less random there), it is easier to work without as we then can use the objects directly as provided by `DcAccounts`
- [x] create `DcAccounts` object on program startup
- [x] migrate existing accounts
- [x] adapt `AccoutsManager` to use new `DcAccounts` api
- [x] make sure, only events of the selected account are processed

some remarks for cleanup: (that part was larger than expected :) 
- members from `ApplicationDcContext` are moved mainly to the already existing `DcHelper` and `DcEventCenter`.
- `DcHelper` got some helpful getters for accounts, eventCenter and notificationCenter. this results in lots of changes in the code, however, these one-lines should be easy to review.
- to the "context" mess: prior to this pr, dcContext also contains a reference to the application context; with this pr, dcContext is independent from application context

the ui is completely unchanged, however, of course, there are several ideas for improvements - but that is out of scope of this pr.


r10s:
i just rebased that to current master (that was just migrated to new contact-request-api)

src/org/thoughtcrime/securesms/connect/AccountManager.java
```diff
@@ -115,144 +50,77 @@ public static AccountManager getInstance() {
...
-            Log.w(TAG, err);
-            Toast.makeText(context, err, Toast.LENGTH_LONG).show();
-        }
+        new SwitchAccountAsyncTask(activity, R.string.switching_account, accounts.getSelectedAccount().getAccountId(), null).execute();
     }
 
 
     // helper class for switching accounts gracefully
 
     private static class SwitchAccountAsyncTask extends ProgressDialogAsyncTask<Void, Void, Void> {
```
r10s, line 105: the `SwitchAccountAsyncTask` might be superfluous now as account switching should be much faster, but i leave that in for now until we actually know that it is _always_ fast.

adbenitez (approved):

src/org/thoughtcrime/securesms/connect/AccountManager.java
```diff
@@ -29,80 +31,13 @@
...
         ApplicationContext appContext = (ApplicationContext)context.getApplicationContext();
-        appContext.dcContext.notificationCenter.removeAllNotifiations();
-        appContext.dcContext.stopIo();
-        appContext.dcContext.unref();
-        appContext.dcContext = new ApplicationDcContext(context);
-        appContext.dcContext.setStockTranslations();
+        DcHelper.getNotificationCenter(context).removeAllNotifiations();
+        appContext.dcContext = appContext.dcAccounts.getSelectedAccount();
+        appContext.notificationCenter = new NotificationCenter(context);
+        appContext.eventCenter = new DcEventCenter(context);
```
Hocuri, line 39: What does this line do? Remove all observers? Why do we need this?

src/org/thoughtcrime/securesms/connect/AccountManager.java
```diff
@@ -115,144 +50,77 @@ public static AccountManager getInstance() {
...
     }
 
 
     // helper class for switching accounts gracefully
 
     private static class SwitchAccountAsyncTask extends ProgressDialogAsyncTask<Void, Void, Void> {
         private final WeakReference<Activity> activityWeakReference;
-        private final @Nullable Account destAccount; // null creates a new account
-        private final @Nullable String deleteDbName;
+        private final @Nullable int destAccountId; // 0 creates a new account
```
Hocuri, line 107: `int` can't be nullable

src/org/thoughtcrime/securesms/connect/AccountManager.java
```diff
@@ -302,15 +172,14 @@ public void showSwitchAccountMenu(Activity activity) {
                 .setSingleChoiceItems(menu.toArray(new String[menu.size()]), presel, (dialog, which) -> {
```
Hocuri, line 172:
```suggestion
                .setSingleChoiceItems(menu.toArray(new String[0]), presel, (dialog, which) -> {
```

See https://stackoverflow.com/questions/53284214/toarray-with-pre-sized-array

src/org/thoughtcrime/securesms/connect/AccountManager.java
```diff
@@ -319,32 +188,34 @@ public void showSwitchAccountMenu(Activity activity) {
...
         ArrayList<String> menu = new ArrayList<>();
-        for (AccountManager.Account account : accounts) {
-            menu.add(account.getDescr());
+        for (int accountId : accountIds) {
+            menu.add(accounts.getAccount(accountId).getNameNAddr());
         }
         int[] selection = {-1};
         new AlertDialog.Builder(activity)
                 .setTitle(R.string.delete_account)
                 .setSingleChoiceItems(menu.toArray(new String[menu.size()]), -1, (dialog, which) -> selection[0] = which)
```
Hocuri, line 202:
```suggestion
                .setSingleChoiceItems(menu.toArray(new String[0]), -1, (dialog, which) -> selection[0] = which)
```

Hocuri (commented):
Read about half the code, didn't test.

src/com/b44t/messenger/DcContext.java
```diff
@@ -188,6 +202,24 @@ public void unref() {
...
+      String ret = "";
+
+      if (!displayname.isEmpty() && !addr.isEmpty()) {
+        ret = String.format("%s (%s)", displayname, addr);
+      } else if (!addr.isEmpty()) {
+        ret = addr;
+      }
+
+      if (ret.isEmpty() || isConfigured() == 0) {
+        ret += " (not configured)";
```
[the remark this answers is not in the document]
r10s: well, it was not translated before (the string was build in AccountManager), and i did not want to change ui-related things in this pr. also, usually, this string should not appear at all, it is more a sign that sth. is wrong.

src/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -61,7 +74,27 @@ public void onCreate() {
...
+      DcAccountsEventEmitter emitter = dcAccounts.getEventEmitter();
+      while (true) {
+        DcEvent event = emitter.getNextEvent();
+        if (event==null) {
+          break;
+        }
+        eventCenter.handleEvent(event);
+      }
+      Log.i("DeltaChat", "shutting down event handler");
+    }, "eventThread").start();
```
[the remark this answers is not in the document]
r10s: the DcEventCenter is recreated on switches, so, that would need adaption as well. better leave it here for now.

src/org/thoughtcrime/securesms/connect/AccountManager.java
r10s: removing all observers is fine as also all activities are discarded. maybe it is not needed and the EventCenter and other objects can be persisted and longer-living, however, changing that is not in scope of this pr, i did not want to change these things and have a pretty narrow scope (the objets were recreated also previously)

src/org/thoughtcrime/securesms/connect/AccountManager.java
r10s: that looks weird to me ... but well, java is weird.

Hocuri (approved):
code lgtm, didn't test

PR #1989, opened 2021-07-25
Fix format string in language "km"

Hocuri:
ask_forward takes a string, not a number


adbenitez (approved):

r10s:
good catch, however, it needs to be changed on transifex, then it will go to the repo on the next `./scripts/tx-pull-translations.sh` roundtrip. i will change it there.

Issue #1990, opened 2021-07-26
contact request tweaks


r10s:
- [x] is the "Chat with ..." line really required? all information are already shown on the screen - it is probably from the old api where we have shown an alter without much context. also, it pop up for mailinglist and one-to-ones only, not for groups. on ios, it was never there and it feels intuitive, at least from my point of view. maybe just drop it and see how it works out.

- [ ] while the accept/block buttons look nice, ui-wise, they do not really fit into the rest of the app nor to material design (they look like ios). maybe use buttons similar to the rest of the app (as in the welcome screen, the fab-button). not a major thing, however, maybe it is also good to play around a bit longer.

- [ ] there was some possible slowness @Hocuri is checking out, cmp. #1994

- [x] contact requests should not be shown in "bold" as for fresh messages

- [x] ~~do we need a badge counter for "Request" - it was added for ios but not for android, cmp https://github.com/deltachat/deltachat-ios/issues/1306 - i do not have a strong opinion on that, however, maybe it is easier to drop it on ios.~~
EDIT: that part is solved, also ios skips the badge now

- [x] ~~wrt wording of "Accept": we already have the string "Start chat" - maybe that is even better, and it is more clear what finally happen (came over that as i do not like the german translations "Akzeptieren" or "Annehmen" - or what else can we use?)~~  
EDIT: let's leave that as is for now and see how that works. it is also the wording used in signal.

- [x] done in core: on upgrade, move old/noticed contact requests to "archived chats" - otherwise, after upgrade ppl may be an otherwise clean chatlist spammed with hundreds of pretty old chats, see https://github.com/deltachat/deltachat-core-rust/issues/2562

- [x] also changed in core: remove badge counter from "Archived Chats" cell, see https://github.com/deltachat/deltachat-core-rust/pull/2566 for reasoning

- [x] add "Archived Chats" to menu

Hocuri:
>  is the "Chat with ..." line really required? all information are already shown on the screen - it is probably from the old api where we have shown an alter without much context. also, it pop up for mailinglist and one-to-ones only, not for groups. on ios, it was never there and it feels intuitive, at least from my point of view. maybe just drop it and see how it works out.

[...]

PR #1991, opened 2021-07-28
tweak 'Archived Chats' and  'Requests'

r10s:
- [x] add "Archived Chats" menu entry the place where the 'Contact Requests' were before -
that way, ppl may find existing contact requests easier.  
moreover, it is now possible to access 'Archived Chats'
without endless scrolling ;)
- [x] add a hint as "If you archive chats, they will be shown here." in case there are no archived chats (also before, the list could be empty, however, typically, the activity was not opened then)
- [x] allow "Archived" and "Request" label in combination (maybe one can do two badge views at some point, in another pr, however, for now, this is good enough; it is not so much in focus anyway.
- [x] do not show requests in bold

targets some points of #1990 - the other points require some more playing around and can be done in another pr.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationListItem.java
@@ -245,7 +245,11 @@ public class ConversationListItem extends RelativeLayout
     if (visibility==DcChat.DC_CHAT_VISIBILITY_ARCHIVED)
     {
       badgeView.setVisibility(View.VISIBLE);
-      badgeView.setText(R.string.chat_archived_label);
+      if (isContactRequest) {
+        badgeView.setText(getContext().getString(R.string.chat_archived_label) + "  " + getContext().getString(R.string.chat_request_label));
+      } else {
+        badgeView.setText(R.string.chat_archived_label);
+      }
       deliveryStatusIndicator.setNone();
       unreadIndicator.setVisibility(View.GONE);
     }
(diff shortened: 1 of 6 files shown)


src/org/thoughtcrime/securesms/ConversationListItem.java
```diff
@@ -245,7 +245,11 @@ private void setStatusIcons(int visibility, int state, int unreadCount, boolean
     if (visibility==DcChat.DC_CHAT_VISIBILITY_ARCHIVED)
     {
       badgeView.setVisibility(View.VISIBLE);
-      badgeView.setText(R.string.chat_archived_label);
+      if (isContactRequest) {
+        badgeView.setText(getContext().getString(R.string.chat_archived_label) + " " + getContext().getString(R.string.chat_request_label));
```
r10s, line 249: maybe one can do two badge views at some point, however, for now, this is good enough. it is not so much in focus anyway - and saves one view per item :)

adbenitez (approved):
the combined "archived request" text could not be fine in other languages (ex. in Spanish it is not good), but separating them with a comma or `/` would probably make de badge look ugly or just be as confusing :)

r10s:
> the combined "archived request" text could not be fine in other languages

yip, i also think, we should do a follow up and create two separate views for it, witch a separate recrangle each (as on ios)

i also like to have [Archive] on the right then. @adbenitez can you file an issue for that and maybe even implement it?

adbenitez:
>  this is how it looks on ios (however, i [suggest](https://github.com/deltachat/deltachat-ios/issues/1320) to swap "Request" and "Archive" badges)

nice, will try to do it then

r10s:
> nice, will try to do it then

great!

PR #1996, opened 2021-07-29
hide question from request panel for now

r10s:
if needed we can do something more specific in the future.

(in did not changed the layout to avoid shifting around constraints)

targets #1990


adbenitez (approved):
using it already, looks nice +1

PR #1997, opened 2021-07-30
Separate request and archive badges

### LINKED ISSUE #1995: Archived request badge issues
archived contact requests should show both badges, currently both badges are combined in one which produce a text that doesn't make sense in some languages, also, iOS already show both badges separated, see screenshot at https://github.com/deltachat/deltachat-android/pull/1991#issuecomment-888906372

adbenitez:
close #1995

### DIFF
--- a/res/layout/conversation_list_item_view.xml
+++ b/res/layout/conversation_list_item_view.xml
@@ -88,10 +88,24 @@
             <LinearLayout
                 android:orientation="horizontal"
                 android:minHeight="21dp"
+                android:showDividers="middle"
+                android:divider="@drawable/badge_divider"
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content">
 
-                <TextView android:id="@+id/badge"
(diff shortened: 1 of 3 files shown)


res/layout/conversation_list_item_view.xml
```diff
@@ -88,10 +88,24 @@
             <LinearLayout
                 android:orientation="horizontal"
                 android:minHeight="21dp"
+                android:showDividers="middle"
+                android:divider="@drawable/badge_divider"
```
r10s, line 92: this is an elegant solution! i was not aware that `android:divider` this exist :)

i just double checked https://developer.android.com/reference/android/widget/LinearLayout - dividers were revamped in API 11, however, seems to exist forever - so, in any case, no issue with out [API 16](https://github.com/deltachat/deltachat-android/pull/1227) :)

r10s (approved):
lgtm, much nicer and clearer than before!

PR #2000, opened 2021-07-30
fix chatlist anr

### LINKED ISSUE #1994: new requests: chatlist updates lead to anr
on recent master, i am getting anr (app not responding, ui frozen) when many messages are downloaded (eg. after getting online). i am pretty sure that this does not happen that often/badly before.

i assume, that this is because by #1975, `getChatlist()` is always called in the main thread and `DcChatlistLoader` was dropped.

there is a hint to the removed loader at https://github.com/deltachat/deltachat-android/pull/1975#issuecomment-886103166 - and i think, it makes sense to not use it on opening activity - but for lots of events received (as on updates), a background thread still seems to be advantageous.

when targeting that: i am not totally sure if `DcChatlistLoader` did actually _debounce_ things, but when fixing that issue, we can maybe do this, similar to https://github.com/deltachat/deltachat-android/pull/1884

not sure if that is a blocker, by using dc_accounts_t, this will be mitigated sometimes, however, i think, we should target that with some priority.

r10s:
fixes #1994 and, as an extra, adds debouncing. chatlist loading should be pretty smart now.

the technique for the debouncing with two booleans is taken from #1884 - however, the first update is not delayed.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationListFragment.java
@@ -400,6 +400,11 @@ public class ConversationListFragment extends Fragment
           loadChatlist();
         }
 
+        // in theory, here is a potential race condition:
+        // if the background thread is exactly here and the main thread checks `inLoadChatlist` at that moment,
+        // one update will be missing.
+        // that can probably be fixed with some other multi-theading techniques, any suggestions welcome :)
+
         inLoadChatlist = false;
       });
     }
--- a/src/org/thoughtcrime/securesms/ConversationListFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationListFragment.java
@@ -177,7 +177,7 @@ public class ConversationListFragment extends Fragment
     reloadTimer.scheduleAtFixedRate(new TimerTask() {
       @Override
       public void run() {
-        Util.runOnMain(() -> loadChatlist());
+        Util.runOnMain(() -> { list.getAdapter().notifyDataSetChanged(); });
       }
     }, 60 * 1000, 60 * 1000);
   }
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -177,7 +178,7 @@ public void onResume() {
     reloadTimer.scheduleAtFixedRate(new TimerTask() {
       @Override
       public void run() {
-        Util.runOnMain(() -> loadChatlist());
+        Util.runOnMain(() -> { list.getAdapter().notifyDataSetChanged(); });
```
r10s, line 181: this just reverts a wrong change, see https://github.com/deltachat/deltachat-android/pull/1975#pullrequestreview-719498036

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -379,7 +380,32 @@ private void handleCreateConversation(int chatId) {
...
+        needsAnotherLoad = false;
+        loadChatlist();
+
+        while (needsAnotherLoad) {
+          Util.sleep(100);
+          needsAnotherLoad = false;
+          Log.i(TAG, "executing debounced chatlist loading");
+          loadChatlist();
+        }
+
```
r10s, line 402: in theory, here is a potential race condition: if the background thread is exactly here and the main thread checks `inLoadChatlist` at that moment, one update will be missing. that can probably be fixed with some other multi-theading techniques, any suggestions welcome :) however, in practice, these boolean may be pretty fine, for incremental search, we're doing the same, and i have not seen a race condition there in practise (would be easy noticable as the last typed character is not searched for)

adbenitez (approved):
besides the potential race condition, lgtm, the chat list works much better now

r10s:
k, then let's merge that in. i have a rought idea how the race can be targeted, but that can be done/discussed in a separate pr.

PR #2001, opened 2021-07-31
fix potential race condition in chatlist loading

r10s:
this should fix the race condition mentioned at #2000 (and in the source code)
by synchronizing getters/setters of the booleans.

also, this unifies code to one loop, nothing special is done anymore for the first load; the 100ms sleep is always done after loading, so, first loading is still instant.

@link2xt as you are much more experienced with that - do you still see a race condition here? or was it okay even before?

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationListFragment.java
@@ -380,34 +380,34 @@ public class ConversationListFragment extends Fragment
     ((ConversationSelectedListener)getActivity()).onCreateConversation(chatId);
   }
 
+  private final Object loadChatlistLock = new Object();
   private boolean inLoadChatlist;
   private boolean needsAnotherLoad;
   private void loadChatlistAsync() {
-    if (inLoadChatlist) {
+    synchronized (loadChatlistLock) {
       needsAnotherLoad = true;
-      Log.i(TAG, "chatlist loading debounced");
-    } else {
+      if (inLoadChatlist) {
+        Log.i(TAG, "chatlist loading debounced");
+        return;
+      }
       inLoadChatlist = true;
-      Util.runOnAnyBackgroundThread(() -> {
+    }
 
-        needsAnotherLoad = false;
-        loadChatlist();
-
-        while (needsAnotherLoad) {
-          Util.sleep(100);
+    Util.runOnAnyBackgroundThread(() -> {
+      while(true) {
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -380,34 +380,34 @@ private void handleCreateConversation(int chatId) {
     ((ConversationSelectedListener)getActivity()).onCreateConversation(chatId);
   }
 
+  private final Object loadChatlistLock = new Object();
   private boolean inLoadChatlist;
   private boolean needsAnotherLoad;
   private void loadChatlistAsync() {
-    if (inLoadChatlist) {
+    synchronized (loadChatlistLock) {
```
link2xt, line 387: If the chatlist is loading at this moment in another thread, we will have to wait for it to finish before entering this `synchronized` instead of simply setting `needsAnotherLoad = true` and returning.

adbenitez: the lock is not acquired while `loadChatlist()` is executing, only in some vars assignments and checking and that should be fast, or maybe I didn't understand something :sweat_smile:

adbenitez (approved):
lgtm

link2xt (changes requested):
<del>I don't see the point of debouncing if you have to wait for `loadChatlist` (and following sleep!) just to set `needsAnotherLoad = true`.</del> Overlooked that the lock does not cover `loadChatlist`.

src/org/thoughtcrime/securesms/ConversationListFragment.java
link2xt: Yes, lock is not acquired, so your thread is waiting at this line. Instead, it should set `needsAnotherLoad = true` and return immediately.

adbenitez: I meant that lock can be acquired here while `loadChatlist()` is executing

link2xt: I overlooked that `synchronized` does not span over the `loadChatlist`

link2xt (approved):

r10s:
nb: the race condition is not so potential-only - it became real, on deleting chats - without that pr, it happend several times to me that the chatlist misses the last update, honestly because of a weird update behavior, but still. this was fixed by this pr (noticed that when using a mobile that did not had this fix yet ;)

PR #2003, opened 2021-08-01
update archived chatlist when contact request is blocked

### LINKED ISSUE #2002: Archived chatlist not updated when an archived contact request is blocked
the issue is not related to the new contact requests, it also happens with normal archived chats, so was an old bug it seems

adbenitez:
close #2002


r10s (approved):
lgtm (open/close by accident).

thanks a lot for figuring that out!

PR #2004, opened 2021-08-01
avoid showing compose panel when pressing block button

### LINKED ISSUE #1999: new requests: flashing input bar after blocking
after hitting "block" (not: "delete"), the input bar appears for a little moment, for a fraction of a second.

reason is probably, that, by calling block(), the chat is no longer a contact request and, by some event, initializeContactRequest() is called and the input bar is shown - as isContactRequest() is false.

it becomes more visible by fixing #1994, as the app is less blocked in main thread - but the bug is unrelated and also on current master.

adbenitez:
stop observing `DC_EVENT_CONTACTS_CHANGED` events before blocking chat so the compose panel is not made visible in `handleEvent()`

close #1999

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -1668,9 +1668,10 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
       messageRequestBottomView.setVisibility(View.GONE);
       composePanel.setVisibility(View.VISIBLE);
     });
-    DcEventCenter eventCenter = DcHelper.getEventCenter(this);
     messageRequestBottomView.setBlockOnClickListener(v -> {
-      eventCenter.removeObserver(DcContext.DC_EVENT_CONTACTS_CHANGED, this);
+      // avoid showing compose panel on receiving DC_EVENT_CONTACTS_CHANGED for the chat that is no longer a request after blocking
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1668,7 +1668,9 @@ public void initializeContactRequest() {
       messageRequestBottomView.setVisibility(View.GONE);
       composePanel.setVisibility(View.VISIBLE);
     });
+    DcEventCenter eventCenter = DcHelper.getEventCenter(this);
     messageRequestBottomView.setBlockOnClickListener(v -> {
+      eventCenter.removeObserver(DcContext.DC_EVENT_CONTACTS_CHANGED, this);
```
r10s, line 1673: even if it may be okay here, in general, i think, it is better not to keep unneeded strong references, so maybe better `DcHelper.getEventCenter(this).removeObserver(DcContext.DC_EVENT_CONTACTS_CHANGED, this);`.

i would also add a comment, sth. as "// avoid showing compose panel on receiving DC_EVENT_CONTACTS_CHANGED for the chat that is no longer a request after blocking"

r10s (approved):
lgtm, interesting, but simple approach!

(i would have checked the state somehow and leave the event - but this is nice as well, and probably even more performant)

PR #2022, opened 2021-08-09
Add simple benchmark & a test framework

Hocuri:
See #2019, #2020 and #2021 for failed attempts at using a benchmark framework. So, for now I'll just go with timing the results manually. 

For trying this out:
- Backup your current account first, maybe there are some bugs in switching accounts.
- You can run benchmarks on either an emulated device or a real device. For better benchmark results, you should run the benchmark on a real device and make sure that the core is compiled in release mode.
- Disable animations on your device, otherwise the test may fail. (Developer options -> Window animation scale, Transition animation scale & Animatior duration scale -> set all three of them to 0x)
- In Android Studio: File -> Sync project with gradle files
- In Android Studio: Run -> Edit configurations -> `+` -> Android Instrumented test -> Either select a specific class or select "All in Module" -> OK -> Select your configuration in the toolbar -> Click on the green "run" button in the toolbar to run the tests

When the benchmark is done, you will get a result like `MEASURED RESULTS (Benchmark) - Going thorough all 10 chats: 11635,11207,11363,11352,11279,11183,11137,11145,11032,11057`, Paste `11635,11207,11363,11352,11279,11183,11137,11145,11032,11057` into a cell in a LibreOffice spreadsheet, do `Data -> Text to columns`, choose `,` as a separator, hit `OK`, and create a diagram:

Screenshot_2021-08-09_12-25-12

You can see that the benchmarks are getting faster over time, probably because of cache and/or branch prediction.


src/org/thoughtcrime/securesms/util/ViewUtil.java
```diff
@@ -137,6 +137,9 @@ public static void fadeIn(final @NonNull View view, final int duration) {
     final SettableFuture future = new SettableFuture();
     if (view.getVisibility() == visibility) {
       future.set(true);
+    } else if (AccessibilityUtil.areAnimationsDisabled(view.getContext())) {
+      view.setVisibility(visibility);
+      future.set(true);
```
r10s, line 142: this fixes a bug that the animation of the send-button wasn't disabled with system animations are disabled.

r10s (approved):

PR #2027, opened 2021-08-11
reload adapter's dcChat when chat is modified

### LINKED ISSUE #1992: can't quote-reply message after accepting a contact request
swipe-to-reply and the reply menu action are hidden in a contact request chat, but after accepting the contact request, the input panel appears but swipe-to-reply and reply menu action are not enabled

adbenitez:
close #1992


adbenitez:
btw, I don't know why `this.dcContext` is not used instead of loading a new context object in https://github.com/deltachat/deltachat-android/pull/2027/files#diff-87aba4c3108902a337014a0669a18c71c15ace6cb5447ac413559d03307cba6dL481

Hocuri (approved):

PR #2031, opened 2021-08-12
Add very basic onboarding and sharing test

Hocuri:
When fixing https://github.com/deltachat/deltachat-android/issues/2040, I'll try to create a failing test first. This PR here is the foundation for this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,12 +174,26 @@ environment.
   Either select a specific class or select "All in Module" / "OK" /
   Select your configuration in the toolbar / Click on the green "run" button in the toolbar to run the tests
 
+## Get the benchmark results
+
 When the benchmark is done, you will get a result like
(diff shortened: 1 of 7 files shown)


Hocuri:
@adbenitez @r10s this has to be merged before #2065, please review :)

adbenitez (approved):
overall it looks good, but I can't test this at the moment, I am a newbie in android unit testing :D

Hocuri:
Apparently you can also run the tests from the command line, but I didn't try this yet: https://developer.android.com/studio/test/command-line. And it the Readme there is an explanation how to do it using Android Studio, please tell me if sth doesn't work or is incomprehensible.

I'll go ahead and merge this, as nobody depends on these tests, it should be uncritical (the only thing that could theoretically create problems are the changes to build.gradle, but only while building, not for end-users, and I think the only way to find out is merge and wait - should be OK though)

PR #2034, opened 2021-08-12
Fix "Can't edit shared image"

### LINKED ISSUE #2032: Can't edit shared image
- Delta Chat version: current master (the bug exists since quite some time already though)
  - Share an image from another app to Delta Chat
  - Select a chat
  - Tap on the pen to edit the image
  - edit the image and hit confirm
- Expected: The edited image should be shown
- Actual: The edit is discarded

Hocuri:
Fix #2032

The problem was that after returning from the image editor, `doReinitializeDraft` was true and therefore we the draft was reinitialized to the originally shared, unedited image.

This was introduced in https://github.com/deltachat/deltachat-android/pull/1770. I im pretty sure that what I changed this time is fine, but I also was sure when implementing #1770…. We had so many regressions in the past in this area that we should thoroughly test this change before merging (and I'm really looking forward to UI tests in this area).

I extended
https://github.com/deltachat/interface/blob/master/user-testing/mailto-links.md
a bit; before merging this PR we should test that everything there still
works.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -429,6 +429,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
       break;
     case ScribbleActivity.SCRIBBLE_REQUEST_CODE:
       setMedia(data.getData(), MediaType.IMAGE);
+      doReinitializeDraft = false;
       break;
     case SMS_DEFAULT:
       initializeSecurity(isSecureText, isDefaultSms);


adbenitez (approved):
works like a charm \o/

Hocuri:
I tested this, and didn't find any problems, so let's merge it. Also, the change only affects what happens when the user comes back from the scribble activity, so I think it's unlikely that this makes things worse (it can't be that much worse than losing your edits at this point, and I think it's unlikely that the image will completely vanish or so).

Hocuri:
Ping @r10s

Hocuri:
This does introduce a very small and rare bug, but I think that's fine and I actually don't think we should try and fix this, because chances seem to be good that we would introduce other bugs:

- Select an image to attach
- Click "Edit image"
- If you want to, make some edits, but it's not necessary.
- In DC select Saved Messages. A draft text will be created.
- Leave DC
- Open DC again from the "Recent apps". You will be back in the scribble activity.
- Hit the ckeckmark in the upper right corner.

Expected: The draft text and the edited image are there.
Actual (after this PR): Only the edited image is there, the draft text is gone.
Actual (before this PR): The draft text and the _unedited_ image are there. Which IMHO was worse though, because the last thing the user did was confirming their edits in the scribble activity. And also, it happened not only with these special steps to reproduce but much more often (everytime you tried to edit a shared image).

r10s:
thanks for the detailed write up!

i also think, that is okay. it is a bit weird and unclear anyway how things get mixed when the user shares "parts together", so, i think, it is fine if the "last action" is taken - esp. as these are rare case and changing behavior may add worse bugs :)

PR #2039, opened 2021-08-20
profile: show tab containing 'status' for device chat

r10s:
the device chat's status contains some useful information
since https://github.com/deltachat/deltachat-core-rust/pull/2613 -
unfortunately, the whole tab was hidden on android.

reminder for myself, again: things are never easy, esp. not the superficial changes ;)

ios and desktop should be fine in this regard.


r10s:
merged that in quickly to move forward, comments still welcome.

PR #2045, opened 2021-08-29
do not download large messages automatically

r10s:
this pr allows to not download large messages automatically

- by default, all messages are downloaded automatically, at "Settings / Chats and Media" is an option to set a limit.

- if a message exceeding the limit is received then, the message is assigned to the correct chat. however, instead of the message, just a placeholder text is shown (that should probably be tweaked a bit)

- the "Download" button then replaces the message with the real content.

- the pr requires the core branch [`draft-dl-ffi2`](https://github.com/deltachat/deltachat-core-rust/pull/2631) to be checked out

- there are tons of ideas how a preview can be achieved, and how media types may be detected, however, all of that is out of scope of this pr ;)

so, for testing, things should be fine, however, there are some open questions in the core, if they're solved. as a first step, maybe we can just add a limit for nauta and try out things there, cc @adbenitez 

some impressions:

  

https://user-images.githubusercontent.com/9800740/131254310-626dbead-fa38-4fb0-aa82-d92b2ee62dcc.mov

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -828,6 +828,11 @@ public class ConversationFragment extends MessageSelectorFragment
         startActivity(intent);
         getActivity().overridePendingTransition(R.anim.slide_from_right, R.anim.fade_scale_out);
       }
+
+      @Override
+      public void onDownloadClicked(DcMsg messageRecord) {
+        dcContext.downloadFullMsg(messageRecord.getId());
+      }
     }
 
     @Override
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -638,6 +638,12 @@ JNIEXPORT jstring Java_com_b44t_messenger_DcContext_getMsgHtml(JNIEnv *env, jobj
 }
 
 
+JNIEXPORT void Java_com_b44t_messenger_DcContext_downloadFullMsg(JNIEnv *env, jobject obj, jint msg_id)
+{
+    dc_download_full_msg(get_dc_context(env, obj), msg_id);
+}
+
+
 JNIEXPORT void Java_com_b44t_messenger_DcContext_deleteMsgs(JNIEnv *env, jobject obj, jintArray msg_ids)
 {
     int msg_ids_cnt = 0;
@@ -1386,6 +1392,12 @@ JNIEXPORT jint Java_com_b44t_messenger_DcMsg_getType(JNIEnv *env, jobject obj)
 JNIEXPORT jint Java_com_b44t_messenger_DcMsg_getState(JNIEnv *env, jobject obj)
 {
     return dc_msg_get_state(get_dc_msg(env, obj));
+}
+
+
+JNIEXPORT jint Java_com_b44t_messenger_DcMsg_getDownloadState(JNIEnv *env, jobject obj)
+{
+    return dc_msg_get_download_state(get_dc_msg(env, obj));
 }
 
 
--- a/res/layout/conversation_item_received.xml
+++ b/res/layout/conversation_item_received.xml
@@ -158,7 +158,7 @@
                 tools:text="Mango pickle lorem ipsum"/>
 
             <Button
-                android:id="@+id/show_full_message"
+                android:id="@+id/msg_action_button"
                 android:visibility="gone"
                 style="@style/Signal.Text.Body"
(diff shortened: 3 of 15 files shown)


r10s:
@adbenitez i think, these 64 KiB for "most worse quality images" are not correct. i did some tests images, all of them are above 40 KiB, several of them 100 KiB, net sizes.

regarding gross sizes, so adding headers and base64 encoding, sth. as 160 KiB sounds more reasonable for "most worse quality images". 

rough checks for "balanced", also here 640 KiB seems to be more reasonable to really catch "most". 

so, maybe the limits 40 KiB / 160 KiB / 640 KiB / 4 MiB / 16 MiB seems to be more practical.

r10s:
> Should the big message be packaged in some way like this so that the other party can confirm some summary of the message before downloading it?

without additional effort, for _encrypted_ messages, we just do not know anything about the message before downloading it completely - so a "generic" fallback as above is needed anyway - and this is where we start with this pr :)

even with unencrypted messages, things are tricky.  
ahh, and also a preview would be nice, sth. as https://blurha.sh

however, there are plans and ideas to target all of that, however, this is something for subsequent prs as that is quite some work and we unfortunately have only very limited development resources.

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -828,6 +828,11 @@ public void onShowFullClicked(DcMsg messageRecord) {
         startActivity(intent);
         getActivity().overridePendingTransition(R.anim.slide_from_right, R.anim.fade_scale_out);
       }
+
+      @Override
+      public void onDownloadClicked(DcMsg messageRecord) {
+        dcContext.downloadFullMsg(messageRecord.getId());
```
adbenitez, line 834: if user tap the download button several times, will core ignore calls to `downloadFullMsg` if the download is already ongoing? or should `onDownloadClicked` ignore calls while it is already downloading? 
(offtopic: I have issues with this buttons I am never sure if tapping them works because they are styled like links without any state change on taps, so if the download failed quickly I am not sure if it failed or I missed the button)

adbenitez (approved):

src/org/thoughtcrime/securesms/ConversationFragment.java
r10s: multiple calls to downloadFullMsg() should be ingored by core.

once the download is in progress, the ui showns the usual tiny loading animation, however, it would be nice to also disable the download button to reflect that state.

at some point, a "cancel download" appearing instead would maybe be nice, however, we just do not have that functionality core-side currently - and it is also not that easy iirc.

r10s: i tweaked the button a bit, it gets disabled now and shows "Downloading..." at that time. would be nice to have that as grey, however, that needs tweaking of some xml; i did not dived into that.

adbenitez: it is much better now +1

adbenitez:
btw, classic email with subject shows the subject + the `[xKiB message] ...`, it might look a bit weird, on the other hand is useful to see the subject before downloading

r10s:
> btw, classic email with subject shows the subject + the [xKiB message] ..., it might look a bit weird, on the other hand is useful to see the subject before downloading

you mean when replying to a not-yet-downloaded message?

otherwise, the [xKiB message] is not sent over the wire, it is used as the text/summary only locally until the message is fully downloaded.

r10s:
apart from that, i think, we can merge the pr in, tiny things can be tweaked later as needed.

adbenitez:
no, I mean like if some mailing list send you a big HTML message, and you have prevent-download enabled, and they send the email with subject "Our feed for you this week", you will see the message like:

```
Our feed for you this week - [xKiB message] ...
```

adbenitez:
> apart from that, i think, we can merge the pr in, tiny things can be tweaked later as needed.

Yes I also think it is fine to merge, actually it is a core thing, and I should have commented in the core PR that was merged anyway already, and I don't think it is a blocker

r10s:
> no, I mean like if some mailing list send you a big HTML message, and you have prevent-download enabled, and they send the email with subject "Our feed for you this week", you will see the message like:
> Our feed for you this week - [xKiB message] ...

ah, i see. tbh, did not thought much about that case - but i think, if we have the subject, it is fine to display that :)

PR #2047, opened 2021-09-05
use loadDataWithBaseURL() instead of loadData()

adbenitez:
fix issue with "jump to section" links


adbenitez:
one thing I don't understand well is that `loadDataWithBaseURL` documentation says:

```
Note that content specified in this way can access local device files (via 'file' scheme URLs) only if baseUrl specifies a scheme other than 'http', 'https', 'ftp', 'ftps', 'about' or 'javascript'.
```

maybe that is related to why "data" scheme was used, but well "data" scheme is not mentioned there so maybe using "data" is as good as "file"

PR #2048, opened 2021-09-05
allow to export all selected attachments

adbenitez:
when selecting multiple messages that contain attachments, for example documents in the "Documents" tab, it was not possible to export all the selected attachments at the same time, user had to select & export one by one, this PR fixes this, allowing to export all selected messages if all of them have attachments

### DIFF
--- a/res/menu/conversation_context.xml
+++ b/res/menu/conversation_context.xml
@@ -30,7 +30,7 @@
         android:icon="?menu_forward_icon"
         app:showAsAction="always" />
 
-    <item android:title="@string/menu_export_attachment"
+    <item android:title="@string/menu_export_attachments"
         android:id="@+id/menu_context_save_attachment"
         android:visible="false"
         app:showAsAction="never" />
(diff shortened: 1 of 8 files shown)


adbenitez:
maybe also the string `R.string.ask_export_attachment` needs to be tweaked, for example, removing the start of the message "Export attachment?" or replacing with something like "Export attachment(s)?"

r10s:
great!

> maybe also the string R.string.ask_export_attachment needs to be tweaked, for example, removing the start of the message "Export attachment?" or replacing with something like "Export attachment(s)?"

i'd add a new string "Export attachment**s**" (this is similar to delete and forward then) and remove the  start of the message "Export attachment?" as you suggested.

adbenitez:
>     * for the menu entry, i'd add a new string "Export attachment**s**" (this is similar to delete and forward then)

I checked and it seems that "Forward message" is always used even if several messages are selected (not noticeable since a forward icon is used instead of the text

adbenitez:
I tackled the missing parts

r10s (approved):

r10s:
works like a charm, thanks a lot for taking care!

PR #2050, opened 2021-09-16
pull in translatable/lasting strings from desktop

r10s:
pull in most recent strings from https://github.com/deltachat/deltachat-desktop/blob/master/_locales/_untranslated_en.json

i left out two strings that are hard to translate/explain and more debugging/testing information and will not last in that form anyway (and are probably outdated before translated significantly :)

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -450,6 +450,11 @@
     <string name="login_smtp_port">SMTP Port</string>
     <string name="login_smtp_security">SMTP Security</string>
     <string name="login_auth_method">Authorization Method</string>
+    <string name="login_socks5_use_socks5">Use SOCKS5</string>
+    <string name="login_socks5_socks5_host">SOCKS5 Host</string>
+    <string name="login_socks5_socks5_port">SOCKS5 Port</string>
+    <string name="login_socks5_socks5_user">SOCKS5 User</string>
+    <string name="login_socks5_socks5_user">SOCKS5 Password</string>
     <!-- Translators: %1$s will be replaced by an e-mail address -->
     <string name="login_oauth2_checking_addr">Checking %1$s</string>
     <string name="login_info_oauth2_title">Continue with simplified setup?</string>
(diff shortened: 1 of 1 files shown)


Simon-Laux:
note: `account_info_hover_tooltip_desktop` is already in the translation set, it only changed as now the account path is more irrelevant it now shows the account id. But sure might make sense to further discuss what should be in that hover info and what not before always changing the string.

res/values/strings.xml
```diff
@@ -450,6 +450,13 @@
     <string name="login_smtp_port">SMTP Port</string>
     <string name="login_smtp_security">SMTP Security</string>
     <string name="login_auth_method">Authorization Method</string>
+    <!-- the word "SOCKS5" here and in the following strings should not be translated in most cases -->
+    <string name="login_socks5">SOCKS5</string>
+    <string name="login_socks5_use_socks5">Use SOCKS5</string>
+    <string name="login_socks5_socks5_host">SOCKS5 Host</string>
+    <string name="login_socks5_socks5_port">SOCKS5 Port</string>
+    <string name="login_socks5_socks5_user">SOCKS5 User</string>
+    <string name="login_socks5_socks5_password">SOCKS5 Password</string>
```
Simon-Laux, line 459:
```suggestion
    <string name="login_socks5_host">SOCKS5 Host</string>
    <string name="login_socks5_port">SOCKS5 Port</string>
    <string name="login_socks5_user">SOCKS5 User</string>
    <string name="login_socks5_password">SOCKS5 Password</string>
```

Simon-Laux:
would have been useful to describe what you changed, so desktop has it easier when pulling in the new translations.

- did not update `account_info_hover_tooltip_desktop`
- did not add `login_socks5_experimental_warning`
- add  `sync_all`, `login_socks5`, `login_socks5_use_socks5`, `login_socks5_host`, `login_socks5_port`, `login_socks5_password`, `unconfigured_account`
- add and change key
   - `login_socks5_login` -> `login_socks5_user`
   - `unconfigured_account_subtitle` -> `unconfigured_account_hint`

r10s:
account-id is debugging-info, nothing that needs to be translated.

Simon-Laux (approved):

PR #2059, opened 2021-09-24
do not say, device-messages can be deleted from the server somehow

### LINKED ISSUE #2058: Question to confirm deletion of device messages "on the server"
- Android version:
Android 11.

- Device:
Fairphone FP2 (Lineage OS 18.1 build RQ3A.210705.001 from 2021-09-17).

- Delta Chat version:
1.23.0 (latest nightly build).

- Expected behavior:
When a user is about to delete a device message, the question should be `Delete 1 message?` only.

- Actual behavior:
Instead, the question is `Delete 1 message here and on the server?`.

-- Open the "Device Messages" chat.
-- Long-tap any message so it is marked/selected.
-- Tap on the rubbish bin icon.

- Screenshots:
N/A. (It is obvious.)

- Logs:
N/A.

- Remark:
This does not occur on Delta Chat for Desktop.

r10s:
it is a bit nitpicking and also a cornercase,
however, yes, it was not correct before.

another option would be to just strike the "... and on your server" which may lead to confusion also on other circumstances - eg. when auto-delete is set to "at once", there is nothing on the server. however, that may raise other issues, we added the "... and on your server" for some reasons, so, i would not change that without further investigations :) - therefore just this pr :)

closes #2058

### DIFF
--- a/src/org/thoughtcrime/securesms/MediaPreviewActivity.java
+++ b/src/org/thoughtcrime/securesms/MediaPreviewActivity.java
@@ -53,6 +53,7 @@ import androidx.loader.content.Loader;
 import androidx.viewpager.widget.PagerAdapter;
 import androidx.viewpager.widget.ViewPager;
 
+import com.b44t.messenger.DcChat;
 import com.b44t.messenger.DcContext;
 import com.b44t.messenger.DcMediaGalleryElement;
 import com.b44t.messenger.DcMsg;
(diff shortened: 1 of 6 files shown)


src/org/thoughtcrime/securesms/MediaPreviewActivity.java
```diff
@@ -342,17 +343,21 @@ private void deleteMedia() {
...
     AlertDialog.Builder builder = new AlertDialog.Builder(this);
-    builder.setMessage(getResources().getQuantityString(R.plurals.ask_delete_messages, 1, 1));
+    builder.setMessage(text);
     builder.setCancelable(true);
 
     builder.setPositiveButton(R.string.delete, (dialogInterface, which) -> {
       new AsyncTask<Void, Void, Void>() {
         @Override
         protected Void doInBackground(Void... voids) {
-          if (mediaItem.msgId == DcMsg.DC_MSG_NO_ID) {
```
r10s, line 353: this is not needed as it was already checked before.

adbenitez (approved):

PR #2061, opened 2021-09-27
Disable group editing actions if user can't edit group (ex. self not in group)

### LINKED ISSUE #2016: Deleting members from group not disabled for contact requests
in the profile of new groups contact requests, it is possible to select some group member and press the "trash" icon to delete them, it seems deleting doesn't work until the contact request is accepted so the remove-member action should be removed from menu in group contact requests

adbenitez:
close #2016

### DIFF
--- a/src/org/thoughtcrime/securesms/ProfileSettingsFragment.java
+++ b/src/org/thoughtcrime/securesms/ProfileSettingsFragment.java
@@ -208,9 +208,6 @@ public class ProfileSettingsFragment extends Fragment
 
   public void onAddMember() {
     DcChat dcChat = dcContext.getChat(chatId);
-    if (!dcChat.canSend()) {
-      return;
-    }
     Intent intent = new Intent(getContext(), ContactMultiSelectionActivity.class);
     intent.putExtra(ContactSelectionListFragment.SELECT_VERIFIED_EXTRA, dcChat.isProtected());
     ArrayList<String> preselectedContacts = new ArrayList<>();
@@ -223,10 +220,6 @@ public class ProfileSettingsFragment extends Fragment
   }
 
   public void onQrInvite() {
-    DcChat dcChat = dcContext.getChat(chatId);
-    if (!dcChat.canSend()) {
-      return;
-    }
     Intent qrIntent = new Intent(getContext(), QrShowActivity.class);
     qrIntent.putExtra(QrShowActivity.CHAT_ID, chatId);
     startActivity(qrIntent);
--- a/src/org/thoughtcrime/securesms/ProfileSettingsFragment.java
+++ b/src/org/thoughtcrime/securesms/ProfileSettingsFragment.java
@@ -205,6 +205,9 @@ public class ProfileSettingsFragment extends Fragment
 
   public void onAddMember() {
     DcChat dcChat = dcContext.getChat(chatId);
+    if (!dcChat.canSend()) {
+      return;
+    }
     Intent intent = new Intent(getContext(), ContactMultiSelectionActivity.class);
     intent.putExtra(ContactSelectionListFragment.SELECT_VERIFIED_EXTRA, dcChat.isProtected());
     ArrayList<String> preselectedContacts = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


adbenitez:
pending: hide "edit" menu option in profile when chat is a contact request or you are no longer member of the group, currently the name is changed even if the contact request is still not accepted, but no "name changed" message is sent, so group status diverge.

I will wait for the broadcast PR to be merged since it does changes in `onCreateOptionsMenu`

src/org/thoughtcrime/securesms/ProfileSettingsFragment.java
```diff
@@ -248,8 +258,7 @@ private void onNewChat() {
     @Override
     public boolean onCreateActionMode(ActionMode mode, Menu menu) {
       mode.getMenuInflater().inflate(R.menu.profile_context, menu);
-      DcChat dcChat = dcContext.getChat(chatId);
-      menu.findItem(R.id.delete).setVisible(!dcChat.isMailingList());
```
adbenitez, line 252: this is not necessary since the member list is not even shown in a mailing list's profile so if we are inside `onCreateActionMode` this can not be a mailing list

adbenitez:
instead of just "disabling" the add/remove group members actions, a toast could be shown, but then that is more strings to translate and I don't know what to say that could be generic enough to use when it is a contact request or when self is not in group, etc.

if you want that, please suggest what to say

r10s:
> instead of just "disabling" the add/remove group members actions, a toast could be shown, but then that is more strings to translate and I don't know what to say that could be generic enough to use when it is a contact request or when self is not in group, etc.

maybe just hide the options instead of showing "Add Members" and "QR Invite Code" buttons that are not working and have no visual clue that they are disabled.  
(hiding would be done by not adding the buttons in `ProfileSettingsAdapter.changeData()`)

(disabling vs. hiding is a tricky ux topic in general; eg. disabled fields give extra thought for the user about why it is not available - which may be wanted. however, in this case, i'd say hiding is also easier and also totally reasonable - also we already hide other things as the "edit" button.

we should also keep in mind, that `canSend()` will more often change when we finally have "instant group joins" as of https://github.com/deltachat/deltachat-core-rust/pull/2508 - on changes, however, as the profile view alrady listens to  DC_EVENT_CHAT_MODIFIED and updates things accordingly, that should be fine, at least as a first glance :)

Hocuri:
+1 for hiding unavailable group actions

src/org/thoughtcrime/securesms/ProfileSettingsFragment.java
```diff
@@ -217,6 +223,10 @@ public void onAddMember() {
   }
 
   public void onQrInvite() {
+    DcChat dcChat = dcContext.getChat(chatId);
+    if (!dcChat.canSend()) {
+      return;
+    }
```
Hocuri, line 229: Same here, as a user I find buttons that don't do anything when clicked pretty frustrating

src/org/thoughtcrime/securesms/ProfileSettingsFragment.java
```diff
@@ -205,6 +208,9 @@ else if(contactId>DcContact.DC_CONTACT_ID_LAST_SPECIAL) {
 
   public void onAddMember() {
     DcChat dcChat = dcContext.getChat(chatId);
+    if (!dcChat.canSend()) {
+      return;
+    }
```
Hocuri, line 213: I think we should hide this button, too, instead of just doing nothing when it's clicked

Hocuri (commented):
Nice! I commented two things that should be changed I think.

@r10s I think that there was some misunderstanding: This PR hides the menu actions, it doesn't just make them grey and unclickable.

Off-topic for this PR: Maybe it would be nice to actually have a function `can_edit()` in the core that checks `is_self_in_chat() && !is_device_talk() && !is_contact_request() && !is_self_talk()`

src/org/thoughtcrime/securesms/ProfileSettingsFragment.java
r10s: yip, this is the button i meant above :)

Hocuri: Ah, now I understand what you meant, so the misunderstanding was only on my side :)

adbenitez:
ok, I also think it is better to just hide the buttons, done :)

r10s (approved):

PR #2064, opened 2021-10-01
Fix back button not working in connectivity view sometimes

Hocuri:
I noticed that sometimes the system back button (the one in the lower
left) does not work in the connectivity view. The top bar back button
worked.

This should fix this.

Also, I moved `getSupportActionBar().setTitle(R.string.connectivity)` to onCreate(). It had been in refresh() so that the title could be set to "Connectivity | Connected", "Connectivity | Connecting" and so on. But as we decided not to do this, there is no reason for it to be in refresh().


r10s (approved):

r10s:
did not test, but code and explanations are reasonable :)

adbenitez (approved):

PR #2071, opened 2021-10-05
allow to select audio files in multi-select mode

### LINKED ISSUE #2069: Audio files can't be selected in multi-selection mode in Docs tab
in the Docs tab, if you long-press some files to multi-select, then try to select an audio file, it doesn't work, instead it is possible to move the seek bar and play the audio

adbenitez:
close #2069


src/org/thoughtcrime/securesms/ProfileDocumentsFragment.java
```diff
@@ -160,7 +160,6 @@ private void handleMediaPreviewClick(@NonNull DcMsg dcMsg) {
   public void onMediaLongClicked(DcMsg mediaRecord) {
     if (actionMode == null) {
       ((ProfileDocumentsAdapter) recyclerView.getAdapter()).toggleSelection(mediaRecord);
-      recyclerView.getAdapter().notifyDataSetChanged();
```
adbenitez, line 163: this seems unnecessary, `notifyDataSetChanged()` is called inside `toggleSelection()` called in the previous line

PR #2077, opened 2021-10-10
improve the account selector dialog

adbenitez:
The goal is to somewhat converge with DC Desktop's account selector.

- [x] improve the account selector's style, showing account avatar, etc.
- [x] ~~hide from the list the current selected account since user can't switch to it nor delete it.~~
- [x] allow to delete account directly from the account selector (X button at the right of each account in the list)
- [x] show badge with the unread/fresh message count for each account
- [x] instead of showing the avatar fallback for "me", show the avatar other contacts will see when you write to them. 

**[OUTDATED]** This is how it looks currently:

deltachat-2021-10-10-003031


adbenitez:
so can we merge this? #2075 depends on this

PR #2079, opened 2021-10-13
Remove "View Profile" and "Add Attachment" from menu

Hocuri:
The menu has grown waaaaay to big, before it has been up to 8 entries,
now it's two entries less.


webratte:
I never realised this menu items.😂
So yes, I also think it's not needed.
There are the paperclip icon. And the tap on the username is also intuitive 👍🏻

PR #2081, opened 2021-10-14
disable fullscreen keyboard

### LINKED ISSUE #2068: Full screen input field (landscape mode) doesn't addapt to dark theme

adbenitez:
close #2068


r10s:
i would not be surprised if that removal break workflows of some users with tiny displays or watches. (i am not suggesting to revert things, however, we'll see :) (i just came over that while writing the changelog)

PR #2082, opened 2021-10-14
don't add members if back button is pressed

### LINKED ISSUE #2067: Members added to group if you mark them for addition but press back button
pressing back (phone button) is doing the same as the :heavy_check_mark: button (confirmation), it should instead do the same action as the cancel button :x:

this is unexpected, since the back button usually means cancel action, users might be added to a group accidentally

adbenitez:
close #2067

### DIFF
--- a/src/org/thoughtcrime/securesms/ContactMultiSelectionActivity.java
+++ b/src/org/thoughtcrime/securesms/ContactMultiSelectionActivity.java
@@ -73,12 +73,6 @@ public class ContactMultiSelectionActivity extends ContactSelectionActivity {
     return false;
   }
 
-  @Override
-  public void onBackPressed() {
-    saveSelection();
-    super.onBackPressed();
-  }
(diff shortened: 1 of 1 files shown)


Hocuri:
> accidentally sending addition messages and disclosing email address of some contacts in a group

Yes, that's a point, there should be _some_ confirmation before sending a message out. I still think that the back button should generally not silently abort anything that's full-screen[1] both because it says so at https://material.io/components/dialogs#full-screen-dialog (at "Confirmation") and because it usually works this way on Android [2] Same goes for the profile editing activity. It's not _that_ important for me though and I already have lots of things on my to-do list, so I won't do that right now, maybe you can do it or maybe I'll do it later.

[1]  it's common (and therefore OK) for the back button to abort things in a non-full-screen dialog, though
[2] in our settings, the users don't have to confirm their actions, either

PR #2084, opened 2021-10-15
select previous account when rolling back account creation

### LINKED ISSUE #2006: after canceling account-creation, always first account is selected
as, from the view of the core,  the "selected account" is the one that is in creation. if creation is cancelled, the selected account gets deleted - and core selected the first account.

to fix that, easier thing will be to remember the selected account id on welcome-entering - and select that account after deleting the account in creation.

this issue is true probably on all uis. 

@link2xt or can we handle that in core by a remembering the account that was active before selection was changed? that would make things also nicer if in the future, we have a list of accounts and one of them is deleted - we would then switch back to the previously selected. i do not think, that we need and endless stack, ui-wise, the "selected before selection" would do the job.  
just asking, handling that in the ui would be okay as well :)

adbenitez:
close #2006


adbenitez:
yeah, some extra care won't hurt :D

PR #2088, opened 2021-10-19
provider app support

r10s:
corresponding yggmail pr: https://github.com/deltachat/AndroidYggmail/pull/15

the goal is to make the interaction as smooth as possible, at best without confirmation, questions etc. the flow is roughly the following:

- start the "provider app", yggmail here, configure account there (just tap "start")

if delta chat is already installed (doesn't need to be running):

- (X) the provider apps sends the new account by a Message object to Delta Chat

if delta chat is not installed:
- the provider app guides the user to downloading Delta Chat (eg. open play store after tapping "start")

if provider app is installed (doesn't need to be running) and delta chat is installed but not yet configured:
- deltachat, when starting unconfigured, sends a broadcast event to possible provider apps
- (X) the provider apps sends the new account by a Message object to Delta Chat

step (X) is identical on both scenarios.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -294,6 +294,16 @@
 
     <service android:name=".service.GenericForegroundService" />
 
+    <service
+        android:name=".service.IPCAddAccountsService"
+        android:enabled="true"
+        android:exported="true"
+        >
+        <intent-filter>
+            <action android:name="chat.delta.addaccount" />
+        </intent-filter>
+    </service>
+
     <receiver android:name=".notifications.MarkReadReceiver"
               android:enabled="true"
               android:exported="false">
--- a/src/org/thoughtcrime/securesms/RegistrationActivity.java
+++ b/src/org/thoughtcrime/securesms/RegistrationActivity.java
@@ -185,6 +185,11 @@ public class RegistrationActivity extends BaseActionBarActivity implements DcEve
             emailInput.setText(emailAddress);
             passwordInput.setText(password);
             onLogin();
+          } else {
+            String errorText = "Companion app auto-configuration failed.";
+            errorText += TextUtils.isEmpty(emailAddress) ? " Missing emailAddress." : "";
+            errorText += TextUtils.isEmpty(password) ? " Missing password." : "";
+            Toast.makeText(this, errorText, Toast.LENGTH_LONG).show();
           }
         }
 
--- a/src/org/thoughtcrime/securesms/RegistrationActivity.java
+++ b/src/org/thoughtcrime/securesms/RegistrationActivity.java
@@ -59,6 +59,7 @@ import static org.thoughtcrime.securesms.connect.DcHelper.CONFIG_SEND_SERVER;
 import static org.thoughtcrime.securesms.connect.DcHelper.CONFIG_SEND_USER;
 import static org.thoughtcrime.securesms.connect.DcHelper.CONFIG_SERVER_FLAGS;
 import static org.thoughtcrime.securesms.connect.DcHelper.getContext;
+import static org.thoughtcrime.securesms.service.IPCAddAccountsService.ACCOUNT_DATA;
 
 public class RegistrationActivity extends BaseActionBarActivity implements DcEventCenter.DcEventDelegate {
 
@@ -175,6 +176,17 @@ public class RegistrationActivity extends BaseActionBarActivity implements DcEve
 
             int certCheckFlags = DcHelper.getInt(this, "imap_certificate_checks");
             certCheck.setSelection(certCheckFlags);
+        } else if (getIntent() != null && getIntent().getBundleExtra(ACCOUNT_DATA) != null) {
+          // Companion app might have sent account data
+          Bundle b = getIntent().getBundleExtra(ACCOUNT_DATA);
+          String emailAddress = b.getString(CONFIG_ADDRESS);
+          String password = b.getString(CONFIG_MAIL_PASSWORD);
+          if (!TextUtils.isEmpty(emailAddress) && !TextUtils.isEmpty(password)) {
+            emailInput.setText(emailAddress);
+            passwordInput.setText(password);
+            updateProviderInfo();
+            onLogin();
+          }
         }
 
         DcHelper.getEventCenter(this).addObserver(DcContext.DC_EVENT_CONFIGURE_PROGRESS, this);
--- a/src/org/thoughtcrime/securesms/RegistrationActivity.java
+++ b/src/org/thoughtcrime/securesms/RegistrationActivity.java
@@ -184,7 +184,6 @@ public class RegistrationActivity extends BaseActionBarActivity implements DcEve
           if (!TextUtils.isEmpty(emailAddress) && !TextUtils.isEmpty(password)) {
             emailInput.setText(emailAddress);
             passwordInput.setText(password);
-            updateProviderInfo();
             onLogin();
           }
         }
--- a/src/org/thoughtcrime/securesms/WelcomeActivity.java
+++ b/src/org/thoughtcrime/securesms/WelcomeActivity.java
@@ -84,7 +84,7 @@ public class WelcomeActivity extends BaseActionBarActivity implements DcEventCen
           // Since android API 26 only explicit broadcasts are allowed for IPC with a few exceptions.
           // As a result we have to send for each companion app we want to support an intent with a
           // specified package
-          intent.setPackage("org.cyberta");
+          intent.setPackage("chat.delta.AndroidYggmail");
           intent.setAction(DC_REQUEST_ACCOUNT_DATA);
           sendBroadcast(intent);
         }
--- a/src/org/thoughtcrime/securesms/WelcomeActivity.java
+++ b/src/org/thoughtcrime/securesms/WelcomeActivity.java
@@ -52,6 +52,7 @@ public class WelcomeActivity extends BaseActionBarActivity implements DcEventCen
     public static final int PICK_BACKUP = 20574;
     private final static String TAG = WelcomeActivity.class.getSimpleName();
     public static final String TMP_BACKUP_FILE = "tmp-backup-file";
+    public static final String DC_REQUEST_ACCOUNT_DATA = "chat.delta.DC_REQUEST_ACCOUNT_DATA";
 
     private boolean manualConfigure = true; // false: configure by QR account creation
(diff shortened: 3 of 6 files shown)


Simon-Laux:
when done, documentation about how to use the api in a markdown file would be nice, but probably only makes sense when the api is a bit settled.

src/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -175,6 +176,17 @@ public void onTextChanged(CharSequence s, int start, int before, int count) { }
...
             certCheck.setSelection(certCheckFlags);
+        } else if (getIntent() != null && getIntent().getBundleExtra(ACCOUNT_DATA) != null) {
+          // Companion app might have sent account data
+          Bundle b = getIntent().getBundleExtra(ACCOUNT_DATA);
+          String emailAddress = b.getString(CONFIG_ADDRESS);
+          String password = b.getString(CONFIG_MAIL_PASSWORD);
+          if (!TextUtils.isEmpty(emailAddress) && !TextUtils.isEmpty(password)) {
+            emailInput.setText(emailAddress);
+            passwordInput.setText(password);
+            updateProviderInfo();
```
cyBerta, line 187: It seems to work also without that line and that would reduce the visual noise a little bit. I'm not sure of all consequences though. @r10s maybe you can have a closer look here?

r10s: you are taking about the single statement `updateProviderInfo()`?

i had a look, it should be okay to not call it, it only updates the ui and should not affect login.

src/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -175,6 +176,17 @@ public void onTextChanged(CharSequence s, int start, int before, int count) { }
...
+            onLogin();
+          }
```
r10s, line 189: if email address and password are not set, we should show this as a toast or so to the user.

no need to have it translatable, it should happen to programmers using the Intent only, however, some hint would be great esp. then :)

src/org/thoughtcrime/securesms/RegistrationActivity.java
cyBerta: yes that line.
I would leave that out then. the yellow info box disappears way too fast to read it's content when yggmail is configured automatically.

r10s: yip, better leave out the call to `updateProviderInfo()` then.

src/org/thoughtcrime/securesms/WelcomeActivity.java
```diff
@@ -60,6 +61,16 @@ public void onCreate(Bundle bundle) {
         DcEventCenter eventCenter = DcHelper.getEventCenter(this);
         eventCenter.addObserver(DcContext.DC_EVENT_CONFIGURE_PROGRESS, this);
         eventCenter.addObserver(DcContext.DC_EVENT_IMEX_PROGRESS, this);
+
+        if (!DcHelper.hasAnyConfiguredContext(this)) {
+          Intent intent = new Intent();
+          // Since android API 26 only explicit broadcasts are allowed for IPC with a few exceptions.
+          // As a result we have to send for each companion app we want to support an intent with a
+          // specified package
+          intent.setPackage("org.cyberta");
```
cyBerta, line 70: I was also thinking about renaming the package structure in AndroidYggmail. I wonder if it's appId should rather be chat.delta.AndroidYggmail.

src/org/thoughtcrime/securesms/WelcomeActivity.java
```diff
@@ -60,6 +61,16 @@ public void onCreate(Bundle bundle) {
         DcEventCenter eventCenter = DcHelper.getEventCenter(this);
         eventCenter.addObserver(DcContext.DC_EVENT_CONFIGURE_PROGRESS, this);
         eventCenter.addObserver(DcContext.DC_EVENT_IMEX_PROGRESS, this);
+
+        if (!DcHelper.hasAnyConfiguredContext(this)) {
+          Intent intent = new Intent();
+          // Since android API 26 only explicit broadcasts are allowed for IPC with a few exceptions.
```
cyBerta, line 67: Btw. a workaround for the need to define the package names of all companion apps we support could be to use a share intent. It has some interesting implications, eg. if the user *would* have installed several companion apps, the user could manually select which one to choose for the account setup. 
Broadcast receivers listening for share intents can also be registered in the AndroidManifest. When registered there, the app can be started from the system - in contrast to broadcast receivers that are registered on runtime. 
However I didn't went this path because the share intent is not inteded to be used like that (from my point of view) and it felt more hackish.

cyBerta:
note: this mechanism is not yet built for severeal companion apps installed on the same device

AndroidManifest.xml
```diff
@@ -294,6 +294,16 @@
 
     <service android:name=".service.GenericForegroundService" />
 
+    <service
+        android:name=".service.IPCAddAccountsService"
+        android:enabled="true"
+        android:exported="true"
+        >
```
Hocuri, line 301: We could think about requiring a custom permission for this, like, `android:permission="..."`. See https://developer.android.com/guide/topics/permissions/defining

This would e.g. have the advantage that a malicious app that somehow exploits the companion-app-system has to declare this permission directly in the manifest so that everyone can quickly see it, while otherwise it could do it with some obfuscated code.

src/org/thoughtcrime/securesms/WelcomeActivity.java
Hocuri: Sounds a lot nicer to me than hardcoding the list of companion apps. And letting the user manually select which companion app to use for the account setup does sound like the way to go.

With "share intent", do you mean "an intent with ACTION_SEND"? Or "an intent with ACTION_VIEW"?

/me also wondering why it's called "Broadcast" if you can only send it to one explicitly-given app, but anyway...

src/org/thoughtcrime/securesms/WelcomeActivity.java
Hocuri: I think I'd slightly prefer `chat.delta.AndroidYggmail`.

src/org/thoughtcrime/securesms/service/IPCAddAccountsService.java
```diff
@@ -0,0 +1,93 @@
...
+import static org.thoughtcrime.securesms.connect.DcHelper.CONFIG_ADDRESS;
+
+import com.b44t.messenger.DcAccounts;
+import com.b44t.messenger.DcContext;
+
+/**
+ * This service is invoked by companion apps aiming to add a new account to Delta Chat
+ */
+
+public class IPCAddAccountsService extends Service {
```
Hocuri, line 35: IPC means "Inter package communication"? Maye this should be commented somewhere?

Hocuri (commented):
I feel like I commented lots of stuff that may be unimportant, so feel free to ignore me where you have a different opinion.

src/org/thoughtcrime/securesms/service/IPCAddAccountsService.java
r10s: for share vs. broadcast: i would be fine with both as long as share is comparable fast and is also invisible to the user: so not UI must pop up in case there is _no_ companion app (which is probably true in >99% of cases currently) or only _one_ companion app. 

as we already that broadcasts work that way and this method is also already a little bit tested, maybe we should stay with that for now, also as we want to get this pr in soon.

if there are _really_ multiple companion apps in the future, we can think that over and also change that. so, for the list of ids where we want to broadcast to, i would not care too much about that just now.

src/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -175,6 +176,21 @@ public void onTextChanged(CharSequence s, int start, int before, int count) { }
...
             int certCheckFlags = DcHelper.getInt(this, "imap_certificate_checks");
             certCheck.setSelection(certCheckFlags);
+        } else if (getIntent() != null && getIntent().getBundleExtra(ACCOUNT_DATA) != null) {
+          // Companion app might have sent account data
+          Bundle b = getIntent().getBundleExtra(ACCOUNT_DATA);
+          String emailAddress = b.getString(CONFIG_ADDRESS);
+          String password = b.getString(CONFIG_MAIL_PASSWORD);
+          if (!TextUtils.isEmpty(emailAddress) && !TextUtils.isEmpty(password)) {
+            emailInput.setText(emailAddress);
+            passwordInput.setText(password);
```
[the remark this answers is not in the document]
r10s: i agree, as both, email and password are set together as provided by the provider app, this sounds not very realistic. if log in fails, will the user "just" enter another password that is unrelated? i doubt so.

but, if we stay with the whitelisting - can we maybe check the sender app-id as well? that would probably make all imaginable attacks a lot harder.

src/org/thoughtcrime/securesms/WelcomeActivity.java
r10s: i commented on that "share alternative" at https://github.com/deltachat/deltachat-android/pull/2088#discussion_r746072280 - in short, i would stay with broadcasts as well for now.

AndroidManifest.xml
cyBerta: That might be sth. for the future, yes. I would leave that out for now, just to avoid one more potential error vector while testing the initial implementation.

Hocuri: Note though that if we do it in the future, it will be a breaking change for any third-party apps and it's possible that we won't do it because we don't want to break other people's working companion apps.

Hocuri (approved):
I'm a bit confused about the current status of this PR as it's still marked as a draft even though it got two reviews already. But anyway, I'll approve so that you can merge it if you want.

src/org/thoughtcrime/securesms/RegistrationActivity.java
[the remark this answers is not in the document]
cyBerta: > but, if we stay with the whitelisting - can we maybe check the sender app-id as well? that would probably make all imaginable attacks a lot harder.

actually it's not that easy. I would also leave that out.

r10s:
@Hocuri even though (or especially as :) you were not involved in all previous discussions and thoughts, your comments were very helpful and welcome!

PR #2089, opened 2021-10-21
improve selected recipients list

adbenitez:
when creating a new group or broadcast list, the selected contacts list was lacking some love, and was different from all other contacts lists, this PR fixes that


adbenitez:
one remaining issue is that the self contact is not put always at the end of the list

PR #2101, opened 2021-10-30
show verified-checkmark in memberlist in group creation

### LINKED ISSUE #2098: verified-checkmark not shown in memberlist on group-creation
with #2089, we have a much nicer list on contact creation, however, the "verified" symbol is still missing.

it was missing before as well, however, so this is no new bug :)

adbenitez:
close #2098


r10s (approved):
nice!

PR #2102, opened 2021-10-30
adapt 'self-profile' hint to reality

r10s:
- the profile hint misses that also the signature
  is sent out, together with name and image

- strike the part about that sent information cannot be retracted -
  that is pretty clear,
  and if we say that here, why not beside every message ;)
  not sure where that comes from,
  i checked similar settings in other messengers,
  there is no comparable hint.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -539,7 +539,7 @@
     <string name="blocked_empty_hint">If you block contacts, they will be shown here.</string>
     <string name="pref_profile_photo_remove_ask">Remove profile image?</string>
     <string name="pref_password_and_account_settings">Password and Account</string>
-    <string name="pref_who_can_see_profile_explain">Your profile image, name and signature will be shown alongside your messages when communicating with other users.</string>
+    <string name="pref_who_can_see_profile_explain">Your profile image, name and signature will be sent together with your messages when communicating with other users.</string>
     <string name="pref_your_name">Your Name</string>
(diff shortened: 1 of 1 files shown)


res/values/strings.xml
```diff
@@ -539,7 +539,7 @@
     <string name="blocked_empty_hint">If you block contacts, they will be shown here.</string>
     <string name="pref_profile_photo_remove_ask">Remove profile image?</string>
     <string name="pref_password_and_account_settings">Password and Account</string>
-    <string name="pref_who_can_see_profile_explain">Your profile image and name will be shown alongside your messages when communicating with other users. Already sent information can not be deleted or removed.</string>
+    <string name="pref_who_can_see_profile_explain">Your profile image, name and signature will be shown alongside your messages when communicating with other users.</string>
```
link2xt, line 542: The signature is not "shown alongside your messages", only in the profile and classic MUAs.

r10s: hm, i regarded the sender profiles as "alongside" :) what to say instead? "Your profile image, name and signature will be sent together with your messages when communicating with other users." ?

link2xt: > "Your profile image, name and signature will be sent together with your messages when communicating with other users." ?

This looks good.

"Shown alongside your messages" was probably referring to the chat view, where your avatar and name are displayed near the message bubble.

r10s: k, i added a commit.

link2xt (approved):

Issue #1566, opened 2020-08-16
Terms "CANCEL" and "DONE" in group picture selection view not localized


gerryfrancis:
- Operating System (Linux/Mac/Windows/iOS/Android):
Android 10 on Fairphone FP2 (Lineage OS 17.1 build QQ3A.200805.001 from 2020-08-14).

- Delta Chat Version:
1.12.2.

- Expected behavior:
The terms "CANCEL" and "DONE" in the group picture selection view are localized when a different language than English is used.

- Actual behavior:
The terms "CANCEL" and "DONE" in the group picture selection view are not localized when a different language than English is used.

- Steps to reproduce the problem:
-- Create a new group and tap on the group avatar space to select a picture.

- Screenshots:
![grafik](https://user-images.githubusercontent.com/51095533/90341092-fd247900-dffc-11ea-8f0b-1fde8386ef17.png)

- Logs:
N/A.

Hocuri:
I can confirm this. The problem is that we are using https://github.com/jdamcd/android-crop (`com.soundcloud.android:android-crop`) for this activity, which is not maintained anymore and will definitely not get any new translations. Maybe we can have a look at what Signal is doing now, maybe they fixed this in the meantime.

Gh0stz0x:
![Screenshot_20230205_203320](https://user-images.githubusercontent.com/74795149/216840701-7268f4c8-b347-4bfc-935e-3fc08491a96b.jpg)

The problem is still present in 1.34.12

AndreaEnzoLattmann:
The problem is still present in 1.36.2

gerryfrancis:
@adbenitez Thanks, I did not have this one on my "radar" anymore! ;)

PR #2103, opened 2021-10-31
QR activities improvements

adbenitez:
- [x] allow to copy QR data to clipboard
- [x] allow to share QR image
- [x] allow to read QR data from clipboard
- [x] allow to scan QR from gallery

### DIFF
--- a/res/menu/qr_show.xml
+++ b/res/menu/qr_show.xml
@@ -2,6 +2,11 @@
 
 <menu xmlns:android="http://schemas.android.com/apk/res/android" xmlns:app="http://schemas.android.com/apk/res-auto">
 
+    <item android:title="@string/menu_share"
+          android:id="@+id/share"
+          android:icon="@drawable/ic_share_white_24dp"
+          app:showAsAction="ifRoom"/>
+
     <item android:title="@string/menu_copy_link_to_clipboard"
           android:id="@+id/copy"
           app:showAsAction="never"/>
--- a/res/menu/qr_show.xml
+++ b/res/menu/qr_show.xml
@@ -7,7 +7,7 @@
           android:icon="@drawable/ic_share_white_24dp"
           app:showAsAction="ifRoom"/>
 
-    <item android:title="@string/menu_copy_link_to_clipboard"
+    <item android:title="@string/menu_copy_to_clipboard"
           android:id="@+id/copy"
           app:showAsAction="never"/>
 
--- a/res/menu/qr_show.xml
+++ b/res/menu/qr_show.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<menu xmlns:android="http://schemas.android.com/apk/res/android" xmlns:app="http://schemas.android.com/apk/res-auto">
+
+    <item android:title="@string/menu_copy_link_to_clipboard"
+          android:id="@+id/copy"
+          app:showAsAction="never"/>
+
+</menu>
--- a/src/org/thoughtcrime/securesms/qr/QrActivity.java
+++ b/src/org/thoughtcrime/securesms/qr/QrActivity.java
@@ -12,6 +12,7 @@ import android.preference.PreferenceManager;
 import android.util.Log;
 import android.view.Menu;
 import android.view.MenuItem;
+import android.widget.Toast;
 
 import androidx.annotation.NonNull;
 import androidx.fragment.app.Fragment;
@@ -203,6 +204,7 @@ public class QrActivity extends BaseActionBarActivity {
                             qrCodeHandler.handleQrData(result.getText());
                         } catch (NotFoundException e) {
                             Log.e(TAG, "decode exception", e);
+                            Toast.makeText(this, getString(R.string.qrscan_failed), Toast.LENGTH_LONG).show();
                         }
                     } catch (FileNotFoundException e) {
                         Log.e(TAG, "can not open file: " + uri.toString(), e);
--- a/src/org/thoughtcrime/securesms/qr/QrActivity.java
+++ b/src/org/thoughtcrime/securesms/qr/QrActivity.java
@@ -152,7 +152,7 @@ public class QrActivity extends BaseActionBarActivity {
                 break;
             case R.id.paste:
                 QrCodeHandler qrCodeHandler = new QrCodeHandler(this);
-                qrCodeHandler.handleOpenPgp4Fpr(Util.getTextFromClipboard(this));
+                qrCodeHandler.handleQrData(Util.getTextFromClipboard(this));
                 break;
         }
 
@@ -200,7 +200,7 @@ public class QrActivity extends BaseActionBarActivity {
                         try {
                             Result result = reader.decode(bBitmap);
                             QrCodeHandler qrCodeHandler = new QrCodeHandler(this);
-                            qrCodeHandler.handleOpenPgp4Fpr(result.getText());
+                            qrCodeHandler.handleQrData(result.getText());
                         } catch (NotFoundException e) {
                             Log.e(TAG, "decode exception", e);
                         }
--- a/src/org/thoughtcrime/securesms/qr/QrActivity.java
+++ b/src/org/thoughtcrime/securesms/qr/QrActivity.java
@@ -4,7 +4,9 @@ import android.Manifest;
 import android.content.pm.PackageManager;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
+import android.view.Menu;
 import android.view.MenuItem;
+import android.widget.Toast;
 
 import androidx.annotation.NonNull;
 import androidx.fragment.app.Fragment;
@@ -16,10 +18,12 @@ import com.google.android.material.tabs.TabLayout;
 
 import org.thoughtcrime.securesms.BaseActionBarActivity;
 import org.thoughtcrime.securesms.R;
+import org.thoughtcrime.securesms.connect.DcHelper;
 import org.thoughtcrime.securesms.permissions.Permissions;
 import org.thoughtcrime.securesms.util.DynamicLanguage;
 import org.thoughtcrime.securesms.util.DynamicNoActionBarTheme;
 import org.thoughtcrime.securesms.util.DynamicTheme;
+import org.thoughtcrime.securesms.util.Util;
 import org.thoughtcrime.securesms.util.ViewUtil;
 
 public class QrActivity extends BaseActionBarActivity {
@@ -86,6 +90,17 @@ public class QrActivity extends BaseActionBarActivity {
     }
 
     @Override
+    public boolean onPrepareOptionsMenu(Menu menu) {
+      menu.clear();
+      if(tabLayout.getSelectedTabPosition() == TAB_SHOW) {
+        getMenuInflater().inflate(R.menu.qr_show, menu);
+      } else {
+        getMenuInflater().inflate(R.menu.qr_scan, menu);
+      }
+      return super.onPrepareOptionsMenu(menu);
+    }
+
+    @Override
     protected void onResume() {
         super.onResume();
 
@@ -108,6 +123,14 @@ public class QrActivity extends BaseActionBarActivity {
             case android.R.id.home:
                 finish();
                 return true;
+            case R.id.copy:
+                Util.writeTextToClipboard(this, DcHelper.getContext(this).getSecurejoinQr(0));
+                Toast.makeText(this, getString(R.string.copied_to_clipboard), Toast.LENGTH_SHORT).show();
+                break;
+            case R.id.paste:
+                QrCodeHandler qrCodeHandler = new QrCodeHandler(this);
+                qrCodeHandler.handleOpenPgp4Fpr(Util.getTextFromClipboard(this));
+                break;
         }
 
         return false;
--- a/src/org/thoughtcrime/securesms/qr/QrActivity.java
+++ b/src/org/thoughtcrime/securesms/qr/QrActivity.java
@@ -1,9 +1,15 @@
 package org.thoughtcrime.securesms.qr;
 
 import android.Manifest;
+import android.app.Activity;
+import android.content.Intent;
 import android.content.pm.PackageManager;
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.net.Uri;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
+import android.util.Log;
 import android.view.Menu;
 import android.view.MenuItem;
 
@@ -14,6 +20,12 @@ import androidx.fragment.app.FragmentStatePagerAdapter;
 import androidx.viewpager.widget.ViewPager;
 
 import com.google.android.material.tabs.TabLayout;
+import com.google.zxing.BinaryBitmap;
+import com.google.zxing.MultiFormatReader;
+import com.google.zxing.NotFoundException;
+import com.google.zxing.Result;
+import com.google.zxing.RGBLuminanceSource;
+import com.google.zxing.common.HybridBinarizer;
 
 import org.thoughtcrime.securesms.BaseActionBarActivity;
 import org.thoughtcrime.securesms.R;
@@ -24,11 +36,16 @@ import org.thoughtcrime.securesms.util.DynamicTheme;
 import org.thoughtcrime.securesms.util.Util;
 import org.thoughtcrime.securesms.util.ViewUtil;
 
+import java.io.FileNotFoundException;
+import java.io.InputStream;
+
 public class QrActivity extends BaseActionBarActivity {
 
     private final DynamicTheme dynamicTheme = new DynamicNoActionBarTheme();
     private final DynamicLanguage dynamicLanguage = new DynamicLanguage();
 
+    private final static String TAG = QrActivity.class.getSimpleName();
+    private final static int REQUEST_CODE_IMAGE = 9;
     private final static int TAB_SHOW = 0;
     private final static int TAB_SCAN = 1;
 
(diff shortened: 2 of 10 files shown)


res/menu/qr_show.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<menu xmlns:android="http://schemas.android.com/apk/res/android" xmlns:app="http://schemas.android.com/apk/res-auto">
+
+    <item android:title="@string/menu_share"
+          android:id="@+id/share"
+          android:icon="@drawable/ic_share_white_24dp"
+          app:showAsAction="ifRoom"/>
+
+    <item android:title="@string/menu_copy_link_to_clipboard"
```
adbenitez, line 10: I used "copy link" instead of "copy" to avoid assumptions that the image would be copied instead the QR data, technically saying the QR content is a link is not incorrect, right? do you think a play "copy" is better?

r10s: i think, "Copy to Clipboard" is better here, it is neutral and does not say anything about the content - in contrast to "copy link", which can be assumed as being wrong as the copied item is a "link" only technical-wise - the user may expect sth. as "https://..." as a link.

also, desktop uses the neutral form (just "Copy" there), same for ios (if we decide to revive https://github.com/deltachat/deltachat-ios/pull (partly))

adbenitez: done

adbenitez:
One thing missing is the "Scan QR code" from the welcome screen to create a temporal account, that needs some refactoring so I think that can be postponed for another PR

src/org/thoughtcrime/securesms/qr/QrActivity.java
```diff
@@ -108,6 +140,24 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+            case R.id.copy:
+                qrShowFragment.copyQrData();
+                break;
+            case R.id.load_from_image:
+                Permissions.with(this)
+                           .request(Manifest.permission.WRITE_EXTERNAL_STORAGE)
+                           .ifNecessary()
+                           .withPermanentDenialDialog(getString(R.string.perm_explain_access_to_storage_denied))
+                           .onAllGranted(() -> Util.selectMediaType(this, "image/*", null, REQUEST_CODE_IMAGE))
+                           .execute();
```
Hocuri, line 155: We can freely decide what we want to do here; we don't need the permission to select an image, but we might need the permission later on when we want to show the recent photo rail.

What you will need to do is replace `WRITE_EXTERNAL_STORAGE` with `READ_EXTERNAL_STORAGE`, but why not use `selectImage()` instead of this whole block? It's always better not to have code duplication.

adbenitez: now that you say it I don't even know why I moved `selectMediaType` to utils instead of using AttachmentManager directly :sweat: :man_facepalming:  will fix that

adbenitez: >  but why not use `selectImage()` instead of this whole block?

done, thanks for pointing that out!

src/org/thoughtcrime/securesms/qr/QrActivity.java
```diff
@@ -13,25 +20,39 @@
...
+import java.io.FileNotFoundException;
+import java.io.InputStream;
+
 public class QrActivity extends BaseActionBarActivity {
 
     private final DynamicTheme dynamicTheme = new DynamicNoActionBarTheme();
     private final DynamicLanguage dynamicLanguage = new DynamicLanguage();
 
+    private final static String TAG = QrActivity.class.getSimpleName();
+    private final static int REQUEST_CODE_IMAGE = 9;
```
r10s, line 49: as this is an app-wide-magic number, maybe use something easier grep'able with more digits (31331 ... sth. that does not exist in code currently :)

however, it was doing for the other selectImage() calls only partly as well, but still.

src/org/thoughtcrime/securesms/qr/QrShowFragment.java
```diff
@@ -119,6 +129,31 @@ public void onReceive(Context context, Intent intent) {
         return view;
     }
 
+    public void shareQr() {
+        try {
+            File file = new File(getActivity().getExternalCacheDir(), "share-qr.png");
```
r10s, line 134: any special reason to prefer [`getExternalCacheDir()`](https://developer.android.com/reference/android/content/Context.html#getExternalCacheDir%28%29) over [`getCacheDir`](https://developer.android.com/reference/android/content/Context#getCacheDir()) - they seem to be widely identical, however, only for the latter the system always takes care about deletion and cleaning up.

also, for the external cache, doc says "Shared storage may not always be available, since removable media can be ejected by the user".

r10s (commented):
this is a great functionality boost \o/

i have a few suggestions annotated, but all in all that works nicely.

for importing images, i did not manage to get the qr code from a "real photo", however, screenshots and exports work.

nb: first, i thought, it may be better to add the share/load icons to the menu, as text may be more descriptive (so, display=never), however, meanwhile, i think it is okay to have an additional icon. and it is also nice to make the new features a bit visible :)

adbenitez:
> for importing images, i did not manage to get the qr code from a "real photo"

I was tempted to add the crop view we have for avatar selection, but that view is an outdated lib so I preferred to postpone that for when I manage to replace that old lib that is causing issues (see #1566) for the nice crop and rotation tool we have in the image editor

adbenitez:
> nb: first, i thought, it may be better to add the share/load icons to the menu, as text may be more descriptive

I also think that showing as text is fine, but ended adding the icons to make it more noticeable and obvious that when you change tab the menu options also change, btw, long pressing the icons you get the text that would be shown in the menu entries

src/org/thoughtcrime/securesms/qr/QrActivity.java
```diff
@@ -125,10 +171,54 @@ public void onRequestPermissionsResult(int requestCode, @NonNull String[] permis
...
+                        bitmap.getPixels(pixels, 0, width, 0, 0, width, height);
+                        bitmap.recycle();
+                        bitmap = null;
+                        RGBLuminanceSource source = new RGBLuminanceSource(width, height, pixels);
+                        BinaryBitmap bBitmap = new BinaryBitmap(new HybridBinarizer(source));
+                        MultiFormatReader reader = new MultiFormatReader();
+                        try {
+                            Result result = reader.decode(bBitmap);
+                            QrCodeHandler qrCodeHandler = new QrCodeHandler(this);
+                            qrCodeHandler.handleOpenPgp4Fpr(result.getText());
```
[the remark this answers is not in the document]
adbenitez: agree, maybe "handleQrData" or so

src/org/thoughtcrime/securesms/qr/QrActivity.java
```diff
@@ -125,10 +171,54 @@ public void onRequestPermissionsResult(int requestCode, @NonNull String[] permis
...
+                        int width = bitmap.getWidth(), height = bitmap.getHeight();
+                        int[] pixels = new int[width * height];
+                        bitmap.getPixels(pixels, 0, width, 0, 0, width, height);
+                        bitmap.recycle();
+                        bitmap = null;
+                        RGBLuminanceSource source = new RGBLuminanceSource(width, height, pixels);
+                        BinaryBitmap bBitmap = new BinaryBitmap(new HybridBinarizer(source));
+                        MultiFormatReader reader = new MultiFormatReader();
+                        try {
+                            Result result = reader.decode(bBitmap);
```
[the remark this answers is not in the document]
adbenitez: currently Desktop also fails silently :/

src/org/thoughtcrime/securesms/qr/QrActivity.java
adbenitez: I will use 46243, that means `i-m-a-g-e` in a phone dial pad / keyboard :D

src/org/thoughtcrime/securesms/qr/QrShowFragment.java
adbenitez: We use `getExternalCacheDir()` with preference over `getCacheDir()` in `PersistentBlobProvider.getExternalDir()` I will use that instead of using `getExternalCacheDir()` direclty

btw, I used `getExternalCacheDir()` because if I use `getCacheDir()` I get an error:

```
java.lang.IllegalArgumentException: Failed to find configured root that contains /data/data/chat.delta.lite/cache/shar
e-qr.png
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at androidx.core.content.FileProvider$SimplePathStrategy.getUriForFile(FileProvider.java:744)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at androidx.core.content.FileProvider.getUriForFile(FileProvider.java:418)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at org.thoughtcrime.securesms.util.FileProviderUtil.getUriFor(FileProviderUtil.java:19)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at org.thoughtcrime.securesms.qr.QrShowFragment.shareQr(QrShowFragment.java:141)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at org.thoughtcrime.securesms.qr.QrActivity.onOptionsItemSelected(QrActivity.java:145)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at android.app.Activity.onMenuItemSelected(Activity.java:4326)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at androidx.fragment.app.FragmentActivity.onMenuItemSelected(FragmentActivity.java:383)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at androidx.appcompat.app.AppCompatActivity.onMenuItemSelected(AppCompatActivity.java:228)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at androidx.appcompat.view.WindowCallbackWrapper.onMenuItemSelected(WindowCallbackWrapper.java:109)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at androidx.appcompat.view.WindowCallbackWrapper.onMenuItemSelected(WindowCallbackWrapper.java:109)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at androidx.appcompat.app.ToolbarActionBar$2.onMenuItemClick(ToolbarActionBar.java:65)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at androidx.appcompat.widget.Toolbar$1.onMenuItemClick(Toolbar.java:207)  
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at androidx.appcompat.widget.ActionMenuView$MenuBuilderCallback.onMenuItemSelected(ActionMenuView.java:779)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at androidx.appcompat.view.menu.MenuBuilder.dispatchMenuItemSelected(MenuBuilder.java:834)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at androidx.appcompat.view.menu.MenuItemImpl.invoke(MenuItemImpl.java:158)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at androidx.appcompat.view.menu.MenuBuilder.performItemAction(MenuBuilder.java:985)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at androidx.appcompat.view.menu.MenuBuilder.performItemAction(MenuBuilder.java:975)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at androidx.appcompat.widget.ActionMenuView.invokeItem(ActionMenuView.java:623)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at androidx.appcompat.view.menu.ActionMenuItemView.onClick(ActionMenuItemView.java:151)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at android.view.View.performClick(View.java:8160)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at android.widget.TextView.performClick(TextView.java:16222)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at android.view.View.performClickInternal(View.java:8137)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at android.view.View.access$3700(View.java:888)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at android.view.View$PerformClick.run(View.java:30236)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at android.os.Handler.handleCallback(Handler.java:938)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at android.os.Handler.dispatchMessage(Handler.java:99)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at android.os.Looper.loop(Looper.java:246)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at android.app.ActivityThread.main(ActivityThread.java:8587)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at java.lang.reflect.Method.invoke(Native Method)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:602)
11-03 19:44:56.925  8562  8562 W QrShowFragment:        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1130)
```

adbenitez: of switched to `getCacheDir()`, the error was because `<cache-path name="cache" path="." />` was missing in `file_provider_paths.xml`

src/org/thoughtcrime/securesms/qr/QrActivity.java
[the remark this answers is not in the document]
adbenitez: I renamed to `handleQrData`

src/org/thoughtcrime/securesms/qr/QrActivity.java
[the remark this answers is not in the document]
adbenitez: there is string `qrscan_failed` ("QR code could not be decoded") that could be used instead of adding a new string, what do you think? I think it is fine, btw, it seems that string is currently unused, good opportunity to recycle all that translation efforts :D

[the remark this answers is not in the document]
adbenitez: I showed a toast instead of a dialog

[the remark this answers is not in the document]
adbenitez: btw, the error is already captured, just that only a log entry was written without user feedback

src/org/thoughtcrime/securesms/qr/QrActivity.java
```diff
@@ -125,10 +171,54 @@ public void onRequestPermissionsResult(int requestCode, @NonNull String[] permis
...
+
+        if (resultCode != Activity.RESULT_OK)
+            return;
+
+        switch (reqCode) {
+            case REQUEST_CODE_IMAGE:
+                Uri uri  = (data != null ? data.getData() : null);
+                if (uri != null) {
+                    try {
+                        InputStream inputStream = getContentResolver().openInputStream(uri);
```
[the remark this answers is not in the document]
adbenitez: that is how it is done in many other parts of the code and in api30 branch, for example:

https://github.com/deltachat/deltachat-android/pull/2087/files#diff-e2cad08e31255783bf3c91276e6048a1f8110c9ebda9f34f5d602120b6c4a9d9R186

adbenitez:
@r10s I think I resolved all your comments/findings

r10s (approved):
lgtm, now.

i am still a bit unsure if the "primary" action of copy+paste isn't more the text thingie, esp. as that works more reliable on import.

however having the option to export the qr code as an image is also great, eg. for printing - and importing _that_ seems to work reliable.

so, let's merge that in, and see how that gets used. we can tweak menu options and ordering later anyway.

src/org/thoughtcrime/securesms/qr/QrShowFragment.java
```diff
@@ -119,6 +129,31 @@ public void onReceive(Context context, Intent intent) {
...
+            bitmap.compress(Bitmap.CompressFormat.PNG, 90, stream);
+            stream.flush();
+            stream.close();
+            Uri uri = FileProviderUtil.getUriFor(getActivity(), file);
+            Intent intent = new Intent(Intent.ACTION_SEND);
+            intent.setType("image/png");
+            intent.putExtra(Intent.EXTRA_STREAM, uri);
+            intent.putExtra(Intent.EXTRA_TEXT, hint);
+            intent.addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
+            startActivity(Intent.createChooser(intent, getString(R.string.chat_share_with_title)));
```
Hocuri, line 147: I just found the method `DcHelper.openForViewOrShare()` which does the same thing but with more special-casing e.g. for older Android versions. (also, it takes a message id, not a file, but we could extract everything after the line `String mimeType = msg.getFilemime();` into an extra function that can then share an arbitrary file.

With luck, this would also mean that we don't have to change `file_provider_paths.xml` because I'm not sure what the security implications of this change are - I hope that it doesn't give other apps full access to our cache directory, but Android tends to surprise me (except if you know @adbenitez?). And also, there is probably a reason that we do the thing with `DcHelper.sharedFiles`.

r10s: yip, using sth. as in openForViewOrShare() sounds like the best way to to that, thanks for pointing that out. sounds not to be that hard.

i agree with having mixed feelings about possible security implications of security of `file_provider_paths.xml` ...

as a quick step to move forward, we can also revert the commit https://github.com/deltachat/deltachat-android/pull/2103/commits/2a2a801b311e7c4929f5063d3340234413528b93 that introduces the issue - seems as if also with getExternalCacheDir things work in general.

Issue #1762, opened 2020-12-29
setConfigFromQr error is not displayed


link2xt:
Scanning invalid `DCACCOUNT` code, such as `DCACCOUNT:http://example.org/`, results in error dialog "Configuration failed. Error: Please enter an email address".

When running `setqr DCACCOUNT:http://example.org/`, there is an error `Cannot set config from QR code: Cannot create account, request to http://example.org/ failed`. It should be displayed in Android UI, but it isn't.

Attached QR code for testing:
![1](https://user-images.githubusercontent.com/18373967/103284464-56a9a500-49ec-11eb-9e71-68edb0bf98be.png)

Relevant line is https://github.com/deltachat/deltachat-android/blob/971c7e1d68129b235221d9d8c86c5cd724a1726a/src/org/thoughtcrime/securesms/WelcomeActivity.java#L166

One bug is that `dcContext.configure()` is called regardless of whether `setQrConfigFromQr` succeeds or not. The other one is that error is not displayed.

adbenitez:
@link2xt the issue is that `dcContext.setConfigFromQr(qrCode)` is returning that it failed, but at that time the error event have not been processed yet by `ApplicationDcContext.handleEvent()` (where the `lastErrorString` is saved) so `dcContext.getCapturedError()` returns empty string, adding an additional call to `dcContext.getCapturedError()` just afterwards I got the error string, it is just a race condition, 

if  `dcContext.setConfigFromQr(qrCode)` returns `false` there was an error, maybe it makes sense to wait for `dcContext.getCapturedError()` (`lastErrorString`) to have a value or so

link2xt:
Or maybe get rid of error capturing once https://github.com/deltachat/deltachat-core-rust/pull/2319 is merged and get back to simply displaying toasts on error events. As I understand this error capturing was created because of annoying network errors which should not be emitted anymore.

r10s:
>  As I understand this error capturing was created because of annoying network errors which should not be emitted anymore.

no, the capturing was created to show _concrete_ alert-dialogs after some actions, as configure, import, scan account etc.  toasts are not-so-great for that and, due to unclear dialog/acticvity-context you cannot easily get "real" dialogs on events.

[...]

PR #2106, opened 2021-11-02
show errors after account scanning failure

r10s:
reading captured error has a race with the thread that sets the error.

this is mitigated by sleep().

see https://github.com/deltachat/deltachat-core-rust/issues/2787
for details and a proper solution.

### DIFF
--- a/src/org/thoughtcrime/securesms/WelcomeActivity.java
+++ b/src/org/thoughtcrime/securesms/WelcomeActivity.java
@@ -6,6 +6,7 @@ import android.app.Activity;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.os.Bundle;
+import android.text.TextUtils;
 import android.text.util.Linkify;
 import android.view.View;
 import android.widget.Button;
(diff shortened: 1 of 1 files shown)


adbenitez:
this hack will fix #1762 there you can find info about the issue and some comments from @link2xt proposing to remove this altogether and use toasts

src/org/thoughtcrime/securesms/WelcomeActivity.java
```diff
@@ -167,7 +169,10 @@ private void startQrAccountCreation(String qrCode)
         DcHelper.getEventCenter(this).captureNextError();
 
         if (!dcContext.setConfigFromQr(qrCode)) {
-            progressError(DcHelper.getEventCenter(this).getCapturedError());
+            Util.sleep(100); // hack to avoid a race condition, see https://github.com/deltachat/deltachat-core-rust/issues/2787 for more details and possible fix
+            String err = DcHelper.getEventCenter(this).getCapturedError();
+            progressError(TextUtils.isEmpty(err) ? "Cannot create account from QR code." : err);
```
adbenitez, line 174: untranslated text

adbenitez (approved):

src/org/thoughtcrime/securesms/WelcomeActivity.java
r10s: yip, but that text normally should not appear at all, if the function fails an error should be set.

(we have tons of untranslated error messages and such things)

r10s:
merging that in, so that the issue is fixed.

will prepare a subsequent pr that uses `get_last_error()` of https://github.com/deltachat/deltachat-core-rust/pull/2788 then.

PR #2113, opened 2021-11-04
Fix photo view rail

### LINKED ISSUE #2112: attach-images-preview missing, at least on some android versions
on android6, after upgrading to a version containing https://github.com/deltachat/deltachat-android/pull/2087, the image-preview on opening the attachment drawer is no longer working.

it stopped working on both, an update and a fresh install.

iirc, also in the past, on first opening, the previews were never shown until the rights were granted when tapping "gallery" - but then, no whitespace was shown, everything was just hidden. this issue seems to be different as the drawer seems to assume there are sufficient rights and shows room for images that do not appear.

@adbenitez reported a similar issue in the Android Dev group.

Hocuri:
Fix #2112

- Don't check for write permission when we don't need it
- Revert one of the changes from Signal. I have no idea why they changed this back then, but it prevents all photos from being shown in the recent photo view rail.

### DIFF
--- a/src/org/thoughtcrime/securesms/database/loaders/RecentPhotosLoader.java
+++ b/src/org/thoughtcrime/securesms/database/loaders/RecentPhotosLoader.java
@@ -25,7 +25,7 @@ public class RecentPhotosLoader extends CursorLoader {
   };
 
   private static final String SELECTION  = Build.VERSION.SDK_INT > 28 ? MediaStore.Images.Media.IS_PENDING + " != 1"
-                                                                      : MediaStore.Images.Media.DATA + " IS NULL";
+                                                                      : null;
 
   private final Context context;
 
--- a/src/org/thoughtcrime/securesms/database/loaders/RecentPhotosLoader.java
+++ b/src/org/thoughtcrime/securesms/database/loaders/RecentPhotosLoader.java
@@ -36,7 +36,7 @@ public class RecentPhotosLoader extends CursorLoader {
 
   @Override
   public Cursor loadInBackground() {
-    if (Permissions.hasAll(context, Manifest.permission.WRITE_EXTERNAL_STORAGE)) {
+    if (Permissions.hasAll(context, Manifest.permission.READ_EXTERNAL_STORAGE)) {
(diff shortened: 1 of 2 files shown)


r10s (commented):
unfortunately, this still does not fix the issue for me.

even if i granted all permissions manually, the image-previews do not appear.

adbenitez (approved):

src/org/thoughtcrime/securesms/database/loaders/RecentPhotosLoader.java
```diff
@@ -36,7 +36,7 @@ public RecentPhotosLoader(Context context) {
 
   @Override
   public Cursor loadInBackground() {
-    if (Permissions.hasAll(context, Manifest.permission.WRITE_EXTERNAL_STORAGE)) {
+    if (Permissions.hasAll(context, Manifest.permission.READ_EXTERNAL_STORAGE)) {
```
r10s, line 39: READ_EXTERNAL_STORAGE sounds reasonable, however, we used WRITE_EXTERNAL_STORAGE before - and also Signal still uses WRITE_EXTERNAL_STORAGE (Signals RecentPhotosLoader.java is quite the same).

so, i do not really get that part (/me currently playing around with the android6 phone and trying to get the preview rail back there again)

Hocuri:
I pushed another commit that fixes an unrelated bug that had the same symptoms. @r10s can you re-try?

The fix is 38776e5 - reverting one of the changes that Signal did when upgrading to API 30. I have no idea why they changed this back then, but it prevents all photos from being shown in the recent photo view rail. And also I have no idea why the photo view rail works for them (OTOH, maybe it doesn't I didn't try Signal on any devices older than Android 10)

src/org/thoughtcrime/securesms/database/loaders/RecentPhotosLoader.java
Hocuri: It gets even weirder, what you are experiencing on android6 has a completely unrelated cause.

But also there, Signal should have the same issue.

Sometimes it would be nice to build Signal and poke around a bit (adding log statements) but it's not building out of the box and it's probably not worth the effort.

r10s (approved):

PR #2118, opened 2021-11-05
reduce selected account's checkmark size a bit

adbenitez:
@gerryfrancis complained in the testing group that the check-mark looks too big compared with the avatar size, I was using the size used in Signal but Signal avatars are bigger than in our account selector


r10s (approved):
:)

gerryfrancis:
@adbenitez @r10s Thank you so much, I will post a screenshot of the result in our Testing group! :)

PR #2132, opened 2021-11-12
CI: fix and improve apk preview workflows

adbenitez:
remove deprecated action and make the generated APK more visible

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -1,0 +1,20 @@
+name: add artifact links to pull request
+on:
+  workflow_run:
+    workflows: ["Upload Preview APK"]
+    types: [completed]
+
+jobs:
+  artifacts-url-comments:
+    name: add artifact links to pull request
+    runs-on: windows-2019
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - name: add artifact links to pull request
+        uses: tonyhallett/artifacts-url-comments@v1.1.0
+        env:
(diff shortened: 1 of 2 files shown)


r10s (approved):
great, thanks for taking care!

btw, i am always failing in finding the "apk artifact" for downloading, probably because i usually do not need it - however, is that only me? :)

adbenitez:
indeed I no longer find the menu that was there in the past to download the artifact, the artifact is uploaded but it seems GitHub UI changed :v I think desktop puts the files directly in the PR somehow I will check that, otherwise it is too hard to find the artifact (well it was in the past now it is completely missing XD)

adbenitez:
ah deltachat-desktop uploads stuff to https://download.delta.chat/desktop/preview/

adbenitez:
found the artifact here at the end: https://github.com/deltachat/deltachat-android/actions/runs/1454807233

also if you remove the `?check_suite_focus=true` from the "Details" link the artifact menu then re-appears, that is pretty weird from GitHub:
 https://github.com/deltachat/deltachat-android/runs/4194649464

r10s:
thanks for figuring that out! easy, when one knows where to search it :)

adbenitez:
@r10s now the apk direct download link will be published in a comment in the corresponding PR, similar to what https://github.com/deltachat/deltachat-pages/ does with the pages previews. 

the nice thing is that people watching this repo will even receive the comment by email and can try the apk direclty without having to search here in GitHub

r10s (approved):
a post with a preview link sounds great!

Hocuri (approved):
Nice!

PR #2138, opened 2021-11-18
[WIP] use the new SVG QR images generated in core

adbenitez:
related core pull request: https://github.com/deltachat/deltachat-core-rust/pull/2815

- [X] add the new API to the java bindings
- [X] show SVG QR
- [X] convert SVG to PNG on sharing
- [ ] set stock string


r10s (approved):

r10s:
k, merging that in, the stock string can be set in another pr.

also, if we find a way to get rid of the hack, we can do that in another pr.

PR #2139, opened 2021-11-20
fix that ndk make didn't work with symlinked core

Simon-Laux:


Hocuri:
Looks fine, but why does it help?
- How do you get a symlinked core?
- Why is `../something` not the same as `$jnidir/something` if the last thing you did was `jnidir=$(pwd); cd some_directory`
- I once read that it's better to use `$PWD` instead of `$(pwd)`

Simon-Laux:
> How do you get a symlinked core?

```
ln -s ../../deltachat-core-rust jni/deltachat-core-rust
```
the first path needs to be replaced with the location of your core repo clone ( in my case it's right next to the `deltachat-android` folder)

> Why is `../something` not the same as `$jnidir/something` if the last thing you did was `jnidir=$(pwd); cd some_directory`

it seems like relative paths are relativ to the current real folder path, not to the symbolic link path. at least for the `cp` command.

> I once read that it's better to use `$PWD` instead of `$(pwd)`

`$(pwd)` works for me, if you want we can try the other one, but I'm not after perfectionism here. I just needed it myself to work with a symlinked core.

Hocuri (approved):
Ok, so I pushed a commit to use `$PWD`.

Do you use MacOS? If not, we might want to wait a bit for @r10s to test this on MacOS, too

adbenitez (approved):

r10s:
`echo $PWD` works on mac, so i would assume the whole pr is fine on mac as well.

PR #2144, opened 2021-11-23
Display last-seen info in profile

adbenitez:
cc @link2xt

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -85,6 +85,8 @@
     <string name="this_month">This month</string>
     <string name="last_week">Last week</string>
     <string name="last_month">Last month</string>
+    <string name="last_seen_at">Last seen at %1$s</string>
+    <string name="last_seen_never">Last seen: never</string>
     <!-- Translators: show beside messages that are "N minutes old". prefer a short string, prefer abbreviations if they're well-known -->
     <plurals name="n_minutes">
         <item quantity="one">%d min</item>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -86,7 +86,7 @@
     <string name="last_week">Last week</string>
     <string name="last_month">Last month</string>
     <string name="last_seen_at">Last seen at %1$s</string>
-    <string name="last_seen_never">Last seen: never</string>
+    <string name="last_seen_never">Last seen: Never</string>
     <!-- Translators: show beside messages that are "N minutes old". prefer a short string, prefer abbreviations if they're well-known -->
     <plurals name="n_minutes">
         <item quantity="one">%d min</item>
--- a/src/org/thoughtcrime/securesms/ProfileSettingsAdapter.java
+++ b/src/org/thoughtcrime/securesms/ProfileSettingsAdapter.java
@@ -19,6 +19,7 @@ import com.b44t.messenger.DcLot;
 import org.thoughtcrime.securesms.connect.DcHelper;
 import org.thoughtcrime.securesms.contacts.ContactSelectionListItem;
 import org.thoughtcrime.securesms.mms.GlideRequests;
+import org.thoughtcrime.securesms.util.DateUtils;
 import org.thoughtcrime.securesms.util.StickyHeaderDecoration.StickyHeaderAdapter;
 
 import java.util.ArrayList;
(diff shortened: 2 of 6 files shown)


src/org/thoughtcrime/securesms/ProfileSettingsAdapter.java
```diff
@@ -235,11 +236,12 @@ public void onBindHeaderViewHolder(HeaderViewHolder viewHolder, int position) {
         txt = context.getString(R.string.profile_shared_chats);
         break;
       case ItemData.TYPE_PRIMARY_SETTING:
-        if(itemDataContact!=null && itemDataContact.isVerified()) {
-          txt = context.getString(R.string.verified_contact);
+        long lastSeen = (itemDataContact!=null? itemDataContact.getLastSeen() : 0);
+        if(lastSeen==0) {
```
r10s, line 240:
```suggestion
        if (lastSeen == 0) {
```

i think, this is the style we are mostly using for new code.

(existing code may use different styles, if pulled from some others sources, in general, we do not change the style to to retain the option of easy diffs on updates, same for tabs :)

res/values/strings.xml
```diff
@@ -85,6 +85,8 @@
     <string name="this_month">This month</string>
     <string name="last_week">Last week</string>
     <string name="last_month">Last month</string>
+    <string name="last_seen_at">Last seen at %1$s</string>
+    <string name="last_seen_never">Last seen: never</string>
```
r10s, line 89:
```suggestion
    <string name="last_seen_never">Last seen: Never</string>
```

usually, we start option names upper case, so maybe this is also better here.

r10s (approved):
checked the source, lgtm, some very minor nitpicking only :)

src/org/thoughtcrime/securesms/ProfileSettingsAdapter.java
```diff
@@ -235,11 +236,12 @@ public void onBindHeaderViewHolder(HeaderViewHolder viewHolder, int position) {
         txt = context.getString(R.string.profile_shared_chats);
         break;
       case ItemData.TYPE_PRIMARY_SETTING:
-        if(itemDataContact!=null && itemDataContact.isVerified()) {
-          txt = context.getString(R.string.verified_contact);
```
Hocuri, line 239: So, we don't show whether a contact is verified anymore?

I think it should somehow be visible in the profile that a contact is verified

Hocuri:
Nice :tada:!

src/org/thoughtcrime/securesms/ProfileSettingsAdapter.java
r10s: it is shown with the blue verified-checkbox, this is what we are doing on desktop/ios as well, so, probably fine.

Hocuri: I couldn't find the blue verified-checkbox anywhere, but maybe that's the actual problem

r10s: it is right of the names on android.

ios and desktop shows them on avatars, which is maybe better. however, as this "verified" stuff seems not to be totally settled yet, i am not suggesting to change things here on android :)

adbenitez: yip that is what does this PR now the verified checkbox is added to the name, currently it is not show in the chat list, should I also display it there? it is a bit confusing with verified groups but IMHO "verified groups" should be rebranded to "protected group" and use a shield icon for protected chats and let the verified icon for verified contacts then display verified icon everywhere for verified contacts

Hocuri: All of this is out of scope for this PR, so consider this approved :)

But at least on my device there is no checkmark shown next to the name
image_2021-11-23_17-10-17

adbenitez: hocuri, check this PR's commits, the checkmark will be shown with this PR it was not show in the DC version you are using that is what this PR changes, it removes the "verified contact" header and instead shows a checkmark

PR #2146, opened 2021-11-24
"scan qr code" wordings

r10s:
see commit messages for details

successor of https://github.com/deltachat/deltachat-android/pull/2138


adbenitez (approved):

Simon-Laux (approved):
lgtm

Hocuri (approved):

r10s:
i added quotes around the group name, be be in sync with similar strings.

nb: androidsvg 1.4 has a bug that does not display "enities"; this seems to be fixed in 1.5, which,  however, is not released yet, see https://github.com/BigBadaboom/androidsvg/issues/214 - however, for us, this is probably okayish for now, and this pr does not make things worse.

PR #2147, opened 2021-11-25
make avatar letter bold and transform it to uppercase

adbenitez:
to make it fit with the new QRs and also that is how they are shown in Desktop


r10s (approved):
looks good! and also great that we synchronize desktop/ios/android that way.

PR #2148, opened 2021-11-25
synchronize unread-badge-color with desktop

r10s:
currently, all three big delta chat uis have different colors for the unread badge:

- desktop is _blue_
- android is _green_
- ios is _red_

esp. for the latter, red seems to be a bad choice as that is easily mixed with error symbols, however, it makes sense to use a similar color for all systems anyway.

in several one-to-one-discussions, there was a majority for _blue_ - so that desktop does not need to do anything to do on that :)
however, blue might also be a good choice as that color is already used on android here and there (headlines etc.) and we were also thinking about chaning the title bar to sth. blueish (however, that is out of scope for now)

i am totally aware that there will ppl that _love_ the current green, these kind of changes are always a matter of taste and cannot be fine for everyone :)

before/after:

Screen Shot 2021-11-28 at 17 31 06 &nbsp; &nbsp; Screen Shot 2021-11-28 at 17 32 19

if we merge that in, i'll create a corresponding pr for ios.

### DIFF
--- a/res/values/colors.xml
+++ b/res/values/colors.xml
@@ -12,6 +12,8 @@
     <color name="primary_alpha_focus">#882090ea</color>
     <color name="primary_alpha33">#552090ea</color>
 
+    <color name="unread_count">#ff3792fc</color>
+
     <color name="white">#ffffffff</color>
     <color name="black">#ff000000</color>
     <color name="gray5">#ffeeeeee</color>
--- a/src/org/thoughtcrime/securesms/ConversationListItem.java
+++ b/src/org/thoughtcrime/securesms/ConversationListItem.java
@@ -278,7 +278,7 @@ public class ConversationListItem extends RelativeLayout
               .textColor(Color.WHITE)
               .bold()
               .endConfig()
-              .buildRound(String.valueOf(unreadCount), getResources().getColor(R.color.green_A700)));
+              .buildRound(String.valueOf(unreadCount), getResources().getColor(R.color.unread_count)));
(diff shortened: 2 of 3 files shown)


adbenitez (approved):
the checkmark in the selected account is green to match the unread message color, maybe change that to blue as well?

r10s:
> the checkmark in the selected account is green to match the unread message color, maybe change that to blue as well?

also thought about that, at the end, i would not care for now. it is also fine to leave that in green i'd say. as you like ;)

Hocuri:
> i am totally aware that there will ppl that _love_ the current green

Yes, me :see_no_evil: but I guess that's life :joy:

Hocuri:
> it is also fine to leave that in green i'd say.

+1 for staying with green checkmarks :+1:

r10s:
trying that out for some days, i am also not totally convinced that blue is the better choice. however, the important thing is to have it consistent over the systems, so, at the end i would not really care.

for green one could also argue that this color is also used for other states (the ticks), also, i think, this fits slightly better to the titlebar.

we can also do a poll for that - however, not sure how stuck desktop is to blue currently cc @Simon-Laux @Jikstra

Simon-Laux:
I don't think that green would look good on desktop, but It's not as if I care much about that little detail.

Hocuri:
> [red] appears too alarming i'd say

+1

r10s:
let's try that out, we can change it anytime :)

PR #2149, opened 2021-11-25
add translation strings from desktop

Simon-Laux:


res/values/strings.xml
```diff
@@ -208,6 +208,7 @@
     <string name="menu_copy_link_to_clipboard">Copy Link</string>
     <string name="menu_copy_text_to_clipboard">Copy Text</string>
     <string name="menu_copy_image_to_clipboard">Copy Image</string>
+    <string name="menu_copy_email_to_clipboard">Copy Email</string>
```
adbenitez, line 211: currently `menu_copy_to_clipboard` is used in Android to copy the contact address, is this really needed? if so then maybe makes sense to use this one instead of "copy to clipboard" in the profile since this is more clear what will be copied

adbenitez (approved):

res/values/strings.xml
r10s: we are saying E-Mail (with minus) everywhere else in the app, so that should be consistent.

r10s, line 211:
```suggestion
    <string name="menu_copy_email_to_clipboard">Copy E-Mail</string>
```

Simon-Laux: I added it for right clicking on email addresses in messages in desktop

PR #2152, opened 2021-11-28
make tracking core versions easier

r10s:
this is the outcome of recent dc discussions, see commit messages for details.

i already used the changed script for the last core update, see https://github.com/deltachat/deltachat-android/commit/c67e61c12ef06a9eafd91bdbf0621f61ef23e0cf


link2xt (approved):
How do you update to the latest tagged version?

r10s:
> How do you update to the latest tagged version?

the script always updates to current master, that may or may not be a tagged version. i call it after a core release, but also in between as needed. if one really would want to upgrade to a special tag/commit, that can be done manually, however, up to now, the script was just fine.

PR #2153, opened 2021-11-28
prepare 1.26.0

r10s:
as this comes shortly after 1.24, as an exception, we keep both device messages:

Screen Shot 2021-11-28 at 21 47 00


Hocuri (approved):

adbenitez (approved):
LGTM

PR #2157, opened 2021-11-30
avoid NullPointerException when selecting thumbnail image

adbenitez:
close #2156

PR #2159, opened 2021-12-01
bump to 1.26.2

r10s:
maybe we should get https://github.com/deltachat/deltachat-android/pull/2157 in, however, i am not deep in these things, maybe  @Hocuri can have a look at it. however, it seem not to be that critical and is probably not a regression.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Delta Chat Android Changelog
 
-## v1.26.1
-2021-11
+## v1.26.2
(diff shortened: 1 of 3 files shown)


adbenitez (approved):
it is fine to release as is already, the fix of the thumbnail can wait, actually I fixed that kinda blindly without knowing why exactly it does happens so it is better if someone else takes a look at that

r10s:
k, thanks for approving and confirming, build is running just now ....

PR #2160, opened 2021-12-02
Add mention

unknownscheme:
Is unnecessary but give to app fresh air


r10s:
what is this pr about? does not look mergeable at a first glance.

PR #2134, opened 2021-11-13
Don't set draft after creating group, instead improve group creation hint

Hocuri:
fix https://github.com/deltachat/deltachat-core-rust/issues/2793

Also see https://github.com/deltachat/deltachat-core-rust/pull/2805

PR #2165, opened 2021-12-08
improve wording of one-to-one-chat hint

r10s:
at #2134 we shortend the hint shown for an empty group massively,
removing greylisting hints and the hint that the recipient can user other mua. also we do not say anything about encryption on group creation.

the result of that shortening is that the creation of a one-to-one chat
is now way too long :)

i suggest shorten the hint in a similar way,
not much would be left then, however, that is maybe not bad :)

suggestions welcome, of course :)

Screen Shot 2021-12-12 at 11 43 52 &nbsp; Screen Shot 2021-12-08 at 12 58 15  
_left: new one-to-one hint in this pr; right: already improved group hint_

EDIT: updated screenshots:

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -359,7 +359,7 @@
     </plurals>
     <string name="chat_no_messages_hint">Send message to %1$s:\n\n• It is okay if %2$s does not use Delta Chat.\n\n• Delivering the first message may take a while and may be unencrypted.</string>
     <!-- The placeholder will be replaced by the name of the recipient in a one-to-one chat. -->
-    <string name="chat_new_one_to_one_hint">Send a first message to %1$s.</string>
+    <string name="chat_new_one_to_one_hint">Send a message. It is okay if %1$s does not use Delta Chat.</string>
     <string name="chat_new_broadcast_hint">In a broadcast list, others will receive messages in their private chats with you.\n\nThe recipients will not be aware of each other.</string>
     <string name="chat_new_group_hint">Others will only see this group after you sent a first message.</string>
     <string name="chat_record_slide_to_cancel">Slide to cancel</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -358,6 +358,8 @@
         <item quantity="other">%d new messages</item>
     </plurals>
     <string name="chat_no_messages_hint">Send message to %1$s:\n\n• It is okay if %2$s does not use Delta Chat.\n\n• Delivering the first message may take a while and may be unencrypted.</string>
+    <!-- The placeholder will be replaced by the name of the recipient in a one-to-one chat. -->
+    <string name="chat_new_one_to_one_hint">Send a first message to %1$s.</string>
     <string name="chat_new_broadcast_hint">In a broadcast list, others will receive messages in their private chats with you.\n\nThe recipients will not be aware of each other.</string>
     <string name="chat_new_group_hint">Others will only see this group after you sent a first message.</string>
     <string name="chat_record_slide_to_cancel">Slide to cancel</string>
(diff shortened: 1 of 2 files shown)


res/values/strings.xml
```diff
@@ -358,6 +358,8 @@
         <item quantity="other">%d new messages</item>
     </plurals>
     <string name="chat_no_messages_hint">Send message to %1$s:\n\n• It is okay if %2$s does not use Delta Chat.\n\n• Delivering the first message may take a while and may be unencrypted.</string>
```
r10s, line 360: this was the old string, it can be deleted once the UI pick up the new one (some string cleanup is needed anyway at some point :)

Hocuri:
As the new string does not tell the user any new information, I'd rather totally leave it away than only saying "Send a first message to …".

The current (old) string tells the user 2 medium-important pieces of information:
- It's OK if the other side doesn't use DC
- The first message may be unencrypted.

It also says "Delivering the first message may take a while" but this information doesn't seem valuable at this point, so we I'd leave this out in any case.

--

The reason that I had proposed to leave out all the medium-important information from the group hint was that there is one highly-important information (you need to send a message so that others see the group) and if the string is shorter, more people will read it. I.e. don't let the medium-important information distract the user from the highly-important information.

(I'm using "highly-important" as "more important than medium-important" here, not as "bad things will happen if the user doesn't know)

--

Conclusion: I'm in favor of:
- Either keeping the current string, but removing the greylisting hint.
- Or completely removing the one-to-one-chat hint.

r10s:
> Either keeping the current string, but removing the greylisting hint (and we could also shorten it in other ways)

so maybe just say:

"Send message to %1$s. It is okay if %2$s does not use Delta Chat."

in contrast to groups with tons of messages, it is _really_ okay if the recipient does not use Delta Chat as things in this case, so that hint may be valuable here.

i'd also like to avoid the doubling of the name somehow, "Send message to %1$s. It is okay if the recipient does not use Delta Chat." - however, that sounds a bit bureaucratic to me, not sure.

adbenitez (approved):

adbenitez:
if user scan QR the message will be encrypted from the beginning, if security is important then user should scan QR from the beginning because otherwise encryption is not safe anyway (from a paranoid point of view) and then it doesn't matters that  much if the first message is unencrypted or not if keys are not verified anyway.

at some point the send button could show the state of the message that will be sent (encrypted/unencrypted) so users can feel safe, currently it is a bit stressful task if you care about encryption you never know for sure if next message will be sent encrypted or not

Hocuri:
What about just "It is okay if %1$s does not use Delta Chat."? I.e. leave away "Send message to %1$s.". This would also prevent the problem with doubling the name.

r10s:
> What about just "It is okay if %1$s does not use Delta Chat."? I.e. leave away "Send message to %1$s.". This would also prevent the problem with doubling the name.

maybe that is okay.

and i agree, that is the most important information, we should not leave that out (as in my initial suggestion).

however, it feels to me as if that misses a bit context, if we just say "It is okay if %1$s does not use Delta Chat."; it would be good to point out _why_ it is okay (it feels to me a bit like if we say "Potatoes are vegetables." - it is true, but why is it important? just my 2 cents, however :)

maybe "Send a first message. It is okay if %1$s does not use Delta Chat." ?

r10s:
@adbenitez oh, just edited the message, so i am not sure, what the "thumb ob belongs to" :)

adbenitez:
> "It is okay if %1$s does not use Delta Chat, just send your message."

+1 :)

Hocuri:
> however, it feels to me as if that misses a bit context

true, to me too

I think I'd slightly prefer "Send a first message. It is okay if %1$s does not use Delta Chat." over the other wording, but they seem both nice.

r10s:
thinking it over after some days, i do not really like the "first message" aspect - the user may have sent a "first message" long time ago, even if the chat appears empty in Delta Chat.

so, maybe just say "Send a message. It is okay if %1$s does not use Delta Chat." - (i'd also slightly prefer that wording as it first says what do to and does not start with cornercases :)

Hocuri (approved):

PR #2169, opened 2021-12-11
remove unused strings

r10s:
there were some left over the time :)

- using the 'Unused Resources' linter in Android Studio
  to get delete candidates

- removing all old 'Contact Request' strings
  as we do not have this anymore in that form - if there are references left on the os, they should be removed.

- checking every other string using ./scripts/grep-string.sh
  that checks for usages on android, ios, node and desktop.

there is still some chance, a string is removed accidentally or will be needed later on, however, we can re-add the string as needed then.

@Simon-Laux @Jikstra maybe you can also have a rough look at the removed strings, but i assume, things should be fine.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -345,7 +345,6 @@
         <item quantity="one">%d new message</item>
         <item quantity="other">%d new messages</item>
     </plurals>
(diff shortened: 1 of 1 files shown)


adbenitez (approved):

r10s:
`chat_no_messages_hint` should be removed as well, esp. as this is a longer text with some work for translators.

PR #2173, opened 2021-12-13
Make ndk-make.sh debug builds (for just one arch) more usable

Hocuri:
Debug builds of the core (like `ndk-make.sh arm64-v8a`) take about a minute, while
a release bulid (`ndk-make.sh`) takes about 15 minutes.

But as most people seem not to use debug builds regularly because
they are tricky to use, here comes an improvement :) If you gave the wrong argument to `ndk-make.sh`, a warning dialog will now pop up inside the app, telling you what you did wrong and what you should do instead.

`ndk-make.sh` writes its argument into the file `ndkArch`.
`getNdkArch()` in `build.gradle` then reads this file and its content is assigned to
`BuildConfig.NDK_ARCH`. `checkNdkArchitecture()` then checks this.

Maybe in another PR I'll also make DC show a helpful error message if someone is completely new to DC and never ran `ndk-make.sh` at all.

### DIFF
--- a/ndk-make.sh
+++ b/ndk-make.sh
@@ -10,9 +10,9 @@
 #
 #     adb shell uname -m
 #
-# or:
-#
-#     adb shell cat /proc/cpuinfo
+# Or you just guess your phone's architecture to be arm64-v8a and if you
+# guessed wrongly, a warning message will pop up inside the app, telling
+# you what the correct one is.
 #
 # The values in the following lines mean the same:
 #
@@ -27,6 +27,9 @@
 # typing `nmake`:
 #
 # nmake() {(cd ../..; ./ndk-make.sh arm64-v8a && ./gradlew installFatDebug; notify-send "install finished")}
+#
+#
+# If anything doesn't work, please open an issue!!
 
 set -e
 echo "starting time: `date`"
@@ -142,5 +145,11 @@ ndk-build
 cd jni
 # Restore old Application.mk:
 echo "$oldDotMk" > Application.mk
+
+if test $1; then
+    echo "NDK_ARCH=$1" > ../ndkArch
+else
+    rm ../ndkArch 2>/dev/null || true # Remove ndkArch, ignore if it doesn't exist
+fi
 
 echo "ending time: `date`"
--- a/src/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -16,10 +16,22 @@
  */
 package org.thoughtcrime.securesms;
 
+import static org.thoughtcrime.securesms.ConversationActivity.CHAT_ID_EXTRA;
+import static org.thoughtcrime.securesms.ConversationActivity.STARTING_POSITION_EXTRA;
+import static org.thoughtcrime.securesms.map.MapDataManager.ALL_CHATS_GLOBAL_MAP;
+import static org.thoughtcrime.securesms.util.RelayUtil.REQUEST_RELAY;
+import static org.thoughtcrime.securesms.util.RelayUtil.acquireRelayMessageContent;
+import static org.thoughtcrime.securesms.util.RelayUtil.getDirectSharingChatId;
+import static org.thoughtcrime.securesms.util.RelayUtil.isDirectSharing;
+import static org.thoughtcrime.securesms.util.RelayUtil.isForwarding;
+import static org.thoughtcrime.securesms.util.RelayUtil.isRelayingMessageContent;
+import static org.thoughtcrime.securesms.util.RelayUtil.resetRelayingMessageContent;
+
 import android.content.Intent;
-import android.text.TextUtils;
 import android.net.Uri;
 import android.os.Bundle;
+import android.text.TextUtils;
+import android.util.Log;
 import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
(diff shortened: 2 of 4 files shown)


ndk-make.sh
```diff
@@ -10,9 +10,9 @@
 #
 #     adb shell uname -m
 #
-# or:
-#
-#     adb shell cat /proc/cpuinfo
```
Hocuri, line 15: `adb shell cat /proc/cpuinfo` creates a very long output, which makes it a lot less helpful.

src/org/thoughtcrime/securesms/ConversationListActivity.java
```diff
@@ -128,6 +130,63 @@ protected void onCreate(Bundle icicle, boolean ready) {
...
+          message = "Apparently you used `ndk-make.sh " + BuildConfig.NDK_ARCH + "`, but this device is " + arch + ".\n\n" +
+                  "You can use the app, but changes you made to the Rust code were not applied.\n\n" +
+                  "To compile in your changes, you can:\n" +
+                  "- Either run `ndk-make.sh " + arch + "` to build only for " + arch + " in debug mode\n" +
+                  "- Or run `ndk-make.sh` without argument to build for all architectures in release mode\n\n" +
+                  "If something doesn't work, please open an issue at https://github.com/deltachat/deltachat-android/issues!!";
+        }
+        Log.e(TAG, message);
+
+        if (!Prefs.getBooleanPreference(this, NDK_ARCH_WARNED, false)) {
```
Hocuri, line 179: The user is only warned on first startup. Not completely sure about this - maybe it should be on every startup, maybe we should show the warning as a toast (which nobody can completely read though) after the first startup.

src/org/thoughtcrime/securesms/ConversationListActivity.java
```diff
@@ -128,6 +130,63 @@ protected void onCreate(Bundle icicle, boolean ready) {
...
+        if (arch.equals("")) {
+          message = "This phone has the unknown architecture " + archProperty + ".\n\n"+
+                  "Please open an issue at https://github.com/deltachat/deltachat-android/issues.";
+        } else {
+          message = "Apparently you used `ndk-make.sh " + BuildConfig.NDK_ARCH + "`, but this device is " + arch + ".\n\n" +
+                  "You can use the app, but changes you made to the Rust code were not applied.\n\n" +
+                  "To compile in your changes, you can:\n" +
+                  "- Either run `ndk-make.sh " + arch + "` to build only for " + arch + " in debug mode\n" +
+                  "- Or run `ndk-make.sh` without argument to build for all architectures in release mode\n\n" +
+                  "If something doesn't work, please open an issue at https://github.com/deltachat/deltachat-android/issues!!";
```
Hocuri, line 175: The text is not linkified - "ndk-make.sh" was also linkified when I enabled linkifying, and it's probably not woth it as most developers will not open an issue from their phone, anyway I guess.

src/org/thoughtcrime/securesms/ConversationListActivity.java
adbenitez: another option could be to properly set the `ndk.abiFilters` property in `build.gradle` and don't have any unnecessary logic in the app for a test usecase

adbenitez: btw, there are apps like https://f-droid.org/packages/com.kgurgul.cpuinfo/ that shows the device's ABI, I think it is unnecessary that DC does this in production releases

adbenitez:
btw, I just realized that when compiling only one arch the previous `.so` files of other architectures at `./libs/*` are not removed, would be nice to also remove arch folders there not only on `./jni/` since `ndk-build` copies them to `./libs`

src/org/thoughtcrime/securesms/ConversationListActivity.java
Hocuri: In production releases the method will probably be optimized away anyway because I am doing `if (BuildConfig.DEBUG)` above.

And, installing an extra app is way more effort for something that should be as intuitive as possible for new developers.

If we set the `ndk.abiFilters`, then installing would fail with a not-very-helpful error message, and I actually quite like it currently that if I don't care about the core, I can just install the debug version to any device and an older (latest compiled) core will be used.

Hocuri:
> because otherwise if you compile the wrong arch you might still get an old version of the `.so` you compiled previously

That's what I tried to explain in the alert dialog message. Also here, I actually like this behavior currently because it means that when developing for the UI and not caring about the core, I can just install the app to any device and get _some_ core, even if it's some days or weeks old (or whenever I last compiled the core for this arch).

adbenitez:
having cores with different code states for different architectures sounds awful to me, like when you want to test something in another device but you forgot to compile the updated core and then waste time testing with an old core and wondering why the bug appears in some phone and not in the other

src/org/thoughtcrime/securesms/ConversationListActivity.java
adbenitez: fine then, but anyway compared to the whole process of compiling core and creating the apk installing an extra app sounds quite easy to me :D

adbenitez (approved):
besides some opinions, lgtm

Hocuri:
> having cores with different code states for different architectures sounds awful to me

That's why I made this PR to add a warning - We can generally talk about changing this, for me it would slightly slow development down. Maybe this warning is enough, maybe not, in any case if others are reading this I'd be curious for your experiences with building DC Android.

PR #2175, opened 2021-12-17
allow to remove all messages from current chat

adbenitez:
to decide: when chat has no messages, what to do? hide the menu item? show an error toast on menu item selected? show a dialog? 

currently the same dialog is show for all cases so when there is no messages the "delete 0 messages?" dialog is shown and does nothing

### DIFF
--- a/res/menu/conversation_clear.xml
+++ b/res/menu/conversation_clear.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <item android:title="@string/menu_clear_chat"
+          android:id="@+id/menu_clear_chat"/>
+
+</menu>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -192,6 +192,7 @@
     <string name="menu_add_attachment">Add Attachment</string>
     <string name="menu_leave_group">Leave Group</string>
     <string name="menu_delete_chat">Delete Chat</string>
+    <string name="menu_clear_chat">Clear Chat</string>
     <string name="ask_delete_named_chat">Are you sure you want to delete \"%1$s\"?</string>
     <string name="menu_delete_messages">Delete Messages</string>
     <string name="delete_contact">Delete Contact</string>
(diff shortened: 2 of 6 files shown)


adbenitez:
this do what is described in https://github.com/deltachat/deltachat-core-rust/issues/1299 but in the Android side, could be tweaked when such core api is added

res/menu/conversation_clear.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <item android:title="@string/menu_delete_messages"
```
adbenitez, line 4: I reused this string but maybe a "clear chat"/"clear history" string could be added

Simon-Laux:
is there a core function for it or is this using ui code to delete each message?

adbenitez:
> ftr, desktop does a similar thing since [deltachat/deltachat-desktop#2909](https://github.com/deltachat/deltachat-desktop/pull/2909) - if there are no issues with that, we can maybe revive this pr.

@r10s so desktop already have this and some users already have complained in community channels about this issue that they can do it in desktop but can't in Android, is something blocking from merging this PR? 
cc @Hocuri

adbenitez:
desktop is using an unstranslated string "clear chat" it seems: 
https://github.com/deltachat/deltachat-desktop/pull/2909/files#diff-5bd8446ffbbe28eb2eacda360fd1fa0640832786a9170440fe58275a4190102bR36

cc @Simon-Laux

r10s:
yip, i think, we can get that in, at some point a core-function would be nice and more performant, but it seems good enough already this way.

to the wording:

whatsapp (ios/android) is using the term "Clear Chat" as well  
telegram (ios) uses "Clear Messages", telegram (macos) uses "Clear Chat History" (not really consistent)
signal does not have this function, but woth "Clear message history" you can clear all chats

so, i'd say "Clear Chat", as already introduced on desktop seems okay, esp. as in the following confirmation we ask explicitly to confirm deletion of "N messages".

so, what's missing is only the string "Clear Chat" :)

adbenitez:
> so, what's missing is only the string "Clear Chat" :)

ok, changed to use "Clear Chat" string

r10s (approved):
lgtm, i think, this can go to 1.36 :)

PR #2178, opened 2021-12-23
don't allow to add POI if user can't send in current chat

adbenitez:
a fix for an issue reported by @link2xt  recently in the DC-Android group

### DIFF
--- a/src/org/thoughtcrime/securesms/map/MapActivity.java
+++ b/src/org/thoughtcrime/securesms/map/MapActivity.java
@@ -306,7 +306,7 @@ public class MapActivity extends BaseActivity implements Observer,
     }
 
     private boolean handleAddPoiClick(LatLng point) {
-        if (chatId == 0) {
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/map/MapActivity.java
```diff
@@ -306,7 +306,7 @@ private boolean handleInfoWindowClick(LatLng point) {
     }
 
     private boolean handleAddPoiClick(LatLng point) {
-        if (chatId == 0) {
+        if (chatId == 0 || !DcHelper.getContext(this).getChat(chatId).canSend()) {
```
Hocuri, line 309: Just something I noticed / FYI: The call to `getChat()` will make clicking on the map to add a POI a little bit slower. But I checked it, `getChat()` can't be avoided here. I mean, it could, but only by loading the chat before the button is clicked, and as we don't want to slow down opening the map either, we would have to do it asynchronously, which wouldn't be worth the effort.

Hocuri (approved):
Didn't test, code lgtm.

PR #2179, opened 2021-12-24
display SVG images in chats

adbenitez:
now that we are using a lib to display the SVG QR codes, reusing that lib to display SVG files also in chats is cheap, I have taken some example from glide's repo to add SVG support as a glide plugin so no change is needed in all parts of the code that load the images with glide

### DIFF
--- a/src/org/thoughtcrime/securesms/glide/svg/SvgDecoder.java
+++ b/src/org/thoughtcrime/securesms/glide/svg/SvgDecoder.java
@@ -1,0 +1,40 @@
+package org.thoughtcrime.securesms.glide.svg;
+
+import static com.bumptech.glide.request.target.Target.SIZE_ORIGINAL;
+
+import androidx.annotation.NonNull;
+import com.bumptech.glide.load.Options;
+import com.bumptech.glide.load.ResourceDecoder;
+import com.bumptech.glide.load.engine.Resource;
+import com.bumptech.glide.load.resource.SimpleResource;
+import com.caverock.androidsvg.SVG;
+import com.caverock.androidsvg.SVGParseException;
+import java.io.IOException;
+import java.io.InputStream;
+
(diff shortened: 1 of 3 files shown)


r10s:
even if initial implementation is cheap for android, overall maintenance may not.

i am overall sceptical if we really want to support that format internally, esp. as the external support may be good enough.

we have to add this initial support on all systems and also, all systems display svg differently, we might get bug reports for that and have to deal with that.

also, that makes it hard to remove svg support again if we decide to go for sth. else for qr-codes (it is always hard to remove a feature, see watch-send ;)

also, i am wondering, if svg is really a "thing", i mean stickers mostly use png or lottie? are there really lots of svg stickers out there? (nb: i was recently wondering if we could have generated the qr-codes in lottie :)

sorry for ranting here, maybe i am too pessimistic :)

r10s:
nb: for the just created two webxdc icons ([[1]](https://github.com/deltachat/interface/blob/master/icons/webxdc-white-border.svg), [2] https://github.com/deltachat/interface/blob/master/icons/webxdc-logo-no-border.svg): one cannot be displayed with ios' system viewer nor with dc desktop (i was not aware that desktop shows svg) nor with github - only with inkscape - for whatever reason.

of course that is only anecdotal and not empirically and can probably be fixed on creation of the logo somehow, but we noticed inconsistencies with svg quote often  - and ppl will not know that and we get the issues ;)

adbenitez:
I am in favor to make it work the same in all platforms, I am fine if SVGs are displayed as file in all DCs.

It is a bit sad that an standard like SVG has such crappy support across platforms :(

adbenitez:
otoh I don't think it is catastrophic and I think there is something wrong with that SVG logo, it doesn't works not even in my FireFox

adbenitez:
closing this for now due to stabilization phase and this PR will probably open new issues related to SVG rendering

PR #2180, opened 2021-12-24
add support for .tgs stickers

adbenitez:


adbenitez:
the stickers are show as statics stickers, still haven't managed to make them auto-play when the image is loaded

src/org/thoughtcrime/securesms/components/ThumbnailView.java
```diff
@@ -257,6 +257,12 @@ protected void onPostExecute(Boolean success) {
           }
         }.executeOnExecutor(AsyncTask.THREAD_POOL_EXECUTOR);
       }
+      if(slide.hasSticker())
```
adbenitez, line 260: it turned out that the lottie stickers were not animated because the transition (`.transition(withCrossFade())`) I just removed the transition for stickers, if someone knows more about glide and have an idea about how to make the lottie drawable work with the transitions, patches are welcome :D

adbenitez:
outgoing tgs stickers are shown fine but incoming .tgs are show as file, have to check if it is a core issue

Simon-Laux:
also shouldn't the other platforms support this format too then?

adbenitez:
it is like for SVG, Desktop supports them but Android doesn't, this is about reception not sending, (well currently there is even no sticker picker on android) so android could start supporting receiving .tgs and showing them as stickers

adbenitez:
> also shouldn't the other platforms support this format too then?

mmh now I understand what you mean, well in that case anyway when you send them they are show as stickers so maybe then send as file as well until supported on all platforms?

Simon-Laux:
If we want it we can make issues on the ios and desktop repo asking them to implement it, too.

at-least for Desktop the display part seems easy enough: https://github.com/airbnb/lottie-web (MIT)

for iOS the "official" lib is licensed Apache 2, I'm not sure if that's compatible with our iOS app, but I'd guess yes.

r10s:
closing stale pr, we can always come back this attempt if needed

PR #2186, opened 2022-01-10
Make it possible to create an encrypted database

Hocuri:
Screenshot_20220110-131339 (1)

Depends on ~~https://github.com/deltachat/deltachat-core-rust/pull/2956~~ https://github.com/deltachat/deltachat-core-rust/pull/2980.

Currently:
- Before Android 6 (API 23), there was no KeyStore yet. Currently, you can still create an encrypted account, but the passphrase is stored in clear-text, because that's how Signal is doing things.
- There is no option to manually set the keyphrase or to unlock via fingerprint. This has the advantage that we never need user input for unlocking the accounts and can simply unlock all accounts on startup and the obvious disadvantage that anyone with access to the unlocked device can open DC.
- Each account is encrypted with a different passphrase.
- When importing, you can chosse to encrypt the account.

Findings:
- It may be useful (better UX) to have an API that tells you whether an opened account is encrypted, but no high-prio. Like, have a function `dc_accounts_is_encrypted(account_id) -> {Unencrypted | Encrypted-Closed | Encrypted-Opened}`
- Switching to encrypted after a context has already been created but not configured is kind of complex, but necessary as we already create an unconfigured context before we know whether the user wants to encrypt; but it's OKish already, and I'm not sure we should create a core function just for this, so maybe it's fine:

```java
    public void switchToEncrypted(Activity activity) {
        DcAccounts accounts = DcHelper.getAccounts(activity);
        DcContext selectedAccount = accounts.getSelectedAccount();
        if (selectedAccount.isConfigured() == 1) {
            throw new RuntimeException("Can't switch to encrypted account if already configured");
        }

        int selectedAccountId = selectedAccount.getAccountId();
        accounts.addClosedAccount();
        accounts.removeAccount(selectedAccountId);

        DcContext newAccount = accounts.getSelectedAccount();
        DatabaseSecret secret = DatabaseSecretProvider.getOrCreateDatabaseSecret(activity);
        newAccount.open(secret.asString());
    }
```

### DIFF
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -313,13 +313,13 @@ JNIEXPORT jlong Java_com_b44t_messenger_DcContext_createContextCPtr(JNIEnv *env,
 }
 
 
-JNIEXPORT jint Java_com_b44t_messenger_DcContext_open(JNIEnv *env, jobject obj, jstring passphrase)
+JNIEXPORT jboolean Java_com_b44t_messenger_DcContext_open(JNIEnv *env, jobject obj, jstring passphrase)
 {
     CHAR_REF(passphrase);
-    jint ret = dc_context_open(get_dc_context(env, obj), passphrasePtr);
+    jboolean ret = dc_context_open(get_dc_context(env, obj), passphrasePtr);
     CHAR_UNREF(passphrase);
     return ret;
-} // TODO should maybe return jboolean
+}
 
 
 JNIEXPORT jboolean Java_com_b44t_messenger_DcContext_isOpen(JNIEnv *env, jobject obj)
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -206,6 +206,12 @@ JNIEXPORT jint Java_com_b44t_messenger_DcAccounts_addAccount(JNIEnv *env, jobjec
 }
 
 
+JNIEXPORT jint Java_com_b44t_messenger_DcAccounts_addClosedAccount(JNIEnv *env, jobject obj)
+{
+    return dc_accounts_add_closed_account(get_dc_accounts(env, obj));
+}
+
+
 JNIEXPORT jint Java_com_b44t_messenger_DcAccounts_migrateAccount(JNIEnv *env, jobject obj, jstring dbfile)
 {
     CHAR_REF(dbfile);
@@ -304,6 +310,21 @@ JNIEXPORT jlong Java_com_b44t_messenger_DcContext_createContextCPtr(JNIEnv *env,
     CHAR_UNREF(dbfile)
     CHAR_UNREF(osname);
     return contextCPtr;
+}
+
+
+JNIEXPORT jint Java_com_b44t_messenger_DcContext_open(JNIEnv *env, jobject obj, jstring passphrase)
+{
+    CHAR_REF(passphrase);
+    jint ret = dc_context_open(get_dc_context(env, obj), passphrasePtr);
+    CHAR_UNREF(passphrase);
+    return ret;
+} // TODO should maybe return jboolean
+
+
+JNIEXPORT jboolean Java_com_b44t_messenger_DcContext_isOpen(JNIEnv *env, jobject obj)
+{
+    return dc_context_is_open(get_dc_context(env, obj));
 }
 
 
--- a/src/org/thoughtcrime/securesms/ApplicationContext.java
+++ b/src/org/thoughtcrime/securesms/ApplicationContext.java
@@ -45,6 +45,7 @@ import java.util.concurrent.TimeUnit;
 //import com.squareup.leakcanary.LeakCanary;
 
 public class ApplicationContext extends MultiDexApplication {
+  private static final String TAG = ApplicationContext.class.getSimpleName();
 
   public DcAccounts             dcAccounts;
   public DcContext              dcContext;
@@ -83,8 +84,10 @@ public class ApplicationContext extends MultiDexApplication {
     for (int accountId : allAccounts) {
       DcContext ac = dcAccounts.getAccount(accountId);
       if (!ac.isOpen()) {
-        DatabaseSecret secret = DatabaseSecretProvider.getOrCreateDatabaseSecret(this);
-        ac.open(secret.asString());
+        DatabaseSecret secret = DatabaseSecretProvider.getOrCreateDatabaseSecret(this, accountId);
+        int res = ac.open(secret.asString());
+        if (res == 1) Log.i(TAG, "Successfully opened account " + accountId);
+        else Log.e(TAG, "Error opening account " + accountId);
       }
     }
     if (allAccounts.length == 0) {
--- a/src/org/thoughtcrime/securesms/ApplicationContext.java
+++ b/src/org/thoughtcrime/securesms/ApplicationContext.java
@@ -86,7 +86,7 @@ public class ApplicationContext extends MultiDexApplication {
       if (!ac.isOpen()) {
         DatabaseSecret secret = DatabaseSecretProvider.getOrCreateDatabaseSecret(this, accountId);
         int res = ac.open(secret.asString());
-        if (res == 1) Log.i(TAG, "Successfully opened account " + accountId);
+        if (res == 1) Log.i(TAG, "Successfully opened account " + accountId + ", path: " + ac.getBlobdir());
         else Log.e(TAG, "Error opening account " + accountId);
       }
     }
--- a/src/org/thoughtcrime/securesms/ApplicationContext.java
+++ b/src/org/thoughtcrime/securesms/ApplicationContext.java
@@ -84,10 +84,15 @@ public class ApplicationContext extends MultiDexApplication {
     for (int accountId : allAccounts) {
       DcContext ac = dcAccounts.getAccount(accountId);
       if (!ac.isOpen()) {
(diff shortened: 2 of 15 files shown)


cyBerta:
Nice, it looks good on a first glance.

I wonder if we should hide the encryption option for pre android 6 to avoid a false sense of security. Otherwise, it should be theoretically possible to encrypt secrets and store key material in the keychain on pre Android M. IIRC an additional step using use asymmetrical encryption to encrypt the AES key and store it in your preferences would be needed. That method should work down to Android 4.3.

src/org/thoughtcrime/securesms/connect/AccountManager.java
```diff
@@ -135,6 +134,22 @@ public void switchAccountAndStartActivity(Activity activity, int destAccountId,
         }
     }
 
+    public void switchToEncrypted(Activity activity) {
```
cyBerta, line 137: If we keep this approach, it would be nice to indicate in the method declaration that this method throws a RuntimeException, (maybe a more specific IllegalStateException?)

r10s:
> Switching to encrypted after a context has already been created but not configured is kind of complex

for that reason, in the chat was discussed to offer a button "create encrypted account" instead of a flag in the account creation settings. that avoids the first account being encrypted, however, for a first try, that is okay.

however, it if now works on android with a switch, of course that is even better. but note, that the switch may be removed completely. if things work out and are not notable slower, we may decide to always create encrypted accounts.

@cyBerta same for ios - a button that creates an encrypted context is probably easier than a switch, and would be sufficient for now.

link2xt:
> that avoids the first account being encrypted, however, for a first try, that is okay.

Why is there a need for "first account"? Core does not create it anymore since https://github.com/deltachat/deltachat-core-rust/pull/2500

r10s:
> Why is there a need for "first account"?

when the app is started for the very first time, the welcome-screen is shown directly - and that already needs a dc_context_t for $reasons.

that could be done differently, of course, but that would require refactorings in an already complex and sensitive area - for a solution as @Hocuri did above.

Hocuri:
> "create encrypted account" instead of a flag in the account creation settings.

I'd actually guess that this would have been about equally complex (less lines, but in more areas), because adding an extra button to the recyclerview list isn't as straightforward as adding a toggle to the XML.

src/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -77,7 +79,15 @@ public void onCreate() {
 
     dcAccounts = new DcAccounts("Android "+BuildConfig.VERSION_NAME, new File(getFilesDir(), "accounts").getAbsolutePath());
     AccountManager.getInstance().migrateToDcAccounts(this);
-    if (dcAccounts.getAll().length == 0) {
+    int[] allAccounts = dcAccounts.getAll();
+    for (int accountId : allAccounts) {
```
cyBerta, line 83: @link2xt @Hocuri I wonder if this is something we want to let the account manager handle, by passing the passphrase to the account manager

cyBerta:
> > wonder if we should hide the encryption option for pre android 6
> 
> it is flagged as experimental, so that's probably fine for now. also, we need the experience of testing encryption on as many devices as possible, esp. on older and slower ones, so i would not remove that option.

Maybe then we should then indicate that the account encryption will have no effect on the data security on pre android 6 devices.

cyBerta:
> @cyBerta same for ios - a button that creates an encrypted context is probably easier than a switch, and would be sufficient for now.

@r10s my understanding of the current api is, that creating new accounts are per default unencrypted as long as we add new accounts with [dc_accounts_add_closed_account](https://github.com/deltachat/deltachat-core-rust/pull/2956/files#diff-7d5e9c113766adfcca5e5b842258ba90cab966d995b7df97eb0e1cafbd18781cR2588.) We can opt-in account encryption by opening a closed dc_context with a passphrase.  The only constraint is that the dc_context was not yet configured. If we open the dc_context without a passphrase we get an unencrypted account.
So why don't we replace the calls to [dc_accounts_add_account](https://github.com/deltachat/deltachat-core-rust/blob/8352c44590b950f0e76665f31b8debc498ac7c22/deltachat-ffi/deltachat.h#L2571) with `dc_accounts_add_closed_account` during the account setup? @Hocuri That should also make the `switchToEncrypted` method kind of superfluous?!

src/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -176,6 +182,10 @@ public void onTextChanged(CharSequence s, int start, int before, int count) { }
 
             int certCheckFlags = DcHelper.getInt(this, "imap_certificate_checks");
             certCheck.setSelection(certCheckFlags);
+            encryptCheckbox.setHeight(0);
```
Hocuri, line 185: I'm setting the height to 0 here instead of setting the visibility to GONE because setting the visibility to GONE doesn't work well with constraint layouts: The `inbox` TextView is constrained to `Top_toBottomOf="@id/encrypt_checkbox"` and if the visibility was GONE, this constraint would probably fail, even though I didn't try it out (it did last time I tried to do this with constraintLayout).

(Also, the visibility is changed everytime the user hits "Advanced", so the visibility would have to be changed in the onClickListener, not here.)

cyBerta: +1 for an easy solution

ftr, there's a Barrier class that should help in these cases: https://developer.android.com/reference/androidx/constraintlayout/widget/Barrier

Hocuri:
> > > wonder if we should hide the encryption option for pre android 6
> > 
> > 
> > it is flagged as experimental, so that's probably fine for now. also, we need the experience of testing encryption on as many devices as possible, esp. on older and slower ones, so i would not remove that option.
> 
> Maybe then we should then indicate that the account encryption will have no effect on the data security on pre android 6 devices.

I think as it's flagged as experimental, that's fine for now. And maybe we'll just prompt the user for a password on pre android 6, then we won't need such a warning.

src/org/thoughtcrime/securesms/WelcomeActivity.java
```diff
@@ -138,12 +143,37 @@ private void startImportBackup() {
...
-                                    .setMessage(String.format(getResources().getString(R.string.import_backup_ask), backupFile))
+                                    .setView(gl)
                                     .setNegativeButton(android.R.string.cancel, null)
-                                    .setPositiveButton(android.R.string.ok, (dialog, which) -> startImport(backupFile, null))
+                                    .setPositiveButton(android.R.string.ok, (dialog, which) -> startImport(backupFile, null, encryptCheckbox.isChecked()))
                                     .show();
+
+
+
+//                            new AlertDialog.Builder(this)
```
cyBerta, line 171: is this still needed / planned to be reused?

jni/dc_wrapper.c
```diff
@@ -307,6 +313,21 @@ JNIEXPORT jlong Java_com_b44t_messenger_DcContext_createContextCPtr(JNIEnv *env,
 }
 
 
+JNIEXPORT jint Java_com_b44t_messenger_DcContext_open(JNIEnv *env, jobject obj, jstring passphrase)
+{
+    CHAR_REF(passphrase);
+    jint ret = dc_context_open(get_dc_context(env, obj), passphrasePtr);
+    CHAR_UNREF(passphrase);
+    return ret;
+} // TODO should maybe return jboolean
```
Hocuri, line 322: Should this return `jboolean`? (@r10s or others)

src/org/thoughtcrime/securesms/WelcomeActivity.java
Hocuri: No, I forgot to remove it

src/org/thoughtcrime/securesms/crypto/KeyStoreHelper.java
```diff
@@ -0,0 +1,208 @@
...
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.KeyGenerator;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.GCMParameterSpec;
+
+public final class KeyStoreHelper {
+
+  private static final String ANDROID_KEY_STORE = "AndroidKeyStore";
+  private static final String KEY_ALIAS         = "SignalSecret";
```
adbenitez, line 46: Signal? why not Delta? :D

adbenitez (approved):

jni/dc_wrapper.c
cyBerta: on DC iOS I return a boolean: https://github.com/deltachat/deltachat-ios/pull/1438/files#diff-a5590eddfe26770d5692a338e715ab8f047049b4b63c9ea802f6de4d43b097c4R127

For now I'm only logging if opening fails though

r10s: i would also return a `jboolean`, that makes the java part easier to read, so `return ret != 0` (ios uses `ret == 1` but i would keep the style used elsewhere here)

Hocuri:
@r10s @cyBerta do you still want to review more thoroughly or shall I merge this in so it can go to the nightlies?

cyBerta (approved):
I haven't built this PR, but codewise it looks good to me.

link2xt:
+1 on merging and testing on nightlies.

PR #2174, opened 2021-12-14
webxdc

r10s:
implement the new webxdc functionality as of https://github.com/deltachat/deltachat-core-rust/pull/2826 being checked out.

- [x] use core for sending and receiving updates
- [x] create basic webxdc view
- [x] create a demo app
- [x] adapt to new zipped webxdc format
- [x] maybe allow querying the name (`selfName()`, similar to `selfAddr()`) 
- [x] show app name in title bar
- [x] nicer in-chat-presentation with icon and app name ...
- [x] ... same, nicer presentation for drafts
- [x] allow configuration from staging area (cmp. `AttachmentManager`, that requires a draft to be set much sooner)
- [x] reuse existing drafts for webxdc messages for sending (and for saving as draft again) (needed to preserve configuration from staging area)
- [x] hide "search" from three-dot-menu (?)
- [x] swipe-to-reply crashes
- [x] show .xdc in files tab

for review: most tricky part is the draft handling - webxdc drafts cannot be just created from scratch, therefore, explicitly for webxdc drafts, we reuse the message-id. i trired to change as few as possible in draft handling, as this is subject to cause troubles. i also reused DocumentSlide for webxdc, that seems to be the less-invasive thing.

# how an app could look like and some more background

<a name="example"></a>  

source code of the app (please forgive my rusty js skills): https://github.com/r10s/webxdc-poll

simulator and a minimal app that can be used to start development: https://github.com/deltachat/webxdc-dev

faq ...

- how can i send/receive an updates? where is an api overview?
  -> https://github.com/deltachat/deltachat-core-rust/blob/web3x/draft/webxdc-dev-reference.md

- i need to identify my own updates!
  -> compare addresses from payload against your own

- i need the sending time or a counter in the payloads!
  -> add time or whatever to the payload

- i need the receiving time of an update
  -> in case you are receiving the update from `setUpdateListener()`,
     the current time is pretty close.
     when using getAllUpdates(), you cannot, we may add that in the future.

- i want to use local storage
  -> you cannot, currently, you need to replay status updates.
     but we may add sth. for that in the future.

seems as if not much "general" metadata is needed, we may enhance that in the future, [some in-depth discussions.](https://github.com/deltachat/deltachat-core-rust/issues/2913)

based on @Simon-Laux's [prototype working without help of core](https://github.com/deltachat/deltachat-android/compare/w30_prototype).

PR #2188, opened 2022-01-16
[WIP] Try for webxdc fix

Hocuri:
Recognizing as a webxdc app works now, but opening it doesn't.

The problem was that on Android 10, `data.getData().toString().endsWith(".xdc")` can't be used to check for the type (you get `content:// URIs, and they don't have a file ending).

The correct file name is computed in `getManuallyCalculatedSlideInfo()`/`getContentResolverSlideInfo()`.

No idea what the new bug is.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -16,6 +16,13 @@
  */
 package org.thoughtcrime.securesms;
 
+import static org.thoughtcrime.securesms.NewConversationActivity.MAILTO;
+import static org.thoughtcrime.securesms.TransportOption.Type;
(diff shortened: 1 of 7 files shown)


r10s:
> No idea what the new bug is.

i assume you are taling about the `"index.html" not found"` error:

maybe `int appMessageId = b.getInt("appMessageId");` in `Webxdc.onCreate()` returns an already outdated webxdc message that was overwritten somewhere else using `setDraft()`. 

i had a similar bug before. i'd add a breakpoint there and see what the message-ids look like.

nb: this new bug is also present on android<10, so, eg. on android9 drafts work without this pr.

apart from that, drafts look great on android10 now!

r10s:
i added a commit to https://github.com/deltachat/deltachat-android/pull/2174 based on this pr, thanks @Hocuri for the findings!

PR #2193, opened 2022-01-18
show webxdc errors when sending fails

### LINKED ISSUE #2190: trying to send a too large .xdc should result in a visible error
at least a toast, so seems an `error!` is missing in the core.

r10s:
this happens only in rare cases as usually the message is always added
and errors happen far later.

but some conditions are checked very soon (webxdc sizes),
and these let an message not even being added.

note, that this pr only shows the error,
it does not change draft handling or so -
it would eg. be better if the message is re-converted to a draft,
however, that would be a larger refactoring and maybe not worth the effort.

closes #2190


Hocuri (approved):
Code LGTM, didn't test

PR #2199, opened 2022-01-22
webxdc: warn when the app tries sendUpdate() and the user cannot send

r10s:
show a warning when the app tries to use sendUpdate() for contact requests or other chats the user cannot send to.

this and https://github.com/deltachat/deltachat-core-rust/pull/3001 already catches many cases and improves things a lot, we'll see how that works out in practice.

wording suggestions welcome :)


adbenitez (approved):
nice!

PR #2203, opened 2022-01-26
tweak webxdc bubble layout

r10s:
this pr aims to do some _easy_ things to improve the layout of webxdc bubbles:

- larger icon - this came up in discussions here and there
- longer summary line - it is now the full bubble width, the summary is only one line for $reasons (changing this is out of scope of this pr)
- as title left of icon and summary below looks weird, also the title is moved down

before / after:

 

if we agree in general on that layout, we can improve things further - as making the bubble as wide as the normal bubbles (tbh, no idea where to tweak things for that). i was also thinking here and there that landscape icons as on telegram may be nice, however, not totally sure. having this square format fits better to "apps" - the landscape is more a preview of some content.

### DIFF
--- a/res/layout/webxdc_view.xml
+++ b/res/layout/webxdc_view.xml
@@ -10,46 +10,37 @@
                       android:clickable="false"
                       android:focusable="false"
                       android:gravity="center_vertical"
-                      android:orientation="horizontal">
+                      android:orientation="vertical">
 
             <androidx.appcompat.widget.AppCompatImageView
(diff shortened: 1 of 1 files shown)


adbenitez (approved):
better than before, still would like that the summary just wrap the text instead of a single line

r10s:
> better than before, still would like that the summary just wrap the text instead of truncating the text

yip, but that is sth. for another release for $reasons ;)

PR #2205, opened 2022-01-27
avoid caching resources between webxdc apps

### LINKED ISSUE #2204: resources leak to other webxdc
as reported by @adbenitez:

> for example:
>
> include jQuery lib in one app (in index.html and package with jquery.min.js file),
>
> open the app,
>
> now create another app that doesn't includes jquery.min.js file, just add the script tag in its index.html, the app will be able to use jquery.min.js from the other instance

and:

> also the opposite: if you first tried a version without including jquery then pushed another modified version to the phone, the new app can't get jquery file until you restart the app, this was driving me crazy because errors pointing to wrong/old line of code etc

unwanted caching can explain the first observation, not sure about the second.

r10s:
closes #2204 

did not test that yet ...


adbenitez (approved):
yay! it works

r10s:
great! thanks for testing!

PR #2206, opened 2022-01-27
Add OnlyFetchMvbox option

Hocuri:
Depends on ~~https://github.com/deltachat/deltachat-core-rust/pull/3014~~ https://github.com/deltachat/deltachat-core-rust/pull/3028

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -607,7 +607,7 @@
     <string name="pref_send_copy_to_self">Send Copy to Self</string>
     <string name="pref_auto_folder_moves">Automatic Moves to DeltaChat Folder</string>
     <string name="pref_auto_folder_moves_explain">Chat conversations are moved to avoid cluttering the Inbox</string>
-    <string name="pref_only_fetch_mvbox_title">Only fetch from the DeltaChat folder</string>
+    <string name="pref_only_fetch_mvbox_title">Only Fetch from DeltaChat Folder</string>
     <string name="pref_only_fetch_mvbox_explain">Ignore all other folders. Regardless of this setting, the Spam folder is sometimes checked for replies to known messages.</string>
     <string name="pref_show_emails">Show Classic E-Mails</string>
     <string name="pref_show_emails_no">No, chats only</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -608,7 +608,7 @@
     <string name="pref_auto_folder_moves">Automatic Moves to DeltaChat Folder</string>
     <string name="pref_auto_folder_moves_explain">Chat conversations are moved to avoid cluttering the Inbox</string>
     <string name="pref_only_fetch_mvbox_title">Only fetch from the DeltaChat folder</string>
-    <string name="pref_only_fetch_mvbox_explain">This will ignore all other folders. Regardless of this setting, the Spam folder is sometimes checked for replies to known messages.</string>
+    <string name="pref_only_fetch_mvbox_explain">Ignore all other folders. Regardless of this setting, the Spam folder is sometimes checked for replies to known messages.</string>
     <string name="pref_show_emails">Show Classic E-Mails</string>
     <string name="pref_show_emails_no">No, chats only</string>
     <string name="pref_show_emails_accepted_contacts">For accepted contacts</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -608,7 +608,7 @@
     <string name="pref_auto_folder_moves">Automatic Moves to DeltaChat Folder</string>
     <string name="pref_auto_folder_moves_explain">Chat conversations are moved to avoid cluttering the Inbox</string>
     <string name="pref_only_fetch_mvbox_title">Only Fetch from DeltaChat Folder</string>
-    <string name="pref_only_fetch_mvbox_explain">Ignore all other folders. Regardless of this setting, the Spam folder is sometimes checked for replies to known messages.</string>
+    <string name="pref_only_fetch_mvbox_explain">Ignore other folders. Requires your server to move chat messages to the DeltaChat folder.</string>
     <string name="pref_show_emails">Show Classic E-Mails</string>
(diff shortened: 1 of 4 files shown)


r10s:
maybe, for now, move that to "Experimental" options, so that we do not again raise wrong expectations again, if that stands in the way of something.

(but that is no blocker and can also be done in a subsequent pr, maybe together with the other pretty minor thing "Watch Send Folder")

Hocuri:
I think that it should be close to "Watch Sent Folder" and "Automatic Moves to DeltaChat Folder", because toggling it also toggles these two. If we want to make clear that it's experimental, maybe write "(Experimental)" before the summary?

res/values/strings.xml
```diff
@@ -607,6 +607,8 @@
     <string name="pref_send_copy_to_self">Send Copy to Self</string>
     <string name="pref_auto_folder_moves">Automatic Moves to DeltaChat Folder</string>
     <string name="pref_auto_folder_moves_explain">Chat conversations are moved to avoid cluttering the Inbox</string>
+    <string name="pref_only_fetch_mvbox_title">Only fetch from the DeltaChat folder</string>
+    <string name="pref_only_fetch_mvbox_explain">This will ignore all other folders. Regardless of this setting, the Spam folder is sometimes checked for replies to known messages.</string>
```
Hocuri, line 611: I'm wondering if we should change this to

```suggestion
    <string name="pref_only_fetch_mvbox_explain">…and ignore all other folders. Regardless of this setting, the Spam folder is sometimes checked for replies to known messages.</string>
```

r10s: i do not like the ellipsis-form - we do not use that somewhere else, it is overall not really used and looks a bit weird. also may make translations a bit harder as a sentence is split.

as the plain "Ignore all other folders." is even shorter, i'd vote for that :)

adbenitez (approved):
looks good, but didn't test

res/values/strings.xml
```diff
@@ -607,6 +607,8 @@
     <string name="pref_send_copy_to_self">Send Copy to Self</string>
     <string name="pref_auto_folder_moves">Automatic Moves to DeltaChat Folder</string>
     <string name="pref_auto_folder_moves_explain">Chat conversations are moved to avoid cluttering the Inbox</string>
+    <string name="pref_only_fetch_mvbox_title">Only fetch from the DeltaChat folder</string>
```
r10s, line 610:
```suggestion
    <string name="pref_only_fetch_mvbox_title">Only Fetch from DeltaChat Folder</string>
```

we "headline capitalize" options meanwhile, that means everythings beside to/and/or/from/etc. becomes uppercase

res/values/strings.xml
```diff
@@ -607,6 +607,8 @@
     <string name="pref_send_copy_to_self">Send Copy to Self</string>
     <string name="pref_auto_folder_moves">Automatic Moves to DeltaChat Folder</string>
     <string name="pref_auto_folder_moves_explain">Chat conversations are moved to avoid cluttering the Inbox</string>
+    <string name="pref_only_fetch_mvbox_title">Only Fetch from DeltaChat Folder</string>
+    <string name="pref_only_fetch_mvbox_explain">Ignore all other folders. Regardless of this setting, the Spam folder is sometimes checked for replies to known messages.</string>
```
Hocuri, line 611: @r10s you wrote that the option description should mention that that requires the help of the server - like this?

```suggestion
    <string name="pref_only_fetch_mvbox_explain">This only works if your server somehow moves chat messages to the DeltaChat folder.</string>
```

src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
```diff
@@ -76,16 +78,28 @@ public void onCreate(Bundle paramBundle) {
     bccSelfCheckbox.setOnPreferenceChangeListener((preference, newValue) -> {
       boolean enabled = (Boolean) newValue;
       dcContext.setConfigInt(CONFIG_BCC_SELF, enabled? 1 : 0);
+      DcHelper.getAccounts(getContext()).startIo();
```
r10s, line 81: why is io not stopped before here?

src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
```diff
@@ -76,16 +78,28 @@ public void onCreate(Bundle paramBundle) {
...
       return true;
     });
 
+    onlyFetchMvboxCheckbox = this.findPreference("pref_only_fetch_mvbox");
+    onlyFetchMvboxCheckbox.setOnPreferenceChangeListener(((preference, newValue) -> {
+      boolean enabled = (Boolean) newValue;
+      DcHelper.getAccounts(getContext()).stopIo();
+      dcContext.setConfigInt(CONFIG_ONLY_FETCH_MVBOX, enabled? 1 : 0);
+      DcHelper.getAccounts(getContext()).startIo();
```
r10s, line 99: we should probably add the line `changes require restarting IO by calling dc_stop_io() and then dc_start_io().` to deltachat.h to make the requirement clear.

src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
Hocuri: Doesn't seem needed, idk why I added it

r10s (approved):
some minor comments ...

src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
r10s: i just had a rough look at core, it seems not to be needed.

src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
Hocuri: I'll do that

PR #2208, opened 2022-01-28
don't force dark mode in webpage for webxdc's webview

adbenitez:
the forced dark mode usually messes the apps background and colors, it is completely unreadable in some cases, changing color of background images to light colors while the texts remains white

### DIFF
--- a/src/org/thoughtcrime/securesms/ConnectivityActivity.java
+++ b/src/org/thoughtcrime/securesms/ConnectivityActivity.java
@@ -15,6 +15,7 @@ public class ConnectivityActivity extends WebViewActivity implements DcEventCent
   @Override
   protected void onCreate(Bundle state, boolean ready) {
     super.onCreate(state, ready);
+    setForceDark();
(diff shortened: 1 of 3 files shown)


r10s (commented):
i would not override the `set()` function to bypass the darkmode. it seems a bit weird to call a `set()` function that does not what it is supposed to at the end.

that could be fixed by another name, however, i'd just call the already outfactored `setForceDark()` not in the `WebViewActivity`  but in the derived classes where it is actually desired - that is probably only `LocalHelpActivity` and `ConnectivityActivity` (i assume also html mails do not look great in forced dark mode)

adbenitez:
good point, after some javascript "development" I am doing weird stuff in java XD will improve it

about html emails, the "show full message" is used also for non-html long messages, and more often for normal messages than for full blow web pages so probably also force it there as currently until some more concrete issues appear

adbenitez:
well, didn't forced for FullMsgActivity at the end as you recommended

r10s (approved):
great!

PR #2212, opened 2022-01-29
tweak default summaries

### LINKED ISSUE #2211: webxdc summaries in drafts
-  in draft mode, the app can already send updates (eg. for configuration) and these updates may contain summaries.  
these summary should be shown in the draft (where currently only the size is shown)

nb: for the summary in the bubbles: there, the string "123 KB Webxdc" is questionable. it is confusing that the size information is not always shown, also, with message-info there is a way to always find out the size. always showing the size seems to be too much noise (we do not do that for stickers as well) - so, maybe just leave the line empty when there is no summary?

r10s:
see commit messages for details.

closes #2211


adbenitez (approved):
lgtm

Simon-Laux (approved):

Issue #2187, opened 2022-01-16
QRCode Display Error


Gh0stz0x:
- Android version: 10
- Device: Huawei Mate 20
- Delta Chat version: 1.26.2 Nightly 2022-01-16
- Expected behavior: When I see an invitation QRCode of a group it should not show quot;  before and after the group name
- Actual behavior: Quot;  before and after the group name
- Steps to reproduce the problem: View the invitation QRCode
- Screenshots: 
![Screenshot_20220116_105805](https://user-images.githubusercontent.com/74795149/149655444-13e9a209-38d7-4993-8df4-ff096ba96eac.jpg)

- Logs:

Hocuri:
Can reproduce, also in other languages. Something with QR code generation is going wrong there. The source string is `Scansionalo per entrare nel gruppo \"%1$s\".`, which actually looks fine.

In 1.26.2, there simply were no quotes, maybe we should simply remove them again.

Gh0stz0x:
> `Scansionalo per entrare nel gruppo \"%1$s\".`

From this string I have eliminated the dot, which was not there in the sentence to be translated.  For the rest I noticed this bug only in the nightly.  In the stable version it does not show up 🤔

Gh0stz0x:
> `Scansionalo per entrare nel gruppo \"%1$s\".`

From this string I have eliminated the dot, which was not there in the sentence to be translated.  For the rest I noticed this bug only in the nightly.  In the stable version it does not show up 🤔

>In 1.26.2, there simply were no quotes, maybe we should simply remove them again.

I really think so

gerryfrancis:
The text shown below the QR code is `Scansional per entrare nel gruppo quot;QRCode_Testquot;.` while it should be `Scansional per entrare nel gruppo &quot;QRCode_Test&quot;.` instead, according to https://developer.android.com/guide/topics/resources/string-resource.html#FormattingAndStyling . (Note the `&` character!)

PR #2214, opened 2022-01-31
Remove quotes from `Scan to join group %1$s` string

Hocuri:
They were shown as `quot;` instead of `"` in the QR code:

image_2022-01-31_14-18-13

fix https://github.com/deltachat/deltachat-android/issues/2187

My first try to actually fix it failed (i.e. show the quote in the QR code), so, maybe just remove it?


r10s:
that is a bug in android svg somewhere? desktop an ios looks fine. but yes, if that fixes a very visible issue, maybe better remove them.

Hocuri:
Superseeded by https://github.com/deltachat/deltachat-android/pull/2215

PR #2215, opened 2022-01-31
hack: remove quotes from qr-code group join string

r10s:
alternative to https://github.com/deltachat/deltachat-android/pull/2214

that should work with all translations and can be easily undone without burden anything to the translators.

closes https://github.com/deltachat/deltachat-android/issues/2187


Hocuri (approved):
Makes sense

PR #2216, opened 2022-01-31
prepare 1.28.0

r10s:
device message draft:


res/values/strings.xml
```diff
@@ -913,4 +913,5 @@
...
     <string name="update_1_24_ios">1.24 at a glance:\n\n• Start an in-chat search from the chat\'s profile\n• Save traffic: Download large files as needed\n• Save nerves: faster, non-blocking QR code joins\n\nMore at 👉</string>
     <string name="update_1_26">1.26 polishing 💅\n\n• After making the QR codes non-blocking in 1.24, they also look much nicer now - on all platforms\n• Avatars and badges look similar across platform\n• Lots of bugs fixed</string>
+    <string name="update_1_28_android">1.28 Highlights:\n\n🔒 Encryption: A new experimental option for keeping your contacts, encryption secrets and text messages encrypted on your device. Only available on account creation and importing backup files.\n\n📫 Write to many mailing list types - can you write to your favorite list now?\n\n💫 Faster inital setup for more e-mail servers and streamlined networking options to prevent accidental misconfiguration\n\n🐜 Maaaaaany user-reported issues fixed</string>
```
link2xt, line 916: "inital" -> "initial"

"can you write to your favorite list now?" does not sound that good after translation. I see you are trying to say that some mailing list types are not supported, but most users don't even know there are different mailing list types. Maybe just say that there is a support for writing to mailing lists now, in contrast to previous version which could only read?

r10s: yip, you're probably right. also the direct german translation would be weird. maybe we should say to the translators that they can be a bit more free on such texts, however, that is another topic for another time :)

PR #2221, opened 2022-02-17
enable `Window.sessionStorage`, `Window.localStorage` and `IndexedDB` APIs

r10s:
after several discussions, we decided to allows these APIs, just natively what the browser/webview provide.

advantages of that approach are:
- existing apps can use the APIs without adaptions, so porting apps are easier
- no re-implementation needed - esp. for `IndexDB` that would probably take weeks~months, plus maintenance 
- as we are not using core, the local storage APIs cannot slow down receiving/sending messages

disadvantages:
- the local storage does not go the backups. it could be done by injecting javascript on exit and get `localStorage` etc. to the database, the other way round on opening. however, that's quite some effort, esp. when it comes to `IndexedDB`, this is probably not worth the effort.

other hints:
- no multi-device-sync possible, however, it is questionable if we would do that if we'd re-implement the APIs - `localStorage` may be used for scrolling positions, themes and so on you may configure differently across devices. also, existing implementations would not expect localStorage to be sth. other than local :)

same should be done for iOS and Desktop, cc @cyBerta , @Simon-Laux - and we need a test.xdc for that.

### DIFF
--- a/src/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -70,6 +70,8 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
     webSettings.setGeolocationEnabled(false);
     webSettings.setAllowFileAccessFromFileURLs(false);
     webSettings.setAllowUniversalAccessFromFileURLs(false);
+    webSettings.setDatabaseEnabled(true);
+    webSettings.setDomStorageEnabled(true);
     webView.addJavascriptInterface(new InternalJSApi(), "InternalJSApi");
 
     // `msg_id` in the subdomain makes sure, different apps using same files do not share the same cache entry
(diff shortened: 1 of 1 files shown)


adbenitez:
for some reason the localStorage state is shared between webxdc instances

src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -70,13 +70,15 @@ protected void onCreate(Bundle state, boolean ready) {
...
+    webSettings.setDatabaseEnabled(true);
+    webSettings.setDomStorageEnabled(true);
     webView.addJavascriptInterface(new InternalJSApi(), "InternalJSApi");
 
     // `msg_id` in the subdomain makes sure, different apps using same files do not share the same cache entry
     // (WebView may use a global cache shared across objects).
     // (a random-id would also work, but would need maintenance and does not add benefits as we regard the file-part interceptRequest() only,
     // also a random-id is not that useful for debugging)
-    webView.loadUrl("webxdc://msg" + appMessageId + ".localhost/index.html");
+    webView.loadUrl("https://msg" + appMessageId + ".localhost/index.html");
```
Simon-Laux, line 81: do we need https for this on android?
on desktop I had to define the webxdc scheme as special "priviliged scheme" to get it working, so could be that this is required.
anyways its an implementation detail so not that important as long as it works and because we (can) intercept all requests on android using `https` works.

adbenitez: yes, `https` is needed on android, check comments in: https://github.com/deltachat/deltachat-android/pull/2221/commits/3df96825aefefc3d55da76b3b1c7be51d9ace612

PR #2223, opened 2022-02-21
adapt js to new webxdc-update-api

r10s:
ui part of https://github.com/deltachat/deltachat-core-rust/pull/3081

### DIFF
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -749,9 +749,9 @@ JNIEXPORT jboolean Java_com_b44t_messenger_DcContext_sendWebxdcStatusUpdate(JNIE
 }
 
 
-JNIEXPORT jstring Java_com_b44t_messenger_DcContext_getWebxdcStatusUpdates(JNIEnv *env, jobject obj, jint msg_id, jint status_update_id)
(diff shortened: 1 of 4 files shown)


r10s:
tested it roughly, seems to work :) requires https://github.com/deltachat/deltachat-core-rust/pull/3081

once this and https://github.com/deltachat/deltachat-core-rust/pull/3081 are merged, we should file corresponding issues/pr for ios/desktop.

r10s:
hm, for the poll webxdc, creating a draft and already voting in the draft, loses the state on the sending device. the other device seems to be okay (did a quick test with two devices only, did not test other apps)

Simon-Laux (approved):
Change is smaller than what I imagined it to be, good job!

adbenitez (approved):

adbenitez:
> hm, for the poll webxdc, creating a draft and already voting in the draft, loses the state on the sending device. the other device seems to be okay (did a quick test with two devices only, did not test other apps)

I checked the code of the poll app, the issue is that you are loading the state with the old `getAllUpdates()` which is always empty now and in the listener given to `setUpdateListener()` you ignore events from `selfAddr` so the app can't restore the state only from what is given to `setUpdateListener()` needs to be reworked to not ignore `selfAddr` updates, in general it always was a good practice to process events even from `selfAddr` instead of what you did in the poll app ;P

r10s:
thanks for checking webxdc-poll. indeed, that app does not use good practices and need to be adapted.

PR #2227, opened 2022-02-28
add finnish translation

r10s:
thanks to @anevaiopvn on transifex, we have a finnish translation now :)


r10s:
@anevaiopvn from transifex validated also the locale name for finnish ("Suomi").

adbenitez (approved):

PR #2233, opened 2022-03-13
Mark already seen messages as seen

link2xt:
Marking the message as seen starts ephemeral timer if it is not
already started.

Due to the bug in the core, sometimes messages were marked as seen
without starting ephemeral timer. This can also happen due to other
errors, like database access timeout.

Always calling markseenMsgs whenever the message is displayed, even if
it is already seen, ensures that ephemeral timer is started
eventually.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -670,7 +670,7 @@ public class ConversationFragment extends MessageSelectorFragment
         int index = 0;
         for(int pos = firstPos; pos <= lastPos; pos++) {
             DcMsg message = ((ConversationAdapter) list.getAdapter()).getMsg(pos);
-            if (message.getFromId() != DC_CONTACT_ID_SELF && !message.isSeen()) {
+            if (message.getFromId() != DC_CONTACT_ID_SELF) {
                 ids[index] = message.getId();
                 index++;
             }


link2xt:
Related bug in the core: https://github.com/deltachat/deltachat-core-rust/issues/3113
Related bugfix in the core: https://github.com/deltachat/deltachat-core-rust/pull/3122
Change similar to this in Desktop: https://github.com/deltachat/deltachat-desktop/commit/b5fb1eaa2f4f0c19cfce63e3d1df5831947a57f3

Hocuri (approved):
Nothing seems to speak against it, but let's still wait a bit for the others to review, as the Android code is full of pitfalls and I don't know this code

adbenitez (approved):

r10s:
i looked in to the core a bit, logically, that seems to be okay.

but there are quite some more SQL commands that are called now and may degrade performance:

1. with that change, _each_ message displayed, ephemeral timer or not, results in an _additional_ SQL-SELECT command for each message [1]  
before, marking 20 messages as seen results in **1** SELECT command, now we have **21** SELECT-commands, that may be noticeable, even if a SELECT is faster than other SQL commands.

2. if the messages actually have an ephemeral timer, started or not, there are even 41 SELECT- or UPDATE-commands instead of 1 SELECT commands. [2]

without measuring, that is probably a  worsening that can be measured. as we are currently in ["_Speed_ and Reliability mode"](https://github.com/orgs/deltachat/projects/48) we should maybe have a look at that, also as as dc_markseen_msgs() is one of the most called functions.

esp. the first point could be fixed easily in core by [adding `ephemeral_timer` to the outer SELECT ](https://github.com/deltachat/deltachat-core-rust/blob/master/src/message.rs#L1290), that would not lead to more SQL commands in case there is no ephemeral timer, with few additional code and complexity.

----

[1] [start_ephemeral_timer() is called for each message](https://github.com/deltachat/deltachat-core-rust/blob/master/src/message.rs#L1324) and results in at least one SELECT for each message

[2] https://github.com/deltachat/deltachat-core-rust/blob/master/src/ephemeral.rs#L289

r10s:
i have to correct myself:

the _"1 SELECT command"_ mentioned above are actualy 20 single SELECT commands - but it could be only one ... but still, the number of SQL commands will increase by this pr, and we should consider optimizing markseen_msgs() in the core. for this pr, however, there is nothing to do, so that can be merged, esp. as this fixes a bug.

Issue #1704, opened 2020-11-03
forward might end activity


r10s:
forward to saved-messages sometimes removes all activities and closes app.

the app is not crashing or so, just the activity is ended.

i could reproduce that currently only on my android 4.4 testphone.

csb0730:
Can confirm this.

just after forwarding the target chat is shown (not only at saved-messages). when I then press the _back button_ then all activities are closed. DC is not returning to main activity or last chat selected.

r10s:
thanks for confirming! we will fix that before 1.14 is released to the stores.

r10s:
this issue seems to be older, eg. it is also in the 1.12.

it can be reproduced even simpler: 

- tap forward-message
- then the "back" key (either in the action bar or the system key) 
- app will close

also, when really forwarding, the chatlist is not updated.

maybe related to `launchMode="singleTask"` we're using, might be the behavior changed at some point.

as this is a pretty old issue and no regression, i remove the blocker tag, however, it should be fixed anyway soon.

csb0730:
I'm using code base of 1.2.x ... much older than 1.12 ;-)

r10s:
> I'm using code base of 1.2.x ... much older than 1.12 ;-)

and 1.2.x also already has this issue?

r10s:
that `A.chatlist -> B.chat -> A.forward -> back` exists the app with launchMode=singleTask seems to be logical to me, as `A.forward` just reuses `A.chatlist`- and removes things on top of that (`B.chat` in this case)

what i do not understand: why does this work on newer androids?

r10s:
> what i do not understand: why does this work on newer androids?

it works because startActivityForResult() behaves differently on newer androids, cmp. https://github.com/deltachat/deltachat-android/pull/1412 

so, we have several solutions to solve that:

- give up singleTask thing, that would require us to track that no more than one Chatlist is present in the activity stack, also may have other side effects. does not sound tempting to me.

- start the following activity manually for android4 - that may be some code / some situations to think of, however, the advantage that we would not run into danger making things worse for non-android4

- maybe also a separate FowardActivity is a solution, similar to ShareActivity, that would not alter the stack. FowardActivity can maybe be an aliased activity, that has to be checked out.

however, i just tried delta chat 1.0 (from beginning of this year - setup has becom 30x faster on that device, btw), that also has the bug in, so it is no regression and this issue should not block upcoming 1.14 release.

Hocuri:
[...]

PR #2246, opened 2022-03-24
Introduce dedicated ConversationListRelayingActivity to fix forwarding bug

Hocuri:
fix https://github.com/deltachat/deltachat-android/issues/1704 by creating an extra activity:

https://github.com/deltachat/deltachat-android/pull/2246/files#diff-5027fc167512e95578b7074b02d0740011435b3d033a2fe66d4893a74d81e0e2R12-R24

Also create some new UI test (and test it with the old UI tests) because we had so many regressions wrt sharing, forwarding and switching activities in the past. The test depends on https://github.com/deltachat/deltachat-core-rust/pull/3150; if it doesn't get merged, I'll have to find another method to disable this animation.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -153,6 +153,12 @@
                 android:value="org.thoughtcrime.securesms.ConversationListActivity" />
     </activity>
 
+    <activity android:name=".ConversationListRelayingActivity"
+              android:label="@string/app_name"
+              android:launchMode="singleTask"
+              android:theme="@style/TextSecure.LightNoActionBar"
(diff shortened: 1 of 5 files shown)


adbenitez (approved):
works like a charm, thanks a lot!!!!

r10s (approved):
let's merge that in soon, so we can find and target potential issues soon as well (seems unlikely, however, as the pr introduces a new activity just for forwarding and does not change existing activity flow imu)

Hocuri:
I must ask, though, what does "imu" mean?

r10s:
> I must ask, though, what does "imu" mean?

"in my understanding"

or "in my opinion" with a typo :)

Issue #1835, opened 2021-03-12
The profile picture is rotated 90° counterclockwise


Gh0stz0x:
- Operating System (Linux/Mac/Windows/iOS/Android): Android 10 - Lineage Os 17.1
- Delta Chat Version: 1.14.5 - nightly
- Expected behavior: When I select an image or take a photograph, I want it to stay where I took it.
- Actual behavior: When I select an image, it rotates 90° counterclockwise
- Steps to reproduce the problem: Go to settings, select the profile, change image by selecting it from the gallery or taking a photograph in portrait mode. The photograph will be rotated 90 ° counterclockwise. This bug was detected on Samsung and Xiaomi smartphones.
- Screenshots: 
![IMG_20210312_231019](https://user-images.githubusercontent.com/74795149/111003910-ee7a2680-837f-11eb-81e1-5f42bcbfe8cd.jpg)
- Logs:

Hocuri:
Strange. I can't reproduce this on my Fairphone 3 w/ Android 10. 
And this happens both when you directly take a photo and when you select one from the gallery? 
Does it happen everytime or just sometimes? 
Did it only happen on LineageOS devices so far?
Does it also happen if the device is in landscape mode?

r10s:
there is a related fix at https://github.com/deltachat/deltachat-core-rust/pull/2063, however, the fix is not release, but it is in the nigtlies (please double-check, that core 1.51.0-alpha.0 is used).

please note, that existing avatars are not rotated correctly after update, you can check that only with new avatars.

however, of course, may also be another issue :)

Gh0stz0x:
@Hocuri It occurs either by taking a new photograph or by selecting a photograph from the gallery.
@r10s The user who reported it to me told me that he uses the nightly. I try to check on a Samsung that I have to reproduce the problem and see which core it uses. 
However, the bug exists on both Android and Lineage.

Gh0stz0x:
I looked on my Samsung log: it uses 
1.51.0-alpha.0 core

Gh0stz0x:
>Does it also happen if the device is in landscape >mode?

@Hocuri I'll give it a try and I'll tell you

Gh0stz0x:
@Hocuri If I do the landscape photography the photograph is "straight".  Now I have asked the two users to test them too.

adbenitez:
[...]

PR #2249, opened 2022-03-27
avoid rotated avatar image when selecting new self-avatar from Gallery

adbenitez:
fixes 50% of #1835, when selecting image from Gallery, that is usually 99% of the use case when you set an avatar, I think it is rare to take the picture at the moment, the issue is still present in that case because the dependency we use to crop the avatars, which is outdated and causes other issues like #1566, need to be replaced by the image editor we have already, then #1835 will be 100% fixed

### DIFF
--- a/src/org/thoughtcrime/securesms/CreateProfileActivity.java
+++ b/src/org/thoughtcrime/securesms/CreateProfileActivity.java
@@ -10,7 +10,6 @@ import android.graphics.Bitmap;
 import android.net.Uri;
 import android.os.AsyncTask;
 import android.os.Bundle;
-import android.provider.MediaStore;
 import android.text.TextUtils;
 import android.util.Log;
 import android.view.KeyEvent;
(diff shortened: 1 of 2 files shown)


adbenitez:
I just copied the way group avatars are set and that was not having this issue when selecting from gallery, since in the case of group avatars, the image is loaded to a bitmap with Glide and Glide takes care to fix orientation based on EXIF data

src/org/thoughtcrime/securesms/CreateProfileActivity.java
```diff
@@ -170,49 +174,38 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+    final Uri output = Crop.getOutput(data);
+    final ProfileMediaConstraints constraints = new ProfileMediaConstraints();
+    GlideApp.with(this)
+            .asBitmap()
+            .load(output)
+            .skipMemoryCache(true)
+            .diskCacheStrategy(DiskCacheStrategy.NONE)
+            .centerCrop()
+            .override(constraints.getImageMaxWidth(this), constraints.getImageMaxHeight(this))
+        .into(new SimpleTarget<Bitmap>() {
```
Hocuri, line 193: Just a note: SimpleTarget is deprecated and "unsafe", but since it's also used in GroupCreationActivity, we can just leave it for now I think, and then mabye one day we get to fix all the deprecation warnings. Or maybe not, as far as we don't get any problems with it.

Hocuri (approved):

PR #2255, opened 2022-04-01
Load chat asynchronously

Hocuri:
Call get_chat_msgs() on a background thread; should speed up chat opening with large databases.

There may still be bugs, but I don't know of any.

### DIFF
--- a/res/layout/conversation_fragment.xml
+++ b/res/layout/conversation_fragment.xml
@@ -22,6 +22,7 @@
         android:layout_height="wrap_content"
         android:layout_gravity="center"
         android:visibility="visible"
+        android:text="@string/loading"
         style="@style/NoMessagesTextView"
         tools:text="@string/chat_new_one_to_one_hint"
         />
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -231,10 +231,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     chatId = getIntent().getIntExtra(CHAT_ID_EXTRA, -1);
 
     Util.runOnBackground(() -> {
-      long startMs = System.currentTimeMillis();
-      final int[] msgs = dcContext.getChatMsgs((int) chatId, 0, 0);
-      Log.i(TAG, "⏰ getChatMsgs(" + chatId + "): " + (System.currentTimeMillis() - startMs) + "ms");
-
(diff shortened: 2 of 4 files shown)


adbenitez (dismissed):
works good, even without the speed improvements to get_chat_msgs this already feels fast

Hocuri:
Discovered by @adbenitez:

> Hocuri, issue: with the async loading of chat messages the "unread messages" marker is lost and chat is open at the end

Hocuri:
I fixed the two bugs adb found, so now let's wait for @r10s to find out whether this PR still improves performance with https://github.com/deltachat/deltachat-core-rust/pull/3194 with his bigger-than-mine database.

r10s:
@Hocuri i am still on checking things and i am also busy with other stuff. however, i think, i can tell more on monday :)

r10s:
k, i played around with this pr and with master, i cannot say that the one feels  faster than the other, my first impression even was that this pr is slower than before.

i put by phone to flight mode and stopped opening 5 chats (some quitebig ones eg. xyiv)  on a quite old phone (moto g4 from 2017) and here are the results:

https://gist.githubusercontent.com/r10s/529e40d14a8a7e1395e482f482fb4b88/raw/ba467b36cdfbabe20629a9e2c02058542b65e24b/gistfile1.txt

(the additional logging was done with https://github.com/deltachat/deltachat-android/compare/temp-get-chat-msgs-logging )

so, my findings:
- getChatMsgs() is meanwhile really fast, only 10% of opening time is spend there
- putting getChatMsgs() to background makes that part a bit slower, this is expected, however
- some of the 5 test chats are loaded a bit slower, others a bit faster, but that is probably within measurement inaccuracy (while the log only shows one test, i did several, but they are all comparable within maybe +/- 5%

so, overall, i could not measure a speed improvement.

as also @Hocuri pointed out, there is some risk in merging this pr, maybe we should just leave things as is.

otoh - i did not run into issues with this pr yet - so if others  measure speed improvements, i would not be against going for it.

r10s:
@adbenitez in Delta Chat:

> "I am testing DC in my new old test phone, and yes the async loading of chat messages can take a lot here, like 10 seconds sometimes, while without the Loading... thing it is 3-5 seconds sometimes"

> "in my more modern phone it is always fast with both, but this old phone has only like 600MB of free RAM :)"

r10s:
k, i think, we mostly agree to forego this pr, at least for now and for 1.30.

PR #2256, opened 2022-04-02
avoid null mime type

### LINKED ISSUE #2197: sending webxdc file via share fails to send
a toast is shown:
 
```
Failed to send message: mime parse error: a slash (/) was missing between the type and subtype
```

adbenitez:
close #2197


r10s (approved):
probably okay as it is converted to `application/webxdc+zip` by core later.

also, other calls do this explicit check for null after calling `MimeTypeMap.getSingleton().getMimeTypeFromExtension()` - at some point, that could be cleaned up a bit, also as there is `PersistentBlobProvider::getMimeTypeFromExtension()` that does exactly the same.

but well, for now, this is good enough if this fixes the bug :)

adbenitez:
yip the first thing I did was to make public PersistentBlobProvider::getMimeTypeFromExtension() and use that but then since it was repeated in several places decided to just do as little and possible and added a note to my "to refactor" list where I have several small refactorings/improvements to do in a "code clean up" PR

PR #2257, opened 2022-04-03
don't jump to parent if parent is not a webxdc

### LINKED ISSUE #2250: clicking any system message jumps to some previous message
this feature was added for webxdc info-messages but it is affecting member added/removed messages as well

adbenitez:
close #2250 

this doesn't solve the issue in 100% of cases, in some rare cases a "member removed/added" etc might has a webxdc as parent, I guess the proper fix must be done in core, perhaps set the parent webxdc as quote and use that instead of parent or make get_parent only return parent for webxdc's info messages


Hocuri (approved):

r10s (approved):
lgtm :)

PR #2259, opened 2022-04-04
pause/destroy webview when activity is paused/destroyed

### LINKED ISSUE #2258: background music keeps playing when webxdc instance is closed
when playing a game like https://github.com/adbenitez/StackUp.xdc with a background music, if you open the game and start to play, then go back to chat, the music keeps playing

adbenitez:
close #2258

(note: at least in my phone, the music is not paused when the webview is paused, but at least when the webxdc is closed destroying the webview stops the music)


r10s (approved):
looks reasonable :) did not test, however.

PR #2266, opened 2022-04-07
tweak confirm-to-forward dialog

adbenitez:
go back to chatlist when back button is pressed or user touches outside the dialog in the dialog asking to confirm to forward message inside chat.

currently if back is pressed, the dialog is canceled but the chat stays open and you have to press back again to go to the chatlist with the "Forward to..." title


r10s (approved):
lgtm, thanks for taking care of details!

PR #2269, opened 2022-04-09
Very small speed improvements

Hocuri:
See individual commit messages.


r10s (approved):
for the "SECURITY" flag: i think, the idea of clearing was at some point that eg. taking a screenshot is possible at once after playing around with the option. however, as the setting even says that his requires to restart the app (which is probably true at least on some systems), it seems reasonable to leave that alone. good catch!

adbenitez (approved):
lgtm, will test if I see any improvements in my chat list loading :D

PR #2271, opened 2022-04-10
Update search results when the chatlist changes

### LINKED ISSUE #1843: Search results not updated after deleting chat
- Operating System (Linux/Mac/Windows/iOS/Android): Android
- Delta Chat Version: master
- Expected behavior: deleted chat is removed from search results
- Actual behavior: weird "ghost" chat is shown as placeholder when you delete it
  - create some test chat,
  - use the search button and search part of the test chat's name to find it,
  - open the test chat in the results list, 
  - tap on the menu and select to delete the chat
  - you are redirected to the search results, but now in the place where the test chat was there is a "ghost" chat with a black avatar with `#` symbol
- Screenshots:
deltachat-2021-03-24-154506

Hocuri:
Fix  #1843

### DIFF
--- a/src/org/thoughtcrime/securesms/search/SearchFragment.java
+++ b/src/org/thoughtcrime/securesms/search/SearchFragment.java
@@ -11,6 +11,7 @@ import androidx.appcompat.app.AlertDialog;
 import androidx.recyclerview.widget.LinearLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
 import android.text.TextUtils;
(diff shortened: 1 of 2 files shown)


r10s (approved):
to fix the bug, it would be sufficient to listen to `DC_EVENT_MSGS_CHANGED`, however, as several information are displayed in the search result, it probably makes some sense to also listen to other updates.

adbenitez (approved):
really nice!!

PR #2275, opened 2022-04-14
show download failures

r10s:
tapping on the message, still allows re-downloading
in case of temporary errors.

https://user-images.githubusercontent.com/9800740/163385728-43a5846e-4006-4018-b7ea-c5f75b2c20ee.MOV

maybe the error message could be nicer, however, not sure if that is worth the effort, at least this is sth. for another time :)

i came over that issue when trying to find out more information about https://github.com/deltachat/deltachat-core-rust/issues/2876


adbenitez (approved):
tested, I like you can still retry

PR #2278, opened 2022-04-23
resend messages

r10s:
requires https://github.com/deltachat/deltachat-core-rust/pull/3238 to be merged before.

https://github.com/deltachat/deltachat-core-rust/pull/3238 is pretty limited, however, even if we revamp most of the internals, the ffi and the UI will probably stay the same - a "Resend" button seems to have advantages in many situations.

see commit messages for some details and reasoning.

### DIFF
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -711,6 +711,16 @@ JNIEXPORT void Java_com_b44t_messenger_DcContext_forwardMsgs(JNIEnv *env, jobjec
 }
 
 
+JNIEXPORT jboolean Java_com_b44t_messenger_DcContext_resendMsgs(JNIEnv *env, jobject obj, jintArray msg_ids)
+{
+    int msg_ids_cnt = 0;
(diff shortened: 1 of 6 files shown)


adbenitez (approved):

adbenitez:
something I noticed is that all info messages are displayed for the receiver, perhaps would be better if they are not displayed, similar to when you use an app in the draft mode and then send it with several updates

r10s:
> something I noticed is that all info messages are displayed for the receiver, perhaps would be better if they are not displayed, similar to when you use an app in the draft mode and then send it with several updates

not sure what you mean here - info messages cannot be resend?! which "all info messages"?

adbenitez:
I did this test:

created a group with a peer A, 
sent 2048 game in the group,
played a bit and scored, an info message is sent in the group,
added peer B to the group,
selected the 2048 game and pressed resent,
peer B received the message with the game, but also an info message with the score is shown for this new member, if there were 100 info messages in an old game instance the new member would get all this noise

r10s:
thanks for clarification! 

ftr, this is the corresponding issue filed by @adbenitez: https://github.com/deltachat/deltachat-core-rust/issues/3251

PR #2280, opened 2022-04-25
adapt strings to new welcome screen

r10s:
this pr is only about fine-tuning the new strings before they can get translated; wrt case: we are using "first letter uppercase for most words" on all buttons and items meanwhile, therefore i slightly adapted that.

suggestions welcome - once this is approved/merged, we can go for layout, positioning, logo (which is less time critical as no external persons involved)

before/after:

Screen Shot 2022-04-25 at 22 00 47 &nbsp; Screen Shot 2022-04-25 at 22 12 36

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -461,9 +461,11 @@
 
 
     <!-- welcome and login -->
+    <string name="welcome_chat_over_email">Chat over E-Mail.</string>
+    <string name="scan_invitation_code">Scan Invitation Code</string>
     <string name="welcome_intro1_message">The messenger with the broadest audience in the world. Free and independent.</string>
     <string name="login_title">Log In</string>
-    <string name="login_header">Log In to Your Server</string>
+    <string name="login_header">Log into your E-Mail Account</string>
     <string name="login_explain">Log in with an existing e-mail account</string>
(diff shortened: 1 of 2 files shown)


res/values/strings.xml
```diff
@@ -495,7 +497,7 @@
     <string name="login_error_server">Please enter a valid server / IP address</string>
     <string name="login_error_port">Please enter a valid port (1–65535)</string>
     <string name="login_error_required_fields">Please enter a valid e-mail address and a password</string>
-    <string name="import_backup_title">Import Backup</string>
+    <string name="import_backup_title">Restore from Backup</string>
```
r10s, line 500: or maybe better `Restore From Backup`?

res/values/strings.xml
```diff
@@ -461,9 +461,11 @@
 
 
     <!-- welcome and login -->
+    <string name="welcome_chat_over_email">Chat over E-Mail.</string>
+    <string name="scan_invitation_code">Scan Invitation Code</string>
     <string name="welcome_intro1_message">The messenger with the broadest audience in the world. Free and independent.</string>
     <string name="login_title">Log In</string>
-    <string name="login_header">Log In to Your Server</string>
+    <string name="login_header">Log into your E-Mail Account</string>
```
r10s, line 468: or maybe better `Log Into Your E-Mail Account`?

(the button all-uppercase comes from the system, however, the line will be also used as title for the subsequent screen)

adbenitez (approved):

PR #2284, opened 2022-04-30
new welcome image and layout

r10s:
this is the outcome of recent design discussions - so, this pr should not start further design discussions but check technical aspects of the implementation :)

before/after (the bad-looking button border and pixelated image is an emulator thingie, this is correct on real devices):

Screen Shot 2022-04-25 at 22 00 47 

i tried to make it nice for several screen sizes, however, it looks best on "typical, bit larger" smartphones used nowadays. smaller phones are fine as well, but spacing is not great there (as there is fewer space :) 

it should already be fine in dark and light mode.

i think, if we merge that in, we will get more feedback on the first testing round wrt devices and layout.

successor of https://github.com/deltachat/deltachat-android/pull/2280


Hocuri (approved):
Screenshot_20220430-155658

Something that could be improved is that for small screen heights, the lower buttons are not shown - instead, the title and image should be hidden.. OTOH there was this problem with the previous screen already, so this PR doesn't make anything worse, and I can't remember anyone complaining.

Code looks good, design too - I still think that the buttons are too grey esp. in dark mode, but one thing at a time.

r10s:
> Something that could be improved is that for small screen heights

wow - this is a really small screen. is this s watch or so?

but anyway,it probably does not worsen things, probably it is even more usable than before as before the image had a fixed height where now it is shrinked on few space.

r10s:
> but I don't like the usage of logos.

i think, no one is exuberantly enthusiastic seeing google or microsoft on the welcome screen.

the idea of adding the logos is to make it clearer to the users that this is "chat over email" and that you can login with your provider etc. there is no need for the user to recognize all logos, but at least some - and this is probably already true.

we tried to add also some small providers to underline, that there is a "big choice" of providers.

the current selection/design is probably only a first iteration, there are still some ideas and it may be revamped. but it is what we have now.

cc @plaintextdesign who can probably explain the user aspect better.

plaintextdesign:
thanks for the ping - and you've captured our thinking nicely! we're really not advertising getting an Outlook account here, the goal of this screen to build the correct mental model for first-time users. you are NOT creating a Delta Chat account, you are using Delta Chat to sign into your (existing) email account. this is really about lowering the barriers for onboarding. to do so, we unfortunately have to use mainstream email provider logos. the bigger, the better :/
the other aspect that I hope we get to introduce soon is an actual collaboration with good email providers, meaning we have better privacy affordances and smoother integrations!
finally, just to add that I am also worried about trademarks. but let's first get there? at least we are not butchering any logos!

r10s:
maybe to the footer or to a kebab menu, however, we'll probably shift things around wrt chat-mail anyways, i think, that is a good timing to reconsider adding a link to the faq

PR #2289, opened 2022-05-03
webxdc: allow internal pages and open some known schemes externally

### LINKED ISSUE #2279: link to internal pages don't work in webxdc
if a webxdc app use internal/separated pages for navigation, for example in your index.html you might have an "about" link pointing to "./about.html" page, this doesn't work on android, IIRC it works on iOS

adbenitez:
close #2279 

this allows to open internal pages and for other link have the same behavior than the viewer of HTML emails, allowing to open web links externally etc.

### DIFF
--- a/src/org/thoughtcrime/securesms/WebViewActivity.java
+++ b/src/org/thoughtcrime/securesms/WebViewActivity.java
@@ -48,7 +48,8 @@ public class WebViewActivity extends PassphraseRequiredActionBarActivity
     webView = findViewById(R.id.webview);
     webView.setWebViewClient(new WebViewClient() {
       // `shouldOverrideUrlLoading()` is called when the user clicks a URL,
-      // returning `true` means, the URL is passed to `loadUrl()`, `false` aborts loading.
+      // returning `true` causes the WebView to abort loading the URL,
+      // returning `false` causes the WebView to continue loading the URL as usual.
       // the method is not called for POST request nor for on-page-links.
       //
       // nb: from API 24, `shouldOverrideUrlLoading(String)` is deprecated and
@@ -64,9 +65,8 @@ public class WebViewActivity extends PassphraseRequiredActionBarActivity
             case "https":
             case "mailto":
             case "openpgp4fpr":
-              openOnlineUrl(url);
-              // URL opened externally, returning `true` causes the WebView to abort loading
-              return true;
+              // open URL externally
+              return openOnlineUrl(url);
           }
         }
         // by returning `true`, we also abort loading other URLs in our WebView;
@@ -271,12 +271,14 @@ public class WebViewActivity extends PassphraseRequiredActionBarActivity
     }
   }
 
-  protected void openOnlineUrl(String url) {
+  protected boolean openOnlineUrl(String url) {
     try {
       startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse(url)));
     } catch (ActivityNotFoundException e) {
       Toast.makeText(this, R.string.no_browser_installed, Toast.LENGTH_LONG).show();
     }
+    // returning `true` causes the WebView to abort loading
+    return true;
   }
 
   protected WebResourceResponse interceptRequest(String url) {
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -105,8 +106,12 @@ public boolean onPrepareOptionsMenu(Menu menu) {
...
 
   @Override
-  protected void openOnlineUrl(String url) {
-    Toast.makeText(this, "Please embed needed resources.", Toast.LENGTH_LONG).show();
+  protected boolean openOnlineUrl(String url) {
+    if (url.startsWith(baseURL +"/")) {
+      // internal page, continue loading in the WebView
+      return false;
+    }
+    return super.openOnlineUrl(url);
```
r10s, line 114: if we want to merge this pr soon for 1.30, i'd be a more reserved here, see comment above.

```suggestion
    Toast.makeText(this, "Please embed needed resources.", Toast.LENGTH_LONG).show();
    return true; // returning `true` causes the WebView to abort loading
```

r10s (changes requested):
> this allows to open internal pages

that's great!

> and for other link have the same behavior than the viewer of HTML emails, allowing to open web links externally etc.

it is not the same behavior as the webxdc-links can include sensitive information gathered from the chat and the webxdc - this is not possible for html-mails.

it may still be okay to allow these links, however, i'd like a broader discussion for that and not sneak that in. iirc, we disallowed external links for $reasons in some discussions. also, if we go for a change, we should make sure to allow that on all platforms then.

therefore, i suggest to limit this pr to the internal links, so we can easily merge that in.

and for the maybe more controversial external links, file an issue, pinging some ppl that are good in such security questions :)

src/org/thoughtcrime/securesms/WebViewActivity.java
```diff
@@ -64,9 +65,8 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
             case "https":
             case "mailto":
             case "openpgp4fpr":
-              openOnlineUrl(url);
-              // URL opened externally, returning `true` causes the WebView to abort loading
-              return true;
+              // open URL externally
```
r10s, line 68: this comment is wrong - the url may be opened internally or externally

adbenitez:
> it is not the same behavior as the webxdc-links can include sensitive information gathered from the chat and the webxdc

yikes I forgot this was the reason it shouldn't be allowed, but that only applies to http/https right? allowing mailto: links should be fine, the verification/group-join links would be nice but probably can also be hijacked to include sensitive info.

so would you say at least mailto is fine? otherwise I will just disable them all

adbenitez:
ok now this only solves the internal page issue and should be easy to merge, we can think about improving the opening of other links at a later point (perhaps showing a dialog displaying the link and offering to open/cancel, this is what telegram does and we probably do it also for the HTML viewer then re-adding what was removed here should be safer)

src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -35,6 +35,7 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
   private static final String TAG = WebxdcActivity.class.getSimpleName();
   private DcContext dcContext;
   private DcMsg dcAppMsg;
+  private String baseURL;
```
r10s, line 38: is that set somewhere?

r10s (commented):
now looks good in general, i am just unsure about if `baseURL` is actually set.

for `mailto:`-protocol - that might be more okay as data leakage would be user-visible (only subject or text can be set), however, to move forward, i'd also put the discussion for that outside this pr.

src/org/thoughtcrime/securesms/WebxdcActivity.java
adbenitez: `*sigh*` after implementing all in this file and testing in deltalab I had to revert the file to be able to checkout upstream/master and for got to re-add this, will fix now

adbenitez: fixed

r10s (approved):
yeah, looks good now! i also tested that really an absolute URL is given to `shouldOverrideUrlLoading()`.

PR #2290, opened 2022-05-04
set translations before starting I/O

### LINKED ISSUE #2288: Set stock translations before starting I/O
Stock translations are set here:
https://github.com/deltachat/deltachat-android/blob/647ffd3dad9cfb8f7750d0cfff59dd856d21d426/src/org/thoughtcrime/securesms/ApplicationContext.java#L139

IO is started here:
https://github.com/deltachat/deltachat-android/blob/647ffd3dad9cfb8f7750d0cfff59dd856d21d426/src/org/thoughtcrime/securesms/ApplicationContext.java#L115

If there are pending MDNs, an MDN may be generated in English because the translation is not set yet:
https://support.delta.chat/t/language-of-return-receipts/2075/3

adbenitez:
close #2288

### DIFF
--- a/src/org/thoughtcrime/securesms/ApplicationContext.java
+++ b/src/org/thoughtcrime/securesms/ApplicationContext.java
@@ -112,6 +112,10 @@ public class ApplicationContext extends MultiDexApplication {
       }
       Log.i("DeltaChat", "shutting down event handler");
     }, "eventThread").start();
+
+    // set translations before starting I/O to avoid sending untranslated MDNs (issue #2288)
+    DcHelper.setStockTranslations(this);
+
     dcAccounts.startIo();
 
     new ForegroundDetector(ApplicationContext.getInstance(this));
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -112,6 +112,17 @@ public void onCreate() {
       }
       Log.i("DeltaChat", "shutting down event handler");
     }, "eventThread").start();
+
+    // set translations before starting I/O to avoid sending untranslated MDNs (issue #2288)
+    DcHelper.setStockTranslations(this);
+    IntentFilter filter = new IntentFilter(Intent.ACTION_LOCALE_CHANGED);
```
r10s, line 118: we could still leave `registerReceiver` at the old place and move only the initial call to `setStockTranslations()` up - that would result in I/O starting sooner and running in parallel with registering receiver with a filter (not sure, however, how expensive that really is, but there may be slow systems and we were surprised by how slow android calls are here and there :)

also, leaving `registerReceiver` would make the change smaller, on startup things are sometimes critical wrt ordering (things may not be initialized etc.) however, i do not see side effects here (which does not mean that they cannot be there :)

therefore, _i_ would probably have changed only the single line - but i also see that it is nice to keep things together.

r10s (approved):
just my 2ct nitpicking, it is probably also fine as it is :) thanks for taking care!

src/org/thoughtcrime/securesms/ApplicationContext.java
adbenitez: makes sense, done

PR #2292, opened 2022-05-05
Add "Debug logging" advanced setting

Hocuri:
To be merged after https://github.com/deltachat/deltachat-core-rust/pull/3296

- [x] localization

### DIFF
--- a/res/xml/preferences_advanced.xml
+++ b/res/xml/preferences_advanced.xml
@@ -89,5 +89,11 @@
 
         <Preference android:key="pref_view_log"
                     android:title="@string/pref_view_log"/>
+
+        <org.thoughtcrime.securesms.components.SwitchPreferenceCompat
+            android:defaultValue="false"
+            android:key="pref_debug_logging_enabled"
+            android:summary="Enable some extra debug logs which can be viewed in 'Device messages'. Will slow down the device by about TODO%."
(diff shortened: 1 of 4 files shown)


res/xml/preferences_advanced.xml
```diff
@@ -89,5 +89,11 @@
 
         <Preference android:key="pref_view_log"
                     android:title="@string/pref_view_log"/>
+
+        <org.thoughtcrime.securesms.components.SwitchPreferenceCompat
```
adbenitez, line 93: can't we reuse the developer option for this? some extra debugging that makes the app slower and perhaps unreliable sounds like a perfect fit :D

Hocuri: Hm, then we don't warn the user as explicitly about the slowndown, and we can't disable the option when the webxdc is deleted, so this "the webxdc is there exactly when the option is on" relationship will be gone.

And the developer option is for people who develop webxdcs (at least currently) while the debug logging option is for people who want to track down issues with deltachat core, which doesn't necessarily belong together.

adbenitez (approved):
a pity this makes things slower if enabled, hope that gets better in the future :(

Hocuri:
Seems not to be necessary after all.

PR #2296, opened 2022-05-11
show context of Webxdc window

r10s:
show the chat name in front of the webxdc name to
- prevent phishing (the window will not show just the name of your bank)
- give the user an idea about where the webxdc sends messages to
  (important eg. when using the same webxdc from different chats)

counterpart of https://github.com/deltachat/deltachat-ios/pull/1578
see also https://github.com/deltachat/deltachat-core-rust/pull/3309

### DIFF
--- a/src/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -87,7 +87,7 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
       String chatName =  WebxdcActivity.this.dcContext.getChat(WebxdcActivity.this.dcAppMsg.getChatId()).getName();
       Util.runOnMain(() -> {
         try {
-          getSupportActionBar().setTitle(chatName + " – " + info.getString("name"));
+          getSupportActionBar().setTitle(info.getString("name") + " – " + chatName);
         } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


r10s:
thinking it over, i am not sure if `ChatName - AppName` is super-nice as it moves the app name maybe far right and the title does not start with the thing you are seeing below.

may sth. as `AppName shared with ChatName` or just shorter `AppName with ChatName` is better, would result in sth. as "Triska with Delta Devs" or "Poll with Friends".

Simon-Laux (approved):

cyBerta:
> thinking it over, i am not sure if `ChatName - AppName` is super-nice as it moves the app name maybe far right and the title does not start with the thing you are seeing below.
> 
> may sth. as `AppName shared with ChatName` or just shorter `AppName with ChatName` is better, would result in sth. as "Triska with Delta Devs" or "Poll with Friends".

not sure if `with` is grammatically the best choice, what about `AppName @ ChatName`or just `AppName - ChatName` ?

r10s:
> AppName @ ChatName

i like that :)

r10s:
after some discussions, we changed the order to `WebxdcName - ChatName`.

if we add an additional DocumentName, that will be prepended or will replace the Webxdc Name, cmp.  https://github.com/deltachat/deltachat-core-rust/pull/3317

Issue #2298, opened 2022-05-15
webxdc: support `source_code_url`


r10s:
see https://github.com/deltachat/deltachat-core-rust/pull/3314

simple implementation:

- if `source_code_url` is set, in the `WebxdcActivity` in the upper right corner the icon "three vertival dots" (as used everywhere else :) should appear. 
- menu entry would just be  "Source Code" currently

in the future, there may be more in this menu, eg. "Full screen" is an obvious idea.

PR #2301, opened 2022-05-17
add 'Source Code' item to webxdc menu

r10s:
the code is a tiny bit more complicated
as we keep the call to getWebxdcInfo() in background
and also avoid a second call to getWebxdcInfo() on startup.

Screen Shot 2022-05-17 at 16 13 39
(the menu is a bit lost, however, it will probably fill up in the future :)

closes https://github.com/deltachat/deltachat-android/issues/2298

### DIFF
--- a/src/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -38,7 +38,7 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
   private DcContext dcContext;
   private DcMsg dcAppMsg;
   private String baseURL;
-  private boolean hasSourceCodeUrl;
+  private String sourceCodeUrl = "";
 
   public static void openWebxdcActivity(Context context, DcMsg instance) {
     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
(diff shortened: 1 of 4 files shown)


src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -149,18 +165,24 @@ public void handleEvent(@NonNull DcEvent event) {
...
   }
 
-  private void updateTitle() {
+  private void updateTitleAndMenu() {
     Util.runOnAnyBackgroundThread(() -> {
       final JSONObject info = this.dcAppMsg.getWebxdcInfo();
       final String docName = JsonUtils.optString(info, "document");
       final String xdcName = JsonUtils.optString(info, "name");
       final String chatName =  WebxdcActivity.this.dcContext.getChat(WebxdcActivity.this.dcAppMsg.getChatId()).getName();
+      final boolean currHasSourceCodeUrl = !JsonUtils.optString(info, "source_code_url").isEmpty();
```
adbenitez, line 178: why not instead of saving a boolean, just save the URL and then avoid loading it each time in the menu? anyway it is an static property of the webxdc metadata

adbenitez (approved):
great!! instead of source code URL I think something like "project site" or "web site" of the app would have been something more general and user friendly then for most projects this would be the source code url but from a point of view of an user you would want to see the app's page not "source code"

r10s:
> great!! instead of source code URL I think something like "project site" or "website" of the app would have been something more general and user friendly then for most projects this would be the source code url but from a point of view of an user you would want to see the app's page not "source code" that might not apply for non-free apps

yip, we can improve in the future as needed. for now it is mainly about developers that see an webxdc and that way easily can go to the source, adapt or fork.

src/org/thoughtcrime/securesms/WebxdcActivity.java
r10s: it is loaded only on tapping, but yes, probably makes sense.

r10s: i changed that, thanks for the tip!

PR #2302, opened 2022-05-17
Add missing translations

OzancanKaratas:
This pull request will add changes that will be explained on each commit.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -905,7 +905,7 @@
     <string name="open_settings">Open Settings</string>
     <!-- iOS device contact alert -->
     <string name="import_device_contacts">Import device contacts</string>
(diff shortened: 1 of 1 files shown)


OzancanKaratas:
@cyBerta or @r10s: Should I add the desktop strings?

r10s:
> @cyBerta or @r10s: Should I add the desktop strings?

no, i think, it is good to keep the scope of the pr small.

r10s (approved):
looks good to me, i'll merge that and also push the strings to transifex so that they can translated there.

not sure if there is some additional magic needed to get them to `deltachat-ios/LANG.lproj/InfoPlist.strings` but as the other strings from `deltachat-ios/Info.plist` are translated already, i am pretty optimistic that it just works on next pulling :)

PR #2303, opened 2022-05-18
exclude com.google.android.gms from apk binaries

adbenitez:
recently DeltaLab got added to [IzzyOnDroid](https://android.izzysoft.de/repo) repo and after being scanned with some apk analyzer got marked with anti-features because GMS blobs are bundled in the apk due to maplibre including it, this is also happening in official Delta Chat.

I excluded gms and all seems to work fine because in practice we don't use google services, and reading maplibre's repo, the class using the `gms` package is `com.mapbox.mapboxsdk.location.engine.GoogleLocationEngineImpl` and in:
https://github.com/maplibre/maplibre-gl-native/blob/ea234edf67bb3aec75f077e15c1c30c99756b926/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/location/LocationComponent.java#L83
it says that for `GoogleLocationEngineImpl` to be used:
> simply add the Google Play Location Services dependency in your build script.

since we don't add it, it is unused, but yet the `com.google.android.gms` was bundled in the apk

Without this Delta Chat might get flagged on f-droid (other apps are getting similar issues, check https://gitlab.com/fdroid/fdroiddata/-/merge_requests/10991 )

as a plus this should make the apk size a bit smaller :)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,7 +66,9 @@ dependencies {
     }
     implementation 'com.annimon:stream:1.1.8' // brings future java streams api to SDK Version < 24
     implementation 'com.codewaves.stickyheadergrid:stickyheadergrid:0.9.4' // glues the current time segment text in the gallery to the top.
-    implementation 'org.maplibre.gl:android-sdk:9.4.0'
+    implementation ('org.maplibre.gl:android-sdk:9.4.0') {
+        exclude group: 'com.google.android.gms'
+    }
 
     testImplementation 'junit:junit:4.12'
     testImplementation 'junit:junit:4.12'


licaon-kter:
Ref: https://gitlab.com/fdroid/fdroiddata/-/commit/a55e368d341bc5407cf77d8c8c1e4c63e7183a7f

r10s (approved):

licaon-kter:
Confirmed this cleans up :+1:

licaon-kter:
@adbenitez right, 626 needs to be disabled too, 621 does not have these classes in `smali/`, this happens when one mixes [2 tasks](https://github.com/vector-im/element-android/issues/6100) at once :(

ref: https://gitlab.com/fdroid/fdroiddata/-/commit/55751671beaac41b9543a9ad9b03082874f7c775

r10s:
@licaon-kter i yesterday tagged 1.30.2 (631), that includes this pr and should hopefully work out :)

f-droid picked the tag already up, https://gitlab.com/fdroid/fdroiddata/blob/master/metadata/com.b44t.messenger.yml , so it is probably in the build queue again - can you say how long f-droid builds take these days until available in the f-droid app?

licaon-kter:
@r10s current cycle has ended and wait for publishing, so this will be in the next one: https://monitor.f-droid.org/builds/running

https://gitlab.com/fdroid/wiki/-/wikis/FAQ#how-long-does-it-take-for-my-app-to-show-up-on-website-and-client :slightly_smiling_face:

r10s:
@licaon-kter thanks a lot for these information in special and for your work on f-droid in general <3

PR #2305, opened 2022-05-24
change upload script to use mailadm.testrun.org instead of testrun.org

missytake:


r10s (approved):
yip, this works, merging this to have sth. working again.

for the now weird domain: maybe we should just download the testrun-releases to download.delta.chat/android- there are also the nightlies and the final releases.

PR #2308, opened 2022-05-25
use `download.delta.chat` also for beta versions

r10s:
nightlies are uploaded there anyway,
so, just everything can go there,
leaving testrun.org for other purposes.

i already tested the script, the `beta` folder was created implicitly and things work out totally as expected :)

@missytake i think, it makes sense to cleanup the `beta` folder similar to the `nightlies` folder, after 1~2 months, betas can be deleted.


adbenitez (approved):

missytake (approved):
looks good :)

PR #2309, opened 2022-05-26
fix behavior of system-back-button for webxdc

r10s:
for webxdc, the system-back-button should do the same
as the back-button in the title bar - just close the webxdc.

without this commit, it loads the previous page, which is often unexpected -
eg. many games use reload() for simplicity,
so you would have press one time 'back' for each round :)

even if there may be situations where this may be useful,
we do not want to encourage webxdc to rely on the system-back-button -
eg. desktop or ios currently just do not have this button.

technically, this pr changes the default -
so, for all classes derived from WebViewActivity, the system-back-button will
now behave as the title-bar-back-button.

for the local-help, this behavior is changed explictly -
i think, for HTML-mails, the 'just close' approach is also better.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConnectivityActivity.java
+++ b/src/org/thoughtcrime/securesms/ConnectivityActivity.java
@@ -43,11 +43,4 @@ public class ConnectivityActivity extends WebViewActivity implements DcEventCent
   public void handleEvent(@NonNull DcEvent event) {
     refresh();
   }
-
-  @Override
-  public void onBackPressed() {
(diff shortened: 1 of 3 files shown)


adbenitez (approved):

adbenitez (approved):
great! I first noticed this with the [test webxdc](https://github.com/webxdc/webxdc-test) when testing the internal page stuff,and with the tower builder game it was a bit annoying as well

r10s:
yeah, i noticed that on hextris, which also just calls reload() for a new round :)

PR #2311, opened 2022-06-01
add desktop strings

Simon-Laux:
related ﻿desktop pr: https://github.com/deltachat/deltachat-desktop/pull/2794

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -372,7 +372,7 @@
     <string name="saved_messages_explain">• Forward messages here for easy access\n\n• Take notes or voice memos\n\n• Attach media to save them</string>
     <!-- this "Saved" should match the "Saved" from "Saved messages" -->
     <string name="saved">Saved</string>
-    <string name="save_as">Saved</string>
+    <string name="save_as">Save As</string>
     <string name="retry_send">Retry to send message</string>
     <!-- reasons for a disabled message composer -->
     <string name="messaging_disabled_not_in_group">You cannot write because you are not in this group. To join, ask another member.</string>
@@ -388,8 +388,6 @@
     <string name="mailing_list_profile_info">Changes on mailing list name and image apply to this device only.</string>
 
     <!-- webxdc -->
-    <!-- atleast desktop needs this string -->
-    <string name="webxdc">Webxdc</string>
     <!-- "Start..." button for an app -->
     <string name="start_app">Start…</string>
     <!-- this is a warning that is shown when one tries to send something to a chat that is not yet accepted. -->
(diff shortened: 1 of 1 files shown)


res/values/strings.xml
```diff
@@ -368,6 +372,7 @@
     <string name="saved_messages_explain">• Forward messages here for easy access\n\n• Take notes or voice memos\n\n• Attach media to save them</string>
     <!-- this "Saved" should match the "Saved" from "Saved messages" -->
     <string name="saved">Saved</string>
+    <string name="save_as">Saved</string>
```
r10s, line 375: typo? copy+paste error?

res/values/strings.xml
```diff
@@ -383,6 +388,8 @@
     <string name="mailing_list_profile_info">Changes on mailing list name and image apply to this device only.</string>
 
     <!-- webxdc -->
+    <!-- atleast desktop needs this string -->
+    <string name="webxdc">Webxdc</string>
```
adbenitez, line 392: I think that name shouldn't be translated, at least android uses it directly

res/values/strings.xml
adbenitez, line 375: save_as == saved ???

adbenitez (approved):

res/values/strings.xml
adbenitez: ah you already spot it :)

res/values/strings.xml
```diff
@@ -882,6 +894,20 @@
     <string name="show_window">Show Window</string>
     <string name="login_socks5_experimental_warning">SOCKS5 support is currently experimental. Please use at your own risk. If you type in an address in the e-mail field, there will be DNS lookup that won\'t get tunneled through SOCKS5.</string>
 
+    <!-- desktop keybindings -->
+    <string name="desktop_keybindings_composer">MessageInput / Composer</string>
```
r10s, line 898: i think, this needs some context for the translators (just add a html-comment above the strings, see other strings with context.

res/values/strings.xml
```diff
@@ -882,6 +894,20 @@
...
+    <string name="desktop_keybindings_global">Global Shortcuts</string>
+    <string name="desktop_keybindings_action_switch_between_chats">Switch between Chats</string>
+    <string name="desktop_keybindings_action_scroll_messages">Scroll Messages</string>
+    <string name="desktop_keybindings_action_jump_to_search">Jump to Search</string>
+    <string name="desktop_keybindings_action_focus_composer">Focus Composer</string>
+    <string name="desktop_keybindings_action_open_help">Open Help</string>
+    <string name="desktop_keybindings_action_open_settings">Open Settings</string>
+    <string name="desktop_keybindings_action_force_refresh_network">Force Refresh Network</string>
+    <string name="desktop_keybindings_action_send_message">Send Message</string>
+    <string name="desktop_keybindings_action_insert_newline">Insert Newline</string>
```
r10s, line 908: i suggest, where doable, just use the existing command strings instead of adding new ones. 

eg. we can use the existing strings for "Send Message",  "Help" and "Open Settings". this would also open an easy way to add even more keybindings (Archived Chats, Gallery, whatever) - otherwise we would have each and every command doubled.

for the other strings, i'd add them more generalized, so remove the `desktop_keybindings_action_` prefix, we usually do not go for that and have eg. already `jump_to_message` - so a `jump_to_search` would make much sense.

/me knows that this approach is not 100% done everywhere :)

res/values/strings.xml
```diff
@@ -882,6 +894,20 @@
...
+    <string name="desktop_keybindings_dialog_title">Keybindings</string>
```
r10s, line 909: same here, just call it keybindings - might be, we can resuse these strings when it comes to tablets :)

r10s (commented):
see comments

r10s (approved):

PR #2314, opened 2022-06-05
show more info messages by adapting spacing to the overall layout

r10s:
the spacing was just a bit too much,
eg. bubble spacing is 6dp atop and abottom.

this was always not-so-good, however, as info-messages are
more in use by webxdc now, it gets more visible.

the new spacing roughly matches the layout on ios.

before/after:

 

nb: some of the info-messages from the screenshot will be cleaned up by https://github.com/deltachat/deltachat-core-rust/pull/3395 , but other series would not :)

### DIFF
--- a/res/layout/conversation_item_update.xml
+++ b/res/layout/conversation_item_update.xml
@@ -8,8 +8,8 @@
     android:focusable="true"
     android:gravity="center"
     android:orientation="vertical"
-    android:paddingTop="13dp"
-    android:paddingBottom="13dp"
+    android:paddingTop="5dp"
+    android:paddingBottom="7dp"
     android:paddingLeft="28dp"
     android:paddingRight="28dp">
 


res/layout/conversation_item_update.xml
```diff
@@ -8,8 +8,8 @@
     android:focusable="true"
     android:gravity="center"
     android:orientation="vertical"
-    android:paddingTop="13dp"
-    android:paddingBottom="13dp"
+    android:paddingTop="5dp"
+    android:paddingBottom="7dp"
```
r10s, line 12: having the same amount atop and abottom leads irregular paddings; i do not want to refactor layout in this pr, therefore just so :)

Hocuri (approved):

adbenitez (approved):
nicer indeed!

PR #2331, opened 2022-07-06
Revise English

rosano:
I want to look over all the English strings. Here are just a few, should we try merging a small amount to see what happens in Transifex (in terms of causing / minimizing notifications for other languages) first? Or should I go through the whole file?

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -913,7 +913,7 @@
 
     <!-- iOS specific strings, developers: please take care to remove strings that are no longer used! -->
     <string name="stop_sharing_location">Stop sharing location</string>
-    <string name="a11y_voice_message_hint_ios">After recording double-tap to send. To discard the recording scrub with two fingers.</string>
+    <string name="a11y_voice_message_hint_ios">After recording, double-tap to send. To discard the recording, scrub with two fingers.</string>
     <string name="a11y_connectivity_hint">Double tap to view connectivity details.</string>
     <string name="login_error_no_internet_connection">No internet connection, login failed.</string>
     <string name="share_account_not_configured">Account is not configured.</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -54,7 +54,7 @@
     <string name="activate">Activate</string>
     <string name="load_remote_content">Load Remote Images</string>
     <!-- Possible answers to the question are: Never, Always, Once -->
-    <string name="load_remote_content_ask">Remote images can be used to track you.\n\nThe setting also allows loading fonts and other content. Even if it is disabled, you may still see embedded or cached images.\n\nLoad remote images?</string>
+    <string name="load_remote_content_ask">Remote images can track you.\n\nThis setting also may load fonts and other content. If disabled, embedded or cached images may appear.\n\nLoad remote images?</string>
     <string name="always">Always</string>
     <string name="once">Once</string>
     <string name="show_password">Show Password</string>
@@ -65,7 +65,7 @@
     <string name="done">Done</string>
     <string name="undo">Undo</string>
     <string name="offline">Offline</string>
-    <!-- Translators: used eg. for the next view, could also be "continue" or so. as used in iOS headers, the string should be as short as possible, though. -->
+    <!-- For the next view or as "continue". Should be as short as possible. -->
     <string name="next">Next</string>
     <string name="error">Error</string>
     <string name="error_x">Error: %1$s</string>
@@ -79,7 +79,7 @@
     <string name="password">Password</string>
     <string name="existing_password">Existing Password</string>
     <string name="now">Now</string>
-    <!-- Translators: used as a headline in sections with actions that cannot be undone. could also be "Caution" or "Cave" or so. -->
+    <!-- Headline for destructive actions with no undo. Could also be "Caution" or "Warning". -->
     <string name="danger">Danger</string>
     <string name="today">Today</string>
     <string name="yesterday">Yesterday</string>
@@ -87,15 +87,15 @@
     <string name="this_month">This month</string>
     <string name="last_week">Last week</string>
     <string name="last_month">Last month</string>
-    <!-- the placeholder will be replaced by a date or time, resulting eg. in "Last seen at 12:13 AM" or "Last seen Nov 12" -->
+    <!-- Placeholder replaced by date or time, resulting in "Last seen at 12:13 AM" or "Last seen Nov 12" -->
     <string name="last_seen_at">Last seen at %1$s</string>
     <string name="last_seen_unknown">Last seen: Unknown</string>
-    <!-- Translators: show beside messages that are "N minutes old". prefer a short string, prefer abbreviations if they're well-known -->
+    <!-- Shown beside messages that are "N minutes old". Prefer short strings, or well-known abbreviations. -->
     <plurals name="n_minutes">
         <item quantity="one">%d min</item>
         <item quantity="other">%d min</item>
     </plurals>
-    <!-- Translators: show beside messages that are "N hours old". prefer a short string, prefer abbreviations if they're well-known -->
+    <!-- Shown beside messages that are "N hours old". Prefer short strings, or well-known abbreviations. -->
     <plurals name="n_hours">
         <item quantity="one">%d hour</item>
         <item quantity="other">%d hours</item>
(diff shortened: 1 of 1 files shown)


rosano:
@r10s I reverted the capitalization. Should I continue revising the rest of the file or would you prefer just doing a small change?

res/values/strings.xml
```diff
@@ -33,13 +33,13 @@
...
-    <!-- In the meaning of "To archive", "Put a chat to the archive". This string does not mean "The Archive". -->
+    <!-- Verb "to archive", as in "put a chat in the archive", not a noun "The Archive". -->
     <string name="archive">Archive</string>
-    <!-- In the meaning of "To unarchive", "Remove a chat from the archive", opposite of the previous string -->
+    <!-- Verb "to unarchive", as in "remove a chat from the archive", opposite of the previous string -->
     <string name="unarchive">Unarchive</string>
     <string name="mute">Mute</string>
     <string name="ephemeral_messages">Disappearing Messages</string>
-    <string name="ephemeral_messages_hint">These settings apply to all chat members using Delta Chat. However, they may copy, save and forward messages or use other e-mail clients.</string>
+    <string name="ephemeral_messages_hint">Removes messages after the specified duration from all members of this chat; this applies only if they use Delta Chat and does not prevent copying, saving, and forwarding messages or using other e-mail clients.</string>
```
r10s, line 42: tbh,i like the old sentence more. there was quite some effort put into make it that short and precise (place is limited!) 

eg. it is pretty clear by the context that it is about "removing messages after some duration".

rosano: Now one character shorter than the original

r10s:
> @r10s I reverted the capitalization as per https://github.com/deltachat/deltachat-android/pull/1979. Should I continue revising the rest of the file or would you prefer just doing a small change?

i think, we figured out the scope of the possible changes, you can continue in this pr if you like. larger rewordings, making things much longer etc. should go to separate prs.

rosano:
In what context does this string appear?
https://github.com/deltachat/deltachat-android/blob/7cc844af95589ad8586730a386c7f304de048d0c/res/values/strings.xml#L290

r10s:
> In what context does this string appear?
> `<string name="videochat_you_invited_hint">You invited to a video chat.</string>`

it is shown as a message on your device when you send a videochat invitation (you need to enable videochat at "experimental features" for this). 
it is shown in the message that says "Tap to Open"

res/values/strings.xml
```diff
@@ -285,24 +285,24 @@
 
     <string name="videochat">Video Chat</string>
     <string name="videochat_invite_user_to_videochat">Invite %1$s to a video chat?</string>
-    <string name="videochat_invite_user_hint">This requires a compatible app or a compatible browser on both ends.</string>
+    <string name="videochat_invite_user_hint">This requires a compatible app or browser on both ends.</string>
     <string name="videochat_contact_invited_hint">%1$s invited to a video chat.</string>
-    <string name="videochat_you_invited_hint">You invited to a video chat.</string>
+    <string name="videochat_you_invited_hint">You were invited to a video chat.</string>
```
adbenitez, line 290: mmmh... I think this is not right, the string is shown to yourself like "you did invite to a video chat" on the receiver side it is "Foo invited to a video chat" look the previous string

r10s:
i also added @link2xt for review as many of the very precise wordings come from him :)

for me, this looks good all in all :)

rosano:
When does this appear? It reads to me as 'The group left' but maybe it means 'You left the group'?

https://github.com/deltachat/deltachat-android/blob/7cc844af95589ad8586730a386c7f304de048d0c/res/values/strings.xml#L714

r10s:
"Group left." is shown as the (hopefully :) last message when the user left a group using the menu command "Leave Group".

rosano:
Ok I'm finished and will leave it for review.

res/values/strings.xml
```diff
@@ -617,7 +617,7 @@
     <string name="pref_watch_sent_folder">Watch Sent Folder</string>
     <string name="pref_send_copy_to_self">Send Copy to Self</string>
     <string name="pref_send_copy_to_self_explain">Required when using this account on multiple devices.</string>
-    <string name="pref_auto_folder_moves">Automatic Moves to DeltaChat Folder</string>
+    <string name="pref_auto_folder_moves">Move automatically to DeltaChat Folder</string>
```
r10s, line 620: the old one looks a bit better to me in the context of the "IMAP Folder Handling" - or is the old one just wrong english?

this is not a blocker, however.

r10s (approved):
i have some minor comments, but all in all this looks very good!

also nice that most strings are shorter now :)

res/values/strings.xml
rosano: The old one wasn't wrong but a bit more awkward and opaque, there's no ambiguity about the second one.

res/values/strings.xml
```diff
@@ -250,16 +250,16 @@
     <string name="title_share_location">Share location with all group members</string>
     <string name="device_talk">Device Messages</string>
     <string name="device_talk_subtitle">Locally generated messages</string>
-    <string name="device_talk_explain">Messages in this chat are generated locally by your Delta Chat app. Its makers use it to inform about app updates and problems during usage.</string>
+    <string name="device_talk_explain">Messages in this chat are generated on your device and come from the makers of Delta Chat to inform about app updates and problems during usage.</string>
```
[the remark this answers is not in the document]
rosano: Maybe like this? Feel free to suggest an alternative: https://github.com/deltachat/deltachat-android/blob/98d7efef677c29b74dc72937b4d76515430a64ce/res/values/strings.xml#L253

res/values/strings.xml
```diff
@@ -315,17 +315,17 @@
...
-    <string name="ask_unblock_contact">Unblock this contact? You will once again be able to receive messages from this contact.</string>
-    <string name="ask_delete_contacts">Delete contacts?\n\nContacts with ongoing chats and contacts from the system\'s address book cannot be deleted permanently.</string>
-    <string name="ask_delete_contact">Delete contact %1$s?\n\nContacts with ongoing chats and contacts from the system\'s address book cannot be deleted permanently.</string>
+    <string name="ask_export_attachment">Exporting attachments will allow other apps on your device to access them.\n\nContinue?</string>
+    <string name="ask_block_contact">Block this contact? You will no longer receive messages from them.</string>
+    <string name="ask_unblock_contact">Unblock this contact? You will then be able to receive messages from them.</string>
+    <string name="ask_delete_contacts">Delete contacts?\n\nContacts with ongoing chats and from the system address book cannot be deleted permanently.</string>
+    <string name="ask_delete_contact">Delete contact %1$s?\n\nContacts with ongoing chats and contacts from the system address book cannot be deleted permanently.</string>
```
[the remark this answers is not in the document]
rosano: Could you clarify what was removed? I don't follow. And what is the other string you refer to?

rosano:
I'm not sure how merging works here with Transifex, does it automatically go there or does one have to do something specific?

res/values/strings.xml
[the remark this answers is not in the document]
r10s: there is now an imbalance between `ask_delete_contacts` and `ask_delete_contact` - one says "Contacts with ongoing chats and from the system address book" the other says "Contacts with ongoing chats and _contacts_ from the system address book".

to fix this, either also remoe "contacts" in `ask_delete_contact`:

```suggestion
    <string name="ask_delete_contact">Delete contact %1$s?\n\nContacts with ongoing chats and the system address book cannot be deleted permanently.</string>
```

or re-add it to the other string. while the latter is shorter, maybe the old one is clearer.

i first read that it is about deleting the system address book, tbh.

so, maybe the old string with the repetition is better and clearer, though a bit longer and maybe not that polished.

but i personally would still be fine with both variants, but there should be no imbalance, see above.

r10s:
> I'm not sure how merging works here with Transifex, does it automatically go there or does one have to do something specific? I'm inclined to ask someone else to do that.

i can merge this; afterwards, i will immediately pull all translations from transifex and push everything again with the change translation source (english). this step is a bit critical, it is easy to loose translations just because of a changed comma in the source :)

res/values/strings.xml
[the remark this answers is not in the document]
r10s: for the short variant: isn't an "or" better?

"Contacts with ongoing chats or from the system address book cannot be deleted permanently."

rosano (commented):
Ok I think everything has been resolved. I went with 'or' before I saw the repeating 'contacts' suggestion. Feel free to change it. Ready to merge.

r10s:
great to have that happen!

after merging, i ran `./scripts/tx-update-changed-sources.sh` so that the changes sources and help texts are available on transifex, without deleting the existing translations.

PR #2332, opened 2022-07-06
Pass APP_ABI to ndk-build instead of creating temporary Application.mk

link2xt:


adbenitez (approved):

adbenitez:
didn't tried it yet, will try now if you want before merging

adbenitez:
it seems to work, now it is fast in subsequent runs

PR #2337, opened 2022-07-09
set draft from mailto data in QR

adbenitez:
depends on https://github.com/deltachat/deltachat-core-rust/pull/3492

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -33,6 +33,7 @@ import android.os.Bundle;
 import android.os.Vibrator;
 import android.provider.Browser;
 import android.text.Editable;
+import android.text.TextUtils;
 import android.text.TextWatcher;
 import android.util.Log;
 import android.util.Pair;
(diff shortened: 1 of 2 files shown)


Hocuri:
Unfortunately, this PR makes the tests `testSimpleForwarding` and `testShareFromScreenshot` fail :/

Quick history: Apparently it's hard getting this sharing stuff right, so I wrote some UI regression tests for this, which you can run as explained at https://github.com/deltachat/deltachat-android#run-ui-tests-and-benchmarks. If this doesn't work or sth. is unclear, please tell me :)

Sooooo, this probably means that this PR introduces new bugs, though I didn't dig into this further. By the way, you could also test the list at https://github.com/deltachat/interface/blob/master/user-testing/mailto-links.md if that's simpler, I guess that some of them will also fail then (and we should go through the list anyway). If you want, you can even expand the list with a test for that what you implemented in the PR here works.

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -752,20 +753,16 @@ public void onSuccess(Boolean result) {
    * @return
    */
   private ListenableFuture<Boolean> initializeDraft() {
-    if (isMailToIntent()) {
```
r10s, line 755: just removing this may be the cause of the regression.

adbenitez:
> Unfortunately, this PR makes the tests `testSimpleForwarding` and `testShareFromScreenshot` fail :/

are that tests not automatic in our CI I guess, can you try again? now I merge the existing draft and new draft text

Hocuri:
Correct, it's not in the CI; would be nice, but probably not that easy since it would require an Android VM somewhere.

Your fix works like a charm and the tests run through now :tada:, nice!

We should also do the manual testing before merging, and probably wait until after the release with merging since it's so easy to introduce regressions here.

It's possible that this partially re-introduced the bug fixed at https://github.com/deltachat/deltachat-android/pull/1770 for complicated reasons [1], but then I noticed that it's already been re-introduced before, maybe by https://github.com/deltachat/deltachat-android/pull/2246. Anyway, this needs a fix, but it seems like this PR here doesn't introduce a regression.

And then I found what created all these problems and made the workaround necessary in the first place: When sharing only a MAILTO: address, or only setting TEXT_EXTRA like you did, `isSharing()` returns `false`, so `handleSharing()` was not called.  I had been working on another fix, but i like your fix better since it also merges drafts.

[1]: See this block:

    if (doReinitializeDraft) {
      initializeDraft();
      doReinitializeDraft = false;
    }

since initializeDraft() now also gets its text from the intent, it might overwrite other text that was shared inbetween. A `resetRelayingMessageContent()` will do the trick, if that is actually a problem, I didn't test this yet.

adbenitez:
> Your fix works like a charm and the tests run through now , nice!

so this can be merged already? it was pending due an upcoming release that was released time ago and this PR got a bit dusty, ping @r10s @Hocuri

r10s:
at a glance, there is still an unsolved regression. so, i would not merge that now as we're aiming for 1.36 soon.

adbenitez:
> i did not review in detail - is there still an unsolved regression?

I am not sure, @Hocuri said the regression was already introduced by other PR:

> Anyway, this needs a fix, but it seems like this PR here doesn't introduce a regression.

it is ok to wait again for a next version, hope I don't forget about this again until again we can't merge it because the next upcoming version :/

adbenitez:
@r10s, @Hocuri said it should be fine to merge this, since it doesn't makes things worse, the issues are already there

src/org/thoughtcrime/securesms/ConversationActivity.java
Hocuri: No, that's fine, it's not the cause of the regression. I don't remember the details of the regression anymore, since it's complicated enough that everytime I don't think about it for some months, I completely forget all the details. But the regression has something to do with sometimes starting the activity using startActivity instead of startActivityForResult.

r10s:
@adbenitez then, let's merge it, @Hocuri is more in that share-forward-mailto-area than me anyway :)

adbenitez:
@r10s actually lets wait for @Hocuri to re-review, there was some unmerged changes in his machine

Hocuri (approved):
LGTM, no changes necessary to merge. :partying_face:!

> there was some unmerged changes in his machine

I went through them now, they were independent simplifications. But 1. they are unrelated to this PR and 2. I don't actually want to merge them because they might make things fail again. Maybe I should create a PR that explains some of the current peculiarities, so that it'll be easier to reduce code complexity together with bug fixes, just like the PR here is doing - good work @adbenitez to remove lines and add a feature in one PR!

> A `resetRelayingMessageContent()` will do the trick, if that is actually a problem, I didn't test this yet.

Doing this might have other implications, and the PR here doesn't introduce any new bugs (as said, it might partly introduce a bug, but as said was already introduced before, anyway), so, let's not do this for now.

PR #2338, opened 2022-07-09
Don't animate the spinning circle if animations are disabled

Hocuri:
This is needed for the UI tests to run (since https://github.com/deltachat/deltachat-core-rust/pull/3150 didn't get merged, but I didn't try to run the UI tests since then).

Plus, it's just the correct behavior; if the animations are disabled system-wide, there should be no animations. A slight downside of this PR is that the circle doesn't make that much sense if it's not spinning, so we could alternatively replace the image with an hourclock or a clock. I don't think that's worth it though, because the kind of users that disables the animations probably won't mind.

Also, this PR shows an error if someone tries to run the tests with animations on (since they will probably fail anyway).

### DIFF
--- a/androidTest/com/b44t/messenger/TestUtils.java
+++ b/androidTest/com/b44t/messenger/TestUtils.java
@@ -19,6 +19,7 @@ import org.thoughtcrime.securesms.ConversationListActivity;
 import org.thoughtcrime.securesms.R;
 import org.thoughtcrime.securesms.connect.AccountManager;
 import org.thoughtcrime.securesms.connect.DcHelper;
+import org.thoughtcrime.securesms.util.AccessibilityUtil;
 import org.thoughtcrime.securesms.util.Prefs;
 import org.thoughtcrime.securesms.util.Util;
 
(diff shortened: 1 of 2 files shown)


Hocuri:
(That's about the spinning circle that's shown in the bottom right corner of a pending message and in the chat list)

adbenitez (approved):

r10s (commented):
> (That's about the spinning circle that's shown in the bottom right corner of a pending message and in the chat list)

while the code looks good, i am unsure if this is the gist of the setting.

all places where i have seen apps reacting to the "accessibility animation switch" [1] is more about  _"larger"_ animation as "switch activities" or "smooth scroll to another message" or not really needed animations as "new fab icon popping up" or other flashes. afaik, gist is to not trigger things eg. due epilepsy.

this pr is about an animation in a very _"small"_ region, i mean this is more like the animation when switch is toggled or so. triggering things seems to be much less probabe - and maybe outweights the mentioned disadvantages of having _some_ visual feedback.

if one has a look at eg. Signal, only large things are disabled.

is there any source that says this kind of animations should be disabled as well? or did we get feedback in that direction?


---

[1] note that the very most andorid ui just have a single switch and do not differ between the options available technically. this is just a 0/1 setting and not three floats or so (cmp various discussions on so on that or search for screenshors)

r10s (approved):

r10s:
let's merge that in as it fixes tests, we can iterate over at a later point, maybe even go for a not animated loading mark (as on ios)

PR #2343, opened 2022-07-14
Optimize translations

OzancanKaratas:
Some quote marks have been removed and a few typos have been fixed.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -199,7 +199,7 @@
     <string name="menu_add_attachment">Add Attachment</string>
     <string name="menu_leave_group">Leave Group</string>
     <string name="menu_delete_chat">Delete Chat</string>
-    <string name="ask_delete_named_chat">Are you sure you want to delete chat named \"%1$s\"?</string>
+    <string name="ask_delete_named_chat">Are you sure you want to delete chat \"%1$s\"?</string>
     <string name="menu_delete_messages">Delete Messages</string>
     <string name="delete_contact">Delete Contact</string>
     <string name="menu_delete_location">Delete this Location?</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -620,7 +620,7 @@
     <string name="pref_auto_folder_moves">Move Automatically to DeltaChat Folder</string>
     <string name="pref_auto_folder_moves_explain">Chat conversations are moved to avoid cluttering the Inbox</string>
     <string name="pref_only_fetch_mvbox_title">Only Fetch from DeltaChat Folder</string>
-    <string name="pref_only_fetch_mvbox_explain">Ignores other folders. Requires your server to move chat messages to the DeltaChat folder.</string>
+    <string name="pref_only_fetch_mvbox_explain">Ignore other folders. Requires your server to move chat messages to the DeltaChat folder.</string>
     <string name="pref_show_emails">Show Classic E-Mails</string>
     <string name="pref_show_emails_no">No, chats only</string>
     <string name="pref_show_emails_accepted_contacts">For accepted contacts</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -706,7 +706,7 @@
 
 
     <!-- system messages -->
-    <string name="systemmsg_group_name_changed">Name of group %1$s changed as \"%2$s\".</string>
+    <string name="systemmsg_group_name_changed">Name of group \"%1$s\" changed as \"%2$s\".</string>
     <string name="systemmsg_group_image_changed">Group image changed.</string>
     <string name="systemmsg_group_image_deleted">Group image deleted.</string>
     <string name="systemmsg_member_added">Member %1$s added.</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -781,7 +781,7 @@
     <string name="qrshow_join_contact_hint">Scan to chat with %1$s</string>
     <string name="qrshow_join_contact_no_connection_toast">No internet connection, can\'t perform QR code setup.</string>
     <string name="qraccount_ask_create_and_login">Create new e-mail address on %1$s and log in there?</string>
-    <string name="qraccount_ask_create_and_login_another">Create new e-mail address on %1$s and log in there?\n\nYour existing account will not be deleted. Use the \"Switch account\" item to switch between your accounts.</string>
+    <string name="qraccount_ask_create_and_login_another">Create new e-mail address on %1$s and log in there?\n\nYour existing account will not be deleted. Use the \"Switch Account\" item to switch between your accounts.</string>
     <string name="qraccount_success_enter_name">Login successful—your e-mail address is %1$s\n\nIf you like, you can now enter a name and an profile image that will be displayed to people you write to.</string>
     <string name="qraccount_qr_code_cannot_be_used">The scanned QR code cannot be used to set up a new account.</string>
     <string name="qraccount_use_on_new_install">The scanned QR code is for setting up a new account. You can scan the QR code when setting up a new Delta Chat installation.</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -199,7 +199,7 @@
     <string name="menu_add_attachment">Add Attachment</string>
     <string name="menu_leave_group">Leave Group</string>
     <string name="menu_delete_chat">Delete Chat</string>
-    <string name="ask_delete_named_chat">Are you sure you want to delete chat with %1$s?</string>
+    <string name="ask_delete_named_chat">Are you sure you want to delete chat named \"%1$s\"?</string>
     <string name="menu_delete_messages">Delete Messages</string>
     <string name="delete_contact">Delete Contact</string>
     <string name="menu_delete_location">Delete this Location?</string>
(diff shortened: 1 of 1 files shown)


res/values/strings.xml
```diff
@@ -199,7 +199,7 @@
     <string name="menu_add_attachment">Add Attachment</string>
     <string name="menu_leave_group">Leave Group</string>
     <string name="menu_delete_chat">Delete Chat</string>
-    <string name="ask_delete_named_chat">Are you sure you want to delete \"%1$s\"?</string>
+    <string name="ask_delete_named_chat">Are you sure you want to delete chat with %1$s?</string>
```
r10s, line 202: this wording does not match when it is a groupname.

res/values/strings.xml
```diff
@@ -706,7 +706,7 @@
 
 
     <!-- system messages -->
-    <string name="systemmsg_group_name_changed">Group name changed from \"%1$s\" to \"%2$s\".</string>
+    <string name="systemmsg_group_name_changed">Name of group %1$s changed as \"%2$s\".</string>
```
r10s, line 709: i think, we should either leave the quotes or remove them; "changed as" does not look correct to me.

res/values/strings.xml
```diff
@@ -755,33 +755,33 @@
     <string name="qrscan_title">Scan QR Code</string>
     <string name="qrscan_hint">Hold your camera over the QR code.</string>
     <string name="qrscan_failed">QR code could not be decoded</string>
-    <string name="qrscan_ask_join_group">Do you want to join the group \"%1$s\"?</string>
+    <string name="qrscan_ask_join_group">Do you want to join the group %1$s?</string>
```
r10s, line 758: why remove the quotes here?

res/values/strings.xml
```diff
@@ -755,33 +755,33 @@
...
     <string name="qrscan_fingerprint_mismatch">The scanned fingerprint does not match the last seen for %1$s.</string>
     <string name="qrscan_no_addr_found">This QR code contains a fingerprint but no e-mail address.\n\nFor an out-of-band-verification, please establish an encrypted connection to the recipient first.</string>
     <string name="qrscan_contains_text">Scanned QR code text:\n\n%1$s</string>
     <string name="qrscan_contains_url">Scanned QR code URL:\n\n%1$s</string>
     <string name="qrscan_fingerprint_label">Fingerprint</string>
     <string name="withdraw_verifycontact_explain">This QR code can be scanned by others to contact you.\n\nYou can deactivate the QR code here and reactivate it by scanning it again.</string>
-    <string name="withdraw_verifygroup_explain">This QR code can be scanned by others to join the group \"%1$s\".\n\nYou can deactivate the QR code here and reactivate it by scanning it again.</string>
+    <string name="withdraw_verifygroup_explain">This QR code can be scanned by others to join the group %1$s.\n\nYou can deactivate the QR code here and reactivate it by scanning it again.</string>
```
r10s, line 765: why remove the quotes here?

res/values/strings.xml
```diff
@@ -755,33 +755,33 @@
...
     <string name="withdraw_qr_code">Deactivate QR Code</string>
     <!-- "could" in the meaning of "possible at some point in the past, but no longer possible today" -->
     <string name="revive_verifycontact_explain">This QR code could be scanned by others to contact you.\n\nThe QR code is not active on this device.</string>
     <!-- "could" in the meaning of "possible at some point in the past, but no longer possible today" -->
-    <string name="revive_verifygroup_explain">This QR code could be scanned by others to join the group \"%1$s\".\n\nThe QR code is not active on this device.</string>
+    <string name="revive_verifygroup_explain">This QR code could be scanned by others to join the group %1$s.\n\nThe QR code is not active on this device.</string>
```
r10s, line 770: why remove the quotes here?

res/values/strings.xml
```diff
@@ -755,33 +755,33 @@
...
     <string name="revive_qr_code">Activate QR Code</string>
     <string name="qrshow_title">QR Invite Code</string>
     <string name="qrshow_x_joining">%1$s joins.</string>
     <string name="qrshow_x_verified">%1$s verified.</string>
     <string name="qrshow_x_has_joined_group">%1$s joined the group.</string>
     <string name="qrshow_join_group_title">QR Invite Code</string>
     <!-- This text is shown inside the "QR code card" with very limited space; please formulate the text as short as possible therefore. The placeholder will be replaced by the group name, eg. "Scan to join group \"Testing group\"" -->
-    <string name="qrshow_join_group_hint">Scan to join group \"%1$s\"</string>
+    <string name="qrshow_join_group_hint">Scan to join group %1$s</string>
```
r10s, line 778: why remove the quotes here?

res/values/strings.xml
```diff
@@ -755,33 +755,33 @@
...
     <!-- This text is shown inside the "QR code card" with very limited space; please formulate the text as short as possible therefore. The placeholder will be replaced by name and address eg. "Scan to chat with Alice (alice@example.org)" -->
     <string name="qrshow_join_contact_hint">Scan to chat with %1$s</string>
     <string name="qrshow_join_contact_no_connection_toast">No internet connection, can\'t perform QR code setup.</string>
-    <string name="qraccount_ask_create_and_login">Create new e-mail address on \"%1$s\" and log in there?</string>
-    <string name="qraccount_ask_create_and_login_another">Create new e-mail address on \"%1$s\" and log in there?\n\nYour existing account will not be deleted. Use the \"Switch account\" item to switch between your accounts.</string>
+    <string name="qraccount_ask_create_and_login">Create new e-mail address on %1$s and log in there?</string>
+    <string name="qraccount_ask_create_and_login_another">Create new e-mail address on %1$s and log in there?\n\nYour existing account will not be deleted. Use the \"Switch Account\" item to switch between your accounts.</string>
```
r10s, line 784: why remove the quotes here?

res/values/strings.xml
r10s, line 202: this wording does not match when it is a groupname. why remove the quotes?

res/values/strings.xml
r10s, line 709: we should either leave the quotes or remove them; i think, the quotes make sense here. "changed as" does not look correct to me. also "Group name" sounds good to me.

r10s (commented):
thanks for the overview.

i commented some places wrt quotes.

for the rewording, reordering and case changes, not sure, cc @rosano 

in general, note that we usually want to start rewording discussions when there is sth. changed in an area; many things depend on opinion :)

res/values/strings.xml
OzancanKaratas: Words are not a placeholder. It does not require the use of quotation marks.

res/values/strings.xml
OzancanKaratas: Whether the variable here represents a group or individual, the deletion will only occur for one person. The person performing the deletion is chatting with the group or the person.

res/values/strings.xml
r10s: > Words are not a placeholder.

i do not really understand that.

using quotes or sth. like that for object names inserted to a sentence is a quite usual pattern. the quotes make clear what the name actually is, where it starts and where it ends. some advantages are recognizability and also tricking users with weird names is harder when there are quotes.

so, if we go for changing that, there must be very good reasons, i cannot see them :)

res/values/strings.xml
r10s: i do not think, the addition makes it clear that the chat is not deleted on other people's devices (i am also not sure if this additional information is needed). but make things worse for groups.

as said,  in general, we prefer rewordings only if there is concrete user feedback (by quite some users!) or if there is something reworked in that area.

res/values/strings.xml
OzancanKaratas: Could this cause a problem as variables only represent a name? I do not think so. When someone calls you by your name, they don't use quotation marks. You can think of it that way too.

res/values/strings.xml
rosano: @r10s Yes the old string is better. This feels a bit haphazard like many changes in this PR.

res/values/strings.xml
rosano: To make it clear whose copy is being deleted, I would suggest "delete your copy of *". This line could be an isolated PR as it might be useful to have it's own discussion.

rosano (commented):
Most of these changes ignore the (important) reasoning behind the quotation marks, and don't provide a strong argument for their removal.

We generally avoid making changes that will cause ripple effects / notifications in our translation system purely for minor cosmetic changes.

I would suggest the author to close this PR and open new ones for
- the case change on line 445, which should be changed, but not super urgent
- the line/discussion about communicating that only your copy is deleted, as we might need to clarify what the project intention is there.

r10s:
i'd also like to take this chance to thank @OzancanKaratas for the work on transifex and also on other prs wrt accessibility 🙏

there is also the idea to have a dedicated Delta Chat group for translations - and maybe alsofor accessible and/or blind people, however, it does not happen yet, maybe take this pr as a chance to evetually go for that.

OzancanKaratas:
@r10s, I am happy to contribute. I would like to say a few things. I have 7 years of experience in software localization. Some of the changes here have been replaced with strings that are easier for translators to understand. I think we should decide on strings one by one instead of closing this pull request. I've already made some of the changes you suggested. What do you think? Shall we reopen?

r10s:
in this pr, there are two strings left that fix upper/lowercase typos and are an improvement:

`tab_audio_empty_hint`, `qraccount_ask_create_and_login_another`

i created a pr for these fixes at: https://github.com/deltachat/deltachat-android/pull/2346 (i cannot reopen this one)

in general, for rewording discussions, please note that they're pretty time consuming and we cannot afford that always, resulting in pr laying around for months and everyone is frustrated :)

therefore, as said, we try to avoid that. typo fixes, of course are another thing.

rosano:
also easier to deal with tiny pull requests that are about one thing, rather than general 'optimizations/revisions' — see #2347 for example

PR #2347, opened 2022-07-16
fix gallery's "files" tab hint

r10s:
depending on the system (desktop), music may be displayed in other tabs; therefore, just remove that hint. 'files' also include 'music' in the remaining cases.

came over that when filing https://github.com/deltachat/deltachat-desktop/issues/2829

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -439,7 +439,7 @@
     <string name="tab_links">Links</string>
     <string name="tab_map">Map</string>
     <string name="tab_gallery_empty_hint">Images and videos shared in this chat will appear here.</string>
-    <string name="tab_docs_empty_hint">Documents, music, and other files shared in this chat will appear here.</string>
+    <string name="tab_docs_empty_hint">Documents and other files shared in this chat will appear here.</string>
     <string name="tab_image_empty_hint">Images shared in this chat will appear here.</string>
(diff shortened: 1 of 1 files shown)


res/values/strings.xml
```diff
@@ -439,7 +439,7 @@
     <string name="tab_links">Links</string>
     <string name="tab_map">Map</string>
     <string name="tab_gallery_empty_hint">Images and videos shared in this chat will appear here.</string>
-    <string name="tab_docs_empty_hint">Documents, music, and other files shared in this chat will appear here.</string>
+    <string name="tab_docs_empty_hint">Documents and other files shared in this chat will appear here.</string>
```
rosano, line 442: If the tab already says document, I would shorten to:

```suggestion
    <string name="tab_docs_empty_hint">Files shared in this chat will appear here.</string>
```

rosano (approved):

res/values/strings.xml
r10s: >If the tab already says 'documents' [...]

the tab will say "Files" :) - see https://github.com/deltachat/deltachat-desktop/issues/2829 , https://github.com/deltachat/deltachat-android/pull/2341 , https://support.delta.chat/t/tabs-for-links-and-files/2142 

so, mentioning some types of files that are not covered by other tabs on any system probably makes some sense.

Issue #2350, opened 2022-07-22
add webxdc apps to home screen


r10s:
for much better integration, there should be an option that allows adding a Webxdc app to the home screen:

- [x] add "Add to Home Screen" to the Webxdc's menu in the upper right corner as well as to the context-overflow-menu when long-tapping a webxdc in the chat
- [x] once selected, an icon is added to the home screen (using Webxdc document name, app name and icon) and a toast as "Done." or so is shown. no need to show a confirmation dialog as an icon can be deleted easily. EDIT: some show a confirmation dialog on their own,
- [x] tapping the icon from the home screen opens the app directly 
- [x] make sure, this does not crash with multi-account, a first attempt may be to start only webxdc from the same account and show a warning otherwise
- [x] back button should go back one level up inside delta, this is the intuitive thing when talking to many ppl and also what other messenger are doing in similar cases [^1]
- [x] deal with deleted webxdc, print a warning (icon removal seems not to be possible on android)

[^1]: also, that seems to be recommended by android ux guides as well, see https://developer.android.com/training/notify-user/navigation for notifications that deep link in a similar way.  
if we would go back to the homescreen on "back button pressed", the question will arise how to stay inside delta, we would need an extra "Show in Chat" item in this case - and the user has to find and to understand it. compared to that, just going back one level as usual seems more intuitive - and if one really wants to go back to the homescreen, one can press the back button several times.

PR #2353, opened 2022-07-23
option to add webxdc apps to home screen

r10s:
this pr adds an option to add webxdc app icons to the launcher ...

... and start webxdc apps from there directly ...

see commit "[recreate back stack, if possible](https://github.com/deltachat/deltachat-android/pull/2353/commits/0d63f3505cc4eb43a3f08339614d7f81cc7e5163)" for a discussion about about back stack.

closes https://github.com/deltachat/deltachat-android/issues/2350

### DIFF
--- a/src/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -259,7 +259,7 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
         .build();
 
       if (ShortcutManagerCompat.requestPinShortcut(context, shortcutInfoCompat, null)) {
-        Toast.makeText(context, R.string.done, Toast.LENGTH_LONG).show();
+        Toast.makeText(context, R.string.done, Toast.LENGTH_SHORT).show();
       } else {
         Toast.makeText(context, "ErrAddToHomescreen: requestPinShortcut() failed", Toast.LENGTH_LONG).show();
       }
--- a/src/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -96,18 +96,24 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
     int appMessageId = b.getInt("appMessageId");
 
     this.dcContext = DcHelper.getContext(getApplicationContext());
-    this.dcAppMsg = this.dcContext.getMsg(appMessageId);
-    // `msg_id` in the subdomain makes sure, different apps using same files do not share the same cache entry
-    // (WebView may use a global cache shared across objects).
-    // (a random-id would also work, but would need maintenance and does not add benefits as we regard the file-part interceptRequest() only,
-    // also a random-id is not that useful for debugging)
-    this.baseURL = "https://acc" + dcContext.getAccountId() + "-msg" + appMessageId + ".localhost";
-
     if (dcContext.getAccountId() != b.getInt("accountId")) {
       Toast.makeText(this, "Switch to belonging account first.", Toast.LENGTH_LONG).show();
       finish();
       return;
     }
+
+    this.dcAppMsg = this.dcContext.getMsg(appMessageId);
+    if (!this.dcAppMsg.isOk()) {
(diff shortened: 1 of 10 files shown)


src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -192,6 +238,36 @@ private void updateTitleAndMenu() {
...
+      Bitmap bitmap = drawable.getBitmap();
+
+      ShortcutInfoCompat shortcutInfoCompat = new ShortcutInfoCompat.Builder(context, "xdc-" + dcContext.getAccountId() + "-" + msgId)
+        .setShortLabel(docName.isEmpty() ? xdcName : docName)
+        .setIcon(IconCompat.createWithBitmap(bitmap))
+        .setIntents(getWebxdcIntentWithParentStack(context, msgId))
+        .build();
+
+      if (ShortcutManagerCompat.requestPinShortcut(context, shortcutInfoCompat, null)) {
+        Toast.makeText(context, R.string.done, Toast.LENGTH_LONG).show();
```
Hocuri, line 262: Since we are just showing "Done", maybe use LENGTH_SHORT.

BTW, some apps go to the home screen after adding a shortcut so that the user can directly see it, but we don't have to do this.

src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -65,7 +96,19 @@ protected void onCreate(Bundle state, boolean ready) {
     int appMessageId = b.getInt("appMessageId");
 
     this.dcContext = DcHelper.getContext(getApplicationContext());
+    if (dcContext.getAccountId() != b.getInt("accountId")) {
+      Toast.makeText(this, "Switch to belonging account first.", Toast.LENGTH_LONG).show();
```
Hocuri, line 100: - s/belonging/associated/
- I guess it's not easily possible to make it automatically switch to the correct account?

Hocuri (approved):

src/org/thoughtcrime/securesms/WebxdcActivity.java
adbenitez: automatically switching to the game's account would be cool

adbenitez: it displays a weird dummy chat, that is a bit ugly

adbenitez (approved):
all in all this is great!!! so cool!!

src/org/thoughtcrime/securesms/WebxdcActivity.java
r10s: LENGTH_SHORT  makes sense, this is also what we are doing at the other places we feedback with a 'Done' toast. i added a commit.

> BTW, some apps go to the home screen after adding a shortcut so that the user can directly see it, but we don't have to do this.

which apps? going back to the homescreen is an idea worth trying out, not totally sure, however. but i would experiment with that maybe in another pr.

src/org/thoughtcrime/securesms/WebxdcActivity.java
r10s: i think, both of "belonging" and "associated" are fine.

however, switching to the account is the nice way to go - with the extra advantage that we do not need the string then :)

but that is sth. for another pr.

src/org/thoughtcrime/securesms/WebxdcActivity.java
r10s: thinking it over, i would not try to go the the homescreen, newer androids offer that option on their own:

Screenshot 2022-08-15 at 11 25 50

here, the user can just drag the icon from the box and ends up in the homescreen, pretty nice ux (stock android 9).

if we mess around with finishing activities etc. we may easily introduce bugs and end up in handling corner cases and checking android versions ...

Hocuri: > which apps? going back to the homescreen is an idea worth trying out, not totally sure, however.

Firefox, when you click "Add to Home Screen" in the three vertical dots menu.

In earlier times, the Play Store also did after installing an app, but it doesn't anymore. All in all, I think that I agree that we shouldn't do this, also because the Play Store apparently decided that they don' think it's good UX.

PR #2354, opened 2022-07-26
show mailinglist post address in profile

r10s:
this pr adds the mailinglist-post-address to the profile's subtitle:

counterpart of https://github.com/deltachat/deltachat-ios/pull/1654

moreover, this pr fixes some "bugs":
- show ephemeral icon in profile for groups (before, it was shown there only for contacts)
- show subtitles for saved-messages, device-chat (i think it was hidden as just most subtitles were hidden at some point in profile)

all in all, the code is a _little_ more straight forward now.


adbenitez (approved):
nice!

Issue #2365, opened 2022-08-19
does not register for `DCACCOUNT:` uri scheme in system


Simon-Laux:
this should work to make the qr code on webxdc.org clickable

also see:
https://github.com/deltachat/deltachat-ios/pull/1640
https://github.com/deltachat/deltachat-desktop/pull/2823

r10s:
closed by https://github.com/deltachat/deltachat-android/pull/2367

PR #2366, opened 2022-08-19
add dclogin scheme to qr code scan

Simon-Laux:
see https://github.com/deltachat/interface/pull/48 for spec.

needs a core with https://github.com/deltachat/deltachat-core-rust/pull/3541 merged.

I started with android implementation mainly to add the strings that they are swiftly discussed and then translated.

### Tasks
- [X] scan as first account
- [X] scan as second account
- [X] strings
- [x]  register DC as `dclogin:` scheme handler systemwide

those intents are beyond me, I can't figure out how to register DC as `dclogin:` scheme handler systemwide. sure it needs to go into `AndroidManifest.xml` but then the java code besides that is to hard for me. (I strongly dislike coding java) so it's better done by someone with real android experience, maybe together with #2365, as those two tasks are very similar.

so feel free to just push java stuff to this branch, I'm tired and annoyed of writing java.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -879,6 +879,8 @@
     <string name="qraccount_success_enter_name">Login successful—your e-mail address is %1$s\n\nIf you like, you can now enter a name and an profile image that will be displayed to people you write to.</string>
     <string name="qraccount_qr_code_cannot_be_used">The scanned QR code cannot be used to set up a new account.</string>
     <string name="qraccount_use_on_new_install">The scanned QR code is for setting up a new account. You can scan the QR code when setting up a new Delta Chat installation.</string>
+    <string name="qrlogin_ask_login">Login to \"%1$s\"?</string>
+    <string name="qrlogin_ask_login_another">Login to \"%1$s\"?\n\nYour existing account will not be deleted. Use the \"Switch Account\" item to switch between your accounts.</string>
     <!-- first placeholder will be replaced by name and address of the inviter, second placeholder will be replaced by the name of the inviter. -->
     <string name="secure_join_started">%1$s invited you to join this group.\n\nWaiting for the device of %2$s to reply…</string>
     <!-- placeholder will be replaced by the name of the inviter. -->
--- a/src/com/b44t/messenger/DcContext.java
+++ b/src/com/b44t/messenger/DcContext.java
@@ -58,6 +58,7 @@ public class DcContext {
     public final static int DC_QR_WITHDRAW_VERIFYGROUP   = 502;
     public final static int DC_QR_REVIVE_VERIFYCONTACT   = 510;
     public final static int DC_QR_REVIVE_VERIFYGROUP     = 512;
+    public final static int DC_QR_LOGIN             = 520;
 
     public final static int DC_LP_AUTH_OAUTH2          =     0x2;
     public final static int DC_LP_AUTH_NORMAL          =     0x4;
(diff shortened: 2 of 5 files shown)


src/com/b44t/messenger/DcContext.java
```diff
@@ -58,6 +58,7 @@ public class DcContext {
     public final static int DC_QR_WITHDRAW_VERIFYGROUP   = 502;
     public final static int DC_QR_REVIVE_VERIFYCONTACT   = 510;
     public final static int DC_QR_REVIVE_VERIFYGROUP     = 512;
+    public final static int DC_QR_LOGIN             = 520;
```
adbenitez, line 61: maybe make the `=` go below the previous one to make it look "prettier"

src/org/thoughtcrime/securesms/qr/QrCodeHandler.java
```diff
@@ -75,6 +75,16 @@ public void handleQrData(String rawString) {
                 builder.setNegativeButton(R.string.cancel, null);
                 builder.setCancelable(false);
                 break;
+
```
adbenitez, line 78:
```suggestion

```

res/values/strings.xml
```diff
@@ -791,6 +791,8 @@
     <string name="qraccount_success_enter_name">Login successful—your e-mail address is %1$s\n\nIf you like, you can now enter a name and an profile image that will be displayed to people you write to.</string>
     <string name="qraccount_qr_code_cannot_be_used">The scanned QR code cannot be used to set up a new account.</string>
     <string name="qraccount_use_on_new_install">The scanned QR code is for setting up a new account. You can scan the QR code when setting up a new Delta Chat installation.</string>
+    <string name="qrlogin_ask_login">Login to \"%1$s\"?</string>
```
r10s, line 794: "Login" is the noun/adjective and "Log in" the verb. this is also how we use that elsewhere.

src/com/b44t/messenger/DcContext.java
r10s: it is aligned with the "majority" of the others, so that seems fine already.

src/org/thoughtcrime/securesms/WelcomeActivity.java
```diff
@@ -256,7 +256,7 @@ private File copyToCacheDir(Uri uri) throws IOException {
         }
     }
 
-    private void startQrAccountCreation(String qrCode)
+    private void startQrAccountOrQrLoginCreation(String qrCode)
```
r10s, line 259: "login creation" is a bit weird. you meant `startQrAccountCreationOrLogin`? 

an alternative may be to  not alter the function name at all, "account creation" is still good enough: it creates and adds a `DcAccounts` object - either by creating a new email account or by logging in to an existent. from the view of the ui, this is a minor detail, the "big" thing is that a new `DcAccount` object is created.

src/org/thoughtcrime/securesms/qr/QrCodeHandler.java
```diff
@@ -75,6 +75,15 @@ public void handleQrData(String rawString) {
                 builder.setNegativeButton(R.string.cancel, null);
                 builder.setCancelable(false);
                 break;
+            case DcContext.DC_QR_LOGIN:
```
r10s, line 78:
```suggestion
                break;

            case DcContext.DC_QR_LOGIN:
```

r10s (approved):
in general, this looks good, some minors.

let's wait for core :)

src/org/thoughtcrime/securesms/qr/QrCodeHandler.java
adbenitez: mmmh it seems my recommendation to remove the empty spaces removed the whole blank line

adbenitez:
> let's wait for core :)

also first  #2367 should be merged into this branch

r10s:
> also first https://github.com/deltachat/deltachat-android/pull/2367 should be merged into this branch

i just rebased.

res/values/strings.xml
r10s: ah, that bit os not fixed, will push that directly on master before uploading the strings to transifex.

PR #2368, opened 2022-08-20
create adaptive shortcuts

adbenitez:


Hocuri (approved):
What does this improve?

r10s:
> What does this improve?

see recent screenshots in the dev group, it makes the webxdc icon larger, but cropped. a matter of taste, somehow :) but i'd also prefer the new layout.

r10s (approved):

Simon-Laux:
Someone should mention this in the webxdc docs, maybe with a small illustration showing a safe area where they can put text or sth. like that.

r10s:
worse are icons that have some borders our shapes on their own, eg. the old calendar icon. here, the doc is already pretty clear that this is discouraged,  "round corners etc. will be added by the implementations as needed". maybe we could be more concrete, maybe say "round corners, circle cut out etc. will be added by the implementations as needed".

PR #2372, opened 2022-08-25
make Recipient.getDcContact() work for one-to-one chats

r10s:
this should either go to master or may become part of https://github.com/deltachat/deltachat-android/pull/2371

### DIFF
--- a/src/org/thoughtcrime/securesms/recipients/Recipient.java
+++ b/src/org/thoughtcrime/securesms/recipients/Recipient.java
@@ -63,7 +63,6 @@ public class Recipient {
   private @Nullable String         profileName;
   private @Nullable String         profileAvatar;
 
-  // either dcChat or dcContact are set
   private @Nullable DcChat dcChat;
   private @Nullable DcContact dcContact;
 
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/recipients/Recipient.java
```diff
@@ -194,11 +194,10 @@ public synchronized String toShortString() {
 
   public int getFallbackAvatarColor() {
     int rgb = 0x00808080;
-    if(dcContact!=null) {
-      rgb = dcContact.getColor();
-    }
-    else if(dcChat!=null){
+    if(dcChat != null) {
```
adbenitez, line 197: isn't contact color equals the chat color of a 1:1 chat???

adbenitez (approved):

adbenitez:
I added the changes to #2371

src/org/thoughtcrime/securesms/recipients/Recipient.java
r10s: _currently_, probably, yes. but it is better to request the semantically correct color from the core.

PR #2373, opened 2022-08-26
use dc_contact_was_seen_recently()

r10s:
use `dc_contact_was_seen_recently()` introduced by https://github.com/deltachat/deltachat-core-rust/pull/3560 (already on android master) instead of ui-implementation.

successor of https://github.com/deltachat/deltachat-android/pull/2371

### DIFF
--- a/src/com/b44t/messenger/DcContact.java
+++ b/src/com/b44t/messenger/DcContact.java
@@ -63,6 +63,7 @@ public class DcContact {
     public native int     getColor       ();
     public native String  getStatus      ();
     public native long    getLastSeen    ();
+    public native boolean wasSeenRecently();
     public native boolean isBlocked      ();
     public native boolean isVerified     ();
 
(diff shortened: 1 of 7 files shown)


src/com/b44t/messenger/DcContact.java
```diff
@@ -50,10 +50,6 @@ public String toString() {
         return getAddr();
     }
 
-    public boolean isSeenRecently() {
-        return DateUtils.isWithin(getLastSeen(), 10, TimeUnit.MINUTES);
```
adbenitez, line 54: since this was removed, now the imports introduced in:
https://github.com/deltachat/deltachat-android/pull/2371/files#diff-9ad8aa37308d8cdfde999a39743f5a70f62cc1488955eba5c677f7a299876f03R3-R6
can be removed also maybe make `DateUtils.isWithin()` private again

adbenitez (approved):

src/com/b44t/messenger/DcContact.java
r10s: great! this also makes the wrapper standalone again (so that it can be used eg. for java-bots or -extensions, iirc @angelo-fuchs was interested in that from time to time :)

PR #2374, opened 2022-08-28
simplify contact list

r10s:
this is the ui part of https://github.com/deltachat/deltachat-core-rust/pull/3562, removing the A-Z decoration, see commit message for further details.

this pr should be merged after we have https://github.com/deltachat/deltachat-core-rust/pull/3562 on android master, however, can already be tested without.

nb: ios does not need any adaptions as the list was always simpler there; probably same for desktop.


adbenitez (approved):
great! already using it!

PR #2377, opened 2022-09-02
reword "Webxdc content"

r10s:
adding "App" to the term makes things much clearer to the user,
was "Webxdc" is about [^1]

eg. on iOS, we have an app selector meanwhile,
calling that _just_ "Webxdc" in the menu or in the title makes it pretty
unclear to the user what this is about.
same for the tab "Webxdc" on desktop or the default description in the summary.
otoh, calling it _only_ "Apps" (as now), may also raise false expectations
and may be mixed with "normal" apps.

so, "Webxdc App" seems to be a good approach.

there is also the term "Mini App" [^2], we could also go for that,
but this is a broader discussion as this would give up the term "Webxdc" partly.
but i would not be against "Mini App" :)

in any case, if there is few space in the UI,
it seems to be okay to abbreviate with just "App" or "Apps" as needed,
if it is otherwise clear that the thing is a "Webxdc App" or "Mini App".

[^1]: we decided meanwhile that using the term "App" should not be a no-go :)
[^2]: the term "Mini App" is aleady known for similar types of apps
(some "small app" running in a "super app")
cmp. https://www.w3.org/TR/mini-app-white-paper/#what-is-miniapp .  
of course, our "Webxdc Apps" are not compatible with other "Mini Apps"
however, this is also not the case between "Mini Apps" of other vendors
which also have different focuses.  
so "Mini App" seems to be a broader term and does not imply compatibiliy.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -155,7 +155,9 @@
     <string name="images_and_videos">Images and Videos</string>
     <string name="file">File</string>
     <string name="files">Files</string>
+    <!-- "Private App" is used to present "Webxdc App" (https://webxdc.org) in a user friendly way. Please stay close to the original term and keep it short (it is used in menus with few screen space). -->
     <string name="webxdc_app">Private App</string>
+    <!-- plural of "Private App"; used to present "Webxdc App" (https://webxdc.org) in a user friendly way. Please stay close to the original term and keep it short (it is used in menus with few screen space). -->
     <string name="webxdc_apps">Private Apps</string>
     <string name="files_and_webxdx_apps">Files and Private Apps</string>
     <string name="unknown">Unknown</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -155,10 +155,9 @@
     <string name="images_and_videos">Images and Videos</string>
     <string name="file">File</string>
     <string name="files">Files</string>
-    <string name="webxdc_app">Webxdc App</string>
-    <!-- Plural of "Webxdc App", used eg. as menu entry or title for an app selector. -->
-    <string name="webxdc_apps">Webxdc Apps</string>
-    <string name="files_and_webxdx_apps">Files and Webxdc Apps</string>
+    <string name="webxdc_app">Private App</string>
+    <string name="webxdc_apps">Private Apps</string>
+    <string name="files_and_webxdx_apps">Files and Private Apps</string>
     <string name="unknown">Unknown</string>
 
     <string name="green">Green</string>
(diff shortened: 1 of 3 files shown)


r10s:
summing up recent chats:

@adbenitez: _"btw the term "Mini Apps" sounds more user-friendly and comes to my mind that small games "Mini Clips" that were popular some years ago, or Facebook mini-games"_

@r10s: _"yip, i'd also like "Mini Apps" - main disadvantage would be that we weaken the term "Webxdc" by that - if we say that this is not an issue, i would prefer "Mini Apps"._  
_("Mini Apps" is more user-friendly, that's true)"_

@adbenitez: _"I would see Webxdc as Flash, you don't call Flash games "Flash", just games to the user is point of view, "Webxdc" would be more the format"_

@r10s: _"+1, this is also how "Mini Apps" are discussed at https://www.w3.org/TR/mini-app-white-paper/#what-is-miniapp imu - they are not necessarily compatible between the different "Super Apps" "_

Simon-Laux:
if we want to do marketing we can't use an already loaded term.
like we are super privacy preserving but the average wechat app is the opposite of that.

adbenitez (approved):
I like more the shorter and easier term "mini app" but this is already an improvement

rosano:
Adding 'apps' definitely makes it more informative than only 'webxdc' – acronyms are not self-explanatory. I like 'mini apps' too, helps set expectations in the right place.

r10s:
ftr: in chat, there is some consent about:

- use "Mini Apps" inside Delta Chat in user-facing areas (where webxdc is more confusing than helping; "Mini Apps" is regarded as more user-friendly in general).

- use "Webxdc Apps" or "Webxdc Mini Apps" outside for marketing purposes

so , "Mini App" is kind of the genre and "Webxdc" is the implementation.

res/values/strings.xml
```diff
@@ -155,6 +155,10 @@
     <string name="images_and_videos">Images and Videos</string>
     <string name="file">File</string>
     <string name="files">Files</string>
+    <string name="webxdc_app">Webxdc App</string>
```
cyBerta, line 158:
```suggestion
    <string name="webxdc_app">Mini App</string>
```

res/values/strings.xml
```diff
@@ -155,6 +155,10 @@
...
+    <!-- Plural of "Webxdc App", used eg. as menu entry or title for an app selector. -->
+    <string name="webxdc_apps">Webxdc Apps</string>
+    <string name="files_and_webxdx_apps">Files and Webxdc Apps</string>
```
cyBerta, line 161:
```suggestion
    <string name="files_and_webxdx_apps">Files and Mini Apps</string>
```

r10s:
not sure if that pop up before, but "Private Apps" may also work for a more user-friendly in-app term of "Webxdc" ... i am not sure if i find that better, just throwing that in :)

adbenitez (approved):
"Private apps" is still better than "webxdc" so +1

PR #2378, opened 2022-09-04
restricted webxdc internet access

r10s:
~~this requires https://github.com/deltachat/deltachat-core-rust/pull/3516 to be on master first.~~ EDIT: done

we we need to know the `internet_access` flag before creation of the html view, this pr moves `getWebxdcInfo()` to the main thread on webxdc creation. doing things in background would be more complicated and error prone (i can easily imagine races if we move html-view-creation to a background thread or change the flag after creation, that would need additional code to be targeted etc.). however, `getWebxdcInfo()` is also called at other places (message list!) from the main thread, so that is probably good enough.

all other things are left in background threads as before.

see https://github.com/deltachat/deltachat-core-rust/pull/3516 for further information.

here is a little demo:

**Download .xdc:** https://github.com/r10s/maps/releases/download/0.0.1/maps.xdc ([source](https://github.com/r10s/maps)) - you need to copy this to the "Saved Messages" chat of a build using this pr - otherwise, you will not see the map.

### DIFF
--- a/src/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -48,6 +48,7 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
   private DcMsg dcAppMsg;
   private String baseURL;
   private String sourceCodeUrl = "";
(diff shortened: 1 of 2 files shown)


adbenitez (approved):
:rocket:

PR #2379, opened 2022-09-07
Revert "create adaptive shortcuts"

r10s:
This reverts commit 35dd8c905b7ff182c92a1bcfcfe03a97edecbb00.

The cut out is not the largets possible circle inside the square icon (as for group images or avatars, diameter==height) but a much smaller circle (diameter==~0.6*height), resulting in too many information being left out.

Therefore, it is better to leave icon layout to the OS.

to get an idea about the amount of cut out, see a circular logo, eg. the hextris one:


Simon-Laux:
I'm for a manifest property or an extra "adaptive" version of the icon in the zip.

Simon-Laux (approved):

PR #2380, opened 2022-09-11
rework system messages

### LINKED ISSUE #2335: Context-free string
@link2xt [explained to me](https://github.com/deltachat/deltachat-android/pull/2331#issuecomment-1176487680) that the following localized string is designed to be replaced in another context:

https://github.com/deltachat/deltachat-android/blob/e14dbd5c3be67db0e29a5a7c2d09e1ecd82f1875/res/values/strings.xml#L714

> > When does this appear? It reads to me as 'The group left' but maybe it means 'You left the group'?
> 
> It's one of those weird strings mentioned in [deltachat/deltachat-core-rust#3429](https://github.com/deltachat/deltachat-core-rust/issues/3429)
> 
> Full sentence is formed by removing the period and attaching the translation of "by %s", e.g. "Group left by you".
> 
> Proper fix is to remove this string and translate the sentence "You left the group" as a whole.

I would recommend to follow his fix by using an entire sentence as a localized string. This makes translation less opaque (people don't need to figure out how the string will be used because it contains all the context), and also avoids encoding English specific grammar into the programming (removing punctuation and replacing placeholders is an example of this).

r10s:
rework system messages and make them easier understandable and easier translatable.

counterpart of https://github.com/deltachat/deltachat-core-rust/pull/3518, which needs to be merged first. also see there for some more details and reasoning.

Closes #2335


r10s:
thanks, @link2xt - for review and for pushing this forward in general. 

i uploaded the strings to transifex, things are waiting for translation there :)

PR #2381, opened 2022-09-13
add missing strings for webxdc selector and webxdc shortcuts on iOS

cyBerta:
Depends partly on #2377 being settled. The webxdc selector strings will probably change after the merge of #2377. 
This PR is mainly to not forget about these untranslated strings currently used on iOS.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -1022,6 +1022,12 @@
     <string name="notifications_avg_minutes">On average every %1$d minutes</string>
     <string name="notifications_avg_hours">On average every %1$d hours</string>
     <string name="last_check_at">Checked at %1$s</string>
+    <!-- iOS webxdc selector -->
+    <string name ="webxdcs">Apps</string>
+    <string name ="webxdc_empty_hint">Received or sent apps will appear here. Tap \"Files\" to select downloaded apps.</string>
+    <!-- iOS webxdc shortcut page -->
+    <string name ="shortcut_share_btn">Click the share button</string>
+    <string name ="shortcut_add_to_home_description">Select - Add to Home Screen - to add the app to your home screen.</string>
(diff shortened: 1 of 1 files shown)


res/values/strings.xml
```diff
@@ -1022,6 +1022,12 @@
     <string name="notifications_avg_minutes">On average every %1$d minutes</string>
     <string name="notifications_avg_hours">On average every %1$d hours</string>
     <string name="last_check_at">Checked at %1$s</string>
+    <!-- iOS webxdc selector -->
+    <string name ="webxdcs">Apps</string>
+    <string name ="webxdc_empty_hint">Received or sent apps will appear here. Tap \"Files\" to select downloaded apps.</string>
```
r10s, line 1027:
```suggestion
    <string name ="webxdc_selector_empty_hint">Private apps received or sent in any chat will appear here. \"Files\" shows private apps just downloaded.</string>
```

cc @rosano

res/values/strings.xml
```diff
@@ -1022,6 +1022,12 @@
...
+    <!-- iOS webxdc shortcut page -->
+    <string name ="shortcut_share_btn">Click the share button</string>
+    <string name ="shortcut_add_to_home_description">Select - Add to Home Screen - to add the app to your home screen.</string>
```
r10s, line 1030:
```suggestion
    <string name ="shortcut_step1_tap_share_btn">Click the share button</string>
    <string name ="shortcut_step2_tap_add_to_home_screeb">Select \"Add to Home Screen\" to add the app to your home screen.</string>
```

having the "step" in the key makes it easier for translators as the strings will appear together in correct order. it is also clearer for us :)

for the quotes around "Add To Home Screen": i know, there are some isszes with them, however, if it is not possible to render this character, we should change that programmatically eg, by `replace('"', ' - ')` - i fear otherwise, some translator will use them anyway.

cyBerta: the key `shortcut_step2_tap_add_to_home_screeb` has an typo.

r10s (approved):

PR #2382, opened 2022-09-15
load contact from database before calling wasSeenRecently()

adbenitez:
this should solve the issue reported by @gerryfrancis in the testing group

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -1580,7 +1580,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
                 || eventId == DcContext.DC_EVENT_MSG_DELIVERED)
                && event.getData1Int() == chatId) {
         DcContact contact = recipient.getDcContact();
-        titleView.setSeenRecently(contact!=null? contact.wasSeenRecently() : false);
+        titleView.setSeenRecently(contact!=null? dcContext.getContact(contact.getId()).wasSeenRecently() : false);
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1580,7 +1580,7 @@ public void handleEvent(@NonNull DcEvent event) {
                 || eventId == DcContext.DC_EVENT_MSG_DELIVERED)
```
r10s, line 1580: why update the state on `DC_EVENT_MSG_DELIVERED`? 

as this is not message coming from the recipient, the last-seen state should not been changed by the fact a message is sent out.

r10s (approved):

src/org/thoughtcrime/securesms/ConversationActivity.java
adbenitez: it is to update the state when you send a message to that contact and it no longer is "seen recently" (ex. to hide the dot after 10 min passed as opposed to the incoming messages that cause it to appear) it is not the ideal solution, but at least to reduce the chances to show the green dot forever (still happens if chat remains quiet without incoming/outgoing messages)

adbenitez: we have issues updating the last-seen indicator as well as message date label being "now" forever if there is not changes or incoming/outgoing messages and the same happens in the chat list, at some point would be nice to have some kind of timer to trigger a refresh if no refreshed after X time

adbenitez: perhaps there could be a core event when a contact's "seen recent" state changes :thinking:

r10s: i see, then, for now, i'd just add a little comment to DC_EVENT_MSG_DELIVERED in this case.

adbenitez: added a comment in the point the listener register for handling this event

r10s (approved):

PR #2386, opened 2022-09-21
separate tab for private apps

r10s:
this pr adds a separate tab for "Private Apps" to the profiles; similar as on desktop (ios has another layout, however, also directly mentioned "Private Apps" meanwhile)


adbenitez (approved):
nice, I think after this we could add also a "voice" tab, what do you think?

r10s:
> nice, I think after this we could add also a "voice" tab, what do you think?

yip, makes sense. i would go for an "Audio" tab that reuses `ProfileDocumentsFragment` with the types DC_MSG_VOICE and DC_MSG_AUDIO;`ProfileDocumentsFragment .showWebxdc` could be upgraded to an an enum with files/webxdc/audio then.

PR #2391, opened 2022-09-30
remove deprecated strings

r10s:
all systems use the new strings introduced by https://github.com/deltachat/deltachat-android/pull/2380 now; the old strings can be removed, esp. as they add quite some work and/or confusion to the translators.


r10s:
as for the other translation update pr, merging this to stay in sync with transifex.

PR #2393, opened 2022-10-03
use less muted color for "recently seen"

r10s:
with the "connection view",
we already have a green indicating sth. as "online", reuse that. (the green used before is a bit too much of a muted color, this does not reflect "activity" well)

i came over that when [playing around with the layout on ios](https://github.com/deltachat/deltachat-ios/pull/1707); ios uses the same color now.

also many other apps use a similar, less muted green for "activity" or "online".

the difference is not huge on android, on ios it was larger - so this pr is mainly to keep things in sync.

before / after:

 

corresponding ios pr using the same color: https://github.com/deltachat/deltachat-ios/pull/1707

once merged, we should also add the color to the interface-repo and change it on desktop (that appears already "more active" by the white border, however)

### DIFF
--- a/res/drawable/ic_circle_status.xml
+++ b/res/drawable/ic_circle_status.xml
@@ -3,6 +3,6 @@
     xmlns:android="http://schemas.android.com/apk/res/android"
     android:shape="oval">
 
-    <solid android:color="@color/green" />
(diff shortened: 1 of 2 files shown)


adbenitez (approved):
better indeed

PR #2394, opened 2022-10-04
hide ephemeral timer in mailing lists

adbenitez:


r10s (approved):

r10s:
ftr: see recent discussions in dev chat; this is also what we are doing on ios as well. for desktop i'll file an issue.

PR #2398, opened 2022-10-17
use Start/End instead of Left/Right in avatar layout

### LINKED ISSUE #2397: RTL: Green dot indicator is stretched
- Android version: 12
- Device:Samsung GS20
- Delta Chat version: 1.34.1
- Expected behavior: The green dot next to a thread would be a small circle as it is in a left-to-right language.
- Actual behavior: If the device is using a right-to-left language, such as persian, it is stretched. 
- Steps to reproduce the problem: Just put the device into persian or arabic or something, then look at the inbox. There needs to be a thread with a green dot next to it in order for you to see it.
- Screenshots:
Screenshot_۲۰۲۲۱۰۱۷-۱۰۵۱۵۵_DeltaLab
- Logs: none

adbenitez:
close #2397

### DIFF
--- a/res/layout/avatar_view.xml
+++ b/res/layout/avatar_view.xml
@@ -21,8 +21,8 @@
         android:layout_width="0dp"
         android:layout_height="0dp"
 
-        app:layout_constraintLeft_toLeftOf="@+id/left_guideline"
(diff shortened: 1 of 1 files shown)


r10s (approved):
great!

Issue #2400, opened 2022-09-29
Messages not coming in anymore (probably due to dead idle tasks caused by network change)


gerryfrancis:
- Operating System (Linux/Mac/Windows/iOS/Android):
Android 11.

- Delta Chat Version:
1.33.0-1.34.5.

- Expected behavior:
Messages are coming in when the app is not focused and WLAN is the active network gateway.

- Actual behavior:
Messages are not coming in when the app is not focused and WLAN is the active network gateway.

- Steps to reproduce the problem:
-- Wait for the issue to occur, but use some other apps in the meantime. (WLAN and mobile data must be enabled on the device.)

- Screenshots:
N/A.

- Logs:
[deltachat-log-20220921-131932_2.txt](https://github.com/deltachat/deltachat-core-rust/files/9675549/deltachat-log-20220921-131932_2.txt)

- Remarks:
I doubt that the change from async-std to tokio has something to do with it. I think the culprit lies in how Delta Chat core recognizes old and new network tasks after a different network gateway is activated by the OS.
![2022-09-26-19-51_2](https://user-images.githubusercontent.com/51095533/193065254-62b9c59d-2f1e-4983-888d-5b2c5f202fac.png)
This switch in Android developer options seems to drive two networks, e.g. LTE and WLAN, in parallel to improve delays for switching between the two networks. When enabled (default setting), the tasks bound to the LTE connection might remain in the back when WLAN is active. Delta Chat might try to hold on to the LTE connection tasks at some time, probably the app cannot distinguish between the "old" and "new" idle tasks properly, and various requests might run into a timeout. I cannot reproduce the issue after I disabled this developer function.
An [Android technote](https://developer.android.com/training/basics/network-ops/reading-network-state) says this:
`All Android apps have a default network. The system decides which should be the default network. The system typically prefers unmetered networks to metered ones and faster networks to slower ones. When an app issues a network request, such as with [HttpsURLConnection](https://developer.android.com/reference/javax/net/ssl/HttpsURLConnection), the system satisfies this request using the default network. Apps can send traffic on other networks, too.`
And further:
`When a new network becomes the default, any new connection the app opens uses this network. At some point later, all remaining connections on the previous default network are forcefully terminated.`
[...]

PR #2403, opened 2022-10-25
use more modern methods to track connectivity changes

r10s:
use [`registerDefaultNetworkCallback()`](https://developer.android.com/reference/android/net/ConnectivityManager#registerDefaultNetworkCallback(android.net.ConnectivityManager.NetworkCallback)) instead of [deprecated CONNECTIVITY_ACTION](https://developer.android.com/reference/android/net/ConnectivityManager#CONNECTIVITY_ACTION).

- there is also a [`registerNetworkCallback()`](https://developer.android.com/reference/android/net/ConnectivityManager#registerNetworkCallback(android.net.NetworkRequest,%20android.app.PendingIntent)) but according to [SO](https://stackoverflow.com/questions/53863034/difference-between-registerdefaultnetworkcallback-and-registernetworkcallback) , that is for less broader usage and won't catch "more".

- it is a bit unclear to me, if the other callbacks are needed, for now, we just log a line. just calling maybe_network everywhere would be too much, as eg. [`onCapabilitiesChanged()`](https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback#onCapabilitiesChanged(android.net.Network,%20android.net.NetworkCapabilities)) is called after [`onAvailable()`](https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback#onAvailable(android.net.Network)) ... but only since API 26.

there should be a review APK build by this pr - maybe this is sufficient for testing? cc @gerryfrancis @Gh0stz0x @link2xt - if not i can also build a "regular" APK tomorrow.

all in all, i am not sure if that fixes the issue. CONNECTIVITY_ACTION is deprecated - but it is still available. and it seems to be still in use by Signal as well. we'll see :)

targets #2400

### DIFF
--- a/src/org/thoughtcrime/securesms/ApplicationContext.java
+++ b/src/org/thoughtcrime/securesms/ApplicationContext.java
@@ -58,6 +58,11 @@ public class ApplicationContext extends MultiDexApplication {
   public NotificationCenter     notificationCenter;
   private JobManager            jobManager;
 
+  private int                   debugOnAvailableCount;
+  private int                   debugOnBlockedStatusChangedCount;
+  private int                   debugOnCapabilitiesChangedCount;
+  private int                   debugOnLinkPropertiesChangedCount;
+
   public static ApplicationContext getInstance(@NonNull Context context) {
     return (ApplicationContext)context.getApplicationContext();
   }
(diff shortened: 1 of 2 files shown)


gerryfrancis:
Awesome, guys!! :heart: I am going to test the APK today and let you know, many thanks to everyone involved! :+1:

gerryfrancis:
Here are two logs. In general, notifications were fired appx. 60-90 seconds after the message had been received:

This one happened after I went from the garage (no Internet connection) to my flat (WiFi), with LTE (mobile data) in between:
[deltachat-log-20221026-215821.txt](https://github.com/deltachat/deltachat-android/files/9873953/deltachat-log-20221026-215821.txt)

And this one ocurred after I relaunched the app with active LTE (mobile data) and WiFi enabled _after_ the launch:
[deltachat-log-20221026-230044.txt](https://github.com/deltachat/deltachat-android/files/9873955/deltachat-log-20221026-230044.txt)

Connections were retrieved every time, so that fix _definitely_ is an improvement! :heart:

r10s:
@link2xt @gerryfrancis thanks for testing! 

i added a commit as suggested by @link2xt that always installs the old hook and adds counters to each output. 

not sure how useful the counters actually are - at least in emulator, the old "Connected" is called twice - which, however, may also be an issue wrt speed if core reinitializes connection or so.

gerryfrancis:
And here are the results of the latest APK:

[deltachat-log-20221028-021316.txt](https://github.com/deltachat/deltachat-android/files/9883853/deltachat-log-20221028-021316.txt)

@r10s @link2xt Again, thank you so much for your hard work! :heart:

gerryfrancis:
@link2xt I agree, thank you! At least I cannot reproduce the issue with the new callback, so let's hope it will help the other users, too. :+1: 

> `NetworkCallback` is always called first, then `Connected` is called twice after some delay.

My configuration has sentbox_watch=1 , maybe that explains the two `Connected` lines.

gerryfrancis:
One more thing in regard to `onCapabilitiesChanged`:

https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback#onCapabilitiesChanged(android.net.Network,%20android.net.NetworkCapabilities)

There it says:

> Starting with [Build.VERSION_CODES.O](https://developer.android.com/reference/android/os/Build.VERSION_CODES#O) this method is guaranteed to be called immediately after [onAvailable(Network)](https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback#onAvailable(android.net.Network)).

No idea if relevant or not, but maybe we should wait for `onCapabilitiesChanged` after `onAvailable` has occurred, before we start any data transfer, it might be a safer approach for the most newer Android versions in the field.

r10s:
> No idea if relevant or not, but maybe we should wait for onCapabilitiesChanged after onAvailable has occurred, before we start any data transfer, it might be a safer approach for the most newer Android versions in the field.

i also thought about calling `maybe_network()` for android >= Build.VERSION_CODES.O in `onCapabilitiesChanged()` instead of `onAvailable()` (older androids seems to be unclear here, if even the doc says that there is no guarantee ;)

however, this would make only sense if changes of [`NetworkCapabilities properties`](https://developer.android.com/reference/android/net/NetworkCapabilities) justify a call to `maybe_network()` in the majority of cases.

also, might be that `onCapabilitiesChanged()` is fired in batches and we would need debouncing, that would increase complexity.

so, maybe merge this in as is and see what the log says. if we have connection problems and see in log that there is a `onCapabilitiesChanged()` we can improve.

does that make sense?

or are there any hints that would make this step useful already? 

anyway, this pr is ready for approval now :)

gerryfrancis:
@r10s I second that, let's merge and find out! :) I know a Delta Chat user who is experiencing this issue. She uses a Google Pixel 3a currently running stock Android 12, so I could install the nightly build that includes the fix on it and report back after some days. Thank you for your time! ❤️

PR #2404, opened 2022-10-26
Also show errors for incoming emails

Hocuri:
Needed for https://github.com/deltachat/deltachat-core-rust/pull/3583/. Before that PR, we never put errors on incoming messages, so this wasn't necessary.

### DIFF
--- a/src/com/b44t/messenger/DcMsg.java
+++ b/src/com/b44t/messenger/DcMsg.java
@@ -1,5 +1,7 @@
 package com.b44t.messenger;
 
+import android.text.TextUtils;
+
 import org.json.JSONObject;
 
 import java.io.File;
@@ -200,7 +202,7 @@ public class DcMsg {
     }
 
     public boolean isFailed() {
-        return getState() == DC_STATE_OUT_FAILED;
+        return (getState() == DC_STATE_OUT_FAILED) || (!TextUtils.isEmpty(getError()));
     }
     public boolean isPreparing() {
         return getState() == DC_STATE_OUT_PREPARING;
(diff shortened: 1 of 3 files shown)


r10s (approved):
lgtm, did not test that, however.

src/com/b44t/messenger/DcMsg.java
```diff
@@ -200,7 +202,7 @@ public long getDateReceived() {
     }
 
     public boolean isFailed() {
-        return getState() == DC_STATE_OUT_FAILED;
+        return (getState() == DC_STATE_OUT_FAILED) || (!TextUtils.isEmpty(getError()));
```
adbenitez, line 205: what does it means when it is `DC_STATE_OUT_FAILED` but `getError()` is empty?
the message will be considered not failed in that case and that doesn't look right :thinking:

adbenitez (approved):

src/com/b44t/messenger/DcMsg.java
r10s: imu, the new condition is about incoming messages that fail. in the past, there was no situation where this could happen, but now DKIM can fail and an incoming message should get an error symbol. as the error can be in any combination with the  [DC_STATE_IN states](https://c.delta.chat/classdc__msg__t.html#a83fbf6e74d09a0b598ccefe9b48bd68c), an DC_STATE_IN_FAILED would be tricky (one could also argue that an explicit DC_STATE_OUT_FAILED is superfluous and UI should just always check the error state, but this is another issue :)

Hocuri: Exactly :)

Hocuri: > (one could also argue that an explicit DC_STATE_OUT_FAILED is superfluous and UI should just always check the error state, but this is another issue :)

Fun fact, this also was exactly my chain of thought while implementing it.

PR #2406, opened 2022-10-29
prepare 1.34.2

r10s:
a bugfix release.

this should be rebased+merged **after** https://github.com/deltachat/deltachat-android/pull/2407

PR #2407, opened 2022-10-29
bump targetSdkVersion to 32

r10s:
using `targetSdkVersion  32` is required for gplay **from Nov 2022** and requires some adaptions, see https://developer.android.com/google/play/requirements/target-sdk#pre12 ; some notes on the steps mentioned there:

- [x] bluetooth: we are not using bluetooth permissions at all
- [x] request the ACCESS_COARSE_LOCATION permission any time you request ACCESS_FINE_LOCATION: we are already doing this this way
- [x] set `android:exported` explicitly for indent filters in the manifest: done by a commit in this pr
- [x] hibernation (what happens if the app is not used for months): should be fine at a first glance
- [x] pending intent mutability: we set the flag with a helper function, see corresponding commit message for details

moreover, this PR updates gradle (i got weird errors with the old version) and androidx, at least the old sharetarget does not compile for API 32 (see https://developer.android.com/jetpack/androidx/releases/appcompat for androidx changelogs (select package on the left))

the updated androidx compat dependencies may also help on https://github.com/deltachat/deltachat-android/issues/2400

once this pr is merged, we can rebase and merge #2406 and do a release.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -63,7 +63,8 @@
               android:launchMode="singleTask"
               android:taskAffinity=""
               android:windowSoftInputMode="stateHidden"
-              android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize">
+              android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize"
+              android:exported="false">
 
         <intent-filter>
             <action android:name="android.intent.action.SEND" />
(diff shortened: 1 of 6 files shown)


adbenitez (approved):

r10s:
thanks for approving. thinking it over, might be that dependencies do not set FLAG_MUTABLE; i'll additionally update the missing dependencies for that reason.

we need some days of testing afterwards.

PR #2410, opened 2022-10-30
fix message bubble corner for RTL languages

adbenitez:


r10s (approved):
thanks a lot for taking care of details, lgtm, let's get that into the 1.34.2 bugfix release ...

PR #2411, opened 2022-10-30
fix crash in share-to-delta and similar issues

r10s:
this issue was reported by the testing group, thanks a lot for all the testers and their great job they're doing 🙏

the [crash](https://gist.github.com) was introduced by 1.34.2 that was released for testing only. 

reason is a wrong `exported` flag - it makes sense to have `exported="true"` here.

the flag was added in https://github.com/deltachat/deltachat-android/pull/2407 as needed for android12 - and i was assuming that the old default was `false` and set that value to all missing ones.  
that assumption seems to be wrong, so we should double-check all flags changed by https://github.com/deltachat/deltachat-android/pull/2407

EDIT: according to https://stackoverflow.com/questions/9274742/is-androidexported-true-really-necessary-for-an-authentication-service , it defaulted to `false` only if there were no intent-filters, otherwise it defaulted to `true`. so we should adapt the other flags as well.

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -177,7 +177,7 @@
               android:theme="@style/TextSecure.LightNoActionBar"
               android:windowSoftInputMode="stateHidden"
               android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize"
-              android:exported="false">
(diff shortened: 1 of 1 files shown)


Hocuri:
:+1:

PR #2416, opened 2022-11-07
clarify meaning of 'left'

r10s:
this came out of transifex discussions, depending on the language there may be quite different words that need this clarification.


r10s:
was approved via message on transifex ;)

Issue #2426, opened 2022-11-15
F-Droid build failed


linsui:
https://monitor.f-droid.org/builds/log/com.b44t.messenger/641#site-footer

```
+ ./ndk-make.sh
starting time: Mon Nov 14 23:18:58 UTC 2022
Setting CARGO_TARGET environment variables.
Full build
-- cross compiling to armv7-linux-androideabi (arm) --
    Updating git repository `https://github.com/async-email/async-imap`
    Updating crates.io index
    Updating git repository `https://github.com/dignifiedquire/rs-async-zip`
    Updating git repository `https://github.com/deltachat/rust-email`
    Updating git repository `https://github.com/async-email/encoded-words`
    Updating git repository `https://github.com/deltachat/lettre`
    Updating git repository `https://github.com/Majored/rs-async-io-utilities`
error: failed to get `async_io_utilities` as a dependency of package `async_zip v0.0.8 (https://github.com/dignifiedquire/rs-async-zip?branch=main#5556c5ac)`
    ... which satisfies git dependency `async_zip` (locked to 0.0.8) of package `deltachat v1.98.0 (/home/vagrant/build/com.b44t.messenger/jni/deltachat-core-rust)`
    ... which satisfies path dependency `deltachat` (locked to 1.98.0) of package `deltachat_ffi v1.98.0 (/home/vagrant/build/com.b44t.messenger/jni/deltachat-core-rust/deltachat-ffi)`

Caused by:
  failed to load source for dependency `async_io_utilities`

Caused by:
  Unable to update https://github.com/Majored/rs-async-io-utilities#6661a0fb

Caused by:
  failed to clone into: /home/vagrant/.cargo/git/db/rs-async-io-utilities-5b754873e3909201

Caused by:
  failed to authenticate when downloading repository

  * attempted to find username/password via git's `credential.helper` support, but failed

  if the git CLI succeeds then `net.git-fetch-with-cli` may help here
  https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli

Caused by:
  failed to acquire username/password from local configuration
```

Could you please take a look? Thanks!

r10s:
we probably need a new android version with a new core101 that includes this fix: https://github.com/deltachat/deltachat-core-rust/pull/3747 (some repo that we were using was just deleted)

linsui:
By the way, I'd like to split the abi in F-Droid buil to reduce the apk size. What version code scheme should we use?

[...]

PR #2427, opened 2022-11-15
add versionCode to debug info

r10s:
that may get handy esp. if the code is patched, cmp https://github.com/deltachat/deltachat-android/issues/2426#issuecomment-1315393104


adbenitez (approved):

r10s:
thanks for quick response <3

PR #2429, opened 2022-11-16
push translation strings from desktop

Simon-Laux:
related desktop pr: https://github.com/deltachat/deltachat-desktop/pull/3013

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -883,7 +883,6 @@
     <string name="notify_name_and_message">Name and message</string>
     <string name="notify_name_only">Name only</string>
     <string name="notify_no_name_or_message">No name or message</string>
-    <string name="notify_bundle_new_messages_in_one_chat">%1$d new messages</string>
     <string name="notify_bundle_new_messages">%1$d messages in %2$d chats</string>
 
 
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -403,7 +403,6 @@
     <!-- broadcast lists -->
     <string name="create_broadcast_list">Create Broadcast List</string>
     <string name="menu_edit_broadcast_list">Edit Broadcast List</string>
-    <string name="menu_broadcast_list_name">Broadcast List Name</string>
     <string name="broadcast_please_enter_broadcast_list_name">Please enter a name for the broadcast list.</string>
     <string name="broadcast_list_warning">Broadcast Lists are always unencrypted because they use blind copy (BCC)</string>
 
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -962,8 +962,6 @@
     <string name="no_spellcheck_suggestions_found">No spelling suggestions found.</string>
     <string name="show_window">Show Window</string>
     <string name="login_socks5_experimental_warning">SOCKS5 support is currently experimental. Please use at your own risk. If you type in an address in the e-mail field, there will be DNS lookup that won\'t get tunneled through SOCKS5.</string>
-    <!-- context: this key is used whenever there is a link to the about deltachat desktop dialog -->
-    <string name="about">About</string>
 
     <!-- title of the "keybindings" dialog (for the keybindings names as such, where possible the normal command strings are used) -->
     <string name="keybindings">Keybindings</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -413,7 +413,7 @@
     <!-- this is a warning that is shown when one tries to send something to a chat that is not yet accepted. -->
     <string name="accept_request_first">Please accept the chat request first.</string>
     <string name="show_webxdc_in_chat">Show Private App in Chat</string>
-    <string name="what_is_webxdc">What is Webxdc?</string>
+    <string name="what_is_webxdc">What are Private Apps?</string>
 
     <!-- map -->
     <string name="filter_map_on_time">Show locations in time frame</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -875,7 +875,6 @@
     <string name="notify_dismiss">Dismiss</string>
     <string name="notify_reply_button">Reply</string>
     <string name="notify_new_message">New message</string>
-    <string name="notify_new_messages">New messages</string>
     <string name="notify_background_connection_enabled">Background connection enabled</string>
     <string name="notify_priority_high">High</string>
     <string name="notify_priority_max">Max</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -87,6 +87,8 @@
     <string name="this_month">This month</string>
     <string name="last_week">Last week</string>
     <string name="last_month">Last month</string>
+    <!-- Placeholder replaced by relative time, resulting in "Last seen 3min ago" -->
+    <string name="last_seen">Last seen %1$s</string>
     <!-- Placeholder replaced by date or time, resulting in "Last seen at 12:13 AM" or "Last seen Nov 12" -->
     <string name="last_seen_at">Last seen at %1$s</string>
     <string name="last_seen_unknown">Last seen: Unknown</string>
@@ -204,6 +206,7 @@
     <string name="menu_add_attachment">Add Attachment</string>
     <string name="menu_leave_group">Leave Group</string>
     <string name="menu_delete_chat">Delete Chat</string>
+    <string name="menu_clear_chat">Clear Chat</string>
     <string name="ask_delete_named_chat">Are you sure you want to delete \"%1$s\"?</string>
     <string name="menu_delete_messages">Delete Messages</string>
     <string name="delete_contact">Delete Contact</string>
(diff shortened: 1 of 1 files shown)


res/values/strings.xml
```diff
@@ -87,6 +87,8 @@
     <string name="this_month">This month</string>
     <string name="last_week">Last week</string>
     <string name="last_month">Last month</string>
+    <!-- Placeholder replaced by relative time, resulting in "Last seen 3min ago" -->
+    <string name="last_seen">Last seen %1$s</string>
```
r10s, line 91: where does the "ago" come from? there seem to be some strings missing: i am pretty sure, this will be hard to translate to some languages, even without. for now, maybe better do what android/ios are doing and use an absolute time and  `last_seen_at` and `last_seen_unknown`

res/values/strings.xml
```diff
@@ -396,11 +400,20 @@
...
+    <string name="broadcast_please_enter_broadcast_list_name">Please enter a name for the broadcast list.</string>
+    <string name="broadcast_list_warning">Broadcast Lists are always unencrypted because they use blind copy (BCC)</string>
+
     <!-- webxdc -->
     <!-- "Start..." button for an app -->
     <string name="start_app">Start…</string>
     <!-- this is a warning that is shown when one tries to send something to a chat that is not yet accepted. -->
     <string name="accept_request_first">Please accept the chat request first.</string>
+    <string name="show_webxdc_in_chat">Show Private App in Chat</string>
+    <string name="what_is_webxdc">What is Webxdc?</string>
```
r10s, line 416: where is that shown? if in UI, ` as we are no longer using the term webxdc in UI, it should probably read `What are "Private Apps"?

res/values/strings.xml
```diff
@@ -426,7 +439,8 @@
     <string name="search">Search</string>
     <string name="search_explain">Search for chats, contacts, and messages</string>
     <string name="search_no_result_for_x">No results found for \"%s\"</string>
-
+    <string name="search_in_chat">Search in Chat</string>
+    <string name="exit_search">Exit Search</string>
```
r10s, line 443: this is used as a hint for the "X" button beside search, right? a transator hint would probably be good in this case. intuitively, i'd also call this "End Search", there is nothing really "exited", the controls and the list stay.

res/values/strings.xml
```diff
@@ -859,12 +876,15 @@
     <string name="notify_dismiss">Dismiss</string>
     <string name="notify_reply_button">Reply</string>
     <string name="notify_new_message">New message</string>
+    <string name="notify_new_messages">New messages</string>
```
r10s, line 879: just a hint, no blocker: there is also `chat_n_new_messages` - this is maybe even nicer?

res/values/strings.xml
```diff
@@ -859,12 +876,15 @@
...
     <string name="notify_background_connection_enabled">Background connection enabled</string>
     <string name="notify_priority_high">High</string>
     <string name="notify_priority_max">Max</string>
     <string name="notify_name_and_message">Name and message</string>
     <string name="notify_name_only">Name only</string>
     <string name="notify_no_name_or_message">No name or message</string>
+    <string name="notify_bundle_new_messages_in_one_chat">%1$d new messages</string>
```
r10s, line 886: "%1$d new messages" already exists as `chat_n_new_messages` and also handles plural forms.

res/values/strings.xml
```diff
@@ -942,6 +962,8 @@
     <string name="no_spellcheck_suggestions_found">No spelling suggestions found.</string>
     <string name="show_window">Show Window</string>
     <string name="login_socks5_experimental_warning">SOCKS5 support is currently experimental. Please use at your own risk. If you type in an address in the e-mail field, there will be DNS lookup that won\'t get tunneled through SOCKS5.</string>
+    <!-- context: this key is used whenever there is a link to the about deltachat desktop dialog -->
+    <string name="about">About</string>
```
r10s, line 966: in german, that would be translated to "Über" which is already ugly - and not even mistranslated as "Ungefähr" which may still easily happy - and we also had that already. we should not open that can of worms again, see more detailed discussions about that eg. just recently at https://github.com/deltachat/deltachat-desktop/issues/2904 - including better working alternatives.

i am a bit picky about that string as it is one that is pretty much visible and directly related to "Delta Chat". i'd like to avoid sth as "they cannot even create a title for their program information" ;)

res/values/strings.xml
Simon-Laux: Momentjs: https://momentjs.com/docs/#/customization/relative-time/

`Last seen vor 36 Minuten` if set to german, so should be translated to "Zuletzt gesehen"

not using relative time makes the thing much less useful, maybe we could add a collon (`:`) to make it more clear?
"Last seen: 3min ago", "Last seen: now", "Last seen: 3hours ago"

If there are languages we moments doesn't support we already have the same problem in the relative timestamps in the messages anyway.

res/values/strings.xml
```diff
@@ -396,11 +400,20 @@
     <string name="mailing_list">Mailing List</string>
     <string name="mailing_list_profile_info">Changes to mailing list name and image apply on this device only.</string>
 
+    <!-- broadcast lists -->
+    <string name="create_broadcast_list">Create Broadcast List</string>
+    <string name="menu_edit_broadcast_list">Edit Broadcast List</string>
+    <string name="menu_broadcast_list_name">Broadcast List Name</string>
+    <string name="broadcast_please_enter_broadcast_list_name">Please enter a name for the broadcast list.</string>
+    <string name="broadcast_list_warning">Broadcast Lists are always unencrypted because they use blind copy (BCC)</string>
```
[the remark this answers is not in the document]
Simon-Laux: "let's hide the fact that broadcast lists are cleartext", is that your suggestion?

res/values/strings.xml
```diff
@@ -396,11 +400,20 @@
...
+
     <!-- webxdc -->
     <!-- "Start..." button for an app -->
     <string name="start_app">Start…</string>
     <!-- this is a warning that is shown when one tries to send something to a chat that is not yet accepted. -->
     <string name="accept_request_first">Please accept the chat request first.</string>
+    <string name="show_webxdc_in_chat">Show Private App in Chat</string>
```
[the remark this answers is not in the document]
Simon-Laux: already used in the same menu:
Bildschirmfoto 2022-11-16 um 18 37 59

right click on webxdc message in chat audit dialog

res/values/strings.xml
Simon-Laux: right, though only plural is needed in this case, because single messages are displayed on as dedicated notifications.

res/values/strings.xml
Simon-Laux: so rather use `global_menu_help_about_desktop`?

Simon-Laux: Bildschirmfoto 2022-11-16 um 18 44 23

res/values/strings.xml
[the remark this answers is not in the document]
r10s: the review is about making things translatable in practice. as i wrote, "let's leave this untranslated" is an option as well, even the first one ;)

res/values/strings.xml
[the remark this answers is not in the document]
r10s: i understand. wondering if that option is really needed, one can easily go to the private app from the info-message (we do not have shortcuts for images etc. as well). so, tbh, i would just remove it.

however, sure, no blocker.

nb: in general, menu items should _never_ contain line breaks.

res/values/strings.xml
r10s: i see, the string "ago" comes from some dependency.

to make that more clear, maybe rename the key to `last_seen_rel`, that also matches nicely to `last_seen_at`

res/values/strings.xml
r10s: better. but i'd remove the first line completely now, that looks pretty redundant. if you really want to have a direct link, add it after "about delta chat". but i suggest to remove it completely here, that shortcut seems to be unneeded, what is the usecase? if even hidden that way? you get to the web easily from the about dialog.

res/values/strings.xml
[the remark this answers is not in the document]
Simon-Laux: > nb: in general, menu items should never contain line breaks.

so we need a desktop issue to make the context menu larger in case the text is too long.

res/values/strings.xml
Simon-Laux: please make an issue for this on the desktop repo if you want this addressed

adbenitez:
@Simon-Laux what is the state of this PR?

adbenitez:
closing this due to inactivity, we can create a new PR when we decide to get back to this

Simon-Laux:
FTR: `last_seen` is now called `last_seen_relative` and addressed in https://github.com/deltachat/deltachat-android/pull/2563 so it needs wo be removed/not done anymore in this pr.

> @Simon-Laux what is the state of this PR?

"Fighting" for every translation string is not my favourite task, so I did not continue it. But it probably makes sense to take up the work again and solve the remaining questions.

r10s:
> Discussing translation strings is not my favourite task, so I did not continue it.

well, there is always some unloved chore to do

PR #2433, opened 2022-12-03
adapt build script for M1/M2 hosts

r10s:


r10s:
> Is it for CI or do you have such host around?

i have an M1 mac as my development machine now. seems, NDK does not fully support M1/M2 as a host, the `uname -m` fails therefore.

PR #2436, opened 2022-12-12
prepare 1.34.7

r10s:
this mainly updates to recent core with several fixes


Hocuri (approved):

adbenitez (approved):

r10s:
let's merge that in, however, we should do a 1.34.8 with a new core to avoid https://github.com/deltachat/deltachat-core-rust/issues/3841 that may at least make the app unusable for some devs ...

PR #2440, opened 2022-12-25
Explicitly set RUSTUP_TOOLCHAIN to 1.64.0

link2xt:
Core is supposed to support all Rust versions above MSRV, there is no need to use the same toolchain for the core across all platforms.


scripts/ndk-make.sh
```diff
@@ -108,12 +108,14 @@ if test -z "$(find "$ANDROID_NDK_ROOT" -name libgcc.a -print -quit)"; then
     echo 'INPUT(-lunwind)' >"$TMPLIB/libgcc.a"
 fi
 
+RUSTUP_TOOLCHAIN=1.64.0
```
r10s, line 111: maybe add a comment that when modifying this line, one should also modify `install-toolchains.sh`. and maybe also say why we are not using `cat rust-toolchain` - at least i tend to forget about that completely as it is touched only very rarely - and at least some other devs are not that deep in rust :)

r10s (approved):
probably makes sense. in all these stacks, ndk, jni and whatnot a slightly older rust toolchain is probably not adding much of issues.

scripts/ndk-make.sh
```diff
@@ -108,12 +108,14 @@ if test -z "$(find "$ANDROID_NDK_ROOT" -name libgcc.a -print -quit)"; then
...
+
 if test -z $1 || test $1 = armeabi-v7a; then
     echo "-- cross compiling to armv7-linux-androideabi (arm) --"
     export CFLAGS=-D__ANDROID_API__=16
     TARGET_CC=armv7a-linux-androideabi16-clang \
     TARGET_AR=llvm-ar \
-    cargo +`cat rust-toolchain` rustc $RELEASEFLAG --target armv7-linux-androideabi -p deltachat_ffi -- -L "$TMPLIB"
+    cargo "+$RUSTUP_TOOLCHAIN" rustc $RELEASEFLAG --target armv7-linux-androideabi -p deltachat_ffi -- -L "$TMPLIB"
```
Hocuri, line 118: I'm +/-0 on this PR, because the downside is that this will mean that we need to bump the Rust version in even more places every few months.

But maybe we can create one file `rust-toolchain` in the dc-android repo that contains the name of the currently used Rust version? This way, we'd only need to update it in one place.

E.g.:

```suggestion
    cargo +`cat ../../rust-toolchain` rustc $RELEASEFLAG --target armv7-linux-androideabi -p deltachat_ffi -- -L "$TMPLIB"
```

flub (approved):

r10s (approved):

PR #2442, opened 2023-01-04
Add metadata/ for F-Droid

link2xt:
fdroiddata/ contains some metadata, but no icon or screenshots at <https://gitlab.com/fdroid/fdroiddata/-/tree/master/metadata/com.b44t.messenger/>

How to add metadata to the project repository is described at <https://f-droid.org/en/docs/All_About_Descriptions_Graphics_and_Screenshots/#in-the-applications-source-repository>

This change adds metadata in the "fastlane" format. Screenshots and icon are copied from the store/ directory and compressed with `oxipng --zopfli`.

### DIFF
--- a/docs/f-droid.md
+++ b/docs/f-droid.md
@@ -4,9 +4,15 @@
   is the Delta Chat page on F-Droid.org,
   the F-Droid-app will show similar information.
 
-- <https://gitlab.com/fdroid/fdroiddata/blob/master/metadata/com.b44t.messenger.yml>
-  contains the description and all meta data shown for Delta Chat on F-Droid;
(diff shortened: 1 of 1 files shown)


r10s (approved):
greaaaat!

i was not aware that we can host the metadata on our own.

so, once this is merged, we have to do a pr against f-droid and remove summary etc. from https://gitlab.com/fdroid/fdroiddata/blob/master/metadata/com.b44t.messenger.yml and https://gitlab.com/fdroid/fdroiddata/-/tree/master/metadata/com.b44t.messenger/ , right?

maybe also add a little note about the metadata at `./docs/f-droid.md` - currently it points to f-droid wrt metadata.

PR #2446, opened 2023-01-08
fade left/right animation when opening the archive

r10s:
this is similar to opening a chat and feels more natural. settings/connectivity/help etc. still appear from the bottom as being "something little different".

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -406,6 +406,7 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
     } else {
       startActivity(intent);
     }
+    overridePendingTransition(R.anim.slide_from_right, R.anim.fade_scale_out);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


Hocuri:
> settings/connectivity/help etc. still appear from the bottom as being "something little different".

for me, they have a zoom in/out animation (doesn't matter though, just noticed).

WRT the PR: To me, this seems as natural as before. WA uses the zoom in/out animation both for chats and for the archive, TG uses the fade left/right animation for both. Let's wait for @adbenitez to comment and if he doesn't have an opinion, either, we can also just merge it.

adbenitez (approved):
I don't have an strong opinion, both ways would be fine, but IMO this animation from the side looks a  bit more modern/material design/elegant than the previous one so +1

r10s:
> for me, they have a zoom in/out animation (doesn't matter though, just noticed).

it think, we were using the default animation before; on my test device, there was sth coming from the bottom.

> WRT the PR: To me, this seems as natural as before. WA uses the zoom in/out animation both for chats and for the archive, TG uses the fade left/right animation for both.

interesting, i always thought wa also comes from the right :) on my test device, this is "from the bottom" on wa - so, indeed seems default animations differ from device to device.

i think, the most important thing ist that the chatlist-archive-chat transitions are consistent.

PR #2447, opened 2023-01-09
prepare 1.34.10

r10s:
Screenshot 2023-01-10 at 11 29 40


r10s:
> Wait, there is a new item in release cycle wrt F-Droid

i have seen that - and still forget about it :)

i'll add that.

i also added a little device message to say sth. about the new archived link.

r10s:
@link2xt i added f-droid changelog and a device message - can you please re-review :)

Issue #1066, opened 2019-10-11
Is Socks Proxy Support in the Roadmap?


Adrianzo:
something in the lines of [this](https://mstajbakhsh.ir/my-contribution-in-k9-android-mail-client/) would be useful for securing group communications. thanks!

r10s:
hi @Adrianzo, welcome!

there are no no concrete plans for the next weeks or months of ppl i know. but these things may change quickly if someone really starts doing things on that. maybe you can help on this issue?

and, if you define "roadmap" as things someone has thought+talked about, yes, this happens already several times :)

in general, please use the support forum for questions and discussion. the support forum is also cross-platform, which also makes much more sense on general questions or feature requests.

https://support.delta.chat (you can log in with your github account)

r10s:
i think, if needed, we can continue the discussion in the [forum](https://support.delta.chat); also as it is not only android-releated

PR #2452, opened 2023-01-18
add socks5 settings to registration activity

### LINKED ISSUE #406: Socks/proxy support
I think the client is missing a very important part where a user can communicate with email servers using socks or http proxies. I personally use Tor via Orbot, but i don't like the way it proxifies apps, I prefer proxifying apps individually. For that, Delta chat needs to support proxy communication.

May we have per-account settings which allow to set socks or http proxy?

### LINKED ISSUE #1742: "Delete messages from device" doesn't free up any memory
- Operating System: Android 7.1.2 (LineageOS 14.1)
- Delta Chat Version: 1.12.5

- Expected behavior:
When DeltaChat automatically deletes messages from the device after a certain time period according to how the user has setup that option in DeltaChat's settings it should not only reflect visually inside DeltaChat but it should actually also free up memory space on the device obviously :-)

- Actual behavior:
deleted messages indeed do not show up inside DeltaChat anymore however the amount of memory used by DeltaChat on the device does not decrease at all. DeltaChat on my device used an enormous amount of 660 MB in total despite only very seldom using DeltaChat and after enabling automatic deletion it still uses 660 MB! I already rebooted the device and even deleted it's cache and Dalvik cache but that made no difference.
Btw I've set DeltaChat to delete messages from device after 1 year and to never delete messages from server. It deleted 2200+ messages when I enabled deleting from device. I also wonder where those 2200+ message came from, I certainly didn't use DeltaChat to that amount. Far from it!
1. check memory used by DeltaChat via Android's settings
2. set DeltaChat option to "Delete messages from device" to a timeframe which would result in DeltaChat needing to delete a decent amount of messages (perhaps including media ideally) so that it should clearly reflect in the amount of memory being used by DeltaChat
3. wait some time and then go back into Android's settings to check the amount of memory being used by DeltaChat now

### LINKED ISSUE #2234: Add SOCKS5 options to advanced account settings
Core and Desktop already have SOCKS5 support, only the UI is missing.

Related "feature request" issues: #406, #1066

### LINKED ISSUE #2405: Can't turn off socks5 proxy after importing backup
- Android version: 10 QP1A.190711.020
- Device: Redmi 9A / M2006C3LG
- Delta Chat version: 1.32.0
- Expected behavior: After importing a backup from desktop client, where I have configured Tor via socks5 proxy, it either works or I can disable the socks5 proxy
- Actual behavior: I can't connect to the socks5 proxy, so I can't connect to the server. Also I can't disable the socks5 proxy, because the deltachat-android configure screen doesn't have the socks5 option.
  - Login on the desktop client
  - configure a socks5 proxy (e.g. localhost:9050 for most tor setups on linux)
  - export a backup
  - open the Android app
  - import the backup
  - see connection errors

adbenitez:
closes #2405
closes #2234
closes #406

(also related: #1066)

### DIFF
--- a/res/layout/registration_activity.xml
+++ b/res/layout/registration_activity.xml
@@ -123,7 +123,7 @@
             android:visibility="gone"
             tools:visibility="visible"
             app:constraint_referenced_ids="encrypt_checkbox, inbox, imap_login, imap_server, imap_port, imap_security_label, imap_security, outbox_view_spacer_top,
-            outbox, smtp_login, smtp_password, smtp_server, smtp_port, smtp_security_label, smtp_security, auth_method_label, auth_method, cert_check_label, cert_check, view_log_button, socks5_label, socks5_switch, socks5_experimental_warning" />
+            outbox, smtp_login, smtp_password, smtp_server, smtp_port, smtp_security_label, smtp_security, auth_method_label, auth_method, cert_check_label, cert_check, view_log_button, socks5_switch" />
 
         <ImageView
             android:id="@+id/advanced_icon"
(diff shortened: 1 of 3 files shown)


adbenitez:
NOTE: I was unable to use a proxy on android, but configuring these options in android with this PR and importing on old desktop version did worked, also inspecting the backup the settings were saved so I suspect that something is failing with recent core versions

adbenitez:
@r10s, ok, adapted to your suggestions, thanks for the feedback!

PR #2456, opened 2023-01-27
Removed create group button from contact info screen

### LINKED ISSUE #2432: Remove "Create Group or Subject"
from nov22 consistency discussions:

_Whatsapp has "create group with contact" in the "shared chats" section. DC Android has it in the contact profile further up, iOS and Desktop don't have it.  Probably best to strike it in Android as it does not appear to be much used._

meganoahj:
closes #2432


r10s (approved):
great! to make this complete, we should also remove `SETTING_NEW_GROUP` and `onNewGroupWith()`

PR #2457, opened 2023-01-29
Quoted message in email 

letyphon:
Hi everybody!

I am not a developer but a simple user.

I really like Delta Chat app. And I would like to use it in a daily basis. But while having a conversation with someone using a simple email app there is something not going on well.

The way that the replied messages appearing in the email inbox page does not relly leave a good impression. I personally would prefer that the quoted part appears under the message not above and is separated by a line from the initial text inatead of the symbol ' > '.

Example of the quotes in the gmail inbox:

resim

How they actually look in the Delta chat app:

image_2023-01-29_00-39-00

So I'd like to know if others feel the same way and if there will be an update on this?

But even if not, I really appreciate your work so far :)

### DIFF
--- a/src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
+++ b/src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
@@ -311,8 +311,11 @@ public class NotificationCenter {
         Util.runOnAnyBackgroundThread(() -> {
 
             DcChat dcChat = dcContext.getChat(chatId);
+            DcMsg dcMsg = dcContext.getMsg(msgId);
+            DcMsg quotedMsg = dcMsg.getQuotedMsg();
+            boolean isGroupMention = quotedMsg != null && dcChat.isGroup() && quotedMsg.isOutgoing();
 
-            if (!Prefs.isNotificationsEnabled(context) || dcChat.isMuted()) {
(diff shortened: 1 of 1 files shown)


webratte:
I absolutely would like it this way. I asked a long time ago for this.
I believe the "normal behaviour" is to put the answer in top of the quoted message in a regular mail.

letyphon:
> I absolutely would like it this way. I asked a long time ago for this. I believe the "normal behaviour" is to put the answer in top of the quoted message in a regular mail.

Look, even here when you click to "reply" button, it puts the quote above and the message under...

My real concern about this actually is that some systems cannot display the original message while we put the quote above. I use "zendesk" software at my work and when people sends a reply via email and when it is quoted in a way that the delta chat does, I cannot see the original message.

Just recently it happened to me, I sent an email via K-9 email to a company and they replied back saying that they did not see my message. Then, ı realized that my settings on K-9 was also set to put the quote above...

So, it is a need more than a preferance.

Hopefully someone will consider this in the fature releases :)

adbenitez:
hi @letyphon this PR seems invalid, it talks about something that has nothing to do with the code modified,
the change you request should be done in deltachat-core-rust repo not here, otherwise the other delta chat versions (desktop, ios and 3rd party clients) will not have the same behavior,

you should start by creating a feature request in our forum where more users can see your proposal and agree/disagree, check:
https://support.delta.chat/

thanks for your feedback!

PR #2458, opened 2023-01-30
Webxdc fixes

Hocuri:
This fixes all the checks in the `test webxdc`, see the group on Delta Chat

### DIFF
--- a/res/raw/webxdc_wrapper.html
+++ b/res/raw/webxdc_wrapper.html
@@ -1,0 +1,107 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <style>
+      html,
+      body {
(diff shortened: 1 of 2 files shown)


r10s (approved):
works as expected, tested with a handful of webxdc.

i am a bit unsure about the change wrt `shouldOverrideUrlLoading()` (via `openOnlineUrl()`), not sure if the method is called at all in the API we're targeting, but that is another issue and the change in `openOnlineUrl()` probably does not make change worse.

r10s:
the pr is meanwhile reverted on master and included in a larger pr with more patches, https://github.com/deltachat/deltachat-android/pull/2465, @Hocuri please double-check :)

PR #2465, opened 2023-02-03
Combined webxdc patches

r10s:
this includes the changes of https://github.com/deltachat/deltachat-android/pull/2458 plus things discussed afterwards.


res/raw/sandboxed_iframe_rtcpeerconnection_check.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
+<head>
+<meta charset="UTF-8">
```
WofWca, line 3:
```suggestion
<!DOCTYPE html>
<html>
<head>
<meta charset="UTF-8">
```

Medium yikes

res/raw/sandboxed_iframe_rtcpeerconnection_check.html
```diff
@@ -0,0 +1,32 @@
...
+    parent.postMessage({ msgType: "result", rtcpcCreationFailed }, location.origin);
+  }
+});
+
+parent.postMessage({ msgType: "ready" }, location.origin);
+</script>
+</head>
+<body>
+</body>
+</html>
```
WofWca, line 32:
```suggestion
</body>
</html>

```

Minor yikes

res/raw/webxdc_wrapper.html
```diff
@@ -0,0 +1,145 @@
...
+          }
+        }
+        console.log("WEBRTC-WG: done");
+
+        try {
+          connections.push(new RTCPeerConnection());
+          console.log("Error: was able to create more than 500 connections");
+          loadingDiv.innerText =
+            "Error: was not able to block webrtc ERROR_A";
+        } catch (error) {
```
WofWca, line 84: Nitpick: doing just `new RTCPeerConnection()` inside the `try {` block would be more correct IMO.

WofWca (changes requested):
Added suggestions to fix my errors (sorry).

Otherwise: did not test, but looks ok.

WofWca (approved):

res/raw/webxdc_wrapper.html
```diff
@@ -0,0 +1,145 @@
...
+            const askIframeToPerformCheck = () => {
+              // Why `"*"`? See the comment below.
+              sandboxedIframe.contentWindow.postMessage("performCheck", "*");
+            };
+            /** @type {(e: MessageEvent) => void} */
+            const messageListener = (e) => {
+              // Checking `event.origin !== location.origin` just in case would be safer,
+              // but `sandbox`ed iframes seem to send messages with `origin` set to `null`,
+              // so we skip the check, relying on the fact that we don't actually load any
+              // untrusted scripts that could `postMessage` before this check is completed.
```
Hocuri, line 98: I'm just wondering if malicious code could continue to be executed when the webxdc is closed (using a serviceworker or so). But worrying about this is probably out of scope for this PR, since it would create a bunch of other problems with the current solution, anyway.

src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -121,6 +122,7 @@ protected void onCreate(Bundle state, boolean ready) {
     webSettings.setJavaScriptEnabled(true);
     webSettings.setAllowFileAccess(false);
     webSettings.setBlockNetworkLoads(!internetAccess);
+    webView.setNetworkAvailable(internetAccess);
```
Hocuri, line 125: Minor remark: [This is only a hint for the WebView](https://developer.android.com/reference/android/webkit/WebView#setNetworkAvailable(boolean)) (similar to our maybe_network()) and doesn't restrict anything.

Simon added it because it cannot hurt, while my personal opinion would be to leave it out, since it doesn't do anything but looks like it does.

src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -193,6 +201,11 @@ protected WebResourceResponse interceptRequest(String rawUrl) {
       if (path.equalsIgnoreCase("/webxdc.js")) {
         InputStream targetStream = getResources().openRawResource(R.raw.webxdc);
         return new WebResourceResponse("text/javascript", "UTF-8", targetStream);
+      } else if (path.equalsIgnoreCase("/sandboxed_iframe_rtcpeerconnection_check_5965668501706.html")) {
+        // It is important to NOT return a malicious file instead of this one.
+        // Consider `iframe.srcdoc` as an alternative.
```
Hocuri, line 206: @WofWca What does this comment mean, I don't understand it.

Hocuri (approved):

res/raw/webxdc_wrapper.html
WofWca: When a webxdc is closed, there's no reference to this window object anymore, as far as I'm aware, so no.

src/org/thoughtcrime/securesms/WebxdcActivity.java
WofWca: Just add a comment then?

src/org/thoughtcrime/securesms/WebxdcActivity.java
WofWca: Sorry, I meant to tell to be careful about changing this and surrounding code. Maybe adding this to the comment could help.

res/raw/webxdc_wrapper.html
```diff
@@ -0,0 +1,145 @@
...
+          ],
+        };
+
+        console.log("WEBRTC-WG: allocating");
+        loadingProgress.value = 0;
+        while (connections.length < 500) {
+          try {
+            connections.push(new RTCPeerConnection(cert));
+            if (connections.length%50 == 0) {
+              loadingProgress.value++;
```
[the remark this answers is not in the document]
WofWca: Good catch, I think you're right.

WofWca (changes requested):
See new comments

res/raw/webxdc_wrapper.html
Simon-Laux, line 67:
```suggestion
              loadingProgress.value = connections.length / 5;
```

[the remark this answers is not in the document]
r10s: it does not go to 100% - but it did not go to 100% before as well (only to 50%).
as we do not know, how long the thing takes, the progress bar is more moving something "up" (we increase every 50 succeeding connections and every failed one)

r10s: nice! so, we do steps of 20 / 40 / 60 / 80 and in between, if needed, up to 20 wait-for-gc.

res/raw/webxdc_wrapper.html
```diff
@@ -0,0 +1,145 @@
...
+        width: 100%;
+      }
+      #progress {
+        width: 100%;
+      }
+    </style>
+  </head>
+  <body>
+    <div id="loading">
+      <progress id="progress" max="100" value="0"></progress>
```
WofWca, line 31:
```suggestion
      <progress id="progress" max="500" value="0"></progress>
```

res/raw/webxdc_wrapper.html
```diff
@@ -0,0 +1,145 @@
...
+        console.log("WEBRTC-WG: allocating");
+        loadingProgress.value = 0;
+        while (connections.length < 500) {
+          try {
+            connections.push(new RTCPeerConnection(cert));
+            if (connections.length%50 == 0) {
+              loadingProgress.value++;
+              await new Promise((res) => setTimeout(res));
+            }
+          } catch (error) {
```
WofWca, line 70:
```suggestion
          try {
            connections.push(new RTCPeerConnection(cert));
            loadingProgress.value++;
            if (connections.length%50 == 0) {
              // Let the the progress bar graphics update.
              await new Promise((res) => setTimeout(res));
            }
          } catch (error) {
```

WofWca (changes requested):
Suggested an alternative progress bar logic.

res/raw/webxdc_wrapper.html
```diff
@@ -0,0 +1,145 @@
...
+            loadingProgress.value++;
+            new Array(1024*1024).fill(0);
+            await new Promise((res) => setTimeout(res, 500));
+            console.log("WEBRTC-WG: waiting for gc");
+          }
```
WofWca, line 75:
```suggestion
          try {
            connections.push(new RTCPeerConnection(cert));
            loadingProgress.value++;
            if (connections.length%50 == 0) {
              // Let the the progress bar graphics update.
              await new Promise((res) => setTimeout(res));
            }
          } catch (error) {
            new Array(1024*1024).fill(0);
            await new Promise((res) => setTimeout(res, 500));
            console.log("WEBRTC-WG: waiting for gc");
          }
```

r10s: picked up the `max="500"` in recent commit, thanks!

res/raw/webxdc_wrapper.html
r10s: combined that with the `max="500"` suggestion from below in the recent commit (to the the progress bar go backward less often)

res/raw/webxdc_wrapper.html
r10s: i also think, it is good enough, for an attack one has to make `console.log` or `innerText` fail - before any code from the webxdc is executed.

src/org/thoughtcrime/securesms/WebxdcActivity.java
r10s: it sets `window.navigator.isOnline` - i think, it is good if that one is `false`, this may help using existing code. but yes, we should add a comment, i do a commit.

r10s: i added a comment.

src/org/thoughtcrime/securesms/WebxdcActivity.java
r10s: i removed the comment with the recent commit as it seems a bit confusing - and also to to not duplicate the comment as it at least partly true also for the  other files.
i also think, it is pretty clear that all that code is sensible :) - otherwise the code is mostly self-explaining.
please feel free to add an updated comment if you still feel it is needed.

src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -122,26 +122,16 @@ protected void onCreate(Bundle state, boolean ready) {
...
-    byte[] bootstrap_bytes = new byte[0];
-    try {
-      bootstrap_bytes = new byte[bootstrap_stream.available()];
-      bootstrap_stream.read(bootstrap_bytes);
-      String bootstrapHtml = new String(bootstrap_bytes);
-      webView.loadDataWithBaseURL(this.baseURL + "/webxdc_bootstrap324567869.html", bootstrapHtml, "text/html", "UTF-8", null);
-    } catch (IOException e) {
-      e.printStackTrace();
-    }
+    webView.loadUrl(this.baseURL + "/webxdc_bootstrap324567869.html");
```
Hocuri, line 134: "/webxdc_bootstrap324567869.html" could be a constant to avoid typos from breaking things (OTOH, the typo would break all non-malicious webxdcs, so no way it goes unnoticed)

Hocuri (approved):

src/org/thoughtcrime/securesms/WebxdcActivity.java
r10s: i think, it is good enough :)

PR #2466, opened 2023-02-04
add CSP to webxdc

r10s:


adbenitez (approved):

Hocuri:
With the changes from https://github.com/deltachat/deltachat-android/pull/2465, the CSP will not be set for the bootstrap html, because it's [loaded using loadDataWithBaseURL](https://github.com/deltachat/deltachat-android/pull/2465/files#diff-412ade266c3378a85632029f2b31418af917def7695df8675bfd30dc5016baebR141) and therefore interceptRequest() is not called.

Is this OK? The solution would be using `loadUrl()` again and in `interceptRequest()` return the correct file, (like we're doing it for `webxdc.js` already).

Simon-Laux:
I think it does not matter much for the bootstrap html, if we need a csp in there we can add it over the meta tag inside of the html?

r10s:
@Hocuri @WofWca was there a reason for switching from `loadUrl()` to `loadDataWithBaseURL()` for the initial page apart from that it was maybe more handy that time?  otherwise, i agree, we should use `loadUrl()` as for the two other internal files (`webxdc.js` and `sandboxed_iframe*.html`), that way everything is routed through `interceptRequest()` which makes the flow cleaner.

apart from that, _at a first glance,_ i also think, we do not really need a CSP for the bootstrap, `index.html` that is then loaded in one of the iframes get its own CSP.

Hocuri:
I don't think there was any particular reason

r10s:
i added a commit to https://github.com/deltachat/deltachat-android/pull/2465 that uses `loadUrl()` also for the bootstrap code so that things are more straight forward and adding CSP is easier.

one advantage why `loadDataWithBaseURL()` might be better is that otherwise the bootstrap code can be reloaded by a webxdc later - but as the bootstrap code is only static code, a copy of the bootstrap code could also be added to the webxdc.

WofWca:
> I think it does not matter much for the bootstrap html

Since we're not sandboxing the app page's iframe, and it's same-origin, it can access the parent, utilizing its lax CSP, so we still need CSP for the wrapper script.

r10s:
thanks for explaining and double-checking, @WofWca

PR #2469, opened 2023-02-07
no automatic remote content loading for contact requests

r10s:
loading remote content in contact requests
now always require explicit consent of the user.
moreover, contact request do not alter
the otherwise used never/once/always settings.

iirc, this was also the original gist (contact requests needed to be confirmed to really use the message that time), however, contact requests have  changed since then and behave mostly like normal messages.

we have a similar thing also for webxdc, where webxdc contact requests also cannot send updates.

these improvements are the result of discussions with @Simon-Laux who is just adding html-view to desltop \o/

counterpart of https://github.com/deltachat/deltachat-ios/pull/1804

### DIFF
--- a/src/org/thoughtcrime/securesms/FullMsgActivity.java
+++ b/src/org/thoughtcrime/securesms/FullMsgActivity.java
@@ -19,9 +19,11 @@ import java.lang.ref.WeakReference;
 public class FullMsgActivity extends WebViewActivity
 {
   public static final String MSG_ID_EXTRA = "msg_id";
+  public static final String IS_CONTACT_REQUEST = "is_contact_request";
   private int msgId;
   private DcContext dcContext;
   private boolean loadRemoteContent = false;
+  private boolean isContactRequest;
 
   enum LoadRemoteContent {
     NEVER,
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/FullMsgActivity.java
```diff
@@ -115,17 +118,24 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+          onceCheckmark = checkmark;
         } else {
-          builder.setNeutralButton(R.string.always, (dialog, which) -> onChangeLoadRemoteContent(LoadRemoteContent.ALWAYS));
-          builder.setNegativeButton((loadRemoteContent? "" : checkmarkPrefix) + getString(R.string.never), (dialog, which) -> onChangeLoadRemoteContent(LoadRemoteContent.NEVER));
-          builder.setPositiveButton((loadRemoteContent? checkmarkPrefix : "") + getString(R.string.once), (dialog, which) -> onChangeLoadRemoteContent(LoadRemoteContent.ONCE));
+          neverCheckmark = checkmark;
         }
 
+        if (!isContactRequest) {
+          builder.setNeutralButton(alwaysCheckmark + getString(R.string.always), (dialog, which) -> onChangeLoadRemoteContent(LoadRemoteContent.ALWAYS));
```
adbenitez, line 134: (side comment: I wonder if this manual concatenation of translated text with a checkmark at the beginning is not wrong for LTR languages)

adbenitez (approved):

src/org/thoughtcrime/securesms/FullMsgActivity.java
r10s: yeah, it is maybe not perfect, but nowadays UIs have checkbox/switches left/right and no one really cares :)

it is as if the checkbox is on the right in english text, would not be worse, imho.

we can change that if there are complains :)

PR #2474, opened 2023-02-10
tweak webxdc

r10s:
remove unused lines; the iframe should include webxdc.js as needed.

they came in by https://github.com/deltachat/deltachat-android/pull/2465/commits/22ae648160df066ad47cdef4db7659e2fc8866ed but i do not see the sense of these lines.

maybe i have overseen sth?

### DIFF
--- a/res/raw/webxdc_wrapper.html
+++ b/res/raw/webxdc_wrapper.html
@@ -129,8 +129,6 @@
           } else {
             loadingDiv.innerHTML = "";
             iframe.src = "index.html";
-            iframe.contentWindow.webxdc_internal = window.webxdc_internal;
-            iframe.contentWindow.webxdc = window.webxdc;
           }
         }
 


WofWca (commented):
Did you test the change?
I thought we're doing this because the `webxdc.js` relies on URL and some other stuff, which is only available in the top frame. This is why we were discussing the "bridging" in case we were gonna go for `sandbox`ed iframes.

res/raw/webxdc_wrapper.html
```diff
@@ -129,8 +129,6 @@
           } else {
             loadingDiv.innerHTML = "";
             iframe.src = "index.html";
-            iframe.contentWindow.webxdc_internal = window.webxdc_internal;
-            iframe.contentWindow.webxdc = window.webxdc;
```
r10s, line 133: in my understanding:

- `window.webxdc` in `webxdc_wrapper.html`  is not set at all, so `null`
- `window.webxdc` in `index.html` may be set when `webxdc.js` is included. from outside, this is reachable as `iframe.contentWindow.webxdc`

in theory, this lines could even overwrite the just loaded webxdc api in the iframe with `null` - probably this just does not happen as javascript is single threaded and `iframe.src` is executed later.

please correct me if i am wrong :)

res/raw/webxdc_wrapper.html
```diff
@@ -129,8 +129,6 @@
           } else {
             loadingDiv.innerHTML = "";
             iframe.src = "index.html";
-            iframe.contentWindow.webxdc_internal = window.webxdc_internal;
```
r10s, line 132: `webxdc_internal` seems not to be used at all, grep returns zero results for me.

r10s:
i tried to comment a bit more in detail. and yes, things work for me. maybe @Hocuri knows more :)

res/raw/webxdc_wrapper.html
WofWca: Makes sense. Perhaps it was just copied from the desktop version.
https://github.com/deltachat/deltachat-desktop/pull/3098/commits/16bdbaa01b634ef4d2bcd98dd86ef712baa157a8#diff-15491a0f9e706d644386a23d0996ebec3a6b94f82a25a220f409272a359fb0c0R149-R150

res/raw/webxdc_wrapper.html
Simon-Laux: that comes from desktop, can be removed on android

Simon-Laux (approved):

Hocuri (approved):
I didn't add these lines and can't comment on whether they are necessary, but seems like these lines weren't necessary.

PR #2479, opened 2023-02-28
Add DcJsonrpcInstance class

link2xt:
TODO:
- [x] Parse responses.
- [x] Dispatch responses via SettableFuture.

The purpose of this is to make it possible to use JSON-RPC calls when the corresponding CFFI is not ready yet, or skip adding CFFI for new calls completely.

This change enables `jsonrpc` core feature in `ndk-make.sh`.

Since core PR https://github.com/deltachat/deltachat-core-rust/pull/4341 is merged, we don't have to switch event processing to use JSON-RPC when we create a JSON-RPC instance. So we can keep event loop using the C API and use JSON-RPC for something simple at first.

For example, we can try to use it for QR-code scanning first to get proper errors on failure. Another option is to use it for event reception, so we receive proper JSON events instead of data1, data2 etc. Another option is to use it instead of the recently added in #2539 `dc_http_response_t` API.

Java API for JSON-RPC instance is ready, but we need a way to dispatch . ~ 
[java.util.concurrent.CompletableFuture](https://docs.oracle.com/javase/8/docs/api/index.html?java/util/concurrent/CompletableFuture.html)(UPDATE: not compatible with Android API 16) looks like something similar to the [`asyncio.Future` we use in the Python JSON-RPC client](https://github.com/deltachat/deltachat-core-rust/blob/514074de8bcd22c3181317e19238131d328ef3a1/deltachat-rpc-client/src/deltachat_rpc_client/rpc.py#L25).


r10s:
why is that needed currently? that would also require core to be compiled with jsonrpc, right?  (i see it is a draft, but still :)

r10s:
tbh, i am unsure about getting jsonrpc in that quickly - i have the feeling it is not yet settled and stable - i see quite some breaking changes and and lots of pr in core, maybe it is wise to postpose that a little longer.

for qr code scanning, i think it is easily doable to have a "classic" ffi (in fact, we already did that at https://github.com/deltachat/deltachat-core-rust/pull/3489 )

still interesting to test in a pr, however, to see how much apk grows (if it is only a few kb, the decision may be easier to have both for some time)

also, i'd like to run the benchmark script of @Hocuri at some point and see how much performance changes by using jsonrpc , eg. for the chatlist.

but again, i do not sth of that as urgent things.

r10s:
a yes, i noticed too late that https://github.com/deltachat/deltachat-android/pull/2483 is actually on master and just shown wrongly as "closed" on github.

so, yes, makes sense to try jsonrpc to get some findings and impressions

hpk42:
IMO it makes sense to start being able to use jsonrpc also with Android (and iOS). 
it's certainly going to take a longer while until a serious attempt at 
porting existing CFFI calls to jsonrpc can/should take place.  
But if there is at least a base mechanism it becomes possible to play with it much better,
also from DeltaLab etc.

adbenitez:
I think this is ready for review (commits need to be squashed)

src/com/b44t/messenger/rpc/Rpc.java
```diff
@@ -0,0 +1,120 @@
...
+import com.google.gson.reflect.TypeToken;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutionException;
+
+public class Rpc {
+    private Map<Integer, CompletableFuture<JsonElement>> requestFutures = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
adbenitez: I don't know, HashMap is used in a similar fashion, maybe @Hocuri can clarify

src/com/b44t/messenger/rpc/Rpc.java
```diff
@@ -0,0 +1,120 @@
...
+        try {
+            return call(method, params).get();
+        } catch (ExecutionException e) {
+            throw (RpcError)e.getCause();
+        } catch (InterruptedException e) {
+            throw new RpcError(e.getMessage());
+        }
+    }
+
+    public int addAccount() throws RpcError {
```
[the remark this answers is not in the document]
adbenitez: yes it is unused, I just add it as example so when adding new API the dev knows how to add them, I could try using addAccount()

src/com/b44t/messenger/rpc/Rpc.java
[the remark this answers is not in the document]
Hocuri: Either way is fine.

> Casting to `Map` interface immediately only reduces the possibilities of usage as far as I understand

That's correct, but since we don't need these other possibilities of usage, in Java it's idiomatic to use the more general type (although we do not have to follow idiomatic Java though).

> like casting to a dynamic trait object in Rust.

Yes, although every variable in Java is actually what would be a dynamic trait object in Rust. So, we have the choice between `Box<dyn ConcurrentHashMap>` or `Box<dyn Map>` here.

src/com/b44t/messenger/rpc/Rpc.java
```diff
@@ -0,0 +1,120 @@
...
+
+    public Rpc(DcJsonrpcInstance dcJsonrpcInstance) {
+        this.dcJsonrpcInstance = dcJsonrpcInstance;
+    }
+
+    private void processResponse() throws JsonSyntaxException {
+        String jsonResponse = dcJsonrpcInstance.getNextResponse();
+        Response response = gson.fromJson(jsonResponse, Response.class);
+
+        if (response.id == 0) { // Got JSON-RPC notification, ignore
```
Hocuri, line 30: For my understanding, what's a JSON-RPC notification?

src/com/b44t/messenger/rpc/RpcError.java
```diff
@@ -0,0 +1,11 @@
+package com.b44t.messenger.rpc;
+
+/**
+ * An exception occurred while processing a request.
+ **/
+public class RpcError extends Throwable {
```
Hocuri, line 6: You should extend Exception or RuntimeException here, not Throwable, because to catch "everything catchable" we often write `catch (Exception e)`. Errors that extend Throwable but not Exception are meant to not be catched at all, like an OutOfMemoryError (idk why it's not meant to be catched, to me it seems like recovering is worth a try, but anyway).

Might also call it `RpcException` instead of `RpcError`, because in Java Errors are usually something unrecoverable, but that's more a matter of taste.

Or we might even get rid of RpcError completely, but then we need some other way of handling errors, prob. returning an object that either contains the result or the error.

build.gradle
```diff
@@ -36,6 +36,7 @@ dependencies {
     implementation 'androidx.constraintlayout:constraintlayout:2.1.4'
     implementation 'com.journeyapps:zxing-android-embedded:3.4.0' // QR Code scanner
     implementation 'com.fasterxml.jackson.core:jackson-databind:2.11.1' // used as JSON library
+    implementation 'com.google.code.gson:gson:2.10.1' // used as JSON library
```
Hocuri, line 39: Hm, so we need one more JSON library? jackson-databind doesn't work? Can we use gson instead of jackson-databind, in order to not increase the number of dependencies?

src/com/b44t/messenger/rpc/Rpc.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        CompletableFuture<JsonElement> future = requestFutures.remove(response.id);
+        if (response.error != null) {
+            future.completeExceptionally(new RpcError(response.error.toString()));
+        } else if (response.result != null) {
+            future.complete(response.result);
+        } else {
+            future.completeExceptionally(new RpcError("Got JSON-RPC response witout result or error: " + jsonResponse));
+        }
```
Hocuri, line 41: There's a lot of code changing exceptions from one form to another here, what if we replaced this whole block with `future.complete(response);`, used `CompletableFuture<Response>` instead of `CompletableFuture<JsonElement>`, and let the callers (like `getResult()`) extract the error on their own.

Also, creating a Throwable is always expensive, since it captures a stack trace. Idk how expensive excactly and if it might be noticeable in our code.

src/com/b44t/messenger/rpc/Rpc.java
```diff
@@ -0,0 +1,120 @@
...
+
+    public void start() {
+        new Thread(() -> {
+                while (true) {
+                    try {
+                        processResponse();
+                    } catch (JsonSyntaxException e) {
+                        e.printStackTrace();
+                    }
+                }
```
Hocuri, line 52:
```suggestion
            while (true) {
                try {
                    processResponse();
                } catch (JsonSyntaxException e) {
                    e.printStackTrace();
                }
            }
```

Hocuri (commented):
I'm wondering if we should do some measurements before merging this (how much bigger does the APK get? How much longer does a native call take now that we have to serialize and deserialize things? How much does stuff like `dc_get_http_response` get faster since we can return the result as JSON instead of needing multiple follow-up calls like `dc_http_response_get_mimetype`?).

build.gradle
adbenitez: if you grep for gson, you will find that we already use it in some part of the code, it seems it is a dependency of some of our dependencies already, I had to explicitly add this because it seems the version used by the dependency was old and didn't have a needed method.

in a follow-up PR we should try to only use one of gson, jackson and org.json (the 3 are used at the moment in existing code besides this PR)

src/com/b44t/messenger/rpc/Rpc.java
adbenitez: some legacy thing I think, in the past the RPC used to send unsolicited notifications with events so "notifications" here means incoming events that come without the client requesting them to the RPC server, we now have a new api to manually poll for events

src/com/b44t/messenger/rpc/Rpc.java
adbenitez: all the callers will want to extract the error/result so it would be too much of a burden and repeated code, they have to do this anyway, and if there is an error they will want to print it in the logs, because it will be a programming error (ex. bad usage of API)

adbenitez: (this are not incoming events that could become a bunch of errors, they are only responses to API calls, so if you call a function and it throws an exception you need to handle it anyway)

Hocuri: But checking for response.result != null is the same amount of code as catching the error

src/com/b44t/messenger/rpc/Rpc.java
Hocuri: Would be nice to either replace `notification` with `event` or remove this block.

adbenitez: done

r10s:
> I'm wondering if we should do some measurements before merging this (how much bigger does the APK get? How much longer does a native call take now that we have to serialize and deserialize things? How much does stuff like dc_get_http_response get faster since we can return the result as JSON instead of needing multiple follow-up calls like dc_http_response_get_mimetype?).

yeah, that'd be interesting, however, in general, for non-time-critical things, jsonrpc probably makes sense anyway.

`dc_get_http_response` would be a perfect example to test that out in this pr, btw - it is not time critical, not overall too critical, has quite some overhead with ffi

for adding accounts, not sure, i would not touch that for now.

also, other than stated at https://github.com/deltachat/deltachat-ios/issues/1847 "the aim is to start by using JSON-RPC for any new API" - i am not sure if that is really the aim for now. eg. if there comes a new method to dc_msg_t, it probably makes sense to add that to ffi still. also, for reading eg. a single record, it has to be proven if that is faster or much easier i'd say.

i see this jsonprc more as something that may sneak in, if it turns out that things are easier and faster with that, this is the way automatically. iirc, jsonrpc could also get a bit more mature before being used everywhere,  iirc there is no mapping between ffi and jsonrpc, low-level things and documentation are also missing. also from that perspective, there is no need to hurry.

adbenitez:
> `dc_get_http_response` would be a perfect example to test that out in this pr, btw - it is not time critical, not overall too critical, has quite some overhead with ffi

ok, done :ballot_box_with_check:

adbenitez:
> also, other than stated at [deltachat/deltachat-ios#1847](https://github.com/deltachat/deltachat-ios/issues/1847) _"the aim is to start by using JSON-RPC for any new API"_ - i am not sure if that is really the aim for now. eg. if there comes a new method to dc_msg_t, it probably makes sense to add that to ffi still.

yip for something like a new method to dc_msg_t wouldn't make sense, mean it like for new classes etc.

r10s:
ftr, as the question of size popped up above: with 1.37, we have some measurements:

> android1.37 is 11.6 mb larger than android1.36 (+24%) :/
checked the reasons a bit:
> plus 6.8 mb for jsonrpc in core (tested by leaving out "--features jsonrpc")
> plus 4.7 mb other core growings - updated dependecies? idk (tested by compiling 1.37 against core112.8
> plus <0.1 mb growing java code
i find that pretty surprisingly, compared to other features we added recently (iroh) -- maybe some dependencies are doubled by jsonrpc or so?

r10s:
> I'm wondering if we should do some measurements before merging this 

ftr, there are some measurements at https://github.com/deltachat/deltachat-core-rust/issues/4463 now

PR #2482, opened 2023-02-28
scripts/ndk-make.sh: unset RUSTFLAGS

link2xt:
Common way to configure `mold` is to set

    RUSTFLAGS="-C linker=clang -C link-arg=-fuse-ld=/usr/bin/mold"

This breaks cross-compilation, as wrong linker is used.


r10s:
> This breaks cross-compilation, as wrong linker is used.

not sure about the comment -  "This" refers to the usage of mold in combination with unset RUSTFLAGS?

or is cross-compilation broken with this pr? (which would be bad) (however, i am just testing #2483 that included this change, so i think, things are fine)

PR #2485, opened 2023-03-03
scripts/ndk-make.sh: enable debug builds explicitly with --debug

link2xt:
Currently F-Droid builds split architecture .apks
by invoking ndk-make.sh with architecture argument. This results in debug builds of the core.

This change makes debug builds explicit.

### DIFF
--- a/scripts/ndk-make.sh
+++ b/scripts/ndk-make.sh
@@ -6,6 +6,11 @@
 #     scripts/ndk-make.sh arm64-v8a
 #
 # Possible values are armeabi-v7a, arm64-v8a, x86 and x86_64.
+#
+# To build the core in debug mode, run with "--debug" argument in the beginning:
+#
+#     scripts/ndk-make.sh --debug arm64-v8a
+#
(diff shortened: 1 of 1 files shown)


Hocuri (approved):

scripts/ndk-make.sh
```diff
@@ -66,6 +71,23 @@ export CARGO_TARGET_X86_64_LINUX_ANDROID_LINKER="$TOOLCHAIN/bin/x86_64-linux-and
 
 export RUSTUP_TOOLCHAIN=$(cat "$(dirname "$0")/rust-toolchain")
 
+if test "$1" = "--debug"; then
+  echo Fast, partial, slow debug build. DO NOT UPLOAD THE APK ANYWHERE.
```
Hocuri, line 75:
```suggestion
  echo Quick debug build that will produce a slower app. DO NOT UPLOAD THE APK ANYWHERE.
```
"Partial" is not true here anymore

PR #2486, opened 2023-03-03
[wip] geckoview experiments

r10s:
see commit messages for details, this pr does not yet sandboxes xdc or hardens geckoview otherwise.

- [x] Build the thing. EDIT: yeah, it builds (283mb instead of 56mb) 
- [x] Open basic GeckoView. EDIT: after disabling Workmanager for now (see commit messages), it starts
- [ ] Make geckoview get files from .xdc, this required sth. as `shouldInterceptRequest`, cmp. https://github.com/mozilla/geckoview/issues/92, https://bugzilla.mozilla.org/show_bug.cgi?id=1721579 ... puhh, that seems to be quite some work of understanding and research, seems a WebExtension is needed for that. or a server.   
EDIT: there is [onLoadRequest()](https://mozilla.github.io/geckoview/javadoc/mozilla-central/org/mozilla/geckoview/GeckoSession.NavigationDelegate.html#onLoadRequest(org.mozilla.geckoview.GeckoSession,org.mozilla.geckoview.GeckoSession.NavigationDelegate.LoadRequest)) ([so](https://stackoverflow.com/questions/55922979/is-there-a-shouldoverrideurlloading-in-geckoview)) - but that seems to allow returning a bool only, not the data 
EDIT: [`onLoadError`](https://mozilla.github.io/geckoview/javadoc/mozilla-central/org/mozilla/geckoview/GeckoSession.NavigationDelegate.html#onLoadError(org.mozilla.geckoview.GeckoSession,java.lang.String,org.mozilla.geckoview.WebRequestError)) allows returning an URL, including [`data:text/html`](https://searchfox.org/mozilla-central/rev/c244b16815d1fc827d141472b9faac5610f250e7/mobile/android/geckoview_example/src/main/java/org/mozilla/geckoview_example/GeckoViewActivity.java#2318).
- [ ] sandbox/harden, target TODOs

nb: geckoview documentation is at https://mozilla.github.io/geckoview/javadoc/mozilla-central/org/mozilla/geckoview/package-summary.html

### DIFF
--- a/assets/geckoview/webxdc/background.js
+++ b/assets/geckoview/webxdc/background.js
@@ -1,0 +1,21 @@
+"use strict";
+
+function setupRedirect(fromUrl, redirectUrl) {
+  browser.webRequest.onBeforeRequest.addListener(
+    details => {
+      console.log(`Extension redirects from ${fromUrl} to ${redirectUrl}`);
+      return { redirectUrl };
+    },
+    { urls: [fromUrl] },
+    ["blocking"]
+  );
+}
+
+setupRedirect(
+  "*://*/webxdc.js",
+  browser.runtime.getURL("webxdc.js")
+);
+setupRedirect(
+  "<all_urls>",
+  browser.runtime.getURL("webxdc.html")
+);
--- a/assets/geckoview/webxdc/manifest.json
+++ b/assets/geckoview/webxdc/manifest.json
@@ -1,0 +1,22 @@
+{
+  "name": "webxdc-extension",
+  "description": "Intercepts requests to XDC resources",
+  "manifest_version": 2,
+  "version": "1",
+  "browser_specific_settings": {
+    "gecko": {
+      "id": "webxdc@delta.chat"
+    }
+  },
+  "background": {
+    "scripts": ["background.js"]
+  },
+  "permissions": [
+    "geckoViewAddons",
+    "nativeMessaging",
+    "webRequest",
+    "webRequestBlocking",
+    "<all_urls>"
+  ],
+  "web_accessible_resources": ["webxdc.js", "webxdc.html"]
+}
--- a/assets/geckoview/webxdc/webxdc.html
+++ b/assets/geckoview/webxdc/webxdc.html
@@ -1,0 +1,1 @@
+Hello from extension!
--- a/assets/geckoview/webxdc/webxdc.js
+++ b/assets/geckoview/webxdc/webxdc.js
@@ -1,0 +1,6 @@
+"use strict";
+
+// TODO inject webxdc API here
+document.body.textContent += ",extension-was-here";
+document.body.style.border = "5px solid red";
+alert("webxdc script loaded");
--- a/build.gradle
+++ b/build.gradle
@@ -97,7 +97,7 @@ android {
 
     defaultConfig {
         versionCode 651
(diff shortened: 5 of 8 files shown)


hpk42:
ftr, https://stackoverflow.com/questions/75508820/how-to-disable-webrtc-in-geckoview/75602901#75602901 has a suggestion how to avoid using WebRTC -- this could reduce the size somewhat significantly

Simon-Laux:
Bildschirmfoto 2023-03-04 um 22 22 51

not sure what these playservice files are, but I bet we don't need them nor the features they are needed for, "fido" sounds like yubikey stuff which we don't need inside of webxdc

Simon-Laux:
### Regarding size of geckoview

> This is sitll strange, because 2 years ago it was reported that libxul.so is 3-10 M: [https://bugzilla.mozilla.org/show_bug.cgi?id=1727150](https://bugzilla.mozilla.org/show_bug.cgi?id=1727150)

No, they talk about a 10mb increase, it actually is zipped around 65mb.

Building gecko view from source also gave me a 130mb file, the same size as I found in the apk above.

I found out that this is the normal mode which contains `Glean`, which is firefox's telemetry library, we can build without it via `--enable-geckoview-lite` and there should also be `Lite` prebuilds already provided by them (though I did not really find them, their omni builds seemed smaller, but docs say omni builds are with telemetry if I remember correctly).


Another easy thing we could disable is HTTP media streaming, it uses exoplayer2, so I don't know if there are really any big size differences, but we could try: `MOZ_ANDROID_HLS_SUPPORT=false`.


### Regarding the custom scheme

I looked into quite some things, but here are the most promising angles so far IMO:
- **nsIProtocolHandler**: someone made a separate lib that got loaded by FF to extend it with a new schema (it's from 2009, so not sure if it still works that way, the wiki link to the method is broken)
    - https://stackoverflow.com/questions/864443/implementing-nsiprotocolhandler-with-delphi
    - https://sourceforge.net/projects/xxm/
- we could look at how internal schemes such as `view_source:`  are implemented
    - when we implement/add a custom one we can either look how we can create an api for it like the ios webview / electron api for registering new URI schemas
    - The other option would be implementing an `webxdc:` schema in native Gecko code and make it speak directly with dc-core somehow (Unix sockets, IPC, ports, somehow also cross-linking to dc-core?).

#### Candidates that won't work for URI scheme:

- `geckoSession.loadString`, `geckoSession.loadData`, `GeckoSession.Loader`
    - can be used to load the initial page, but not more
    - `GeckoSession.Loader` can set headers, that might come in useful

- WebResponse, `ContentDelegate#onExternalResponse(GeckoSession, WebResponse)`
    - the only method that gets a webresponse is the external response method, which is used for downloading files in the geckoview example project

- `onLoadRequest`
    - only usable for allow/deny of web requests, we **can not replace** response there.

- webextensions `protocol_handlers`
    - as far as I can tell from these links (https://stackoverflow.com/questions/22729686/set-custom-protocol-handler-in-firefox, https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/protocol_handlers) this is only for schemes that open something, not for loading a resource in into the browser. ;(

r10s:
closing as it we won't go for that soon - otherwise we can always pick this up

PR #2489, opened 2023-03-05
Add "All Media" option to global menu

adbenitez:
add a global profile to browse all media/documents/attachments, games etc. really useful to find attachments when user can't remember in which chat was sent, also useful for quickly reviewing and removing big attachments to save space

(note: most stuff was already in place to support this, thanks to that fellow collaborator from the past!)

### DIFF
--- a/src/org/thoughtcrime/securesms/ProfileActivity.java
+++ b/src/org/thoughtcrime/securesms/ProfileActivity.java
@@ -97,17 +97,22 @@ public class ProfileActivity extends PassphraseRequiredActionBarActivity
 
     setSupportActionBar(this.toolbar);
     ActionBar supportActionBar = getSupportActionBar();
-    supportActionBar.setDisplayHomeAsUpEnabled(false);
-    supportActionBar.setCustomView(R.layout.conversation_title_view);
-    supportActionBar.setDisplayShowCustomEnabled(true);
-    supportActionBar.setDisplayShowTitleEnabled(false);
-    Toolbar parent = (Toolbar) supportActionBar.getCustomView().getParent();
-    parent.setPadding(0,0,0,0);
-    parent.setContentInsetsAbsolute(0,0);
+    if (isGlobalProfile()) {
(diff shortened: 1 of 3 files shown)


adbenitez:
btw, the "help" menu option could be removed since it is already available in the settings

src/org/thoughtcrime/securesms/ProfileActivity.java
```diff
@@ -97,17 +97,21 @@ protected void onCreate(Bundle bundle, boolean ready) {
...
-    Toolbar parent = (Toolbar) supportActionBar.getCustomView().getParent();
-    parent.setPadding(0,0,0,0);
-    parent.setContentInsetsAbsolute(0,0);
-
-    titleView = (ConversationTitleView) supportActionBar.getCustomView();
-    titleView.setOnBackClickedListener(view -> onBackPressed());
-    titleView.setOnClickListener(view -> onEnlargeAvatar());
+    if (isGlobalProfile()) {
+      supportActionBar.setDisplayHomeAsUpEnabled(true);
+      supportActionBar.setHomeActionContentDescription(getString(R.string.back));
```
r10s, line 102: is that really needed? i would expect that the default text just reads "back". at least we are not doing that at all the other placed where we set the back-button.

maybe better remove this line here - and if we really think, it is important, do a separate pr and set that at all places.

r10s (approved):
nice! 

such an impact with such few lines of changes - interestingly,  the string "All Media" was even already there :)

only a few nitpicks, see comments below. for the "help" button - yip, i also think that can be removed, "help" is pretty common in the settings on android, but we can also do that in a subsequent pr.

src/org/thoughtcrime/securesms/ProfileActivity.java
adbenitez: it is needed because android sets "navigate up" as content description, the other back buttons are part of our custom view

PR #2491, opened 2023-03-06
Add fallback implementations for sendmmsg, recvmmsg and getauxval

link2xt:
These functions are not available in related NDK,
so we provide a fallback implementation for them.


flub (approved):

r10s (approved):
great, with that, android compiles! :)

PR #2493, opened 2023-03-08
add multi-device-setup

r10s:
requires https://github.com/deltachat/deltachat-core-rust/pull/4007 to be checked out
counterpart of https://github.com/deltachat/deltachat-ios/pull/1822

- [x] wrapper
- [x] basic UI
- [x] show progress, hide QR code once scanned (this is also nice feedback)
- [x] keep transfer screens on (i think, the show screen is okay already, not sure about receiver)
- [x] add a hint not to leave the app / the transfer screens
- [x] add permanent notification for sender
- [x] add permanent notification for receiver
- [x] create a dedicated activity also for the receiver
- [x] use "X" in the upper left corner instead of "<-" to make visually clear, things are aborted when leaving
- [x] protect sending backups by system credentials (as for export-backup or export-keys)
- [x] add a hint that QR codes copied to clipboard are invalidated on leaving the activity (in the "Abort transfer?" alert, only in case a code was copied)
- [x] improve wording and layout, make things translatable
- [x] add a real progress bar
- [ ] fix activity leave bug: android7 -> add another device -> home screen -> start delta via icon (opens chatlist, that is already wrong - we should stay in add another device) -> back button -> start delta via icon -> does no longer start sth. at all  EDIT: other startup screen had the same issue, see comments and code in `PassphraseRequiredActionBarActivity` - we should rebase on that or mimic its behavior

sender:

 &nbsp; 

receiver:

 &nbsp;


r10s:
hm, maybe better to  put 1-3 directly below each other

adbenitez:
> hm, maybe better to put 1-3 directly below each other

was thinking exactly that while looking the picture

res/values/strings.xml
```diff
@@ -465,6 +465,28 @@
...
+    <!-- Multi Device -->
+    <string name="multidevice_title">Add Another Device</string>
+    <string name="multidevice_same_network_hint">Make sure both devices are on the same Wi-Fi or network</string>
+    <string name="multidevice_install_dc_on_other_device">Install and start Delta Chat 1.36 or newer on your other device (https://get.delta.chat)</string>
+    <string name="multidevice_tap_scan_on_other_device">During setup, tap “Scan QR Code” and scan the code shown here</string>
+    <!-- Shown inside a "QR code card" with very limited space; please formulate the text as short as possible therefore. The placeholder will be replaced by name and/or address eg. "Scan to set up another device for Alice (alice@example.org)" -->
+    <string name="multidevice_qr_subtitle">Scan to set up another device for %1$s</string>
+    <string name="multidevice_receiver_scanning_ask">Copy the account from the other device to this device? Both devices will work independently afterwards.</string>
+    <string name="multidevice_exporting_database">Exporting database…</string>
+    <string name="multidevice_creating_collection">Creating collection…</string>
```
r10s, line 478: not sure about `Creating collection / Collection created / Receiving Collection / Collection received` which is "conceptually the same thing as creating a tar file" (@dignifiedquire)

on alternative idea was `Preparing content` - but there `Receiving content` does not really work (as afterwards it takes minutes to get the real content)

`Preparing your account data for transfer to second device`,  `Preparing account data for export` was another idea, but a but long (i think, it should be really short as of now).

my alternative idea was `Creating index / Index created / Receiving index / Index received`, btw, as "Index" seems to be a more known thing and easier to translate - still not sure if that would fit :)

res/values/strings.xml
```diff
@@ -465,6 +465,28 @@
     <!-- the explanation is shown (1) as a modal dialog with the buttons "Cancel" and "Continue" as well as (2) as a device message -->
     <string name="aeap_explanation">You changed your email address from %1$s to %2$s.\n\nIf you now send a message to a verified group, contacts there will automatically replace the old with your new address.\n\nIt\'s highly advised to set up your old email provider to forward all emails to your new email address. Otherwise you might miss messages of contacts who did not get your new address yet.</string>
 
+
+    <!-- Multi Device -->
+    <string name="multidevice_title">Add Another Device</string>
```
r10s, line 470: "Add Second Device" may also be a handy name, even if it is little less fitting in case of a third device.

"Link Device" was sorted out somehow, as this implies a bit that one device depends on the other, as of whatsapp or signal.

res/values/strings.xml
adbenitez: `Creating collection…` should be absolutely changed for something else, maybe:
* Preparing for transfer…
* Getting ready for transfer…
* Preparing account for transfer…

res/values/strings.xml
adbenitez: I think it is fine as it is, but if you are looking to change it, "Add Second Device"  and "Add New Device" look fine also

dignifiedquire: What about `Copy to new device`?

res/values/strings.xml
r10s: what would be "receiving colllection"/"collection received" then? "preparing to receive..."/"receiving prepared"?

adbenitez (approved):

res/values/strings.xml
r10s: "Copy to new device" sounds nice at a first glance, however, somehow i had the feeling it makes things less clear and less  catchy.

thinking a bit longer about it, i think, it is about the changed subject - "Add Device" has the subject "Device" - "Copy to new Device" has only an implicit subject.

res/values/strings.xml
r10s: maybe just say "Preparing account..."/"Account prepared." on both sides? and also change to "Export account." the term "Account" is pretty well used already and we would not add a new term.

src/org/thoughtcrime/securesms/WelcomeActivity.java
```diff
@@ -355,6 +362,13 @@ public void handleEvent(@NonNull DcEvent event) {
 
         if (eventId== DcContext.DC_EVENT_IMEX_PROGRESS ) {
             long progress = event.getData1Int();
+            if (progressDialog == null || notificationController == null) {
+                // IMEX runs in BackupTransferActivity
+                if (progress == 1000) {
+                    finish();  // transfer done - remove ourself from the activity stack (finishAffinity is available in API 16, we're targeting API 14)
+                }
+                return;
+            }
```
Hocuri, line 371: Is my understanding correct here?

So, when we do IMEX in BackupTransferActivity, the WelcomeActivity stays in the background so that the user can go back if something doesn't work, and when it's done, we call finish().

But then, why do we need `progressDialog == null || notificationController == null`? Wouldn't `progressDialog == null` be enugh?

src/org/thoughtcrime/securesms/qr/BackupProviderFragment.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    @Override
+    public void onPrepareOptionsMenu(Menu menu) {
+        menu.findItem(R.id.copy).setVisible(qrImageView.getVisibility() == View.VISIBLE);
+        super.onPrepareOptionsMenu(menu);
+    }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
```
Hocuri, line 113:
```suggestion
    public boolean onOptionsItemSelected(@NonNull MenuItem item) {
```

res/values/strings.xml
r10s: finally, after some one-to-one conversations, decided to go for "Second Device" at https://github.com/deltachat/deltachat-android/pull/2493/commits/5f271d8d036361f2593952e91944ba404385bd6c , this is a but catchier and  was also the intuitive wording used in core and on desktop first; let's give it a try, i do not think it make huge differences :)

res/values/strings.xml
r10s: go for that for now, see https://github.com/deltachat/deltachat-android/pull/2493/commits/46871bcbc1365b83d683ddd5b823ddcc1ecd5130

res/values/strings.xml
```diff
@@ -465,6 +465,35 @@
...
+    <!-- Shown beside progress bar, stay short -->
+    <string name="multidevice_exporting_account">Exporting account…</string>
+    <!-- Shown beside progress bar, stay short -->
+    <string name="multidevice_preparing_account">Preparing account…</string>
+    <!-- Shown beside progress bar, stay short -->
+    <string name="multidevice_account_prepared">Account prepared.</string>
+    <!-- Shown beside progress bar, stay short -->
+    <string name="multidevice_waiting_for_receiver">Waiting for receiver…</string>
+    <!-- Shown beside progress bar, stay short -->
+    <string name="multidevice_connecting">Connecting…</string>
```
Hocuri, line 488: Nit: So, we have two "Connecting…" strings - is this really necessary?

res/values/strings.xml
```diff
@@ -465,6 +465,35 @@
...
     <string name="aeap_explanation">You changed your email address from %1$s to %2$s.\n\nIf you now send a message to a verified group, contacts there will automatically replace the old with your new address.\n\nIt\'s highly advised to set up your old email provider to forward all emails to your new email address. Otherwise you might miss messages of contacts who did not get your new address yet.</string>
 
+
+    <!-- Multi Device -->
+    <!-- "Second Device" can also be translated as "Another Device", if that is catchier in the destination language. However, make sure to use the term consistently. -->
+    <string name="multidevice_title">Add Second Device</string>
+    <string name="multidevice_same_network_hint">Make sure both devices are on the same Wi-Fi or network</string>
+    <string name="multidevice_install_dc_on_other_device">Install and start Delta Chat 1.36 or newer on your other device (https://get.delta.chat)</string>
+    <string name="multidevice_tap_scan_on_other_device">During setup, tap “Scan QR Code” and scan the code shown here</string>
```
Hocuri, line 474: I don't think users will understand what "During setup" means.

My favorite would be `Install Delta Chat, tap "Scan QR Code" ...`. This 1. hints at the fact that it needs to be done on another device 2. hints at the fact that it has to be done on the screen that turns up after installation.

Hocuri (commented):
I didn't finish reviewing yet, but here come the first few comments - the comment on strings.xml is the only really discussion-worthy one, the others are just things I noticed but that would be fine as-is.

res/values/strings.xml
r10s: idea is to underline that you have to scan the code from inside delta chat

Hocuri: Another possibility would be `Open Delta Chat, ...`, then. I still like `Install Delta Chat, ...` a very small bit better, though.

Hocuri (commented):
I finished my review - apart from discussing out the wording, this LGTM :tada:

jni/dc_wrapper.c
```diff
@@ -39,7 +39,7 @@ int sendmmsg(int sockfd, struct mmsghdr *msgvec, unsigned int vlen,
 int recvmmsg(int sockfd, struct mmsghdr *msgvec, unsigned int vlen,
              int flags, struct timespec *timeout)
 {
-    if (flags != -1) {
+    if (flags != 0) {
```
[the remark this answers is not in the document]
r10s: [`recvmsg()` is a standard unix function](https://man.freebsd.org/cgi/man.cgi?query=recvmmsg&manpath=FreeBSD+11.0-RELEASE), that is not available in api16, but needed by core.

therefore, @link2xt added an own implementation. with recent core changes, it turns out that the handling of `flags` was a bit different. this is fixed by [this change](https://github.com/deltachat/deltachat-android/pull/2493/commits/cf571e1964c618dc5905f08234c3f853cce414f2)

build.gradle
```diff
@@ -97,7 +97,7 @@ android {
 
     defaultConfig {
         versionCode 649
-        versionName "1.35.1"
+        versionName "1.35.2"
```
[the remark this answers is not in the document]
r10s: yes, we created betas from this pr.

res/layout/backup_provider_fragment.xml
```diff
@@ -0,0 +1,110 @@
...
+		android:layout_margin="16dp"
+		/>
+
+	<LinearLayout
+		android:id="@+id/top_text"
+		android:layout_width="match_parent"
+		android:layout_height="wrap_content"
+		android:layout_margin="16dp"
+		android:orientation="vertical"
+		android:visibility="gone">
```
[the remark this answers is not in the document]
r10s: cool, i was always playing around with `android:visibility` during layout :)

res/values/strings.xml
r10s: no, you are right, usually we duplicate strings only if there is a need. changed that by recent commit.

src/org/thoughtcrime/securesms/WelcomeActivity.java
r10s: maybe, it would be sufficient, yes. it is just to be safe in case there are some maybe weird code paths followed before by user tapping around widely :) user may have eg. started + aborted backup or configure before - and tbh, i cannot oversee all cases :)

res/values/strings.xml
r10s: i changed that with the recent commit to:

Screenshot 2023-03-24 at 11 12 52

Hocuri (approved):

src/org/thoughtcrime/securesms/qr/BackupTransferActivity.java
```diff
@@ -0,0 +1,238 @@
...
+                        }
+                    }
+                } catch (Exception e) {
+                    e.printStackTrace();
+                }
+            }).start();
+        }
+    }
+
+    protected <T extends Fragment> T initFragment(@IdRes int target,
```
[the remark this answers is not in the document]
r10s: `initFragment` will get removed by https://github.com/deltachat/deltachat-android/pull/2508/commits/e7f9d6d87a337989eea32bc3324c57c90a7ea86e

PR #2496, opened 2023-03-13
set background image for QrShowFragment

adbenitez:
the current "QR Show" fragment looks a bit ugly/empty with the white background, this PR sets the nicer default chat background

### DIFF
--- a/res/layout/qr_show_fragment.xml
+++ b/res/layout/qr_show_fragment.xml
@@ -1,14 +1,21 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+<RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
 	android:layout_width="match_parent"
-	android:layout_height="match_parent"
-	android:orientation="vertical">
+	android:layout_height="match_parent">
+
+    <org.thoughtcrime.securesms.components.ScaleStableImageView
+        android:id="@+id/background"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        tools:ignore="ContentDescription"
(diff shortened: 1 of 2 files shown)


r10s:
a matter of taste a bit, but yes, more fancy  💅 i think, i like it :)

for dark theme, did you consider to use the dark default background?

ftr, this is how it looks like:

Screenshot 2023-03-14 at 12 11 13

WofWca:
> for dark theme, did you consider to use the dark default background?

Be aware that some QR readers don't recognize inverted-color QRs. See, for example, https://gitlab.torproject.org/tpo/applications/tor-browser/-/issues/41049

r10s:
> Be aware that some QR readers don't recognize inverted-color QRs. See, for example, https://gitlab.torproject.org/tpo/applications/tor-browser/-/issues/41049

yip, we are aware of that, the QR code is not changed and stays black on white, including a white border. it is only about the background image, that also exist in a dark variant

adbenitez:
> for dark theme, did you consider to use the dark default background?

it automatically changes to the dark background, we use these special android folders that automatically selects the right background for light/dark modes and device orientation

r10s:
this seems not to be sufficient. i could not manage to see the background in dark mode - neither by an os supporting darkmode nor by switching manually. the chat-background, however, follows all these modes, so we should probably peed what is done there.

adbenitez:
@r10s mmmh, odd, in my phone it works even with the QrShowFragment in display, buy I think I know what is the difference: `DynamicTheme` but it is hard for me to fix if it is not broken in my device :/

r10s:
i can have a look at that at some later point

adbenitez:
never mind, it turned out that the folder stuff I mentioned earlier I did it in DeltaLab :man_facepalming:  so in official DC it is still the old way setting background_hd_dark manually, will push a fix

r10s (approved):
lgtm now

PR #2497, opened 2023-03-13
tweak main menu and settings

adbenitez:


r10s (approved):
nice, as bespoken.

ftr, the "connectivity" was at its position because it can be regarded as a non-setting, similar to "help". however, i think, the expected position of an "advanced" item outweights this.

similar for "help" in the main menu: this removal moves "settings" to a more expected position; "help" itself is pretty much known as being one of the last items in the settings.

so, let's give this a try

PR #2503, opened 2023-03-21
add app icon to webxdc info messages

### LINKED ISSUE #2425: add webxdc icon to info message
left of webxdc info messages, the webxdc icon should be shown, padding and size similar to an emoji.

see desktop 1.33.1 to get the idea:

Screenshot 2022-11-13 at 23 38 15

cmp. https://github.com/deltachat/deltachat-ios/issues/1745, https://github.com/deltachat/deltachat-desktop/pull/2994

adbenitez:
close #2425

### DIFF
--- a/res/layout/conversation_item_update.xml
+++ b/res/layout/conversation_item_update.xml
@@ -35,6 +35,7 @@
                 android:layout_width="18sp"
                 android:layout_height="18sp"
                 android:layout_marginEnd="2dp"
+                android:layout_marginRight="2dp"
                 android:layout_gravity="center_vertical"
                 android:scaleType="centerCrop"
                 android:visibility="gone" />
--- a/res/layout/conversation_item_update.xml
+++ b/res/layout/conversation_item_update.xml
@@ -22,19 +22,36 @@
         android:gravity="center"
         android:orientation="vertical">
 
-        <!--Links should be clickable, see https://github.com/deltachat/deltachat-android/issues/1546-->
-        <org.thoughtcrime.securesms.components.emoji.EmojiTextView
-            android:id="@+id/conversation_update_body"
-            style="@style/Delta.Text.UpdateHeader"
+        <LinearLayout
             android:layout_width="wrap_content"
             android:layout_height="wrap_content"
-            android:autoLink="web"
-            android:gravity="center"
-            android:textColor="?attr/conversation_item_update_text_color"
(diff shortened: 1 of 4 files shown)


src/org/thoughtcrime/securesms/ConversationUpdateItem.java
```diff
@@ -80,6 +88,22 @@ public DcMsg getMessageRecord() {
   }
 
   private void setGenericInfoRecord(DcMsg messageRecord) {
+    DcMsg parentMsg = messageRecord.getParent();
```
Hocuri, line 91: This does two database requests: one for finding the parent, one for loading it from the database -> probably making scrolling less smooth.

Maybe we should cache the parent message in DcMsg, or do the loading asynchronously, or both.

res/layout/conversation_item_update.xml
```diff
@@ -22,19 +22,36 @@
...
             android:paddingBottom="3dp"
-            tools:text="Gwen Stacy added to group" />
+            android:gravity="center"
+            android:orientation="horizontal">
+
+            <androidx.appcompat.widget.AppCompatImageView
+                android:id="@+id/app_icon"
+                android:layout_width="18sp"
+                android:layout_height="18sp"
+                android:layout_marginEnd="2dp"
```
Hocuri, line 37:
```suggestion
                android:layout_marginEnd="2dp"
                android:layout_marginRight="2dp"
```
for Android 4.1

src/org/thoughtcrime/securesms/ConversationUpdateItem.java
r10s: we can avoid the additional request by checking for `dc_msg_get_info_type(infoMsg) == DC_INFO_WEBXDC_INFO_MESSAGE` - and only if that condition is true, load the parent.

this is also what ios and probably also desktop are doing.

adbenitez: > we can avoid the additional database request by checking for [`dc_msg_get_info_type() == DC_INFO_WEBXDC_INFO_MESSAGE`](https://c.delta.chat/classdc__msg__t.html#ab3eeb59143e67fda31d3648bd775f77b)- and only if that condition is true, load the parent.

done :heavy_check_mark:

Hocuri (approved):
Woop woop :partying_face:

r10s (approved):
nice!

PR #2506, opened 2023-03-22
Fix sharing, again

### LINKED ISSUE #2505: Direct sharing / chat shortcuts are broken
- Android version: 11+, I think also 10
- Delta Chat version: Current master
- Actual behavior: An empty chat with a hastag as the profile image is shown.

Either: Run `./gradlew connectedFatDebugAndroidTest` on the command line to run the automatic tests.
Or: Click "Share" in another app, then click one of the Delta Chat chats that are sometimes recommended in the first line by the Android system.
Or: Long-click on Delta Chat in a supporting launcher and click on one of the proposed chats.

- Remark: Running `./gradlew connectedFatDebugAndroidTest` on the command line to run the automatic UI tests is always worth it, if you are lucky it just works. If not, chances are that you just found a regression, which happens reeeally often in the sharing code.

Probably introduced by 8962476a.

Hocuri:
Fix #2505 and make the automated tests pass again by simply reverting https://github.com/deltachat/deltachat-android/commit/8962476a21c67fd943a496f759422a4b2e6cd3e3.

The second commits fixes an NPE that somehow sneaked into the tests.

The "manual sharing tests" don't pass again, esp. https://github.com/deltachat/interface/blob/master/user-testing/mailto-links.md#regression-test-for-the-issue-fixed-by-1770 still doesn't work, but this is for another PR.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -158,11 +158,11 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
 {
   private static final String TAG = ConversationActivity.class.getSimpleName();
 
-  public static final String ACCOUNT_ID_EXTRA        = TAG + ".account_id";
-  public static final String CHAT_ID_EXTRA           = TAG + ".chat_id";
-  public static final String FROM_ARCHIVED_CHATS_EXTRA = TAG + ".from_archived";
-  public static final String TEXT_EXTRA              = TAG + ".draft_text";
-  public static final String STARTING_POSITION_EXTRA = TAG + ".starting_position";
+  public static final String ACCOUNT_ID_EXTRA        = "account_id";
+  public static final String CHAT_ID_EXTRA           = "chat_id";
+  public static final String FROM_ARCHIVED_CHATS_EXTRA = "from_archived";
+  public static final String TEXT_EXTRA              = "draft_text";
+  public static final String STARTING_POSITION_EXTRA = "starting_position";
 
   private static final int PICK_GALLERY        = 1;
   private static final int PICK_DOCUMENT       = 2;
--- a/src/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -76,7 +76,7 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
   private static final String OPENPGP4FPR = "openpgp4fpr";
   private static final String NDK_ARCH_WARNED = "ndk_arch_warned";
   public static final String CLEAR_NOTIFICATIONS = "clear_notifications";
-  public static final String ACCOUNT_ID_EXTRA = TAG + ".account_id";
+  public static final String ACCOUNT_ID_EXTRA = "account_id";
 
   private final DynamicTheme    dynamicTheme    = new DynamicNoActionBarTheme();
   private final DynamicLanguage dynamicLanguage = new DynamicLanguage();
--- a/src/org/thoughtcrime/securesms/ShareActivity.java
+++ b/src/org/thoughtcrime/securesms/ShareActivity.java
@@ -261,7 +261,7 @@ public class ShareActivity extends PassphraseRequiredActionBarActivity implement
     Intent composeIntent;
     if (chatId != -1) {
       composeIntent = getBaseShareIntent(ConversationActivity.class);
-      composeIntent.putExtra(EXTRA_CHAT_ID, chatId);
+      composeIntent.putExtra(ConversationActivity.CHAT_ID_EXTRA, chatId);
       RelayUtil.setSharedUris(composeIntent, resolvedExtras);
       startActivity(composeIntent);
     } else {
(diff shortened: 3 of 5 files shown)


src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -158,11 +158,11 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
 {
   private static final String TAG = ConversationActivity.class.getSimpleName();
 
-  public static final String ACCOUNT_ID_EXTRA        = TAG + ".account_id";
-  public static final String CHAT_ID_EXTRA           = TAG + ".chat_id";
-  public static final String FROM_ARCHIVED_CHATS_EXTRA = TAG + ".from_archived";
-  public static final String TEXT_EXTRA              = TAG + ".draft_text";
-  public static final String STARTING_POSITION_EXTRA = TAG + ".starting_position";
+  public static final String ACCOUNT_ID_EXTRA        = "account_id";
```
adbenitez, line 161: why the solution is to make all EXTRA be the same in all classes instead of using the appropriate EXTRA of the target activity when creating an intent? by reading this PR I don't really see the place that were causing the issue (setting a wrong extra) this is not a good signal IMHO

Hocuri: Because I just reverted the commit that caused the bug.

The really correct solution would IMO be to have one definition of the variable in some central place, and then import it from everywhere. A wrong import shouldn't be able to cause a bug.

The cause of the bug is that `ShareActivity` is:
```java
      composeIntent = getBaseShareIntent(ConversationActivity.class);
      composeIntent.putExtra(EXTRA_CHAT_ID, chatId);
      RelayUtil.setSharedUris(composeIntent, resolvedExtras);
      startActivity(composeIntent);
```

which doesn't use ConversationActivity's "chat_id", but ShareActivity's.

adbenitez: > which doesn't use ConversationActivity's "chat_id", but ShareActivity's.

that is the real bug then

adbenitez: > A wrong import shouldn't be able to cause a bug.

putting all extras in a single class would do as well, but I think intent creator must be careful anyway to use the extras of the same class you are launching, it doesn't makes much sense that you use the extras of other classes

adbenitez:
how do I test this is fixed? I compiled it but opening the chats from the menu that appears in launcher when long-pressing, keeps opening an empty buggy chat :(

src/org/thoughtcrime/securesms/ConversationActivity.java
Hocuri: If we only fix this single bug, then I'd be afraid that https://github.com/deltachat/deltachat-android/commit/8962476a21c67fd943a496f759422a4b2e6cd3e3 introduced even more bugs which we'll then also release.

> I think intent creator must be careful anyway to use the extras of the same class you are launching

Well, obviously someone was not careful when using the wrong variable here, whenever possible one shouldn't have to rely on others or oneself being careful (I mean, that's what "software quality" usually boils down to, with writing tests, avoiding code duplication, ...)

---

I'm didn't get which solutions would be OK for you - would you go with just merging this PR if it fixes the bug? This PR plus using the correct variable there (I tried to keep the changes minimal and go back to a known-good state right before a release, but changing the variable here wouldn't be a problem)? This PR plus using the same variable from everywhere?

Hocuri:
That's really weird, I now tested both in the emulator and on a real device that this works. Maybe installing the app didn't work? I sometimes had the problem that I needed to first uninstall the app from the device before changes were actually applied.

Hocuri:
In https://github.com/deltachat/deltachat-android/issues/2505, I wrote 3 ways of reproducing the bug, do all 3 of them still fail for you with this PR?

adbenitez:
I compiled again this branch and tested and in fact it works, maybe I compiled the wrong branch the last time, sorry ;)

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -393,7 +393,8 @@ public void scrollToBottom() {
     }
 
     void setLastSeen(long lastSeen) {
-        getListAdapter().setLastSeen(lastSeen);
+        ConversationAdapter listAdapter = getListAdapter();
```
adbenitez, line 396: @r10s did a PR to fix this, you two should compare which solution is better and revert this or close #2512

src/org/thoughtcrime/securesms/ConversationActivity.java
adbenitez: yes, this PR is fine, but I think also fixing the original issue and use the right variable should be done in this PR as well

adbenitez (approved):
approving, but please first fix also the part that was using the variable from the wrong class (even if it will not cause an issue now that all vars have the same value)

src/org/thoughtcrime/securesms/ConversationFragment.java
Hocuri: Both seem to be fine, but @r10s' solution looks more "careful", so let's go with this one

PR #2510, opened 2023-03-24
fix leaving Delta Chat during backup transfer

### LINKED ISSUE #2507: leaving setup-second-device activity leads to weird app state
android7 -> add another device -> home screen -> start delta via icon (opens chatlist, that is already wrong - we should stay in add another device) -> back button -> start delta via icon -> does no longer start sth. at all

other startup screen had the same issue, see comments and code in PassphraseRequiredActionBarActivity - we should rebase on that or mimic its behavior

r10s:
this fixes leaving Delta Chat during backup transfer by removing other activities that would abort BackupTransfer by getting called with onNewIntent() otherwise.

moreover, the permantent notification is removed more reliably and possible crashes in  the permantent notification closing (races ...) are catched.

see code + comments for details.

successor of #2508 and #2493

closes #2507

### DIFF
--- a/src/org/thoughtcrime/securesms/ApplicationPreferencesActivity.java
+++ b/src/org/thoughtcrime/securesms/ApplicationPreferencesActivity.java
@@ -144,6 +144,8 @@ public class ApplicationPreferencesActivity extends PassphraseRequiredActionBarA
     Intent intent = new Intent(this, BackupTransferActivity.class);
     intent.putExtra(BackupTransferActivity.TRANSFER_MODE, BackupTransferActivity.TransferMode.SENDER_SHOW_QR.getInt());
     startActivity(intent);
+    overridePendingTransition(0, 0); // let the activity appear in the same way as the other pages (which are mostly fragments)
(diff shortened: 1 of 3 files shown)


src/org/thoughtcrime/securesms/ApplicationPreferencesActivity.java
```diff
@@ -144,6 +144,8 @@ public void showBackupProvider() {
     Intent intent = new Intent(this, BackupTransferActivity.class);
     intent.putExtra(BackupTransferActivity.TRANSFER_MODE, BackupTransferActivity.TransferMode.SENDER_SHOW_QR.getInt());
     startActivity(intent);
+    overridePendingTransition(0, 0);
```
r10s, line 147: this is to make the activity appear in the same way as the other pages (which are mostly fragments)

adbenitez (approved):
lgtm, didn't tried it yet tho

r10s:
> lgtm, didn't tried it yet tho

i just pushed beta 1.35.4 that contains this fix

r10s:
let's merge that in, if there as still issues, they can go to another pr

PR #2514, opened 2023-03-26
send backup progress tweaks

r10s:
this pr fixes some things wrt progress bars and crashes when the activity is left during preparing, see commits for details.

moreover, there was a recent request to add a confirmation dialog before the qr code is shown: not sure, how important that really is, however, it turns out, that we already show such a dialog for probably most security-aware ppl using a system-screen-lock (i added that, but have disabled it for testing :)
so, this pr adds a more on-point message to the screen-lock dialog  - and, if the system screen lock is disabled, we show the same message in an alert (that could be made nicer, but that is sth. post 1.36 - also, i assume most ppl use a screen lock)

Screenshot 2023-03-26 at 18 59 33 &nbsp; Screenshot 2023-03-26 at 19 05 20

successor of #2493

### DIFF
--- a/res/layout/activity_registration_2nd_device_qr.xml
+++ b/res/layout/activity_registration_2nd_device_qr.xml
@@ -91,6 +91,7 @@
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
         android:text="@string/multidevice_experimental_hint"
+        android:gravity="center"
         android:textColor="?attr/emoji_text_color"
         android:layout_marginLeft="16dp"
         android:layout_marginRight="16dp"
--- a/res/layout/activity_registration_2nd_device_qr.xml
+++ b/res/layout/activity_registration_2nd_device_qr.xml
@@ -68,7 +68,7 @@
     <RelativeLayout
         android:layout_width="match_parent"
         android:layout_height="0dp"
-        android:layout_weight="7"
+        android:layout_weight="10"
         android:layout_marginStart="24dp"
         android:layout_marginEnd="24dp"
         android:orientation="vertical">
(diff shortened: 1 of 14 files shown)


r10s:
wondering if this "Preparing account..." couldn't just be the "One moment..." we're using elsewhere

res/layout/activity_registration_2nd_device_qr.xml
```diff
@@ -86,4 +86,15 @@
         android:layout_width="match_parent"
         android:layout_height="0dp"/>
 
+    <TextView
+        android:id="@+id/bottom_text"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
```
adbenitez, line 92: this is not centered horizontally

res/layout/backup_provider_fragment.xml
```diff
@@ -108,4 +108,17 @@
 		android:layout_marginBottom="16dp"
 		android:contentDescription="@string/qrscan_title" />
 
+	<TextView
+		android:id="@+id/bottom_text"
```
adbenitez, line 112: maybe also needs to be centered horizontally with `android:gravity="center_horizontal"` or so

adbenitez (approved):

res/layout/activity_registration_2nd_device_qr.xml
r10s: i know - the rest of the text is not centered horizontally either.

i tried both, i think both variants are not bad and are reasonable

res/layout/backup_provider_fragment.xml
r10s: added a commit

PR #2523, opened 2023-03-27
start with light/dark theme depending on system theme

### LINKED ISSUE #2492: Need SplashScreen
`Can you add SplashScreen ?`

_For example with the DeltaChat logo with a dark background, or change the "SplashScreen" color to dark if the user is using a dark theme_

**Every launch of "this" app is a disgusting white light in my eyes.**

adbenitez:
close #2492

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -105,7 +105,7 @@
     <activity android:name=".ConversationListActivity"
               android:label="@string/app_name"
               android:launchMode="singleTask"
-              android:theme="@style/TextSecure.LightNoActionBar"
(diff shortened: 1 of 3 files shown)


r10s:
great! 

nb: that's after 1.36 i'd say. i have often seen to many pitfalls with these xml across android versions :)

adbenitez:
so 1.36 is already out (even 1.36.2!?) what about merging this? :D

r10s (approved):
did not try, but looks simple code-wise.

if there are issues with android4, ppl will complain on nightlies, with chance

PR #2524, opened 2023-03-28
fix some typos in older changelog history

r10s:


adbenitez (approved):
it is never too late :)

PR #2536, opened 2023-04-07
add 'load remote images' strings as needed by desktop

r10s:


r10s:
got direct approvement from @Simon-Laux :)

PR #2538, opened 2023-04-08
Set ndkVersion in build.gradle

link2xt:


build.gradle
```diff
@@ -86,6 +86,14 @@ android {
...
     flavorDimensions "none"
     compileSdkVersion 32
+
+    // Set NDK version to strip native libraries.
+    // Even though we compile our libraries outside Gradle with `scripts/ndk-make.sh`,
+    // without ndkVersion `./gradlew clean` followed by `./gradlew assembleDebug --warning-mode=all` emits the following warning:
+    //   > Task :stripFatDebugDebugSymbols
+    //   Unable to strip the following libraries, packaging them as they are: libanimation-decoder-gif.so, libmapbox-gl.so, libnative-utils.so.
+    // See <https://issuetracker.google.com/issues/237187538> for details.
+    ndkVersion "23.2.8568313"
```
r10s, line 96: wondering what happens if we compile for a different ndk in future and forget to update this line ...

r10s (approved):
thanks for taking care!

build.gradle
r10s: sounds reasonable, thanks!

PR #2541, opened 2023-04-11
Show verified icon exactly if the 1:1 chat is verified

Hocuri:
Part of https://github.com/deltachat/deltachat-core-rust/issues/4188

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationTitleView.java
+++ b/src/org/thoughtcrime/securesms/ConversationTitleView.java
@@ -107,9 +107,6 @@ public class ConversationTitleView extends RelativeLayout {
       else {
         DcContact dcContact = dcContext.getContact(chatContacts[0]);
         subtitleStr = dcContact.getAddr();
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/ConversationTitleView.java
```diff
@@ -107,9 +107,6 @@ else if( dcChat.isDeviceTalk() ) {
       else {
         DcContact dcContact = dcContext.getContact(chatContacts[0]);
         subtitleStr = dcContact.getAddr();
-        if (dcContact.isVerified()) {
-          imgRight = R.drawable.ic_verified;
-        }
```
Hocuri, line 112: I'd actually say that this was a bug: We showed a verified icon whenever we had a verified key of this contact, even if the messages in the chat were not verified with this key.

adbenitez: yes, it was wrong

adbenitez (approved):

PR #2542, opened 2023-04-11
Update to MapLibre 10

### LINKED ISSUE #2361: Remove compile time dependency on GMS
Follow-up to #2349, runtime dependency and non-free components are removed, but GMS is still downloaded during the build.

See https://gitlab.com/fdroid/fdroiddata/-/merge_requests/11527#note_1053804549

A quick solution to get DeltaChat to F-Droid is to use a patched version of mapbox-events-android, like here:
https://github.com/johan12345/EVMap/commit/ed80d7b968826701d59b6590dfba7cf5a9611d74
Patched version is at https://github.com/johan12345/mapbox-events-android

Proper solution is to upstream fixes to MapLibre and introduce a free build flavour, suggested at https://gitlab.com/fdroid/fdroiddata/-/merge_requests/11527#note_1053885392

link2xt:
Fixes #2361

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -65,9 +65,7 @@ dependencies {
     }
     implementation 'com.annimon:stream:1.1.8' // brings future java streams api to SDK Version < 24
     implementation 'com.codewaves.stickyheadergrid:stickyheadergrid:0.9.4' // glues the current time segment text in the gallery to the top.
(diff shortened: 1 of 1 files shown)


licaon-kter:
Does the app work fine on 4 if you use the override?

PR #2544, opened 2023-04-15
Migrate build.gradle from Groovy to Kotlin

link2xt:
Kotlin is the recommended default for Gradle build files now:
- https://developer.android.com/build/migrate-to-kotlin-dsl
- https://blog.gradle.org/kotlin-dsl-is-now-the-default-for-new-gradle-builds

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,0 +1,256 @@
+import java.util.Properties
+
+plugins {
+    id("com.android.application") version "7.4.2"
+}
+
+repositories {
+    google()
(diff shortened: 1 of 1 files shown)


r10s:
> This will need adjustments on F-Droid side
> Or we finally add a proper "fdroid" flavour, and "nighly" flavour.

is this pr actually needed for some feature? otherwise, not calling for trouble and postponing the change until we cannot go away without is an option as well.

just saying, i am not against the change

PR #2549, opened 2023-04-26
Rework how activities interact when forwarding/sharing

### LINKED ISSUE #2509: Disappearing drafts, again

Hocuri:
It always seemed weird to me that there can be multiple running ConversationActivity's. In WA and TG, there is just one chat activity active at a time, and when you press back, you _alway_ get to the chat list.

Now, I noticed that having multiple ConversationActivity's is the cause for at least three bugs on master:

1. - Share to the same chat from within DC
   - Go back
   - The shared text is gone
   
2. - Share to another chat from within DC
   - Open the image
   - Click Back
   - The image is gone

3. #2509

In the past, it was also the reason for #2032 and made #1770 necessary.

@adbenitez also agreed that WA's and TG's behavior is nicer. Soooo, that's what this PR introduces.

This is a pretty big change, so this PR probably introduces new bugs, but fixing three bugs at once, making the logic a lot easier (`doReinitializeDraft`, I'm looking at you), and introducing the behavior that users know from other messengers seems worth it. Plus, I assume that the future bugs will be easier to fix, because there isn't a whole bunch of complex logic that could break with every change you make.

Fixes #2509.

---

**Pecularities I didn't fix because they doesn't cause a bug, and I didn't want to introduce one by changing a running system: (maybe I'll fix some of them later, I'll see)**

---

`isSharing()` returns `false` if there is a `TEXT_EXTRA`, but no `IS_SHARING` set. So, it's possible that `getSharedText()` returns a non-null result, but `isSharing()` returns `false`. This diff would fix it:

```diff
--- a/src/org/thoughtcrime/securesms/util/RelayUtil.java
+++ b/src/org/thoughtcrime/securesms/util/RelayUtil.java
@@@ -31,7 -31,7 +31,8 @@@ public class RelayUtil 
  
      public static boolean isSharing(Activity activity) {
          try {
--            return activity.getIntent().getBooleanExtra(IS_SHARING, false);
++            Intent intent = activity.getIntent();
++            return intent.getBooleanExtra(IS_SHARING, false) || intent.hasExtra(TEXT_EXTRA);
```

---

When sharing text, `initializeDraft()` is actually called twice while the activity starts: Once directly from `onCreate()` an once from `handleSharing()`. Right now,
- When sharing a text, the first `initializeDraft()` sets it as a draft in the compose field.
- When sharing an image, `handleSharing()` sets it as a draft in the core, and the second `initializeDraft()` loads this into the compose field.

---

`DIRECT_SHARING_CHAT_ID` isn't actually used anymore. It's read, but `setDirectSharing()`, which would write it, is unused as of #1852. Instead, the intent created by getChooserTargets() directly sets `EXTRA_CHAT_ID`. So, `DIRECT_SHARING_CHAT_ID` and related logic could be removed.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -19,7 +19,6 @@ package org.thoughtcrime.securesms;
 import static org.thoughtcrime.securesms.ConversationActivity.CHAT_ID_EXTRA;
 import static org.thoughtcrime.securesms.ConversationActivity.STARTING_POSITION_EXTRA;
 import static org.thoughtcrime.securesms.map.MapDataManager.ALL_CHATS_GLOBAL_MAP;
-import static org.thoughtcrime.securesms.util.RelayUtil.REQUEST_RELAY;
 import static org.thoughtcrime.securesms.util.RelayUtil.acquireRelayMessageContent;
 import static org.thoughtcrime.securesms.util.RelayUtil.getDirectSharingChatId;
 import static org.thoughtcrime.securesms.util.RelayUtil.isDirectSharing;
@@ -66,6 +65,7 @@ import org.thoughtcrime.securesms.util.DynamicLanguage;
 import org.thoughtcrime.securesms.util.DynamicNoActionBarTheme;
 import org.thoughtcrime.securesms.util.DynamicTheme;
 import org.thoughtcrime.securesms.util.Prefs;
+import org.thoughtcrime.securesms.util.RelayUtil;
 import org.thoughtcrime.securesms.util.SendRelayedMessageUtil;
 
 public class ConversationListActivity extends PassphraseRequiredActionBarActivity
@@ -209,10 +209,6 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
     refreshAvatar();
     refreshTitle();
     handleOpenpgp4fpr();
-    if (isDirectSharing(this)) {
-      openConversation(getDirectSharingChatId(this), -1);
-    }
-
     if (isDirectSharing(this)) {
       openConversation(getDirectSharingChatId(this), -1);
     }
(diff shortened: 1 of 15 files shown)


src/org/thoughtcrime/securesms/ConversationListActivity.java
```diff
@@ -212,10 +212,6 @@ private void refresh() {
     if (isDirectSharing(this)) {
       openConversation(getDirectSharingChatId(this), -1);
     }
-
-    if (isDirectSharing(this)) {
-      openConversation(getDirectSharingChatId(this), -1);
-    }
```
Hocuri, line 218: I just removed 3 lines that accidentally were in there twice.

src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -597,13 +596,13 @@ public void onDestroyActionMode(ActionMode mode) {
...
       Activity activity = getActivity();
-      if (activity != null) {
+      if (activity instanceof ConversationListActivity) {
         ((ConversationListActivity) activity).refreshTitle();
       }
 
     } else if (event.getId() == DcContext.DC_EVENT_SELFAVATAR_CHANGED) {
       Activity activity = getActivity();
-      if (activity != null) {
+      if (activity instanceof ConversationListActivity) {
```
Hocuri, line 605: I fixed a small crash here: sometimes `activity` is a ConversationListArchiveActivity, not a ConversationListActivity.

adbenitez:
testing this PR a bit, sharing seems to work like a charm, but when I tried to forward got this issue with the date labels

image_2023-04-27_11-24-34

r10s:
we had that issue with the date labels before, when working in that area. i do not remember exactly, but it seems worth doing some research in old issues/prs on that topic

Hocuri:
I made a fix: cb953c866

PR #2555, opened 2023-05-05
add `sendMessage()` api for exporting data to saved-messages

r10s:
`sendMessage` takes an object with `blob`, `fileName`, `text` as parameters, currently, the message is always sent to the saved-messages chat, for the future, we consider to allow sending to the webxdc's source chat as well, therefore the broader name

TODO:
- [ ] settle api, maybe return a promise so we can let the user confirm when it comes to sending to another chat

at https://github.com/webxdc/webxdc-test/pull/9 there is a test for the `sendMessage` api

targets https://github.com/deltachat/deltachat-core-rust/issues/4350

### DIFF
--- a/src/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -1,5 +1,7 @@
 package org.thoughtcrime.securesms;
 
+import static com.b44t.messenger.DcContext.DC_GCL_FOR_FORWARDING;
+
 import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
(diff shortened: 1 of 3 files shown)


Simon-Laux:
I think returning a promise makes sense, so apps could show a progress/busy spinner and hide it again once the export is complete.

src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -332,6 +339,28 @@ public boolean sendStatusUpdate(String payload, String descr) {
       return true;
     }
 
+    @JavascriptInterface
+    public boolean sendMessage(String payload) {
+      Log.i(TAG, "sendMessage");
+      try {
+        JSONObject jsonObject = new JSONObject(payload);
+        byte[] data = Base64.decode(jsonObject.getString("__blobBase64"), Base64.NO_WRAP | Base64.NO_PADDING);
+        String[] fileName = jsonObject.getString("fileName").split("\\.");
```
adbenitez, line 348: this looks dangerous, maybe use https://docs.oracle.com/javase/6/docs/api/java/lang/String.html#split%28java.lang.String,%20int%29 to split at most in two strings, and probably also check the case where the filename has no dot "." (no extension) at all

adbenitez:
closing in favor of #2571

PR #2562, opened 2023-05-15
streamline map wording

r10s:
this pr syncs the wording of "All Location" to what is used elsewhere in the app and common in apps in general (omitting the verb "show" before objects).

see commit messages for more details.

before / after:

 &nbsp; &nbsp;

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -157,6 +157,7 @@
     <string name="switch_camera">Switch Camera</string>
     <string name="toggle_fullscreen">Toggle Full Screen Mode</string>
     <string name="location">Location</string>
+    <string name="locations">Locations</string>
     <string name="gallery">Gallery</string>
     <string name="images_and_videos">Images and Videos</string>
     <string name="file">File</string>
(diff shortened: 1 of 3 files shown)


res/values/strings.xml
```diff
@@ -157,6 +157,7 @@
     <string name="switch_camera">Switch Camera</string>
     <string name="toggle_fullscreen">Toggle Full Screen Mode</string>
     <string name="location">Location</string>
+    <string name="location">Locations</string>
```
Hocuri, line 160: Is this by accident - two strings `location`, one of them in singular, one in plural?

Hocuri (approved):
Apart from the duplicate `location` string, this LGTM

res/values/strings.xml
r10s: thanks, of course, this is an error, i'll fix that

res/values/strings.xml
```diff
@@ -157,6 +157,7 @@
     <string name="switch_camera">Switch Camera</string>
     <string name="toggle_fullscreen">Toggle Full Screen Mode</string>
     <string name="location">Location</string>
+    <string name="locations">Locations</string>
```
Hocuri, line 160: I don't understand, why add this string that seems to be used nowhere?

r10s: it is used in some ios experiments, where the word "Map" is not really fitting when showing a number or "None" beside (not yet there, but planned): 

 

also, as we're talking about "sending/deleting/all/whatnot locations" also elsewhere, this seems to be a more natural fit.

PR #2565, opened 2023-05-22
Remove white icon around foreground image

### LINKED ISSUE #2557: Loading screen shows logo background in dark mode
- Android version: 13
- Device: Pixel 4a
- Delta Chat version: 1.36.5

Screenshot_20230507-015626

Hocuri:
Fix #2557.

I just used gimp to remove the white background; I believe that it was around the icon by mistake.

This is how it looks now: (the `dev` won't be in the released version)

Screenshot_20230522_154413


adbenitez (approved):
good, but using GIMP isn't as nice as just doing the same generation of the images again without the white background, how were they generated anyway? :thinking:

r10s:
yes, using gimp to change the background will look frayed at the edges, we should use the svg source in the [interface](https://github.com/deltachat/interface) repo for that, and create a new png or so as needed.

the "dev" icon, however does not exist as svg, but there i'd say it is also not really important

r10s (approved):

gerryfrancis:
> Won't this also make the upper left part of the logo that's a bit brighter partly-transparent?

I just checked, an opacity value of 0.5 seems to be just right:

grafik
grafik

Besides, you would recognize some irregular gradations of the asymmetrical rectangles, I think.

Hocuri:
In the second image, the edges still look quite white, though - maybe regenerating from svg is the best option (btw, the logos with the `dev` badge are not important, since most users won't see them anyway, so doing this for the normal logo is enough)

gerryfrancis:
@Hocuri You are right, thank you for the information, I will try to regenerate the images from the SVG icon. 👍

PR #2567, opened 2023-05-22
Adapt Padlock and Location indicator to screen size

### LINKED ISSUE #2566: Size of padlock, checkmarks etc. not adjusted to font size (in Android settings)
- Android version:
Android 11.

- Device:
Fairphone FP2 (Lineage OS 18.1 build RQ3A.211001.001 from 2022-04-20).

- Delta Chat version:
1.36.5.

- Expected behavior:
The font used to display icons/symbols (e.g. padlock, checkmark(s)) next to date/time is adjusted to the size, too, after it has been changed by the user.

- Actual behavior:
The font used to display icons/symbols (e.g. padlock, checkmark(s)) next to date/time is not adjusted to the size after it has been changed by the user.

-- Change the font size of your Android device.
-- Open any chat you sent/received an encrypted message (to compare the size of the padlock icon).

- Screenshots:
grafik
grafik

- Logs:
N/A.

Hocuri:
Fix #2566 - well, mostly: This enlarges the padlock (and the location indicator if it's present), but not the checkmark(s), because I didn't manage to also enlarge the checkmark(s) within a bit more than half an hour of work (and I still don't have an idea how it might work, though surely possible).

When using the app like this, this feels good as-is.

This PR with "Largest" resp. "Normal" text size:

Screenshot_20230522-235058

Screenshot_20230522-235028

### DIFF
--- a/res/layout/conversation_item_footer.xml
+++ b/res/layout/conversation_item_footer.xml
@@ -10,8 +10,8 @@
 
     <ImageView
         android:id="@+id/footer_secure_indicator"
-        android:layout_width="12dp"
-        android:layout_height="11dp"
+        android:layout_width="12sp"
+        android:layout_height="11sp"
         android:src="@drawable/msg_encr_out"
         android:visibility="gone"
         android:layout_gravity="center_vertical|end"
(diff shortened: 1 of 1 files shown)


r10s (approved):
thanks!

for the checkmarks, i think, they're not that bad in the different sizes, maybe good enough.

Issue #2569, opened 2023-05-27
Copy image/gif message does not work


Simon-Laux:
**Expected**: When selecting a message and taping "save to clipboard" in the 3dot menu the image is copied to clipboard and I can paste it into an other app or chat

**Actual**: Some text is pasted instead: (translated form of) "Image" or "Gif"

Users expect this, just got this report from a user. If it does not work for images then hide the option, but it would be better if it would work.

r10s:
the broader picture should take into account that there is _multi selection_ as well as _text below pictures/stickers_ etc.

the gist of the function is to copy text and summary of all selected messages - including sender names etc. that way, one has a "cheap" export of the view and can continue from there.

the information "Image" or "Voice Message"  is on purpose therefore, an export may look like:

```
Alice:
Image - Me and my family

Bob:
Thanks for the information!
```

of course, we can consider to offer exporting image data or so additionally, but that would probably quite some work, we may also have to start dealing with mixed content etc., so, not sure if that is worth the effort.

maybe a "copy to clipboard" for images in the enlarged image view? is copying image data to clipboard a  thing used at scale in practise, btw? i know it exist, but i have not seen/used that much.

Simon-Laux:
I think as a first step it should be called "copy text". This would then solve the "bug" here because it won't create false expectations.

Then as second step we could introduce a "copy image/media" or "copy message" (including attachment) when selecting a single message and hiding that option when multiple messages are selected.



> You can add more than one ClipData.Item object to a clip. This lets users copy and paste multiple selections as a single clip. For example, if you have a list widget that lets the user select more than one item at a time, you can copy all the items to the clipboard at once. To do this, create a separate ClipData.Item for each list item, and then add the ClipData.Item objects to the ClipData object.
>
> ~ https://developer.android.com/develop/ui/views/touch-and-input/copy-paste


While it is possible to store multiple "clip items" into the clipboard, deltachat does not support sending multiple attachments in one message yet, so pasting it again into deltachat would probably only paste the first image and all the text. But still it might be useful for "exporting" multiple messages into some note or word app.
But I haven't looked into what the minimum sdk requirements are for storing multiple images and texts into the clipboard, maybe that option would need to be hidden on old devices if the api is not available.

r10s:
as said, +1 for renaming

r10s:
closing this issue, the bug is fixed by #2570, having some additional enhancement is not directly planned currently. feature requests can go to the forum, unless someone directly takes care

PR #2570, opened 2023-05-28
reword 'Copy to Clipboard' to 'Copy Text'

r10s:
message list allows multiple messages
being copied to clipboard at the same time;
if these messages as not just text, a summary is added, sth. as "Voice Messages" or "Forwarded: Image - this is my family" etc.

this is on purpose and the gist of the action,
renaming the action to 'Copy Text' makes that more clear.

targets #2569


Simon-Laux (approved):
LGTM, thanks for caring

PR #2571, opened 2023-05-29
add webxdc.sendToChat() API

### LINKED ISSUE #2573: Implement export API for WebXDC
for exporting `webxdc.sendToChat()` API should be available, for details and discussion on the API specs see: https://support.delta.chat/t/planned-webxdc-apis-for-processing-files/2570

iOS: https://github.com/deltachat/deltachat-ios/issues/1887
Desktop: https://github.com/deltachat/deltachat-desktop/issues/3241

adbenitez:
close #2573

### DIFF
--- a/res/raw/webxdc.js
+++ b/res/raw/webxdc.js
@@ -49,7 +49,7 @@ window.webxdc = (() => {
             data.text = message.text;
         }
         if (message.file) {
-            let isString = value => typeof value === 'string' || value instanceof String;
+            let isString = (value) => typeof value === 'string' || value instanceof String;
             if (!message.file.name || !isString(message.file.base64)) {
                 return Promise.reject("provided file is invalid, you need to set both name and base64 content");
             }
--- a/res/raw/webxdc.js
+++ b/res/raw/webxdc.js
@@ -49,8 +49,7 @@ window.webxdc = (() => {
             data.text = message.text;
         }
         if (message.file) {
-            let isString = (value) => typeof value === 'string' || value instanceof String;
-            if (!message.file.name || !isString(message.file.base64)) {
+            if (!message.file.name || typeof message.file.base64 !== 'string') {
                 return Promise.reject("provided file is invalid, you need to set both name and base64 content");
             }
             data.base64 = message.file.base64;
--- a/res/raw/webxdc.js
+++ b/res/raw/webxdc.js
@@ -42,17 +42,20 @@ window.webxdc = (() => {
 
     sendToChat: async (message) => {
         const data = {};
+        if (!message.text && !message.file) {
+            return Promise.reject("Invalid empty message, at least one of text or file should be provided");
+        }
         if (message.text) {
             data.text = message.text;
         }
         if (message.file) {
+             if (!message.file.name || typeof message.file.base64 !== "string") {
+                  return Promise.reject("provided file is invalid, you need to set both name and base64 content");
+            }
             data.base64 = message.file.base64;
             data.name = message.file.name;
         }
 
-        if (!data.text && !data.name) {
-            return Promise.reject("Invalid empty message, at least one of text or file should be provided");
-        }
 
         const errorMsg = InternalJSApi.sendToChat(JSON.stringify(data));
         if (errorMsg) {
--- a/res/raw/webxdc.js
+++ b/res/raw/webxdc.js
@@ -39,5 +39,28 @@ window.webxdc = (() => {
     sendUpdate: (payload, descr) => {
       InternalJSApi.sendStatusUpdate(JSON.stringify(payload), descr);
     },
+
+    sendToChat: (content) => {
+        var data = {};
+
+        if (content.text) {
+            data.text = content.text;
+        }
+
+        if (content.file) {
+            var file = content.file;
+            var reader = new FileReader();
+            reader.onload = function(readerEvt) {
+                var binaryString = readerEvt.target.result;
+                data.blob = btoa(binaryString);
+                data.name = file.name;
+                data.type = file.type;
+                InternalJSApi.sendToChat(JSON.stringify(data));
+            };
+            reader.readAsBinaryString(file);
+        } else {
+            InternalJSApi.sendToChat(JSON.stringify(data));
+        }
+    },
   };
 })();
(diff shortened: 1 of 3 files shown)


adbenitez:
@Simon-Laux you can add the Promise stuff if you want, the java side returns true/false if it was successful or not so take the reply into account for the promise, also feel free to switch to the https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsDataURL API, I am ok with the current `readAsBinaryString()` so I will not change it myself

src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -337,5 +338,38 @@ public String getStatusUpdates(int lastKnownSerial) {
...
+            fileName = jsonObject.getString("name");
+            mimeType = jsonObject.getString("type");
+        }
+
+        if (jsonObject.has("text")) {
+            text = jsonObject.getString("text");
+        }
+
+        if (data == null && TextUtils.isEmpty(text)) { // invalid empty draft
+            return false;
```
Simon-Laux, line 364: that's not a good error

res/raw/webxdc.js
```diff
@@ -39,5 +39,28 @@ window.webxdc = (() => {
...
+                data.blob = btoa(binaryString);
+                data.name = file.name;
+                data.type = file.type;
+                InternalJSApi.sendToChat(JSON.stringify(data));
+            };
+            reader.readAsBinaryString(file);
+        } else {
+            InternalJSApi.sendToChat(JSON.stringify(data));
+        }
+    },
```
Simon-Laux, line 64:
```suggestion
    sendToChat: async (content) => {
        const data = {};

        if (content.text) {
            data.text = content.text;
        }

        if (content.file) {
            data.blob = await new Promise((resolve, reject) => {
                const reader = new FileReader()
                reader.onload = () => {
                  /** @type {string} */
                  let data = reader.result
                  resolve(data.slice(data.indexOf(data_start) + data_start.length))
                }
                reader.readAsDataURL(content.file)
                reader.onerror = () => reject(reader.error)
              })           
            data.name = content.file.name;
            data.type = content.file.type;
        }
        const result = InternalJSApi.sendToChat(JSON.stringify(data));
        if(!result){
            return Promise.reject("There was an error")
        }
    },
```

adbenitez: how the user uses sendToChat() then? since it is the only of the existing API functions that is declared with async keyword, does that need any special treatment? or can it be normally called as any non-async function and it is ensured to be executed? without needing `await` keyword?

res/raw/webxdc.js
```diff
@@ -39,5 +39,25 @@ window.webxdc = (() => {
...
+            data.text = message.text;
+        }
+        if (message.file) {
+            data.base64 = message.file.base64;
+            data.name = message.file.name;
+        }
+
+        if (!data.text && !data.name) {
+            return Promise.reject("Invalid empty message, at least one of text or file should be provided");
+        }
```
Simon-Laux, line 55:
```suggestion
        if (!data.text && !data.file) {
            return Promise.reject("Invalid empty message, at least one of text or file should be provided");
        }
        if (message.text) {
            data.text = message.text;
        }
        if (message.file) {
             if (!message.file.name && !message.file.base64) {
                  return Promise.reject("Invalid empty message, at least one of text or file should be provided");
            }
            data.base64 = message.file.base64;
            data.name = message.file.name;
        }

```

Simon-Laux: fixed it

res/raw/webxdc.js
```diff
@@ -39,5 +39,29 @@ window.webxdc = (() => {
...
+    sendToChat: async (message) => {
+        const data = {};
+        if (!message.text && !message.file) {
+            return Promise.reject("Invalid empty message, at least one of text or file should be provided");
+        }
+        if (message.text) {
+            data.text = message.text;
+        }
+        if (message.file) {
+            let isString = value => typeof value === 'string' || value instanceof String;
```
r10s, line 52: i read this as the definition of an anonymous function that is assigned to isString. if this this is correct, for better understanding, should't there be some brackets around the parameter list?

```suggestion
            let isString = (value) => typeof value === 'string' || value instanceof String;
```

src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -337,5 +338,34 @@ public String getStatusUpdates(int lastKnownSerial) {
...
+        }
+        if (jsonObject.has("text")) {
+            text = jsonObject.getString("text");
+        }
+
+        if (TextUtils.isEmpty(text) && TextUtils.isEmpty(name)) {
+            return "provided file is invalid, you need to set both name and base64 content";
+        }
+
+        DcHelper.share(WebxdcActivity.this, data, "application/octet-stream", name, text);
```
r10s, line 363: i think, this line closess the  webxdc because of `launchMode="singleTask"` of some subsequent activities? if so, maybe add a little comment as that is not obvious.

```suggestion
        // this ends the WebxdcActivity as by `launchMode="singleTask"` an existing activity is restarted and removes all activities on top
        DcHelper.share(WebxdcActivity.this, data, "application/octet-stream", name, text);
```

i assume, btw, that the WebxdcActivity is not terminated my the chat-selection but on step further (did not test yet, however)

r10s: that will result in the promise being rejected? that's probably fine then, even good for testing.

r10s: i added a commit to https://github.com/webxdc/webxdc-test/pull/11 that prints the state of the promise - and, indeed, the promise is _resolved_ in that case (seems to happen implicitly when the function returns).

i think, this is okay: we're saying that the webxdc does not know anything about the sending state, it just drops a file, and that succeeded. but we need to update the docs, saying that the promise might be resolved.

r10s (approved):
lgtm, also tested a bit, all works as expected :)

res/raw/webxdc.js
```diff
@@ -39,5 +39,29 @@ window.webxdc = (() => {
...
+    sendToChat: async (message) => {
+        const data = {};
+        if (!message.text && !message.file) {
+            return Promise.reject("Invalid empty message, at least one of text or file should be provided");
+        }
+        if (message.text) {
+            data.text = message.text;
+        }
+        if (message.file) {
+            let isString = (value) => typeof value === 'string' || value instanceof String;
```
Simon-Laux, line 52: the types for this will not allow `String` so why do you want to allow it here?

PR #2574, opened 2023-05-31
implement support for <input type=file> HTML tag in webxdc

### LINKED ISSUE #2572: <input type=file> should work in webxdc apps
it is supported by iOS and Desktop already, only Android is missing support, with this webxdc apps already have a way to import files in a classic well-known way used by web developers in existing web apps

adbenitez:
close #2572

### DIFF
--- a/src/org/thoughtcrime/securesms/WebViewActivity.java
+++ b/src/org/thoughtcrime/securesms/WebViewActivity.java
@@ -13,6 +13,7 @@ import android.view.MenuItem;
 import android.webkit.WebResourceRequest;
 import android.webkit.WebResourceResponse;
 import android.webkit.WebView;
+import im.delight.android.webview.AdvancedWebView;
 import android.webkit.WebViewClient;
 import android.widget.ImageView;
 import android.widget.Toast;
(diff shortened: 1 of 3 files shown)


src/org/thoughtcrime/securesms/WebViewActivity.java
```diff
@@ -156,8 +157,8 @@ public void onResume() {
 
   @Override
   protected void onDestroy() {
-    super.onDestroy();
     webView.destroy();
+    super.onDestroy();
```
adbenitez, line 161: this is unrelated just because it seemed to be more correct, in theory I guess you should first destroy the webview before calling this activity's destroy because then maybe the rest of the code maybe is not executed? but I am not sure how this works

r10s:
great you found [AdvancedWebView](https://github.com/delight-im/Android-AdvancedWebView)!

however, i am a bit ambivalent in pulling in a mostly unmaintained dependency in an already hard and critical area - plus that the additional layer may cause problems as it contains code in mostly all things WebViews do, partly for targeting android 2.x

maybe it is better to set `onShowFileChooser()` in our `WebViewActicity` directly and use  https://github.com/delight-im/Android-AdvancedWebView/blob/master/Source/library/src/main/java/im/delight/android/webview/AdvancedWebView.java as a pattern only - i assume this will result in a few dozen lines only as we need to target android5 only

adbenitez:
closed in favor of #2575

PR #2575, opened 2023-05-31
implement onShowFileChooser based on AdvancedWebView

r10s:
this is what i did to check out the effort, works on first tests, alternative to https://github.com/deltachat/deltachat-android/pull/2574

@adbenitez feel free to commit on top of this pr or close it

### DIFF
--- a/src/org/thoughtcrime/securesms/WebViewActivity.java
+++ b/src/org/thoughtcrime/securesms/WebViewActivity.java
@@ -339,13 +339,11 @@ public class WebViewActivity extends PassphraseRequiredActionBarActivity
         try {
           if (data.getDataString() != null) {
             dataUris = new Uri[]{Uri.parse(data.getDataString())};
-          } else {
-            if (data.getClipData() != null) {
-              final int numSelectedFiles = data.getClipData().getItemCount();
-              dataUris = new Uri[numSelectedFiles];
-              for (int i = 0; i < numSelectedFiles; i++) {
-                dataUris[i] = data.getClipData().getItemAt(i).getUri();
-              }
+          } else if (data.getClipData() != null) {
+            final int numSelectedFiles = data.getClipData().getItemCount();
+            dataUris = new Uri[numSelectedFiles];
+            for (int i = 0; i < numSelectedFiles; i++) {
+              dataUris[i] = data.getClipData().getItemAt(i).getUri();
             }
           }
         } catch (Exception e) {
--- a/src/org/thoughtcrime/securesms/WebViewActivity.java
+++ b/src/org/thoughtcrime/securesms/WebViewActivity.java
@@ -139,7 +139,7 @@ public class WebViewActivity extends PassphraseRequiredActionBarActivity
         intent.addCategory(Intent.CATEGORY_OPENABLE);
         intent.setType("*/*");
         intent.putExtra(Intent.EXTRA_ALLOW_MULTIPLE, fileChooserParams.getMode() == FileChooserParams.MODE_OPEN_MULTIPLE);
-        startActivityForResult(Intent.createChooser(intent, getString(R.string.select)), REQUEST_CODE_FILE_PICKER);
+        WebViewActivity.this.startActivityForResult(Intent.createChooser(intent, getString(R.string.select)), REQUEST_CODE_FILE_PICKER);
         return true;
       }
     });
--- a/src/org/thoughtcrime/securesms/WebViewActivity.java
+++ b/src/org/thoughtcrime/securesms/WebViewActivity.java
@@ -1,5 +1,6 @@
 package org.thoughtcrime.securesms;
 
+import android.app.Activity;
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/WebViewActivity.java
```diff
@@ -122,6 +127,22 @@ public WebResourceResponse shouldInterceptRequest(WebView view, WebResourceReque
...
+      @Override
+      @RequiresApi(21)
+      public boolean onShowFileChooser(WebView webView, ValueCallback<Uri[]> filePathCallback, WebChromeClient.FileChooserParams fileChooserParams) {
+        if (mFileUploadCallbackSecond != null) {
+          mFileUploadCallbackSecond.onReceiveValue(null);
+        }
+        mFileUploadCallbackSecond = filePathCallback;
+        Intent intent = new Intent(Intent.ACTION_GET_CONTENT);
+        intent.addCategory(Intent.CATEGORY_OPENABLE);
+        intent.setType("*/*");
```
Simon-Laux, line 140: Is there something in fileChooserParams for this too?

r10s: after checking: there is `WebChromeClient.FileChooserParams.getAcceptTypes()`, not sure if that is 100% compatible to java, however - i would expect AdvancedWebView to use that as well otherwise - so, maybe play safe for now and focus on mimicing AdvancedWebView and improve in a subsequent pr at some later point

r10s: EXTRA_MIME_TYPES comes from an intent, but we do not get an intent from onShowFileChooser(), so that would not help anyways as far as i understand it

src/org/thoughtcrime/securesms/WebViewActivity.java
```diff
@@ -34,6 +37,8 @@ public class WebViewActivity extends PassphraseRequiredActionBarActivity
                                           WebView.FindListener
 {
   private static final String TAG = WebViewActivity.class.getSimpleName();
+  protected static final int REQUEST_CODE_FILE_PICKER = 51426;
+  protected ValueCallback<Uri[]> mFileUploadCallbackSecond;
```
adbenitez, line 41: filePathCallback would be a better name, the "second" part comes from another usage in AdvancedWebview

r10s: definitely, yes

src/org/thoughtcrime/securesms/WebViewActivity.java
```diff
@@ -122,6 +127,22 @@ public WebResourceResponse shouldInterceptRequest(WebView view, WebResourceReque
...
+      public boolean onShowFileChooser(WebView webView, ValueCallback<Uri[]> filePathCallback, WebChromeClient.FileChooserParams fileChooserParams) {
+        if (mFileUploadCallbackSecond != null) {
+          mFileUploadCallbackSecond.onReceiveValue(null);
+        }
+        mFileUploadCallbackSecond = filePathCallback;
+        Intent intent = new Intent(Intent.ACTION_GET_CONTENT);
+        intent.addCategory(Intent.CATEGORY_OPENABLE);
+        intent.setType("*/*");
+        intent.putExtra(Intent.EXTRA_ALLOW_MULTIPLE, fileChooserParams.getMode() == FileChooserParams.MODE_OPEN_MULTIPLE);
+        startActivityForResult(Intent.createChooser(intent, getString(R.string.select)), REQUEST_CODE_FILE_PICKER);
```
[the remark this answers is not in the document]
r10s: does compiler complain about that? why is it needed?

adbenitez (approved):

src/org/thoughtcrime/securesms/WebViewActivity.java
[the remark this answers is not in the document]
adbenitez: no but it is clearer for people using a dummy plain text editor like me :D to avoid ambiguity and clearly see from where it is that method from, better explicit than implicit

#AndroidStudioKillsMyMachine

src/org/thoughtcrime/securesms/WebViewActivity.java
r10s: ah, yes, i understand. so passing `getAcceptTypes()` to EXTRA_MIME_TYPES and/or `setType()` - but yes, this is sth. for another iteration i'd say, again, there might be reasons, `AdvancedWebView` did not do it if it is apparently simple :)

src/org/thoughtcrime/securesms/WebViewActivity.java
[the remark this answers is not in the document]
r10s: i see :)

as it is needed for having `filePathCallback` as membername as well as parameter anyways (see next commit), this even improves consistency.

PR #2576, opened 2023-06-01
webxdc update "send to chat" api

Simon-Laux:


res/raw/webxdc.js
```diff
@@ -41,26 +41,74 @@ window.webxdc = (() => {
...
+          "Error from sendToChat: Invalid empty message, at least one of text or file should be provided"
+        );
+      }
+      /** @type {(file: Blob) => Promise<string>} */
+      const blob_to_base64 = (file) => {
+        const data_start = ";base64,";
+        return new Promise((resolve, reject) => {
+          const reader = new FileReader();
+          reader.readAsDataURL(file);
+          reader.onload = () => {
```
adbenitez, line 56: shouldn't the readAsDataURL be called after setting the event listener? even if due to some JS weirdness this works, isn't easier to read/understand if the listener is set before the event is triggered?

adbenitez (approved):
great stuff!

adbenitez:
tested with https://github.com/webxdc/webxdc-test/pull/14 and it seems to work fine

PR #2578, opened 2023-06-02
unify and shorten sendToChat() errors

r10s:
cleaning up a bit, see commit messages for details. came over that while working on https://github.com/deltachat/deltachat-ios/pull/1886 ...

### DIFF
--- a/res/raw/webxdc.js
+++ b/res/raw/webxdc.js
@@ -43,9 +43,7 @@ window.webxdc = (() => {
     sendToChat: async (message) => {
       const data = {};
       if (!message.file && !message.text) {
-        return Promise.reject(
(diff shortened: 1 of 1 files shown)


res/raw/webxdc.js
```diff
@@ -70,16 +68,14 @@ window.webxdc = (() => {
       if (message.file) {
```
r10s, line 68: wondering if the type script annotation `/** @type {{file_name: string, file_message: string} | null} */` belongs to `message.file` and if it is correct - we do nowhere have the parameters or variables named `file_name`  or `file_message`

adbenitez (approved):

PR #2586, opened 2023-06-08
move the file chooser implementation to WebxdcActivity class

### LINKED ISSUE #2585: crash opening HTML message in Android 4
an Android 4.2 user reported that opening an HTML message fails after the recent addition of https://github.com/deltachat/deltachat-android/pull/2575

adbenitez:
successor of https://github.com/deltachat/deltachat-android/pull/2575

close #2585

### DIFF
--- a/src/org/thoughtcrime/securesms/WebViewActivity.java
+++ b/src/org/thoughtcrime/securesms/WebViewActivity.java
@@ -309,9 +309,4 @@ public class WebViewActivity extends PassphraseRequiredActionBarActivity
   protected WebResourceResponse interceptRequest(String url) {
     return null;
   }
-
-  @Override
-  public void onActivityResult(int reqCode, int resultCode, final Intent data) {
-    super.onActivityResult(reqCode, resultCode, data);
-  }
 }
--- a/src/org/thoughtcrime/securesms/WebViewActivity.java
+++ b/src/org/thoughtcrime/securesms/WebViewActivity.java
@@ -1,6 +1,5 @@
 package org.thoughtcrime.securesms;
 
-import android.app.Activity;
 import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.Intent;
@@ -11,8 +10,6 @@ import android.util.Log;
 import android.view.Menu;
 import android.view.MenuInflater;
 import android.view.MenuItem;
-import android.webkit.ValueCallback;
-import android.webkit.WebChromeClient;
 import android.webkit.WebResourceRequest;
 import android.webkit.WebResourceResponse;
 import android.webkit.WebView;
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/WebViewActivity.java
```diff
@@ -127,22 +122,6 @@ public WebResourceResponse shouldInterceptRequest(WebView view, WebResourceReque
         return super.shouldInterceptRequest(view, request);
       }
     });
-    webView.setWebChromeClient(new WebChromeClient() {
-      @Override
-      @RequiresApi(21)
-      public boolean onShowFileChooser(WebView webView, ValueCallback<Uri[]> filePathCallback, WebChromeClient.FileChooserParams fileChooserParams) {
-        if (WebViewActivity.this.filePathCallback != null) {
-          WebViewActivity.this.filePathCallback.onReceiveValue(null);
-        }
```
adbenitez, line 136: I discarded this because it doesn't match the behavior of browsers, it is simulating a "nothing selected" event upon opening the chooser, which doesn't happen in real browsers, not even when the user cancels the file chooser (don't select anything)

src/org/thoughtcrime/securesms/WebViewActivity.java
```diff
@@ -333,26 +312,6 @@ protected WebResourceResponse interceptRequest(String url) {
...
-            }
-          }
-        } catch (Exception e) {
-          e.printStackTrace();
-        }
-      }
-      filePathCallback.onReceiveValue(dataUris);
-      filePathCallback = null;
-    }
     super.onActivityResult(reqCode, resultCode, data);
```
r10s, line 315: i assume, the whole `@Override onActivityResult()` can be removed here as it does nothing else than calling `super.onActivityResult()`

src/org/thoughtcrime/securesms/WebViewActivity.java
r10s: hm, i would leave it as is. iiuc, the even goes to previous selector, not to the current one. `WebViewActivity.this.filePathCallback.onReceiveValue(null)` is called only if there was previously a file request started that gets aborted. the 3 lines make sure, a request is always somehow ended.

the code comes from the AvancedWebView, so, unless we're really sure what we're doing better, i'd leave it and focus on shifting things in this pr. if you still think, it is wrong, please open a subsequent pr.

adbenitez: but why would you send such event to the previous callback when calling a new one? the old one already received the call, also the callback is set to null every time the response is received so I think this doesn't even happens never :thinking:

src/org/thoughtcrime/securesms/WebViewActivity.java
adbenitez: yes, I was thinking that as well, but since it was like that even before adding this code, I thought it was intentional, will remove it then

src/org/thoughtcrime/securesms/WebViewActivity.java
r10s: > but why would you send such event to the previous callback when calling a new one?

https://developer.android.com/reference/android/webkit/WebChromeClient#onShowFileChooser says that for canceling a request, we should pass `null` - so, if forever reason, onShowFileChooser() is triggered again before we close the dialog, we should cancel the previous callback

i agree that it should not happen normally

adbenitez: ok, I reverted it, it should not hurt to have in the end

r10s (approved):
lgtm now, thanks for fixing!

PR #2590, opened 2023-06-19
prepare 1.38.0

r10s:


CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # Delta Chat Android Changelog
 
+## v1.38.0
+2023-06
+
+* improve group membership consistency
+* fix verification issues because of email addresses compared case-sensitive sometimes
+* fix empty lines in HTML view
+* fix empty links in HTML view
```
adbenitez, line 9: duplicated line? :thinking:

src/org/thoughtcrime/securesms/ConversationListActivity.java
```diff
@@ -101,12 +101,12 @@ protected void onCreate(Bundle icicle, boolean ready) {
     // it is not needed to keep all past update messages, however, when deleted, also the strings should be deleted.
     DcContext dcContext = DcHelper.getContext(this);
     DcMsg msg = new DcMsg(dcContext, DcMsg.DC_MSG_TEXT);
-    msg.setText("1.37 is out:\n\n"
+    msg.setText("1.38 is out:\n\n"
       + "📎 Removed upper size limit of attachments\n\n"
       + "🗜️ PNG images (screenshots!) are now compressed as well\n\n"
       + "⌨️ For Devs: Better apps for chats with sendToChat() and importFiles() (What is Webxdc? -> https://webxdc.org)\n\n"
       + "... and more at https://get.delta.chat/#changelogs");
```
adbenitez, line 108: shouldn't this message go to the strings.xml file for translation? :thinking:

adbenitez (approved):

src/org/thoughtcrime/securesms/ConversationListActivity.java
r10s: maybe, however, we do not always localize these messages ... and that one seems quite technical anyways ...

r10s: but yes, before weighten for and against, i just added a commit and make it translatable, still makes sense

PR #2592, opened 2023-06-21
add strings for improved sendToChat title

r10s:
the current dialog that is show on sendToChat() reads `Share with`.

the idea is to add the filename and also adapt verb to "send" instead of "share" (although i am +-0 on that, it seems more fitting)

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -413,6 +413,9 @@
     <string name="start_app">Start…</string>
     <!-- this is a warning that is shown when one tries to send something to a chat that is not yet accepted. -->
     <string name="accept_request_first">Please accept the chat request first.</string>
+    <!-- placeholder will be replaced by a file name -->
+    <string name="send_file_to">Send \"%1%s\" to…</string>
+    <string name="send_message_to">Send Message to…</string>
 
     <!-- map -->
     <string name="filter_map_on_time">Show locations in time frame</string>


res/values/strings.xml
```diff
@@ -413,6 +413,9 @@
     <string name="start_app">Start…</string>
     <!-- this is a warning that is shown when one tries to send something to a chat that is not yet accepted. -->
     <string name="accept_request_first">Please accept the chat request first.</string>
+    <!-- placeholder will be replaced by a file name -->
+    <string name="send_file_to">Send \"%1%s\" to…</string>
+    <string name="send_file_to">Send Message to…</string>
```
Simon-Laux, line 418: the 2 entries have the same key otherwise looks good to me

res/values/strings.xml
```diff
@@ -413,6 +413,9 @@
     <string name="start_app">Start…</string>
     <!-- this is a warning that is shown when one tries to send something to a chat that is not yet accepted. -->
     <string name="accept_request_first">Please accept the chat request first.</string>
+    <!-- placeholder will be replaced by a file name -->
+    <string name="send_file_to">Send \"%1%s\" to…</string>
```
hpk42, line 417: will there be there some cutting if the file name is very long?

res/values/strings.xml
hpk42, line 418: dumb question: is it correct that there are two duplicate names for the two strings?

res/values/strings.xml
adbenitez: also files shared from other apps sometimes don't even have a filename, this might introduce some weird names on display from data shared from other apps

res/values/strings.xml
r10s: no, this was my fault :)

res/values/strings.xml
r10s: > will there be there some cutting if the file name is very long?

that could be done by the implementation. but even if it is not done and the text is cut at the end, sth. as `Send "this-is-a-very-long-filename-that...` still seems better than `Share with`

> also files shared from other apps sometimes don't even have a filename

this is only about sending messages from webxdc, and there we require a file name.

Simon-Laux (approved):

res/values/strings.xml
r10s: thanks!

PR #2600, opened 2023-06-25
try to fix popup/dialog background in webview

adbenitez:
this somehow fixes: https://github.com/webxdc/calendar/issues/74

but I don't understand that properties, or the impact in other views, the removal of that line, at first glance after some testing it seems to have no adverse effects,

@Hocuri, @r10s  maybe you can tell


Hocuri (approved):
No, I have no idea why this changes anything; could be a bug in Android.

Let's hope that this doesn't introduce new bugs, but since it fixes one, it's probably fine to merge.

PR #2602, opened 2023-06-26
docs: remove the link to GitHub releases from the readme

link2xt:
There are no GitHub releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 This is the Android client for [Delta Chat](https://delta.chat/).
 It is available on [F-Droid](https://f-droid.org/app/com.b44t.messenger) and
 the [Google Play Store](https://play.google.com/store/apps/details?id=chat.delta).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,8 +3,6 @@
 This is the Android client for [Delta Chat](https://delta.chat/).
 It is available on [F-Droid](https://f-droid.org/app/com.b44t.messenger) and
 the [Google Play Store](https://play.google.com/store/apps/details?id=chat.delta).
-The APK can also be downloaded from [GitHub](https://github.com/deltachat/deltachat-android/releases)
```
r10s, line 6: or say `The APK can also be downloaded from [get.delta.chat](https://get.delta.chat)`

r10s (approved):
thanks for taking care. we could also point to get.delta.chat, however, both variants are better than the wrong link :)

PR #2608, opened 2023-06-28
Fix NullPointerException

Hocuri:
The exception was:

```
java.lang.NullPointerException: Attempt to invoke virtual method 'int com.b44t.messenger.DcMsg.getType()' on a null object reference
                                                                                                    	at org.thoughtcrime.securesms.ConversationActivity$5.doInBackground(ConversationActivity.java:1124)
                                                                                                    	at org.thoughtcrime.securesms.ConversationActivity$5.doInBackground(ConversationActivity.java:1115)
                                                                                                    	at android.os.AsyncTask$3.call(AsyncTask.java:394)
```

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -1121,7 +1121,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
 
           // for WEBXDC, drafts are just sent out as is.
           // for preparations and other cases, cleanup draft soon.
-          if (msg.getType() != DcMsg.DC_MSG_WEBXDC) {
+          if (msg == null || msg.getType() != DcMsg.DC_MSG_WEBXDC) {
             dcContext.setDraft(dcChat.getId(), null);
           }
 


r10s (approved):

src/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1121,7 +1121,7 @@ protected Void doInBackground(Object... param) {
 
           // for WEBXDC, drafts are just sent out as is.
           // for preparations and other cases, cleanup draft soon.
-          if (msg.getType() != DcMsg.DC_MSG_WEBXDC) {
+          if (msg == null || msg.getType() != DcMsg.DC_MSG_WEBXDC) {
```
r10s, line 1124: from the logic and the flow, if is maybe easier to understand if we move `if (msg.getType() != DcMsg.DC_MSG_WEBXDC) { ...` (including comment) three three lines down to the `if(msg!=null)` branch

Hocuri: I don't see where `setDraft()` is called twice in the current code (and therefore think that if we don't call it here anymore, it won't be called at all):

There is a call `dcContext.setDraft(dcChat.getId(), msg);`, but it's in the `else` branch of `if (action==ACTION_SEND_OUT)`, so it's not called in the code flow we're talking about.

r10s: you're right, i lost track of the brackets :)

adbenitez (approved):

PR #2609, opened 2023-06-29
fix some linter warnings

adbenitez:


adbenitez:
> @adbenitez Any reason it is still a draft? Looks like a low hanging fruit for rebasing, marking as ready for review and merging.

the only reason was that I have several more small fixes like this, there are several more linter warnings etc. but indeed several PRs can be created

r10s:
this should be merged to stable, or?

otoh, we can just give up stable by ignoring it ... :)

r10s (approved):

adbenitez:
> this should be merged to stable, or?
> 
> otoh, we can just give up stable by ignoring it ... :)

darn it, yet another confusion because default branch is master when creating PRs and I often forget to switch, but it is easy to change in the github UI, what we could do is to set stable as default branch for now until we decide to finally include the 1:1 verified chats that I hope is soon so we get ride of this divergences

PR #2610, opened 2023-06-30
Fix crash in the "Full Msg" view by downgrading Gson

Hocuri:
@adbenitez figured this out, I'm just creating the PR because we tested it on my laptop.

Thanks :tada:!

The issue was:
- On Android 4.2 or older:
- Open an HTML message in Full Msg View
- Enable loading remote images

-> The app crashes.

We only had an Android 4.1 Emulator to test it, but assume that it's the same issue.

It's a known issue in gson 2.10.1: https://github.com/google/gson/issues/2310


adbenitez (approved):
:tada: thanks!

PR #2615, opened 2023-07-13
Update ndk-make.sh initial comment to suggest an app for determining CPU architecture

pjvleeuwen:
Providing an alternative that does not require ADB / connecting a device to laptop. When searching online for "android determine CPU architecture" many non-OSS apps where suggested. I found this app in f-droid to be useful. Maybe somebody else is helped by this.

### DIFF
--- a/scripts/ndk-make.sh
+++ b/scripts/ndk-make.sh
@@ -15,6 +15,8 @@
 #
 #     adb shell uname -m
 #
+# Use an app like: https://f-droid.org/packages/com.kgurgul.cpuinfo/
+#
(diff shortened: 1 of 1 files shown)


scripts/ndk-make.sh
```diff
@@ -15,6 +15,8 @@
 #
 #     adb shell uname -m
 #
+# Use an app like: https://f-droid.org/packages/com.kgurgul.cpuinfo/
```
adbenitez, line 18: indeed I also like that app and have used it and recommended it to know the architecture of the phone

adbenitez (approved):
thanks a lot for contributing! :star_struck:

PR #2621, opened 2023-07-23
change button order of verification-broken alert

r10s:
the old order was:

```
End-to-end encryption cannot be guaranteed anymore...

[More Info]  [OK] [Scan QR code]
```

Having "OK" in the middle of the buttons is weird.

1. _If_ the main purpose of the dialog would be to "Scan QR code", the text should point that out more, also the opening button. Then, the button position would be correct, however, it has to read "Cancel" then.

2. _However_, the main purpose of the idalog is to "Show Info", "Scan QR code" is only an optional offering and the button a shortcut only. And finishing "Show Info" is better done by a simple "OK". Then, the "OK" should be rightmost.

This PR implements the second option, the buttons read

```
End-to-end encryption cannot be guaranteed anymore...

[More Info]  [Scan QR code] [OK]
```

afterwards. Or, if there is few horizontal space,

```
End-to-end encryption cannot be guaranteed anymore...

          [OK]
[Scan QR Code]
  [Learn More]
```


Hocuri (approved):
I had implemented it that way because it's what WhatsApp does. However, looking at your screenshots, the new button order does feel more natural way, so, let's go for it - I guess WhatsApp just didn't put too much effort into this dialog.

Thanks!

r10s:
> I guess WhatsApp just didn't put too much effort into this dialog

yeah! we're becoming better that WhatsApp!

PR #2637, opened 2023-08-15
improve mailto Intent handling in ShareActivity

### LINKED ISSUE #2633: starting conversation with contact from the phone contacts app doesn't work
if you save your contacts's emails in the contacts app of the phone, there are buttons to call them and also to write an email to them, if you click the email button, Delta Chat opens but instead of opening a chat with that contact, the chat list is show and "share with" is displayed at the top, analyzing the android Intent it seems like a normal mailto:... intent

adbenitez:
close #2633

### DIFF
--- a/src/org/thoughtcrime/securesms/NewConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/NewConversationActivity.java
@@ -17,7 +17,6 @@
 package org.thoughtcrime.securesms;
 
 import android.content.Intent;
-import android.net.MailTo;
 import android.net.Uri;
 import android.os.Bundle;
 import android.util.Log;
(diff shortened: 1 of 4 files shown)


src/org/thoughtcrime/securesms/NewConversationActivity.java
```diff
@@ -75,19 +65,14 @@ private void handleIntent() {
...
-                onContactSelected(DcContact.DC_CONTACT_ID_NEW_CONTACT, recipient);
+          if(MailtoUtil.isMailto(uri)) {
+            String textToShare = MailtoUtil.getText(uri);
+            String[] recipientsArray = MailtoUtil.getRecipients(uri);
+            if (recipientsArray.length >= 1) {
+              if (!textToShare.isEmpty()) {
+                getIntent().putExtra(TEXT_EXTRA, textToShare);
               }
+              onContactSelected(DcContact.DC_CONTACT_ID_NEW_CONTACT, recipientsArray[0]);
             } else {
```
Hocuri, line 76: Nothing here changed semantically, you just extracted methods. Is my understanding correct?

Hocuri (approved):
Nice! I checked that the old sharing tests still pass, and pushed a new regression test for this bug. The code looks OK, though of course none of us knows it very well - even better that we have the tests.

One bug less :tada:!

src/org/thoughtcrime/securesms/NewConversationActivity.java
Hocuri, line 76: Nothing here changed semantically, you just extracted methods. Is my understanding correct?

adbenitez: yes I only extracted stuff that needed to be used in the ShareActivity class so it is not duplicated code

PR #2647, opened 2023-09-05
build: enable -Werror flag when building the core

link2xt:
If this flag was enabled, CI would have caught
incorrect core in 1.40.0 release tag
that caused crashing at start F-Droid 1.40.0 release.


r10s (approved):
nice!

Issue #2599, opened 2023-06-23
Upgrade targetSdkVersion and compileSdkVersion to 33


link2xt:
We currently target SDK version 32 since it was upgraded in #2407.

Due to recent incident with deprecated JCenter repository #2595 it would be nice to upgrade at least some affected dependencies to versions that are published outside of JCenter. One such dependency is exoplayer, the version 2.9.6 that we depend on currently is published on JCenter and new version 2.18.7 which is published in Google's Maven repository requires SDK 33: #2598 

SDK upgrade guide is published at https://developer.android.com/google/play/requirements/target-sdk
There is no dedicated list of things to upgrade when migrating from API level 32 to API level 33, but there is a guide for migration from Android 12 (API level 31) to Android 13 (API level 33): https://developer.android.com/about/versions/13/behavior-changes-13

There is a number of changes to make:
- [x] For Wi-Fi use-cases the applications should request `NEARBY_WIFI_DEVICES` instead of `ACCESS_FINE_LOCATION` permission. Quick grep shows that we only request `ACCESS_FINE_LOCATION` in src/org/thoughtcrime/securesms/mms/AttachmentManager.java when attaching the location, but we may need to update this comment and test if backup transfer still works: https://github.com/deltachat/deltachat-android/blob/1343f5399a6d9c6264c772e2f2008919f4490519/src/org/thoughtcrime/securesms/qr/BackupTransferActivity.java#L222-L223 Most likely it is not needed as we are not using Wi-Fi Direct and the documentation says "This change affects your app only if you call a Wi-Fi API.", but we are not using Android Wi-Fi APIs.  
EDIT: should be fine, i adapted the comment and commented at https://github.com/deltachat/deltachat-android/pull/2649/commits/84e8b7ef51851b1d86e31ba3a22852210fd85aa4
- [x] Use `READ_MEDIA_IMAGES`, `READ_MEDIA_VIDEO` and `READ_MEDIA_AUDIO` instead of `READ_EXTERNAL_STORAGE` when running on Android 13. Another option suggested is to use [photo picker](https://developer.android.com/training/data-storage/shared/photopicker), but it is only supported on Android 4.4 (API level 19) which Delta Chat supports API level 16.
- [x] https://developer.android.com/about/versions/13/behavior-changes-13#playback-controls - no idea if this affects us EDIT: i did not find any hint that this affects us, we're not using playback controls in notifications
[...]

PR #2649, opened 2023-09-06
set targetSdkVersion to 33

r10s:
this PR updates targetSdkVersion to 33, compiling and building works, however, this kind if updates usually also comes with behaviour changes (see  https://developer.android.com/about/versions/13/behavior-changes-all ) that need to be checked carefully:

concrete things that do not work at a first glance:

- [x] attach audio or gallery (instead of permission request, the fallback "Permission required" info is shown)

- [x] the "quick image attach bar" is ~~not shown~~ empty as the permissions are granted, but reading photos fails

- [x] check remaining occurrences of READ_EXTERNAL_STORAGE and WRITE_EXTERNAL_STORAGE: EDIT: done, we keep the permissions in the code, however bypass the check on api33 

- [x] check if writing files (backups, keys, export images, ...) work EDIT: log: OK , key: OK , backup: OK , export attachment (imasge): OK

- [x] ["Privacy": "Runtime permission for notifications"](https://developer.android.com/about/versions/13/behavior-changes-all#notification-permission): "Android 13 (API level 33) introduces a runtime [notification permission](https://developer.android.com/guide/topics/ui/notifiers/notification-permission): [POST_NOTIFICATIONS](https://developer.android.com/reference/android/Manifest.permission#POST_NOTIFICATIONS). This change helps users focus on the notifications that are most important to them."  
also see [below](https://github.com/deltachat/deltachat-android/pull/2649#issuecomment-1733929228)

- [x] add a device message if the user taps "deny notifications" 

- [x] handle "denied notifications" in the system settings

general aspects (unticked items need to be checked, please add a summary-IMPACT next to the items then):

- [x] ["Performance and battery": "Task Manager"](https://developer.android.com/about/versions/13/behavior-changes-all#fgs-manager): Apps must be able to [handle this user-initiated stopping](https://developer.android.com/guide/components/foreground-services#handle-user-initiated-stop)." - IMPACT: things seem to work as expected, when delta is restarted, the permanent notification reappears (users can disable it in the settings). system looks like the following: <br> 

- [x] ["Performance and battery": "Improve prefetch job handling using JobScheduler"](https://developer.android.com/about/versions/13/behavior-changes-all): IMPACT: JobScheduler seems to be unused by us

- [x] ["Performance and battery": "High Priority Firebase Cloud Message (FCM) Quotas"](https://developer.android.com/about/versions/13/behavior-changes-all#fcm-quotas): IMPACT we do not use FCM 

- [x] ["Performance and battery": "Battery Resource Utilization"](https://developer.android.com/about/versions/13/behavior-changes-all#battery-resource-utilization): there is a new battery mode "restricted" where "foreground services" cannot be launched (so, our permantent notification) - we should make sure, that we do not crash in this case - IMPACT: looks good at a first glance, a beta test with different, real devices  will tell more

- [x] ["Privacy": "Hide sensitive content from clipboard"](https://developer.android.com/about/versions/13/behavior-changes-all#copy-sensitive-content) - IMPACT: should not affect us

- [x] ["Security": "Intent filters block non-matching intents"](https://developer.android.com/about/versions/13/behavior-changes-all#intents) - IMPACT: i cannot really make sense of the link, however, in practise, things work on first tries :)

- [x] ["Security": "Migrate away from shared user ID"](https://developer.android.com/about/versions/13/behavior-changes-all): IMPACT: we're not using android:sharedUserId

- [x] ["User experience": "Dismissible foreground service notifications"](https://developer.android.com/about/versions/13/behavior-changes-all#dismissible-fgs-notifs) - not sure if this is the same as _"Performance and battery": "Task Manager"_ above - IMACT: assuming, the foreground service is still active if the notification is dismissed, this is a nice feature. iirc, hiding/dismissing foreground notifications was also possible before. however, if it turns out, the dismissed notification affects the service, we should call [setOngoing()](https://developer.android.com/reference/android/app/Notification.Builder#setOngoing(boolean)) 

- [x] ["Core functionality": "Legacy copy of speech service implementation removed"](https://developer.android.com/about/versions/13/behavior-changes-all#speech-service) - IMPACT: we're not using RecognitionService or SpeechService

closes https://github.com/deltachat/deltachat-android/issues/2599

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -14,7 +14,12 @@
 
     <!-- dangerous permissions - we need to as the user with a PermissionsRequest -->
     <uses-permission android:name="android.permission.READ_CONTACTS" />
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
+    <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
+    <uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
+    <uses-permission android:name="android.permission.READ_MEDIA_AUDIO" />
+    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" android:maxSdkVersion="29" />
+
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION"/>
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -17,7 +17,7 @@
     <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
     <uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" android:maxSdkVersion="29" />
-
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION"/>
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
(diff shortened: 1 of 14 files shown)


adbenitez:
I am trying this, for me I had notifications disabled for $reasons (to avoid annoying notifications from my dev/test DC installation) after upgrading I could not enable notifications, in system settings the toggle to enable/disable notifications was disabled, after reverting to target 32 when opening the app I see the popup dialog requesting notifications permissions, maybe something changed there and the way the app is requesting notifications permissions is not compatible with target 33

adbenitez (approved):
:rocket:

PR #2653, opened 2023-09-18
Ask for the name when creating a broadcast list

Hocuri:
Before https://github.com/deltachat/deltachat-core-rust/pull/4644, when creating a broadcast list, the UIs didn't ask for a name since the name wasn't shown to the recipients and therefore not that important.

As of https://github.com/deltachat/deltachat-core-rust/pull/4644, broadcast lists will create their own chat for the recipients, showing the broadcast lists's name, so we should ask the user for a name when creating a broadcast list.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -376,7 +376,7 @@
     </plurals>
     <!-- The placeholder will be replaced by the name of the recipient in a one-to-one chat. -->
     <string name="chat_new_one_to_one_hint">Send a message to %1$s; they can even receive it without Delta Chat.</string>
-    <string name="chat_new_broadcast_hint">In a broadcast list, the recipients will receive messages in a read-only chat.\n\nThey will not be see the other recipients.</string>
+    <string name="chat_new_broadcast_hint">In a broadcast list, recipients will receive messages in a read-only chat with you.</string>
     <string name="chat_new_group_hint">Others will only see this group after you sent a first message.</string>
     <string name="chat_record_slide_to_cancel">Slide to cancel</string>
     <string name="chat_record_explain">Tap and hold to record a voice message, release to send</string>
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -376,7 +376,7 @@
     </plurals>
     <!-- The placeholder will be replaced by the name of the recipient in a one-to-one chat. -->
     <string name="chat_new_one_to_one_hint">Send a message to %1$s; they can even receive it without Delta Chat.</string>
-    <string name="chat_new_broadcast_hint">In a broadcast list, others will receive messages in their private chats with you.\n\nThe recipients will not be aware of each other.</string>
+    <string name="chat_new_broadcast_hint">In a broadcast list, the recipients will receive messages in a read-only chat.\n\nThey will not be see the other recipients.</string>
     <string name="chat_new_group_hint">Others will only see this group after you sent a first message.</string>
     <string name="chat_record_slide_to_cancel">Slide to cancel</string>
     <string name="chat_record_explain">Tap and hold to record a voice message, release to send</string>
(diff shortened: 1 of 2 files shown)


res/values/strings.xml
```diff
@@ -376,7 +376,7 @@
     </plurals>
     <!-- The placeholder will be replaced by the name of the recipient in a one-to-one chat. -->
     <string name="chat_new_one_to_one_hint">Send a message to %1$s; they can even receive it without Delta Chat.</string>
-    <string name="chat_new_broadcast_hint">In a broadcast list, others will receive messages in their private chats with you.\n\nThe recipients will not be aware of each other.</string>
+    <string name="chat_new_broadcast_hint">In a broadcast list, the recipients will receive messages in a read-only chat.\n\nThey will not be see the other recipients.</string>
```
r10s, line 379: i think, it should be either "They will not _be seen by_ the other recipients." or  "They will _not see_ the other recipients."

however, to exclude discussions wrt BCC and encryption, maybe just say "In a broadcast list, recipients will receive messages in a read-only chats with you" - i think, this is exactly what is important for the one creating the list. for the recipient things are clear anyway as there is no member list

Hocuri: :+1:, just "chat" instead of "chats":

```suggestion
    <string name="chat_new_broadcast_hint">In a broadcast list, recipients will receive messages in a read-only chat with you</string>
```

Hocuri: I do think they are the "same" from the view of a user, since they look exactly the same for each recipient (or I'm understanding something wrongly)

r10s: > I do think they are the "same" from the view of a user

i try to make clear what i was meaning by the following example:

1. the chats are the same form the view of the broadcast-recipients - "Me and the Sender" - broadcast receivers probably won't think much about that anyways (they're soft of passive consumers)

2. but from the view of the broadcast-sender, all recipient's chats are different on the recipient's devices - "RecipientA and Me", "RecipientB and Me", "RecipientC and Me" - which is again different from the sender's chat as "Me and RecipientA, RecipientB and RecipientC"

as the string is shown only in the case of 2., my idea was to use the plural form.

however, maybe this is a weird view and not helpful. as there is so much to explain, it is probably already a bad idea :) as said, it is fine as it is now, no need to change things unless you share this idea :) feel free to ignore my comment completely :)

Hocuri:
@r10s Since we won't have a `master` and `stable` branch anymore now. I guess this means that Android `master` goes with Core `main` again (and the nightlies are built like this), and we merge this together with the Core PR?

r10s:
idea is that current "stabe" will be new "main" (as old "master") and use stable cores then. releases and translations will be done from "main"

core-master can be used in feature-branches (however, before being merged, should not contain a submodule usually)

so, to sum up: everything back to what we did the last years before introducing "stable/master"

res/values/strings.xml
r10s: unresolving - the current state does not reflect @Hocuri nor mine suggestion - which are 99% the same beside plural/singular. i would be fine with both

r10s: i applied the suggestion of @Hocuri - to move forward, but also, maybe the plural of my suggestion opens more questions than it answers, i meanwhile would also slightly prefer it :)

r10s (approved):
does as expected, thanks.

the avatar is currently not persisted, however, this is solved probably by https://github.com/deltachat/deltachat-core-rust/pull/4644 

nb: i rebased the pr

Hocuri:
Copying the "Delta Chat Dev" comment here:

I just spent a few hours trying to make avatars work with broadcast lists, and, it's surprisingly much that has to change. 

One thing that's not easy to fix:

In groups, we're sending out a system message everytime we add a member or change the avatar. I would prefer if we didn't need to do this for broadcast lists, esp. it's weird if everyone else gets a message after a member change in a broadcast list.

However, then we will need to either (i) send the avatar with every outgoing message or (ii) remember whether there are members in the group that didn't get the avatar yet.

That's why we decided to just leave Core as-is, and keep disallowing to change the profile image of broadcast lists for now (i.e. make [https://github.com/deltachat/deltachat-android/pull/2653](https://github.com/deltachat/deltachat-android/pull/2653#) only about the chat name, and still don't allow to change the avatar). This way, we can be sure not to introduce any new bugs right before releasing and while we would like to concentrate on OTF.

PR #2655, opened 2023-09-18
ignore title click when relaying content

adbenitez:
don't open connectivity activity when title is clicked if the program is relaying content (forwarding, sharing, etc)


Hocuri (approved):
Thanks!

r10s:
the unrelated API change in combination with different APIs on stable/master and the need to merge to stable and from there to master still make troubles, btw https://github.com/deltachat/deltachat-core-rust/pull/4314#issuecomment-1732535864 -

PR #2659, opened 2023-09-24
style: add dark theme for webxdc wrapper

WofWca:
~~With webxdc~~ it affects the "loading" stage

Related: https://github.com/deltachat/deltachat-desktop/pull/3411

I have not tested it live. ~~I only checked the color on the help page with dev tools on Desktop. The contrast is better than it is in light mode.~~

### DIFF
--- a/res/raw/webxdc_wrapper.html
+++ b/res/raw/webxdc_wrapper.html
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
+    <meta name="color-scheme" content="dark light">
     <style>
       html,
       body {


res/raw/webxdc_wrapper.html
```diff
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
+    <meta name="color-scheme" content="dark light">
```
r10s, line 4: this is for avoiding some flickering? probably okay as long as this affects the loader only and not the loaded iframe.

for the webxdc as such, dark mode cannot be forced from outside, cmp. https://github.com/deltachat/deltachat-android/pull/2208

assets/help/cs/help.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html><head><meta charset="UTF-8" /><meta name="viewport" content="initial-scale=1.0" /><link rel="stylesheet" href="../help.css" /></head><body><ul id="top">
+<html><head><meta charset="UTF-8" /><meta name="color-scheme" content="dark light"><meta name="viewport" content="initial-scale=1.0" /><link rel="stylesheet" href="../help.css" /></head><body><ul id="top">
```
r10s, line 2: why are the changes in help.html needed? does it make a difference? for me, also without this PR, i have dark mode on the help pages, even without flickering on opening.

res/raw/webxdc_wrapper.html
adbenitez: I guess this is to avoid the white screen with the FILL500 loading, I have same concerns as r10s wrt if this apply also to the contained iframe

WofWca: > as long as this affects the loader only and not the loaded iframe

That's right, it doesn't affect the inside of the `<iframe>`

assets/help/cs/help.html
WofWca: Hmmm yeah I checked just now, it really is already dark. I guess it's because of [this call to `setForceDark`](https://github.com/deltachat/deltachat-android/blob/920889c1193e826bc2851f2e5a45e6400b439b32/src/org/thoughtcrime/securesms/LocalHelpActivity.java#L16)?
But [the docs say that `setForceDark` is deprecated](https://developer.android.com/reference/android/webkit/WebSettings#setForceDark(int)), so I think this is a better way to go about this. Well, except for the part where I change the generated files.

WofWca: I moved it to CSS.

WofWca: Plus removed the `setForceDark` (again, I haven't tested it).

r10s (changes requested):
let's concentrate on webxdc in this pr and remove everyrhing else.

it is correct that `setForceDark()` is deprecated, however, the alternative with `<meta name="color-scheme" content="dark light">` may be worse supported on some systems or chrome versions (after all experience, i would at least not be surprised). 

also, there is at least one more usage of `setForceDark()` in the code, so we should replace them alltogether, however, we have time with that, currently, everything is working smoothly

r10s:
i cherry-pick'ed the commit from `stable` (we're about to rename `master` to `feat/verified-chats` and use `stable` as new default `main`, sorry for confusion)

r10s:
ftr, this `<meta name="color-scheme" content="dark light">` does not fix the flickering at all as the iframe where the webxdc is loaded in is always white and always present. so, still in darkmode, 95% of the screen are white.

to fix that, one may have to create the iframe hidden and show it later, but that also may result in issues, so, not sure if many further investigation are is worth it

WofWca:
I'm sorry

r10s:
@adbenitez already made a fix at https://github.com/deltachat/deltachat-android/pull/2713 :)

PR #2661, opened 2023-09-24
improve landscape mode for webxdc

### LINKED ISSUE #2652: webxdc-toolbar too big in landscape mode
some webxdc games need to be played in landscape mode, in landscape mode, the toolbar + system status bar takes a big part of the screen making it hard/impossible to play such games

adbenitez:
close #2652

### DIFF
--- a/res/values-land/attrs.xml
+++ b/res/values-land/attrs.xml
@@ -1,5 +1,4 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-  <!-- in small devices (e.i. not tablets) in landscape height is low, better enter fullscreen  -->
-  <bool name="webxdcFullscreen">true</bool>
+  <bool name="isLandscape">true</bool>
 </resources>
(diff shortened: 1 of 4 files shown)


r10s:
works great!

two questions pop up:

- what if the webxdc is _entered_ from landscape mode (this easily happens eg. when leaving a webxdc from landscape shortly, eg. the editor - if reentering, you get another view)

- what shall we do with tablet  that have landscape as their default mode? always removing the toolbar in this case may force ppl to portait mode to access eg. the menu. it probably would be better to just ignore the toolbar-less behaviour for tablets

adbenitez:
I reworked it with a simpler approach, now it should work as expected on orientation changes and if open directly on landscape, and it should never enter fullscreen mode in tablets or devices with big enough height, could not test since there is no tablet around tho

res/values-sw600dp/attrs.xml
adbenitez, line 1: check https://stackoverflow.com/a/9308284/4557005

res/values-land/attrs.xml
adbenitez, line 1: the "land" (landscape) qualifier has lower precedence than "sw600dp" so this will be ignored if it is in landscape but sw600dp (smaller width >= 600dp)  matches, see https://developer.android.com/guide/topics/resources/providing-resources.html#AlternativeResources

r10s (approved):

r10s:
nb:tested on phones and tablets emulators, works like a charme! so much better, also for the editor!

r10s (approved):

PR #2662, opened 2023-09-25
set rounded corners for webxdc app icons

adbenitez:


res/layout/webxdc_view.xml
```diff
@@ -12,12 +12,14 @@
                       android:gravity="center_vertical"
                       android:orientation="vertical">
 
-            <androidx.appcompat.widget.AppCompatImageView
+            <com.google.android.material.imageview.ShapeableImageView
                 android:id="@+id/webxdc_icon"
+                android:padding="0.1dp"
```
adbenitez, line 17: I don't know if I can add `<!-- -->` comments here, would be nice to tell that this is a hack needed to avoid the transparent background if padding is zero, for some weird reason I guess a bug in material design lib

res/values/styles.xml
```diff
@@ -113,4 +113,7 @@
         <item name="colorControlActivated">#ff2090ea</item>
     </style>
 
+    <style name="ShapeAppearanceOverlay.App.CornerSize10Percent" parent="">
+        <item name="cornerSize">10%</item>
```
r10s, line 117: i would use ~ 7% here, so that the _width_ of the curve between the icon and the bubble is roughly always the same; this looks more even.

10% vs 7%:

Screenshot 2023-09-25 at 13 56 48 Screenshot 2023-09-25 at 13 57 13

r10s (approved):
lgtm, i would only slightly adapt the percentage

for `com.google.android.material...` i double checked, we're using that dependency already a lot, so that should not add anything; it should also be fine for android4

PR #2663, opened 2023-09-25
disable web API for fullscreen

adbenitez:
disable web API for fullscreen mode since it is not supported by the webview


adbenitez:
it turns out this is not needed because document.fullscreenEnabled is false already in android webview without any modifications, apps just need to check that property instead of relaying on hacks like checking for document.exitFullscreen

PR #2667, opened 2023-09-25
add more context to dialog title shown after sendToChat()

### LINKED ISSUE #2594: add more context to dialog title shown after `sendToChat()`
the title should be set to `send_file_to` or `send_message_to`. in case the RelayHelper is not used from `sendToChat()`, the title should stay as is.

for more context and screenshots, see https://github.com/deltachat/deltachat-ios/pull/1907 and https://github.com/deltachat/deltachat-android/pull/2592

adbenitez:
close #2594

### DIFF
--- a/src/org/thoughtcrime/securesms/util/RelayUtil.java
+++ b/src/org/thoughtcrime/securesms/util/RelayUtil.java
@@ -14,6 +14,7 @@ public class RelayUtil {
     private static final String FORWARDED_MESSAGE_IDS   = "forwarded_message_ids";
     private static final String SHARED_URIS             = "shared_uris";
     private static final String IS_SHARING              = "is_sharing";
+    private static final String SHARING_TITLE           = "sharing_title";
     private static final String DIRECT_SHARING_CHAT_ID = "direct_sharing_chat_id";
 
     public static boolean isRelayingMessageContent(Activity activity) {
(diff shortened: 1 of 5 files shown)


r10s (approved):
the the getter/setter use the wording "shared title" while the extra is called "sharing title" - this is unnecessarily confusing, better use the same wording; so eg. use `SHARED_TITLE` as the name for the extra.

otherwise lgtm.

src/org/thoughtcrime/securesms/util/RelayUtil.java
```diff
@@ -79,6 +80,14 @@ public static String getSharedText(Activity activity) {
         }
     }
 
+    public static String getSharedTitle(Activity activity) {
+        try {
+            return activity.getIntent().getStringExtra(SHARING_TITLE);
```
adbenitez, line 85:
```suggestion
            return activity.getIntent().getStringExtra(SHARED_TITLE);
```

src/org/thoughtcrime/securesms/util/RelayUtil.java
```diff
@@ -123,6 +132,10 @@ public static void setSharedText(Intent composeIntent, String text) {
         composeIntent.putExtra(IS_SHARING, true);
     }
 
+    public static void setSharedTitle(Intent composeIntent, String text) {
+        composeIntent.putExtra(SHARING_TITLE, text);
```
adbenitez, line 136:
```suggestion
        composeIntent.putExtra(SHARED_TITLE, text);
```

adbenitez:
> the the getter/setter use the wording "shared title" while the extra is called "sharing title" - this is unnecessarily confusing, better use the same wording; so eg. use `SHARED_TITLE` as the name for the extra.

good point, fixed!

PR #2669, opened 2023-09-26
remove ts fragments from webxdc.js

r10s:
main reason is, that these annotations cover the code, making it hard to follow what is going on.

maybe opionionated, however,
in general, .ts is not needed in small projects,
and an additional burden for devs to keep them up to date


r10s:
approved verbally

PR #2673, opened 2023-09-26
update progress only for the actually playing AudioSlide

### LINKED ISSUE #1288: audio seekbar may show wrong position
when an audio is played and scrolled up, other audios have the progress bar moving like if it is playing. maybe because of android list recycling elements.

reported on "Testing Delta Chat" group.

r10s:
closes #1288

### DIFF
--- a/src/org/thoughtcrime/securesms/components/AudioView.java
+++ b/src/org/thoughtcrime/securesms/components/AudioView.java
@@ -122,7 +122,7 @@ public class AudioView extends FrameLayout implements AudioSlidePlayer.Listener 
 
     if (seekBar.getProgress() + 5 >= seekBar.getMax()) {
       backwardsCounter = 4;
-      onProgress(0.0, -1);
+      onProgress(audioSlidePlayer.getAudioSlide(), 0.0, -1);
     }
   }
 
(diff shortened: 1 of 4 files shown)


src/org/thoughtcrime/securesms/components/AudioView.java
```diff
@@ -79,6 +79,7 @@ public void setAudio(final @NonNull AudioSlide audio, int duration)
   {
     controlToggle.displayQuick(playButton);
     seekBar.setEnabled(true);
+    seekBar.setProgress(0);
```
adbenitez, line 82: this fixes the wrong scrolled/advanced seekbar on recycling the message view, but sadly due to the way AudioSlidePlayer work, there is no per-message progress that can be tracked statically, so a paused message will have its seekbar reset to initial position if the item is recycled, this would be solved when we add a audio mini-player,

overall it is better not to see old items getting a wrong seekbar than a paused audio losing its state because we scrolled to the old chat history

src/org/thoughtcrime/securesms/mms/AudioSlide.java
```diff
@@ -47,6 +47,13 @@ public AudioSlide(Context context, Uri uri, long dataSize, String contentType, b
     super(context,  new UriAttachment(uri, null, contentType, AttachmentDatabase.TRANSFER_PROGRESS_STARTED, dataSize, 0, 0, null, null, voiceNote));
   }
 
+  @Override
+  public boolean equals(Object other) {
```
adbenitez, line 51: this fixes the issue of two different audio messages being considered equal and hence the seekbar and progress listener being set to an old audio message that is not the one being played

r10s (commented):
lgtm

adbenitez (approved):

PR #2674, opened 2023-09-27
use system camera as default

r10s:
the system camera has usually more features (flash, physical zoom, filter) and is less buggy than the mostly unmaintained built-in one. also, the system camera seems to be more supported (isDeviceSupported() is not needed) and does not require to ask for permission

there is still a switch in "advanced settings" to use the old built-in camera, however, at some point we may remove the old code completely

@adbenitez can you please verify that on your device the permission is not needed by disabling it beforehand?


hpk42:
indeed, i hasve increasingly often switchedf to system camera and then shared an image from there, so being able to use system camera these days from dc is nice!

PR #2676, opened 2023-09-28
display reactions details

### LINKED ISSUE #2675: show reaction details
a tap in the reaction line (added in #2401) shall open reaction details (cmp. signal, whatsapp)

some implementation also allow to send reactions from there, however we do not want that but go for a unique way for sending reactions that is the same if there are no reactions at all (long tap, see #2402).

successor of #2401

adbenitez:
close #2675

### DIFF
--- a/res/layout/reaction_recipient_item.xml
+++ b/res/layout/reaction_recipient_item.xml
@@ -1,0 +1,81 @@
+<?xml version="1.0" encoding="utf-8"?>
+<org.thoughtcrime.securesms.reactions.ReactionRecipientItem
+        xmlns:android="http://schemas.android.com/apk/res/android"
+        xmlns:app="http://schemas.android.com/apk/res-auto"
+        xmlns:tools="http://schemas.android.com/tools"
+        android:layout_width="match_parent"
+        android:layout_height="60dp"
+        android:orientation="horizontal"
+        android:gravity="center_vertical"
+        android:focusable="true"
+        android:background="?attr/conversation_list_item_background"
+        android:paddingLeft="16dp"
+        android:paddingStart="16dp"
+        android:paddingRight="10dp"
+        android:paddingEnd="10dp">
+
+  <org.thoughtcrime.securesms.components.AvatarImageView
+      android:id="@+id/contact_photo_image"
+      android:layout_width="40dp"
+      android:layout_height="40dp"
+      android:foreground="@drawable/contact_photo_background"
+      android:cropToPadding="true"
+      app:layout_constraintBottom_toBottomOf="parent"
+      app:layout_constraintStart_toStartOf="parent"
+      app:layout_constraintTop_toTopOf="parent"
+      tools:src="@color/blue_600"
+      android:contentDescription="@null" />
+
+    <LinearLayout android:layout_width="0dp"
+                  android:layout_height="wrap_content"
+                  android:layout_weight="1"
+                  android:layout_marginLeft="11dp"
+                  android:layout_marginStart="11dp"
+                  android:paddingRight="16dp"
+                  android:paddingEnd="16dp"
+                  android:orientation="vertical">
+
+        <org.thoughtcrime.securesms.components.emoji.EmojiTextView
+            android:id="@+id/name"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:drawablePadding="5dp"
+            android:ellipsize="marquee"
+            android:singleLine="true"
+            android:fontFamily="sans-serif"
+            android:textSize="16sp"
+            tools:text="Frieeeeeeedrich Nieeeeeeeeeetzsche" />
+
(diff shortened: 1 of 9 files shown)


adbenitez:
preview:

image

r10s:
it even updates when reaction change! 👌

res/values/strings.xml
```diff
@@ -722,6 +722,9 @@
     <string name="emoji_symbols">Symbols</string>
     <string name="emoji_flags">Flags</string>
 
+    <!-- Reactions -->
+    <string name="reactions_details_title">Reactions</string>
```
r10s, line 726: just call it `reactions` - if we need to become more specific, we do that as needed

res/layout/reaction_recipient_item.xml
```diff
@@ -0,0 +1,81 @@
...
+        android:id="@+id/reaction"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:gravity="center"
+        android:textSize="22dp"
+        android:fontFamily="sans-serif"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        tools:text=":)" />
```
r10s, line 79: the padding right to the reaction should be the same as left of the avatar

to me, it looks too small

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -867,6 +869,12 @@ public void onShowFullClicked(DcMsg messageRecord) {
       public void onDownloadClicked(DcMsg messageRecord) {
         dcContext.downloadFullMsg(messageRecord.getId());
       }
+
+      @Override
+      public void onReactionClicked(DcMsg messageRecord) {
+        ReactionsDetailsFragment dialog = new ReactionsDetailsFragment(messageRecord.getId());
+        dialog.show(((FragmentActivity) getActivity()).getSupportFragmentManager(), null);
```
r10s, line 876: the cast to `FragmentActivity` activity is not needed say android studio, and indeed it works without.

at line 50, `import androidx.fragment.app.FragmentActivity;` can be removed as well then

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -680,6 +680,10 @@ private void setReactions(@NonNull DcMsg current) {
         reactionsView.clear();
       } else {
         reactionsView.setReactions(reactions.getReactions());
+        reactionsView.setOnClickListener(v -> {
+          if (eventListener == null) return;
+          eventListener.onReactionClicked(current);
```
r10s, line 685: the following style is used most often in this project ...

```suggestion
          if (eventListener != null) {
            eventListener.onReactionClicked(current);
          }
```

... and it is indeed easier to read in many cases as there are no different exit cases

src/org/thoughtcrime/securesms/reactions/ReactionRecipientsAdapter.java
```diff
@@ -0,0 +1,105 @@
...
+
+import java.util.ArrayList;
+
+public class ReactionRecipientsAdapter extends RecyclerView.Adapter
+{
+  @SuppressWarnings("unused")
+  private final static String TAG = ReactionRecipientsAdapter.class.getSimpleName();
+
+  private @NonNull ArrayList<Pair<Integer, String>> contactsReactions = new ArrayList<>();
+  private final LayoutInflater                li;
```
r10s, line 24: while using short abbreviations if they're used only at a scope of few lines is okay (loops), for member variables, this is not-so-good style - eg. many will read this as "list item" at a first glance :)

better be verbose here and call the things what it is:


```suggestion
  private final LayoutInflater                layoutInflater;
```

(usages need to be adapted as well)

r10s (approved):
lgtm, good job! some minor comments only

r10s:
another tiny thing: the reactions details should not open in multiselect-mode - instead, just select/deselect items (pass the tap to the parent or so)

adbenitez:
> another tiny thing: the reactions details should not open in multiselect-mode - instead, just select/deselect items (pass the tap to the parent or so)

done :heavy_check_mark:

r10s:
this pr needs a rebase on stable before being able to being merged

adbenitez:
mmmh github shows no conflict :thinking:

r10s:
> mmmh github shows no conflict 🤔

this is how github looks to me:

Screenshot 2023-09-30 at 16 06 30

adbenitez:
LOL :dizzy_face:  this is how it looks to me:
image

adbenitez:
well merging worked :thinking: maybe some github glitch on your side

r10s:
anyway, it worked :)

PR #2677, opened 2023-09-28
show 'verified by SELF' as such

r10s:
core was changed to return SELF from `get_verifier_id()` at https://github.com/deltachat/deltachat-core-rust/pull/4754 - until then, core will return always a contact-id,
which is equal to the current state


r10s:
once this is merged and the string is uploaded, we should create an issue for desktop

adbenitez (approved):

PR #2680, opened 2023-09-29
sending reactions

### LINKED ISSUE #2402: send reactions
sending reactions should be similar to what is discusssed on https://github.com/deltachat/deltachat-ios/issues/1728, however, as we by default use the multi-select-paradigm of android, we should keep the main title bar menu as is and show an additional reaction "area".

this is also what whatsapp is doing:

- the real set of reactions needs to be discussed, assume a handful of default reactions plus the option to pick any emoji.
- if the user already reacted, their reaction should be hilited

cmp. https://github.com/deltachat/deltachat-core-rust/pull/3644, https://github.com/deltachat/deltachat-desktop/pull/2964

## timing

there is no urgency, first we'll need release that uses a core that processes reactions (core >= 98), also #2401 should be implemented first

r10s:
next steps, roughly:

- [x] hide the picker by default, open/close it together with the ActionMode
- [x] hide the picker on scrolling away more than a few pixels - alternatively, make the picker sticky
- [x] make taps on the picker actually send the reaction
- [x] move the picker to the correct position, nearby the bubble
- [x] if there is already an reaction, highlight it - a tap will remove it

closes #2402

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationFragment.java
@@ -63,6 +63,7 @@ import org.thoughtcrime.securesms.connect.DcEventCenter;
 import org.thoughtcrime.securesms.connect.DcHelper;
 import org.thoughtcrime.securesms.database.Address;
 import org.thoughtcrime.securesms.mms.GlideApp;
+import org.thoughtcrime.securesms.reactions.AddReactionView;
 import org.thoughtcrime.securesms.recipients.Recipient;
 import org.thoughtcrime.securesms.reactions.ReactionsDetailsFragment;
 import org.thoughtcrime.securesms.util.AccessibilityUtil;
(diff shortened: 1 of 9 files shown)


adbenitez (approved):

src/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -251,6 +254,11 @@ public void onNewIntent() {
             reloadList();
             updateLocationButton();
         }
+
+        // hack (alternatives welcome): we position the AddReactionView offscreen to "-666dp"
+        // to get the real width/height before it is shown;
+        // after first layout we can hide it so that eg. screen readers are not confused
+        addReactionView.hide();
```
adbenitez, line 261: doesn't `View.INVISIBLE` works like that? the docs[*] say:
> This view is invisible, but it still takes up space for layout purposes.

so I assume the layout size is calculated, this would avoid that weird hack

`*` https://developer.android.com/reference/android/view/View.html#INVISIBLE

r10s: thanks for the hint, i'll check that out!

r10s:
merging that in now, it is still a bit away from shipping publicly as we (1) want an official version that display things correctly before sending and (2) give other OS time to catch up

also, i'd like to (3) improve the layout, however, this is not a blocker, it is not bad as it is currently

PR #2681, opened 2023-09-29
allow to select key to import instead of using Download folder

### LINKED ISSUE #2658: [core 1.123.0] Import keys from the file rather than directory
Core 1.123.0 will allow using secret key `.asc` file as `param1` path for `DC_IMEX_IMPORT_SELF_KEYS`: https://github.com/deltachat/deltachat-core-rust/pull/4737

Directory selection dialog should be replaced with a file selection dialog.

This is going to solve the problem reported on the forum: https://support.delta.chat/t/deltachat-fails-to-import-keys/2592

adbenitez:
close #2658

### DIFF
--- a/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
+++ b/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
@@ -420,7 +420,6 @@ public class AttachmentManager {
     return deck;
   }
 
-
   public static @Nullable String getFileName(Context context, Uri uri) {
     String result = null;
     if (uri.getScheme().equals("content")) {
--- a/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
+++ b/src/org/thoughtcrime/securesms/mms/AttachmentManager.java
@@ -65,7 +65,6 @@ import org.thoughtcrime.securesms.permissions.Permissions;
 import org.thoughtcrime.securesms.providers.PersistentBlobProvider;
 import org.thoughtcrime.securesms.scribbles.ScribbleActivity;
 import org.thoughtcrime.securesms.util.MediaUtil;
-import org.thoughtcrime.securesms.util.StorageUtil;
 import org.thoughtcrime.securesms.util.ThemeUtil;
 import org.thoughtcrime.securesms.util.ViewUtil;
 import org.thoughtcrime.securesms.util.guava.Optional;
@@ -419,6 +418,25 @@ public class AttachmentManager {
     SlideDeck deck = new SlideDeck();
     if (slide.isPresent()) deck.addSlide(slide.get());
     return deck;
+  }
+
+
+  public static @Nullable String getFileName(Context context, Uri uri) {
+    String result = null;
+    if (uri.getScheme().equals("content")) {
+      Cursor cursor = context.getContentResolver().query(uri, new String[]{OpenableColumns.DISPLAY_NAME}, null, null, null);
+      try {
+        if (cursor != null && cursor.moveToFirst()) {
+          result = cursor.getString(cursor.getColumnIndexOrThrow(OpenableColumns.DISPLAY_NAME));
+        }
+      } finally {
+        if (cursor != null) cursor.close();
+      }
+    }
+    if (result == null) {
+      result = uri.getLastPathSegment();
+    }
+    return result;
   }
 
   public static void selectDocument(Activity activity, int requestCode) {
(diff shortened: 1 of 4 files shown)


src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
```diff
@@ -330,11 +371,15 @@ private void exportKeys() {
                           .show();
                     }
                     else {
+                        String explain = getActivity().getString(R.string.pref_managekeys_import_explain_tips);
+                        if (Build.VERSION.SDK_INT < 30) { // in older Android we always import from Download folder
+                          explain = getActivity().getString(R.string.pref_managekeys_import_explain_path, DcHelper.getImexDir().getAbsolutePath()) + "\n\n" + explain;
+                        }
                       new AlertDialog.Builder(getActivity())
```
r10s, line 378: sth. is wrong with indentation here - `String explain ...` and `new AlertDialog ...` should have the same indentation

src/org/thoughtcrime/securesms/preferences/ListSummaryPreferenceFragment.java
```diff
@@ -100,10 +100,17 @@ protected void initializeListSummary(ListPreference pref) {
 
   protected ProgressDialog progressDialog = null;
   protected int            progressWhat = 0;
-  protected String         imexDir = "";
+  protected String         pathName = ""; // path/filename displayed to user on finish
   protected boolean        imexUserAborted = false;
```
r10s, line 104: - these 4 variables are helpers for the second `startImex()` variation and should be moved down (so before `protected void startImex(int what, String imexPath, String pathName)`)

- i find the comment confusing. better drop the comment and improve the variable name to sth. as `pathAsDisplayedToUser` - then things are also clear when the variable is accessed
```

File: src/org/thoughtcrime/securesms/mms/AttachmentManager.java [hunk 3]
```diff
@@ -421,6 +420,24 @@ public boolean isAttachmentPresent() {
     return deck;
   }
 
+  public static @Nullable String getFileName(Context context, Uri uri) {
```
r10s, line 423: thanks for diving into that mess 🤿

r10s (dismissed):
some comments

r10s:
wait: i am just checking desktop:

why no just do the same as there:

- open file selector first, if supported, then ask "Import secret keys from "foobar.asc"?\n\n• Existing secret keys will not be deleted ..."  
(if file selector is not supported do the same as before and ask "Import secret keys from "/path/to/download/folder"?\n\n• Existing secret keys will not be deleted ..."  

- after import say "Secret keys imported from "foobar.asc"

the tips do not really explain that a file selector follows, so it feels more natural afterwards.

also, that way, no new strings are needed

adbenitez:
> why no just do the same as there

done :heavy_check_mark:

r10s (approved):

monperrus:
Hi all, I'm trying to import a public key of a contact, but cannot find the feature, is that possible? thanks!

r10s:
Delta Chat uses [Autocrypt](https://docs.autocrypt.org) and does not want the users be think about keys at all :) - therefore, no, this is not possible, however, this was considered here and then and  of course, this is a topic sometimes.

in general, for feature requests and such discussions, better use the [support forum](https://support.delta.chat)

monperrus:
thanks @r10s 

the feature request is https://support.delta.chat/t/import-and-use-of-existing-gpg-keys/3072/2

importing public GPG keys of contacts is necessary in order to be interoperable with protonmail and gmx/mailevlope.

iequidoo:
> the feature request is https://support.delta.chat/t/import-and-use-of-existing-gpg-keys/3072/2
> 
> importing public GPG keys of contacts is necessary in order to be interoperable with protonmail and gmx/mailevlope.

Btw, i even created a task for this previously: https://github.com/deltachat/deltachat-core-rust/issues/5698. Probably it should be closed and better discussed on the forum first, but the fact is that Delta Chat already imports keys from attachments in incoming emails, so technically it's very doable, just allow the user to trigger the same logic by hand. Though anyway it's a new functionality that should be maintained and covered with tests.

monperrus:
> technically it's very doable, just allow the user to trigger the same logic by hand

great, that would be so good for encryption interoperability. thanks!

PR #2684, opened 2023-09-30
fix welcome title bar

### LINKED ISSUE #2682: title bar on "Welcome Screen"
after the refactoring of https://github.com/deltachat/deltachat-android/pull/2671, there is an unexpected title bar on the welcome screen reading "Delta Chat"

while we should check, why this happens and if there is an underlying bug, we may decide to keep the title bar:

- we are using a similar title bar on ios already - the titles can read "Welcome to Delta Chat" resp. "Add Account" later
- we can have a real back button in case of "Add Accout"
- we can have a kebab menu; this may be useful at some later point when it comes to instant-accounts and we do not have lots of buttons directly on the screen

r10s:
this was overseen at #2671

closes #2682


adbenitez (approved):
danke!!!

r10s:
bitte!!!

PR #2686, opened 2023-09-30
pause background music when playing voice messages

### LINKED ISSUE #2091: Pause background music when playing voice message
**Expected**: When another app plays music (like spotify or vlc) the music should be paused when starting to play a voice message.

**Actual**: Voice message is played without pausing the other music, which means that the voice message's content is not clearly understandable most of the time.

There is also the question whether to resume the music after the voice message has finished, but we don't need to tackle it in this issue:
Sometimes you may want to keep the pause, for example when you're listening to a chat full of voice messages, other times you might want the music to continue afterwards again so you don't have to manually press the resume button it.

adbenitez:
close #2091

### DIFF
--- a/src/org/thoughtcrime/securesms/components/AudioView.java
+++ b/src/org/thoughtcrime/securesms/components/AudioView.java
@@ -6,6 +6,7 @@ import android.content.res.ColorStateList;
 import android.graphics.PorterDuff;
 import android.graphics.Rect;
 import android.graphics.drawable.AnimatedVectorDrawable;
+import android.media.AudioManager;
 import android.os.Build;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/components/AudioView.java
```diff
@@ -223,6 +224,16 @@ public void onClick(View v) {
...
+          AudioManager am = (AudioManager)getContext().getSystemService(Context.AUDIO_SERVICE);
+          AudioManager.OnAudioFocusChangeListener focusChangeListener = new AudioManager.OnAudioFocusChangeListener() {
+              @Override
+              public void onAudioFocusChange(int focusChange) {
+                  if (focusChange == AudioManager.AUDIOFOCUS_LOSS) {
+                      pauseButton.performClick();
+                  }
+              }
+          };
+          am.requestAudioFocus(focusChangeListener, AudioManager.STREAM_MUSIC, AudioManager.AUDIOFOCUS_GAIN);
```
r10s, line 236: ftr, `requestAudioFocus(listener, streamType, durarionHint)` is deprecated. instead, `requestAudioFocus(AudioFocussRequeset)` shall be used: https://developer.android.com/reference/android/media/AudioManager#requestAudioFocus(android.media.AudioManager.OnAudioFocusChangeListener,%20int,%20int)


however, i am fine with the old function, it may take a long time until it actually is deleted, and so we can shift this work to the future (and even save some if at the end the new replacement will be replaced by sth else :)

r10s (approved):
nice!

Issue #2546, opened 2023-04-19
webxdc app doesn't load if you tap it 2+ times very quickly


WofWca:
- Android version:
- Device:
- Delta Chat version: 1.36.4
- Expected behavior: Only one instance of the webxdc app launches, no matter how many times I tap on it, and the app loads.
- Actual behavior: Two instances of the app get launched, and they both get stuck at 50%.
- Steps to reproduce the problem:
  1. Send a webxdc app to a chat
  2. Tap rapidly 2+ times (before the app window blocks the view so you can't tap it anymore)
- Screenshots:
- Logs:

Here's the code where the progress bar is updated:

https://github.com/deltachat/deltachat-android/blob/2611a677fc7ad000a8bc8a80234a7f3be9bfdf3d/res/raw/webxdc_wrapper.html#L67

r10s:
> Two instances of the app get launched

are you sure, really two instances get launched? [^1]

the behaviour you describes happens to me as well sometimes, however the reason was not two instances of the webxdc, but that FILL500 get stock sometime. the slower a phone is, the easier it is to get to that state. it could get to the issue also when eg. exit a webxdc quick and re-enter

[^1]: you can check this if if you hit "back" on the webxdc, you see a exit-animation and are in the same webxdc again. next "back" will bring you to the chat

adbenitez:
![image](https://github.com/deltachat/deltachat-android/assets/24558636/7d89fd6d-135f-4333-b7a1-26f74a082185)

this is what happens when you open and go back in the webxdcActivity, onDestroy() is called rather slowly when you are already back and the conversation, and if you open the app again fast, as you can see at the end of the logs, I managed to launch a new instance of the activity before the onDestroy() of the other one was called, so at the time the new one is running, the onDestroy() of the old one has not run yet, meaning that webxdc.destroy() has not being called yet so the fill500 is still in place

something like this might be what happened for wofwca

WofWca:
> are you sure, really two instances get launched?

Yes. There are two apps that I have to close.
If you have a fancy phone, try with two fingers XD

> but that FILL500 get stock sometime

Yes, it happens to me as well, I believe it's because the old instance's process doesn't get cleaned up in time and is reused by the new app launch.

r10s:
k, so there are two independent issues

[...]

PR #2689, opened 2023-09-30
no more than one webxdc start within 500ms

r10s:
it seems, fast tapping may result in starting a webxdc twice on some systems, https://github.com/deltachat/deltachat-android/issues/2546, i cannot reproduce this, but the timing-check introduced by this PR may help

@WofWca there will be an `.apk` created by CI soonish - can you please check if that fixes your issue and answer in this PR then?

### DIFF
--- a/src/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -63,13 +63,15 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
   private boolean internetAccess = false;
 
   public static void openWebxdcActivity(Context context, DcMsg instance) {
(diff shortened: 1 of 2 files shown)


adbenitez (approved):

WofWca:
> can you please check if that fixes your issue and answer in this PR then?

I still have easily managed to launch two instances on my older phone, because it took more than 500ms to load the webxdc app, but yeah, when it gets warmed up then it becomes less than 500ms so I can't launch two.

r10s:
> I still have easily managed to launch two instances on my older phone, because it took more than 500ms to load the webxdc app, but yeah

the 500ms are not related to webxdc startup, of course that can easily take longer with FILL500.

the 500ms are related to not allowing a tap until the new activity covers the button - so, it seems your system is not able to open activities within 500ms

however, it's merged and it seems to  improves things. let's close this issue, we cannot do more as still no android dev can reproduce  the issue on any phone (btw, i _only_ have old phones for testing :)

PR #2690, opened 2023-09-30
improve video error handling and logging

### LINKED ISSUE #1358: show a clear error message, if video recording is not available

r10s:
closes #1358 by picking up ideas from #1358

this pr does not translate error messages, let's first settle them in english by practise ... also, which ones really appear in the wild


adbenitez (approved):
nice! another old issue fixed :100:

PR #2692, opened 2023-10-01
sanitize filename before creating it in the blobs dir

### LINKED ISSUE #2248: Can't share images with "invalid" characters in name
- Delta Chat version: Current master
  - Create an image with an "invalid" character in the file name like `/` or `\`. (For some reason, my built-in camera app sometimes creates names like `IMG_3874\.jpg`.)
  - Click "share", select DC, and select a chat. Expected: The image is set as a draft. Actual: It's not.

Ping @r10s can you check whether there is a similar issue in iOS, and if not how you avoid it?

**Reason**
There's just never any code that would remove the invalid character. The Android code directly creates a file in the blob dir (`handleSharing()` -> `createMessage()` -> `getRealPathFromUri()` -> `getBlobdirFile()`) and sets it on the message with `set_file()`, which is a simple setter. Then when it calls `set_draft()`, the core errors `failed to set draft: Blob has a badname /data/.../IMG_394\.jpg`.

**Proposed solution**

Expose `BlobObject::create()` via the FFI and let Android use it instead of `getBlobdirFile()`. Ping @link2xt @flub as this would be a Core change.

**Alternative solutions**

- In DC-Android: Directly fix `getBlobdirFile()`, implementing a logic similar to what `BlobObject::create()` does.
- In DC-Core: Let `set_draft()` or `set_file()` repair the filename, copying the file (or moving if we are sure it's not needed under the old name)

adbenitez:
close #2248

### DIFF
--- a/src/org/thoughtcrime/securesms/util/FileUtils.java
+++ b/src/org/thoughtcrime/securesms/util/FileUtils.java
@@ -1,9 +1,49 @@
 package org.thoughtcrime.securesms.util;
 
+import android.text.TextUtils;
+
 import java.io.FileDescriptor;
+import java.nio.charset.StandardCharsets;
 
 public class FileUtils {
 
   public static native int getFileDescriptorOwner(FileDescriptor fileDescriptor);
 
+  public static String sanitizeFilename(String name) {
+    if (TextUtils.isEmpty(name) || ".".equals(name) || "..".equals(name)) {
+      return "(invalid)";
+    }
+    final StringBuilder res = new StringBuilder(name.length());
+    for (int i = 0; i < name.length(); i++) {
+      final char c = name.charAt(i);
+      if (isValidFilenameChar(c)) {
+        res.append(c);
+      } else {
+        res.append('_');
+      }
+    }
+    return res.toString();
+  }
+
+  private static boolean isValidFilenameChar(char c) {
+    if ((0x00 <= c && c <= 0x1f)) {
+      return false;
+    }
+    switch (c) {
+      case '"':
+      case '*':
+      case '/':
+      case ':':
+      case '<':
+      case '>':
+      case '?':
+      case '\\':
+      case '\0':
+      case '|':
+      case 0x7F:
+        return false;
+      default:
+        return true;
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/util/FileUtils.java
```diff
@@ -1,9 +1,49 @@
...
+    }
+    return res.toString();
+  }
+
+  private static boolean isValidFilenameChar(char c) {
+    if ((0x00 <= c && c <= 0x1f)) {
+      return false;
+    }
+    switch (c) {
+      case '"':
```
adbenitez, line 33: maybe this is too strict? (I took it from the FAT filesystem) feel free to tweak and remove the ones you think is fine to allow

src/org/thoughtcrime/securesms/util/FileUtils.java
```diff
@@ -1,9 +1,49 @@
...
+        res.append(c);
+      } else {
+        res.append('_');
+      }
+    }
+    return res.toString();
+  }
+
+  private static boolean isValidFilenameChar(char c) {
+    if ((0x00 <= c && c <= 0x1f)) {
```
r10s, line 29: why not just use

```suggestion
    if (c <= 0x1f) {
```

anyways, this check already covers the 'case '\0'` below

src/org/thoughtcrime/securesms/util/FileUtils.java
```diff
@@ -1,9 +1,49 @@
...
+    switch (c) {
+      case '"':
+      case '*':
+      case '/':
+      case ':':
+      case '<':
+      case '>':
+      case '?':
+      case '\\':
+      case '\0':
```
r10s, line 41: `\0` is already covered by the range check above

src/org/thoughtcrime/securesms/util/FileUtils.java
r10s: i think, it is okay. we're just using the sanitised name and not failing

r10s (approved):
some minors

src/org/thoughtcrime/securesms/util/FileUtils.java
adbenitez: fixed, danke!

PR #2694, opened 2023-10-01
fix account switching issues

### LINKED ISSUE #2611: Wrong avatar in notification
- Android version: 13
- Device: Galaxy S22
- Delta Chat version: 1.38.2 (DeltaLab)
- Expected behavior: No avatar would be displayed.
- Actual behavior: Sometimes— _only_ sometimes, though very frequent— a message from an avatarless non-contact arrives, only to have one of my friends' pictures put next to it. It doesn't make sense to me. See the screenshot: my friend and his wife next to a car wash bulk mail. Makes no sense to me.
- Steps to reproduce the problem: I don't know
- Screenshots: 
Screenshot_۲۰۲۳۰۷۰۱_۰۹۰۹۳۶_Neo Launcher

- Logs:

adbenitez:
- when switching accounts by clicking in a notification, it sometimes happens that the chatlist is not refreshed
- notification sometimes display wrong avatar

close #2611

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationListFragment.java
@@ -203,7 +203,6 @@ public class ConversationListFragment extends Fragment
   }
 
   public void onNewIntent() {
-    loadChatlistAsync();
     initializeFabClickListener(actionMode != null);
   }
 
--- a/src/org/thoughtcrime/securesms/ConversationListFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationListFragment.java
@@ -203,6 +203,7 @@ public class ConversationListFragment extends Fragment
   }
 
   public void onNewIntent() {
+    loadChatlistAsync();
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -203,6 +203,7 @@ public void onPause() {
   }
 
   public void onNewIntent() {
+    loadChatlistAsync();
```
r10s, line 206: this will force an update of the chatlist on every return to the chatlist, not only on tapping notifications.

we try to avoid that for performance reasons - and instead keep the list  updated in hidden state on events - that results in an incredible fast returning where no redraw is needed.

therefore, if really needed, we should force the update more selectively, eg using `RELOAD_LIST` flag

adbenitez:
```suggestion
```

r10s (approved):

PR #2695, opened 2023-10-01
api33: update webview darkmode, enable webxdc darkmode for older androids

r10s:
`setForceDark()` will be a no-op function in api33, use css-attributes instead

~~moreover, this pr enables webxdc darkmode for older androids - for whatever reason, `setForceDark()` was not called in `WebxdcActivity`~~ EDIT: it was not called for reasons as it not sets dark/light and let the css decide what to do but really changes colors, which usually does not work out, see 2048 app as an example

targets #2599

### DIFF
--- a/src/org/thoughtcrime/securesms/ConnectivityActivity.java
+++ b/src/org/thoughtcrime/securesms/ConnectivityActivity.java
@@ -35,7 +35,8 @@ public class ConnectivityActivity extends WebViewActivity implements DcEventCent
   }
 
   private void refresh() {
-    String connectivityHtml = DcHelper.getContext(this).getConnectivityHtml();
+    final String connectivityHtml = DcHelper.getContext(this).getConnectivityHtml()
+                                      .replace("</style>", " html { color-scheme: dark light; }</style>");
     webView.loadDataWithBaseURL(null, connectivityHtml, "text/html", "utf-8", null);
   }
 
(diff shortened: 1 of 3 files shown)


src/org/thoughtcrime/securesms/ConnectivityActivity.java
```diff
@@ -35,7 +35,8 @@ public boolean onPrepareOptionsMenu(Menu menu) {
   }
 
   private void refresh() {
-    String connectivityHtml = DcHelper.getContext(this).getConnectivityHtml();
+    final String connectivityHtml = DcHelper.getContext(this).getConnectivityHtml()
+                                      .replace("</style>", " html { color-scheme: dark light; }</style>");
```
r10s, line 39: this is a bit hackish, but a similar approach as on desktop/ios

src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -101,6 +101,7 @@ private static Intent[] getWebxdcIntentWithParentStack(Context context, int msgI
   @Override
   protected void onCreate(Bundle state, boolean ready) {
     super.onCreate(state, ready);
+    setForceDark();
```
r10s, line 104: this was just missing - @adbenitez or do you know of a reason that it was not there?

src/org/thoughtcrime/securesms/ConnectivityActivity.java
adbenitez: I guess this should be added to core sooner or later

adbenitez (approved):

PR #2697, opened 2023-10-02
check areNotificationsEnabled() before adding notification

r10s:
this is best-practise according to
https://developer.android.com/develop/ui/views/notifications/notification-permission

moreover, use the same check for displaying the settings.

successor of https://github.com/deltachat/deltachat-android/pull/2695

### DIFF
--- a/src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
+++ b/src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
@@ -337,14 +337,17 @@ public class NotificationCenter {
                 return;
             }
 
+            NotificationManagerCompat notificationManager = NotificationManagerCompat.from(context);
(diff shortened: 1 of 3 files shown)


src/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -337,15 +337,18 @@ public void addNotification(int accountId, int chatId, int msgId) {
                 return;
             }
 
+            NotificationManagerCompat notificationManager = NotificationManagerCompat.from(context);
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU && !notificationManager.areNotificationsEnabled()) {
+                return;
```
adbenitez, line 342: shouldn't this go below the next `if` ?? otherwise in-chat sounds will not play :thinking:

adbenitez (approved):

PR #2699, opened 2023-10-03
resume settings activity on language changes

### LINKED ISSUE #2698: language not changed for the settings activity immediately
`DynamicLanguage.onResume()` is not called for `ApplicationPreferencesActivity` on language change.

this results in rtl-changes not to happen, and also @gerryfrancis reported that the action bar was not translated (the latter i cannot reprocuce)

the reason is that the `DynamicLanguage.onResume()` is triggered by `BaseActionBarActivity.onResume()` - but the settings page is not resumed by just a fragment is changed.

triggering `DynamicLanguage.onResume()` after language fixes that issue

r10s:
this applies changed immediately,
so esp. right-to-left-changes are correct -
but also weird issues wrt wrong language in the title bar on some api should be fixed.

little downside is that after changing the language, you're back to the "main settings page"
and do not stay in the "appearance page";
but you can regard this also as a feature, one tap less.

(only devs want to stay in the language-selection after language-selection - most users will go on with other things
also - other apps reset much more on language changes)

(reason is that the activity is resumed on language changes and the underlying intent does not track fragments currently)

closes #2698

### DIFF
--- a/src/org/thoughtcrime/securesms/ApplicationPreferencesActivity.java
+++ b/src/org/thoughtcrime/securesms/ApplicationPreferencesActivity.java
@@ -122,7 +122,7 @@ public class ApplicationPreferencesActivity extends PassphraseRequiredActionBarA
       DynamicTheme.setDefaultDayNightMode(this);
       recreate();
     } else if (key.equals(Prefs.LANGUAGE_PREF)) {
-      recreate();
+      dynamicLanguage.onResume(this);
     }
   }
 


adbenitez (approved):
epic one liner :ok_hand:

src/org/thoughtcrime/securesms/ApplicationPreferencesActivity.java
```diff
@@ -122,7 +122,7 @@ public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, Strin
       DynamicTheme.setDefaultDayNightMode(this);
       recreate();
     } else if (key.equals(Prefs.LANGUAGE_PREF)) {
-      recreate();
+      dynamicLanguage.onResume(this);
```
r10s, line 125: nice thing is that this results in easy maintenance - exactly the same is done for other activities on the language change

PR #2702, opened 2023-10-03
improve screen reader support in conversations

### LINKED ISSUE #2470: Voice message player UX issues with screen readers
from https://support.delta.chat/t/compact-media-view/2433

Dear developers
Could you implement an alternative, i.e. more compact, media view like Telegram s?
The fact is that, for example, the appearance of the voice message interface that we have in Delta Chat is not very convenient for blind users.
Below is a small comparison:

1.    Telegram. A voice message only has one line on the screen (of course, this is from the point of view of TalkBack and Jieshuo screen readers). When I focus on a voice message, the screen reader tells me it s a voice message. Also announces its duration, tells when it was sent or received, and says the word Slider.
    If I double tap on a voice message it starts playing. I can also control the playback time slider using the appropriate screen reader gestures (just like in the Delta Chat interface).
2.    Delta Chat, Signal, and if I m not mistaken, Element.
    Each voice message takes up a lot of screen space. Separately - playback button, separately - playback time slider, separately - message information.
    I understand that my comparison can be confusing, but if you try to interact with Telegram and with Delta Chat using the screen reader TalkBack or Jiesuo, you will understand everything.
    I will be very grateful for your attention to my request. Let me know if testing is needed.

adbenitez:
close #2470 

treat voice messages, audios, videos, documents, webxdc apps, images and stickers as a single unit for screen readers, being able to "activate"/open each of that attachments from the action double-tapping the message without carefully having to select small play buttons etc.

also now the message body is read when a text message is selected instead of just saying "message" and user having to find the text with their finger to read it.

videos are now announced as videos instead of "image"

### DIFF
--- a/src/org/thoughtcrime/securesms/BaseConversationItem.java
+++ b/src/org/thoughtcrime/securesms/BaseConversationItem.java
@@ -3,6 +3,7 @@ package org.thoughtcrime.securesms;
 import android.content.Context;
 import android.util.AttributeSet;
 import android.view.View;
+import android.view.accessibility.AccessibilityManager;
 import android.widget.LinearLayout;
 import android.widget.TextView;
 
@@ -77,6 +78,8 @@ public abstract class BaseConversationItem extends LinearLayout
     return batchSelected.isEmpty() && (messageRecord.isFailed());
   }
 
+  protected void onAccessibilityClick() {}
+
   protected class PassthroughClickListener implements View.OnLongClickListener, View.OnClickListener {
 
     @Override
@@ -103,6 +106,9 @@ public abstract class BaseConversationItem extends LinearLayout
 
     public void onClick(View v) {
       if (!shouldInterceptClicks(messageRecord) && parent != null) {
+        if (batchSelected.isEmpty() && ((AccessibilityManager) context.getSystemService(Context.ACCESSIBILITY_SERVICE)).isTouchExplorationEnabled()) {
+          BaseConversationItem.this.onAccessibilityClick();
+        }
         parent.onClick(v);
       } else if (messageRecord.isFailed()) {
         View view = View.inflate(context, R.layout.message_details_view, null);
--- a/src/org/thoughtcrime/securesms/BaseConversationItem.java
+++ b/src/org/thoughtcrime/securesms/BaseConversationItem.java
@@ -3,7 +3,6 @@ package org.thoughtcrime.securesms;
 import android.content.Context;
 import android.util.AttributeSet;
 import android.view.View;
-import android.view.accessibility.AccessibilityManager;
 import android.widget.LinearLayout;
 import android.widget.TextView;
 
@@ -17,6 +16,7 @@ import com.b44t.messenger.DcMsg;
 import com.b44t.messenger.rpc.Rpc;
 
 import org.thoughtcrime.securesms.connect.DcHelper;
+import org.thoughtcrime.securesms.util.Util;
 
 import java.util.HashSet;
 import java.util.Set;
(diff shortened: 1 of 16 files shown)


src/org/thoughtcrime/securesms/BaseConversationItem.java
```diff
@@ -103,6 +106,9 @@ protected class ClickListener implements View.OnClickListener {
 
     public void onClick(View v) {
       if (!shouldInterceptClicks(messageRecord) && parent != null) {
+        if (batchSelected.isEmpty() && ((AccessibilityManager) context.getSystemService(Context.ACCESSIBILITY_SERVICE)).isTouchExplorationEnabled()) {
```
r10s, line 109: i would create a `Util.isTouchExplorationEnabled()` or so and do the check with a try/catch

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -305,6 +307,36 @@ protected void setInteractionState(DcMsg messageRecord, boolean pulseHighlight)
     }
   }
 
+  private void setContentDescription(DcMsg messageRecord) {
```
r10s, line 310:
```suggestion
  private void setContentDescription() {
```

parameter is not used

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -305,6 +307,36 @@ protected void setInteractionState(DcMsg messageRecord, boolean pulseHighlight)
...
+    String desc = "";
+    if (groupSenderHolder.getVisibility() == View.VISIBLE) {
+      desc = groupSender.getText() + "\n";
+    }
```
r10s, line 314: add an empty line to mark end of logic statements and to make sure no one thinks an `else` is missing :)

```suggestion
    }

```

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -463,6 +498,9 @@ else if (hasDocument(messageRecord)) {
       documentViewStub.get().setDocument(new DocumentSlide(context, messageRecord));
       documentViewStub.get().setDocumentClickListener(new ThumbnailClickListener());
       documentViewStub.get().setOnLongClickListener(passthroughClickListener);
+      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT) {
+        documentViewStub.get().setImportantForAccessibility(View.IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS);
```
r10s, line 502: why can't that be set in the corresponding `.xml`? (`document_view.xml` or `conversation_item_*.xml`)

same for the other types

r10s (commented):
some minors

src/org/thoughtcrime/securesms/ConversationItem.java
adbenitez: because it is only supported for >=kitkat and I am not sure what will happen if I set an "invalid" value there, and duplicating layout files in a layout-vXX folder would be more inconvenient

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -193,6 +194,7 @@ public void bind(@NonNull DcMsg                   messageRecord,
     setReactions(messageRecord);
     setFooter(messageRecord, locale);
     setQuote(messageRecord);
+    if (Util.isTouchExplorationEnabled(context)) setContentDescription();
```
r10s, line 197: maybe a matter of taste, however, it is only a suggestion :)

```suggestion
    if (Util.isTouchExplorationEnabled(context)) {
      setContentDescription();
    }
```

src/org/thoughtcrime/securesms/util/Util.java
```diff
@@ -344,4 +345,12 @@ public static boolean isClickedRecently() {
...
     return false;
   }
+
+  public static boolean isTouchExplorationEnabled(Context context) {
+    try {
+      return ((AccessibilityManager) context.getSystemService(Context.ACCESSIBILITY_SERVICE)).isTouchExplorationEnabled();
+    } catch (Exception e) {
+      return false;
+    }
+  }
```
r10s, line 355: we're no calling this function not only on a single click, but now also for every message, this easily sums up to some 10 calls per view

as it is unclear how expensive `getSystemSerivce()` actually is, i would cache this call somewhere

there is some discussion at https://stackoverflow.com/questions/32417398/is-context-getsystemservice-an-expensive-call - however, regarding all the different android versions, probably no one can really tell :)

ah, and there is also the hint to go over the applicationContent to be sure ...

r10s (approved):

PR #2703, opened 2023-10-03
use only one `audioFocusChangeListener`

### LINKED ISSUE #2700: cannot play same the voice message twice in succession
to reproduce, play a voice message, hit pause and try to hit "play" again - this does not work

funnily, "play" does work again, if you play another voice message in between

this bug was probably introduced by #2686 and was not released

r10s:
use only one `audioFocusChangeListener`
as otherwise the "old" will immediately pause
when we restart playing.

moreover, this pr switches to the new `requestAudioFocus()` api

closes #2700

### DIFF
--- a/src/org/thoughtcrime/securesms/components/AudioView.java
+++ b/src/org/thoughtcrime/securesms/components/AudioView.java
@@ -6,6 +6,8 @@ import android.content.res.ColorStateList;
 import android.graphics.PorterDuff;
 import android.graphics.Rect;
 import android.graphics.drawable.AnimatedVectorDrawable;
+import android.media.AudioAttributes;
+import android.media.AudioFocusRequest;
 import android.media.AudioManager;
 import android.os.Build;
 import androidx.annotation.NonNull;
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/components/AudioView.java
```diff
@@ -39,6 +41,7 @@ public class AudioView extends FrameLayout implements AudioSlidePlayer.Listener
   private final @NonNull TextView        title;
 
   private @Nullable AudioSlidePlayer   audioSlidePlayer;
+  AudioManager.OnAudioFocusChangeListener audioFocusChangeListener;
```
adbenitez, line 44:
```suggestion
  private AudioManager.OnAudioFocusChangeListener audioFocusChangeListener;
```

r10s: thanks!

adbenitez (approved):
:ok_hand:

PR #2708, opened 2023-10-04
allow to resend attachments from chat profile

### LINKED ISSUE #2707: add option to resend apps/docs from chat profile
this is specially useful to re-send webxdc apps from the chat profile for new chat members, without having to select all apps in the conversation view, where they can be mixed with a lot of text messages

adbenitez:
close #2707


r10s (approved):
nice!

PR #2713, opened 2023-10-04
properly display webxdc's loading screen in dark mode

adbenitez:
without this, only the progress bar has dark background but the rest of the screen is still white


WofWca (approved):
I haven't tested it, but the solution looks reasonable.

r10s (approved):
assuming, the display mode does not have effect on FILL500 ongoings (and i do not think so :), this lgtm

PR #2715, opened 2023-10-04
save map preferences per account, not only per chat id

### LINKED ISSUE #2714: map screen location and zoom value is not multi-account ready
the map preferences are saved per chat but not taking into account the account id, so they override the saved preferences (ex. the current map view coordinates and zoom level) of chats in other accounts that have same chat id

adbenitez:
close #2714


r10s (approved):
as this fixes an actual bug, lgtm - otherwise i've would have said, that we should not spend much energy to mapbox anymore if https://github.com/deltachat/deltachat-ios/pull/1912 is coming closer :)

PR #2718, opened 2023-10-04
remove dead code

r10s:
the screen-lock functionality was removed in
https://github.com/deltachat/deltachat-android/pull/1942 ; the getters/setters were in use a little longer for a migration device messsage, however, meanwhile this migration is removed as well.


adbenitez (approved):
:broom:

PR #2719, opened 2023-10-04
search in "attach contact"

r10s:
this pr uses the more powerful `ContactSelectionActivity` for attaching contacts

the old `BlockedAndShareContactsActivity` is now only used for blocked contacts, however, is not renamed to keep the pr more readable and on-point. we can do the rename in a later follow-up - or use `ContactSelectionActivity` there as well

for testing: this pr affects all contact-list-activities (new, share, forward, block)


adbenitez (approved):

adbenitez:
really nice!!!

r10s:
❤️

PR #2721, opened 2023-10-04
allow to select all media

### LINKED ISSUE #2720: allow to select all files in chat media screen
this would allow to remove all big attachments etc.

adbenitez:
close #2720


r10s:
the toobar is not updated if before only one item was selected:

- select one image
- "select all"

-> the toolbar/menu still shows "info", "share", "show in chat" - which should be hidden in multi-select mode

this is true for gallery and audio, probably also files and apps (did not test the latter two)

adbenitez:
good catch! :100:  I pushed a fix

r10s (approved):

PR #2724, opened 2023-10-05
rename blocked-contacts-activity

r10s:
`BlockedAndShareContactsActivity` is no longer used for sharing contacts, this was always a misleading name
and is done since #2719 by `AttachContactActivity`.


adbenitez (commented):
:broom:

adbenitez (approved):

PR #2727, opened 2023-10-06
set a default label for new issues (also to make it more obvious for users that it's only bug reporting)

hpk42:
according to https://docs.github.com/en/communities/using-templates-t…o-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository this should work


adbenitez (approved):
makes sense, cool!

r10s:
good idea!

then it's also easier to argue to close things that are not bugs

PR #2728, opened 2023-10-06
allow to properly click in audio messages in multi-select mode

### LINKED ISSUE #2725: cannot select voice message by tapping inside bubble
- open a chat with at least one voice message and one other message
- long-tap the other message to enter multi-select mode
- try to select the voice message as well by tapping it _inside the bubble_ 

-> this does not work, only taps outside the bubble can be used for selecting voice messages

~~i assume, this bug is introduced by recent changes in that area~~ nope, the bug is there "forever"

---

a similar bug is in the profile view:

- open a chat with two voice messages, go to "profile" and then "audio"
- long-tap the first voice message  to enter multi-select mode
- try to select the other voice message by tapping _inside the seek bar_ 

-> this does not work, only taps outside the seekbar can be used for selecting voice messages

adbenitez:
also allow to long-press in play button to select message

close #2725

### DIFF
--- a/res/layout/audio_view.xml
+++ b/res/layout/audio_view.xml
@@ -96,4 +96,14 @@
 
     </LinearLayout>
 
+    <LinearLayout
+        android:id="@+id/interception_mask"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:background="@android:color/transparent"
+        android:clickable="true"
+        android:focusable="true"
+        android:orientation="horizontal"
+        android:visibility="gone" />
+
 </merge>
--- a/res/layout/audio_view.xml
+++ b/res/layout/audio_view.xml
@@ -96,6 +96,7 @@
 
     </LinearLayout>
 
+    <!-- transparent overlay to easily block clicks in audio player's components in multi-select mode -->
     <LinearLayout
         android:id="@+id/interception_mask"
         android:layout_width="match_parent"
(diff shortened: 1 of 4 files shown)


res/layout/audio_view.xml
```diff
@@ -96,4 +96,14 @@
 
     </LinearLayout>
 
-</merge>
\ No newline at end of file
+    <LinearLayout
```
r10s, line 99: this is the hack you told  me?

can you please add a `<!-- comment -->` above this tag to explain the rough ongoings? or a link that explains them?

adbenitez: added a comment

src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -839,7 +838,7 @@ private int getAvailableMessageBubbleWidth(@NonNull View forView) {
   @Override
   public void onAccessibilityClick() {
     if (mediaThumbnailStub.resolved())    mediaThumbnailStub.get().performClick();
-    else if (audioViewStub.resolved())    audioViewStub.get().performClick();
+    else if (audioViewStub.resolved())    audioViewStub.get().toggleState();
```
r10s, line 841:
```suggestion
    else if (audioViewStub.resolved())    audioViewStub.get().togglePlay();
```

r10s (approved):

PR #2729, opened 2023-10-07
cleanup code, fix some linter warnings

adbenitez:


src/org/thoughtcrime/securesms/crypto/KeyStoreHelper.java
```diff
@@ -170,7 +167,7 @@ public static class SealedData {
     }
 
     @SuppressWarnings("unused")
-    public SealedData() {}
+    public SealedData() { /* needed by JsonUtils.fromJson() */ }
```
r10s, line 170: if it is needed and used - why then suppress "unused" warning?

src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -281,19 +285,19 @@ protected WebResourceResponse interceptRequest(String rawUrl) {
...
+      + "style-src 'self' 'unsafe-inline' blob: ; "
+      + "font-src 'self' data: blob: ; "
+      + "script-src 'self' 'unsafe-inline' 'unsafe-eval' blob: ; "
+      + "connect-src 'self' data: blob: ; "
+      + "img-src 'self' data: blob: ; "
+      + "media-src 'self' data: blob: ;"
+      + "webrtc 'block' ; "
+    );
+    headers.put("X-DNS-Prefetch-Control", "off");
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
```
r10s, line 300: we're not opening the whole activity if we're not running on `>= Build.VERSION_CODES.LOLLIPOP` , however, to remove a warning, the change makes sense.

consider adding also the header construction inside the if:

```suggestion
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.LOLLIPOP) {
      Map<String, String> headers = new HashMap<>();
      headers.put("Content-Security-Policy",
          "default-src 'self'; "
        + "style-src 'self' 'unsafe-inline' blob: ; "
        + "font-src 'self' data: blob: ; "
        + "script-src 'self' 'unsafe-inline' 'unsafe-eval' blob: ; "
        + "connect-src 'self' data: blob: ; "
        + "img-src 'self' data: blob: ; "
        + "media-src 'self' data: blob: ;"
        + "webrtc 'block' ; "
      );
      headers.put("X-DNS-Prefetch-Control", "off");
```

r10s (approved):
some minor comments.

i assume, all changes were done by just tapping suggestions from AndroidStudio? cause i only looked loosely over the math changes wrt min/max :) (but looked good to me :)

src/org/thoughtcrime/securesms/crypto/KeyStoreHelper.java
adbenitez: because it is used indirectly, to JsonUtils.fromJson() SealedData.class is passed, then some json marshal/unmarshal lib uses the class with some reflection to parse json and create instances of the given class, so the IDE can't really know it is actually used via reflection

r10s: thanks for explanation. then removing "unused" things seems dangerous, if IDE does not really know (just saying in general, this PR seems to be fine as mostly TAG things are removed :)

adbenitez:
> math changes wrt min/max

that was suggested by the IDE, it is quite smart I had also to analyze the statement, was a bit surprised Android studio could even realize these small logic details

r10s (approved):

src/org/thoughtcrime/securesms/crypto/KeyStoreHelper.java
adbenitez: I only removed the "unused" annotation where it was not needed (because there was obvious usages in other places and the IDE didn't complain) removing it here for example, would make the IDE then create a warning, so what I did was remove the false positives so it is more clear where there are real "unused" annotations and that they are important and not just leftovers

PR #2732, opened 2023-10-09
fix permissions for ~ android 11

### LINKED ISSUE #2726: Cannot grant permission to access storage (impossible to restore from backup)
- Android version:
Android 11.

- Device:
Fairphone FP2 (Lineage OS 18.1 build RQ3A.211001.001 a4d837813c from 2023-09-20).

- Delta Chat version:
1.41 (latest nightly build).

- Expected behavior:
To be able to grant permission to Delta Chat to access storage.

- Actual behavior:
Permission cannot be granted to Delta Chat to access storage.

-- Launch the app and select `Restore from Backup`.
-- When asked to grant permission to access storage, tap `CONTINUE`. (You will be guided to the app permission settings for Delta Chat in Android, but unfortunately "Storage" is not available and, therefore, cannot be selected.)

- Screenshots:

-- Expected:
grafik

-- Actual:
grafik
grafik

- Logs:
N/A.

r10s:
fixes #2726 

- [x] fix import backup from welcome activity on android 11
- [x] check _other file access_ points on android 11
- [x] check other androids (i assume, android10 is same as android11, so the change should be at sdk29 already, but that needs to be double checked and tested)  
    - EDIT: tried at least import-after-fresh-install, export-image, camera-roll on:
        - android5 (sdk21): ok
        - android9 (sdk28): ok
        - android10 (sdk29): ok
        - android11 (sdk30): ok
        - _android12 (sdk31): cannot test, emulator issues_
        - android12 (sdk32): ok
        - android13 (sdk33): ok
    - EDIT: tried backup-after-fresh-install-without-any-rights-granted:
        - android5 (sdk21): ok
        - android9 (sdk28): ok
        - android10 (sdk29): ok
        - android11 (sdk30): ok
        - android12 (sdk32): ok
        - android13 (sdk33): ok

_other file access points_ are:
- export media from chat
- export media from gallery
- export keys
- export log
- export backup
- import key
- import backup on welcome screen

### DIFF
--- a/src/org/thoughtcrime/securesms/ShareActivity.java
+++ b/src/org/thoughtcrime/securesms/ShareActivity.java
@@ -138,6 +138,7 @@ public class ShareActivity extends PassphraseRequiredActionBarActivity implement
   private void requestPermissionForFiles(List<Uri> streamExtras) {
     Permissions.with(this)
             .request(Manifest.permission.WRITE_EXTERNAL_STORAGE, Manifest.permission.READ_EXTERNAL_STORAGE)
+            .alwaysGrantOnSdk33()
             .ifNecessary()
             .withPermanentDenialDialog(getString(R.string.perm_explain_access_to_storage_denied))
             .onAllGranted(() -> resolveUris(streamExtras))
--- a/src/org/thoughtcrime/securesms/permissions/Permissions.java
+++ b/src/org/thoughtcrime/securesms/permissions/Permissions.java
@@ -80,6 +80,8 @@ public class Permissions {
 
     private boolean condition = true;
 
+    private boolean alwaysGranted = false;
+
     PermissionsBuilder(PermissionObject permissionObject) {
       this.permissionObject = permissionObject;
     }
@@ -97,6 +99,13 @@ public class Permissions {
     public PermissionsBuilder ifNecessary(boolean condition) {
       this.ifNecesary = true;
       this.condition  = condition;
+      return this;
+    }
+
+    public PermissionsBuilder alwaysGrantOnSdk33() {
+      if (Build.VERSION.SDK_INT >= 33) {
+        alwaysGranted = true;
+      }
       return this;
     }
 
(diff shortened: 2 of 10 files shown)


adbenitez:
about Android 10: 
https://stackoverflow.com/questions/64221188/write-external-storage-when-targeting-android-10
https://developer.android.com/about/versions/10/privacy/changes#scoped-storage

r10s:
seems as if sdk29 is fine already. so, this pr really fixes only android11 on-point, from the logic, the older androids should be unchanged

r10s:
@adbenitez for logic review, i think, that paths of other android-versions are not changed, seems calming

r10s:
back to draft: android12 seems special ...

adbenitez:
> back to draft: android12 seems special ... ooh no, was on a wrong branch :) have to check the things again ...

oh :( talk about android fragmentation

r10s:
so, finally, this is ready for review; i double checked the android versions mentioned above ...

src/org/thoughtcrime/securesms/permissions/Permissions.java
```diff
@@ -100,6 +102,20 @@ public PermissionsBuilder ifNecessary(boolean condition) {
       return this;
     }
 
+    public PermissionsBuilder alwaysGrantOnSdk30() {
```
adbenitez, line 105: I would just have a generic function `alwaysGrantOnSdk(int version)` and then call `Permissions.alwaysGrantOnSdk(30)` etc.

src/org/thoughtcrime/securesms/ShareActivity.java
```diff
@@ -138,6 +138,7 @@ private boolean needsFilePermission(List<Uri> uris) {
   private void requestPermissionForFiles(List<Uri> streamExtras) {
     Permissions.with(this)
             .request(Manifest.permission.WRITE_EXTERNAL_STORAGE, Manifest.permission.READ_EXTERNAL_STORAGE)
```
adbenitez, line 140: `WRITE_EXTERNAL_STORAGE` is ".alwaysGrantOnSdk30()" in other places, does this mean that it will be requested for SDK 30 here? what are the implications?

adbenitez (approved):
great! thanks for fixing and enduring having to try in all that emulators! overall the code looks good, I can't test it myself other than in sdk 33

src/org/thoughtcrime/securesms/preferences/ChatsPreferenceFragment.java
```diff
@@ -266,7 +266,8 @@ public boolean onPreferenceClick(Preference preference) {
 
   private void performBackup() {
     Permissions.with(getActivity())
-            .request(Manifest.permission.WRITE_EXTERNAL_STORAGE, Manifest.permission.READ_EXTERNAL_STORAGE)
+            .request(Manifest.permission.WRITE_EXTERNAL_STORAGE)
```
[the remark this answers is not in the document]
r10s: > doesn't core needs READ_EXTERNAL_STORAGE to read the folder content and generate the backup names

good point, then we'll better leave this unchanged - but i'll add a little comment

r10s:
quite some stuff :)

Screenshot 2023-10-12 at 15 49 41

src/org/thoughtcrime/securesms/ShareActivity.java
r10s: i assume, that should be `alwaysGrantOnSdk30()` as well :)

r10s: hm, not sure. in contrast to the other places, this uses some `file:` protocol, that may indeed require these rights (similar to the camera-roll, that also requires READ_EXTERNAL_STORAGE between sdk30..sdk32 (implicitly requested by galleryPermissions()). let's leave this as it was before (so, only change this for sdk33 where the permissions are gone.

might be, that we would better use galleryPermissions() for sdk33 here - but i anyway assume that "files" are playing less and less roles on newer androids, so, probably everything's fine. if not - we can iterate, if things are wrong on testing

src/org/thoughtcrime/securesms/permissions/Permissions.java
r10s: yeah, let's leave this as is for now. it looks more explicit that 33/30 is no typo in a parameter :)

PR #2735, opened 2023-10-10
allow to backup/export all accounts at once

adbenitez:
- [x] adapt backup/export dialog
- [x] hide the "backup all" offering if user has only one account
- [x] implement "export all"

~~:warning:  before this gets merged, would be nicer if core sets the account address in the backup filename~~ (DONE: https://github.com/deltachat/deltachat-core-rust/issues/4816)


adbenitez:
that would get too complicated too soon, if the device doesn't have enough space then backup one by one instead of using all, anyway selecting only a subset also needs to know account size to make an informed decision and then that is more features

adbenitez:
for exporting all accounts it gets a bit complicated, because there is no blocking way to call "create backup" per account, the core API for backups is async and you can't just loop over an array of accounts and call the "create backup" API, progress is received via core event and you would need to start the backup of the next account in the "progress 1000" of the previous one, etc.

would be nice to have an API that blocks until the backup is done or an API to "backup all" accounts similar to the startIO for all accounts, that would be even better/easier for the UI

cc @link2xt

r10s (approved):
lgtm

PR #2741, opened 2023-10-12
update dcContext in ConversationListAdapter.changeData()

### LINKED ISSUE #2740: chatlist displaying wrong chats from other account in multi-account setup
after clicking on a notification, the chatlist sometimes is not rendered properly, the chat ids array is correct and the chat summary is correct, but the chat avatar and name is taken from the previously selected account, 

the reason is that ConversationListAdapter saves the DcContext/account on initial instance creation but it never updates it, then when it receives a new chatlist from the new selected account, it gets the chats IDs but try to load the chat objects from the previously selected account

adbenitez:
close #2740

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListAdapter.java
+++ b/src/org/thoughtcrime/securesms/ConversationListAdapter.java
@@ -199,7 +199,7 @@ class ConversationListAdapter extends RecyclerView.Adapter {
       dcChatlist =  new DcChatlist(0, 0);
     } else {
       dcChatlist = chatlist;
-      dcContext = DcHelper.getAccounts(context).getAccount(chatlist.getAccountId());
+      dcContext = DcHelper.getAccounts(context.get()).getAccount(chatlist.getAccountId());
     }
     notifyDataSetChanged();
   }
--- a/src/org/thoughtcrime/securesms/ConversationListAdapter.java
+++ b/src/org/thoughtcrime/securesms/ConversationListAdapter.java
@@ -34,6 +34,7 @@ import org.thoughtcrime.securesms.connect.DcHelper;
 import org.thoughtcrime.securesms.mms.GlideRequests;
 import org.thoughtcrime.securesms.util.ViewUtil;
 
+import java.lang.ref.WeakReference;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Locale;
@@ -50,7 +51,7 @@ class ConversationListAdapter extends RecyclerView.Adapter {
   private static final int MESSAGE_TYPE_THREAD         = 2;
   private static final int MESSAGE_TYPE_INBOX_ZERO     = 3;
 
-  private final           Context              context;
+  private final WeakReference<Context>         context;
   private @NonNull        DcContext            dcContext;
   private @NonNull        DcChatlist           dcChatlist;
   private final @NonNull  GlideRequests        glideRequests;
@@ -88,7 +89,7 @@ class ConversationListAdapter extends RecyclerView.Adapter {
                           @Nullable ItemClickListener clickListener)
   {
     super();
-    this.context        = context;
+    this.context        = new WeakReference<>(context);
     this.glideRequests  = glideRequests;
     this.dcContext      = DcHelper.getContext(context);
     this.dcChatlist     = new DcChatlist(0, 0);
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/ConversationListAdapter.java
```diff
@@ -51,8 +50,8 @@ class ConversationListAdapter extends RecyclerView.Adapter {
   private static final int MESSAGE_TYPE_THREAD         = 2;
   private static final int MESSAGE_TYPE_INBOX_ZERO     = 3;
 
-  private final WeakReference<Context>         context;
```
r10s, line 54: why is the `WeakReference` to the application context removed? i am not sure, what it is for, iirc, this is to solve hard issues wrt garbage collection (to break cyclic references, even if rare in java)

so, i would not remove that without reasons and deep understanding.

src/org/thoughtcrime/securesms/ConversationListAdapter.java
```diff
@@ -195,7 +189,12 @@ interface ItemClickListener {
   }
 
   void changeData(@Nullable DcChatlist chatlist) {
-    dcChatlist = chatlist == null ? new DcChatlist(0, 0) : chatlist;
+    if (chatlist == null) {
+      dcChatlist =  new DcChatlist(0, 0);
+    } else {
+      dcChatlist = chatlist;
+      dcContext = DcHelper.getAccounts(context).getAccount(chatlist.getAccountId());
```
adbenitez, line 196: is it more optimal to check if chatlist.getAccountId() == previous acccountId instead of always calling `getAccount()`?

src/org/thoughtcrime/securesms/ConversationListAdapter.java
adbenitez: I removed it because it is never the case that context is null, it is always non-null, didn't realize it had other meaning, was thinking it was some superfluous wrapper similar to the "Stub" class is used in other places, will revert that

src/org/thoughtcrime/securesms/ConversationListAdapter.java
```diff
@@ -195,7 +195,12 @@ interface ItemClickListener {
...
```
r10s: that would be my suggestion:

```suggestion
    Context context = this.context.get();
    if (context == null) {
      return;
    }

    if (chatlist == null) {
      dcChatlist =  new DcChatlist(0, 0);
    } else {
      dcChatlist = chatlist;
      if (dcContext.getAccountId() != chatlist.getAccountId()) {
        dcContext = DcHelper.getAccounts(context).getAccount(chatlist.getAccountId());
      }
```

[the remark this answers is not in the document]
adbenitez: can't commit suggestion because it was done to an outdated state of the file, will need to manually pick your suggestion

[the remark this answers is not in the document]
adbenitez: @r10s in the end used DcHelper.getContext() to get the single global instance from ApplicationContext, this is more in-line with what is done in other parts of the code and avoids creating a new DcContext even if the account changed (the DcContext instance from ApplicationContext is reused)

src/org/thoughtcrime/securesms/ConversationListAdapter.java
```diff
@@ -195,7 +195,12 @@ interface ItemClickListener {
   }
 
   void changeData(@Nullable DcChatlist chatlist) {
-    dcChatlist = chatlist == null ? new DcChatlist(0, 0) : chatlist;
+    if (chatlist == null) {
+      dcChatlist =  new DcChatlist(0, 0);
+    } else {
+      dcChatlist = chatlist;
+      dcContext = DcHelper.getContext(context.get());
```
r10s, line 202: k, that is guaranteed be the same account as in the chatlist as the account is switched? 

still, we need to check context.get() for != null (this is the idea of a weak reference - the contained object may be freed, for whatever reason, usually this ConversationListAdapter object will be freed as well soonish):


```suggestion
    Context context = this.context.get();
    if (context == null) {
      return;
    }
    if (chatlist == null) {
      dcChatlist =  new DcChatlist(0, 0);
    } else {
      dcChatlist = chatlist;
      dcContext = DcHelper.getContext(context);
```

adbenitez: > k, that is probably guaranteed be the same account as in the chatlist as the account is switched?

yes, at the time `changeData()` is called in `ConversationListFragment` the new DcChatlist that is passed is from the current selected account

r10s (approved):
lgtm :)

PR #2744, opened 2023-10-13
move reaction pills together with swiped message bubble

adbenitez:
fix bug and recycle unused code! :recycle:

when you swipe a message to reply it, the message bubble moves, but if the message has reactions, the reaction pills remain in the same static position and the user avatar touches the reaction pills in groups,

this pull request fixes that issue

image


r10s (approved):

r10s:
did not test, but that looks easy enough :)

thanks a lot for figuring out that!

PR #2749, opened 2023-10-20
update translations

r10s:
this also updates the english string to 'Verified by me', see commit message for some more details.


adbenitez (approved):
I find it a bit confusing to see someone's profile avatar, name, picture and then a line "verified by me" it is not clear who is "me" because can be read as spoken by that person, (the signature is also displayed close by etc)

r10s:
let's see if it works, we can change that easily, it does not even affect other languages (as eg. in german,the "by me" seems better in any case)

PR #2750, opened 2023-10-22
remove unnecessary attribute isLandscape

adbenitez:
isLandscape attribute was introduced initially as part of a complex logic that would set it to true only if in landscape mode AND device screen size was not big (ex. tablets) but later it was decoupled and a dedicated "isBigScreen" was introduced, then isLandscape was just telling if the device is in landscape mode, for that there is no need for this trick using an attribute, screen orientation can be checked directly and is already available in the onConfigurationChanged() event that calls setScreenMode() so there is no need to check an extra attribute


r10s (approved):
simplified logic, very nice!

PR #2753, opened 2023-10-23
Merge "improve mailto Intent handling in ShareActivity" into main

Hocuri:
That's https://github.com/deltachat/deltachat-android/pull/2637 which was (probably accidentally) dropped out when renaming `stable` to `main` and `main` to `feat/verified-1to1`.

@r10s would you like this to go into 1.41? It might be better to wait until 1.42, since sharing is an area with lots of regressions, and 1.42 is not that far out.

### DIFF
--- a/androidTest/com/b44t/messenger/uitests/offline/SharingTest.java
+++ b/androidTest/com/b44t/messenger/uitests/offline/SharingTest.java
@@ -205,6 +205,25 @@ public class SharingTest {
     onView(withHint(R.string.chat_input_placeholder)).check(matches(withText("Veeery important draft")));
   }
 
+  /**
+   * Regression test:
+   *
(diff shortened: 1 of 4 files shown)


r10s:
thanks for refiling this pr! as there are already so many moving targets, it indeed makes sense to not also fiddle around with "share", if there is nothing urgent, i would postpone this after 1.42

adbenitez (approved):
oops! another victim of the master/stable confusion :sweat: 

this fixes an user-reported bug! thanks for resurrecting! 

(btw I already released this in deltalab since some time, and I am using it as well, so far so good!)

r10s:
> btw I already released this in deltalab since some time, and I am using it as well, so far so good!

i see, then maybe merging is okay. also as it mainly affects mailto and not the whole share-activity (as @adbenitez just explained :)

PR #2754, opened 2023-10-23
Commits that adapt to the next core version

Hocuri:
These commits were left in the `feat/verified-1to1` branch, and they adapt DC Android to core-rust main.

I created this PR here so that we can merge it when core-rust main becomes the new stable.

### DIFF
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -227,7 +227,8 @@ JNIEXPORT jlong Java_com_b44t_messenger_DcAccounts_createAccountsCPtr(JNIEnv *en
 {
     CHAR_REF(osname);
     CHAR_REF(dir);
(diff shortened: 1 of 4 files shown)


r10s:
the core-changes that this PR is waiting for are not included in 1.126.1, right?

is there sth. blocking to get that in? (nb: it is a bit unfortunate that we changed parameters, this makes updating on all systems a bit harder and needs more synchronisation; i know, usually we try to avoid that, just saying :)

r10s:
i added a commit that removes some stock strings that no longer exist (came over that while adaption ios at https://github.com/deltachat/deltachat-ios/pull/1943 )

r10s (approved):
builds with https://github.com/deltachat/deltachat-core-rust/pull/4881 , however, let's s till wait for a new core before merging

PR #2755, opened 2023-10-23
use CONFIG_CONFIGURED_ADDRESS instead of CONFIG_ADDRESS

adbenitez:
if you try to change account configuration ex. change address and password and it fails, still the new tried address will be shown in the account overview in settings and some other places

### DIFF
--- a/src/org/thoughtcrime/securesms/preferences/widgets/ProfilePreference.java
+++ b/src/org/thoughtcrime/securesms/preferences/widgets/ProfilePreference.java
@@ -64,7 +64,7 @@ public class ProfilePreference extends Preference {
   public void refresh() {
     if (profileAddressView == null) return;
 
-    final String address = DcHelper.get(getContext(), DcHelper.CONFIG_ADDRESS);
+    final String address = DcHelper.get(getContext(), DcHelper.CONFIG_CONFIGURED_ADDRESS);
     String profileName  = DcHelper.get(getContext(), DcHelper.CONFIG_DISPLAY_NAME);
 
     if(profileName==null || profileName.isEmpty()) {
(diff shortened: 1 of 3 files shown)


adbenitez:
also probably makes sense to load the configured values when opening the registration activity instead of showing previous tried fields

r10s:
hm, not sure if that is the correct way to fix the issue.

iirc, the UI should not even use the configured_ vars, they are not even documented. i think this needs a closer look

adbenitez:
> iirc, the UI should not use the configured_ vars

well right now from the current core API that is the correct way to get the correct configured values, and the same goes for bots, the non `configured_*` parameters are a clumsy way to pass parameters to `configure()` before starting configuration, instead of a long list of arguments, maybe should be reworked to pass some kind of object holding the configuration params instead of permanently saving temporary configuration params, these params can't hold the right values unless after `configure()` ends it resets the values to the current valid values, and then in configuration screen that would cause issues, otherwise an api to "reset to configured values" after closing the configuration screen is needed, in any case that is on core-land and would need changes in core, the point is that according the current core API the correct values are the `configured_*` ones and in any case with the current core API the UI will need to use that values, directly or getting them and reseting the non-configured values with their configured variants's values on closing configuration screen

cc @link2xt

r10s:
> also probably makes sense to load the configured values when opening the registration activity instead of showing previous tried fields

>  the point is that according the current core API the correct values are the configured_* ones

yes.

however, most of them are not about to be displayed to the user - eg. you may not have set a PORT, however, there will be one in the configiured_ view

the idea is not to have them set into the corresponding fields - instead, if you only set email address + password, after login and reopening the dialog, there should not be everything filled out as that would make subsequent configure harder (eg. things are probed etc, and a subsequent login may have different auto-generated variables)

r10s:
i did a PR https://github.com/deltachat/deltachat-core-rust/pull/4853 for documenting `configured_addr` - which was indeed also in use like that before, eg for AEAP. if https://github.com/deltachat/deltachat-core-rust/pull/4853 is okay, we can merge this PR, that actually fixes issues, thanks a lot for taking care!

r10s (approved):
i did PR https://github.com/deltachat/deltachat-core-rust/pull/4853 for documenting `configured_addr` - which was indeed also in use like that before, eg for AEAP. if https://github.com/deltachat/deltachat-core-rust/pull/4853 is okay, we can merge this PR, that actually fixes issues, thanks a lot for taking care!

adbenitez:
an issue that is still open: if you try to change configuration, then it fails, then close configuration screen then re-open some days later, the old tried values are still there instead of the last correct tried values, that seems to be an issue that should be fixed in core

PR #2758, opened 2023-10-26
make back button work in webxdc

r10s:
this pr switches the "hardware" back button
from always terminating the app
to going back in history first,
only if the history is empty, the app is terminated.

this enables a webxdc to use document.history as additional navigation as usual.

the "back" upper left of the screen still always exits the app.


r10s:
it was disabled for reasons at https://github.com/deltachat/deltachat-android/pull/2309 , so, as we're about doing a release, we should not change that now

PR #2765, opened 2023-10-28
allow to deactivate QR codes

### LINKED ISSUE #2763: add “deactivate QR” menu option in secure-join QR screens
currently it is possible to scan self-created group-invite QRs and contact QR to deactivate it but it is hard to discover this feature, a more user-friendly approach is to also make it available in the menu

adbenitez:
close #2763


r10s (approved):
nice

PR #2770, opened 2023-10-29
show verifier name+address

r10s:
only address is easily confusing, only name may be tricked :)

Screenshot 2023-10-29 at 17 08 14<br>

cmp https://github.com/deltachat/deltachat-ios/pull/1949


adbenitez (approved):
you was faster! was about to start looking into this, that was a simple fix!

PR #2771, opened 2023-10-29
start all backups

r10s:
this starts all backups at the same time,
overwriting different progress-dialogs as needed.
when the last "done" arrives, the progress dialog is closed. the progress itself, is taken from the first account (this could be improved, but maybe it is just good enough)

best review without whitespace as indentation has changes.

needs lots of testing.


adbenitez (approved):
thanks a lot!!!

PR #2778, opened 2023-10-30
remove QR button in group creation screen

### LINKED ISSUE #2774: remove "QR Invite Code" from group-creation
to streamline the different group types, real group creation should be deferred until the member list is actually set, so that we know if we want to create a verified-group or not (or maybe can ask the user then).

the "QR Invite Code" in the group-creation activity stands in the way of this  deferring as it already creates the group (which is anyway weird as also "Cancel" stops working then)

current theory is that "QR Invite Code" during group-creation  is not used widely anyway, so it is not worth breaking other things.

adbenitez:
close #2774


Hocuri (approved):
Nice!

PR #2779, opened 2023-10-30
see all contact list in verified groups "add member", explain why they can't be added

### LINKED ISSUE #2773: improve adding members to verified-groups
this issues is about the add-member-dialogs when a group is out-of-preparation:

to streamline the different group types, use the same contact lists when adding members to verified-groups and normal-groups.

if the contact to be added to a verified group is not verified (and cannot be added therefore), show a AlertDialog and explain what should be done to add  this member, sth. as (wording improvements and shortenings welcome):

```
To guarantee end-to-end-encryption,
you can only add contacts with a green verification mark to this group.

You may meet USER_NAME in person and scan their QR Code.

[Learn more]    [Scan QR Code] [OK]
```

adbenitez:
close #2773

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -904,6 +904,7 @@
     <!-- translators: "setup" is the "encryption setup" here, as in "Autocrypt Setup Message" -->
     <string name="contact_setup_changed">Changed setup for %1$s.</string>
     <string name="verified_group_explain">Verified groups guarantee end-to-end encryption. Only contacts with a green verification checkmark can be added to them.</string>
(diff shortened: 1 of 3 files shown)


adbenitez:
this PR now makes it harder to find the verified contacts in the contact list to add to the group, we should consider at some point adding a menu option to filter out non-verified contacts

r10s:
> this PR now makes it harder to find the verified contacts in the contact list to add to the group, we should consider at some point adding a menu option to filter out non-verified contacts

let's see in practise. by the "last seen" sorting most of the verified contacts are sorted up anyways, so maybe everything's just fine already.

r10s (approved):

PR #2782, opened 2023-10-31
offer to create verified group

### LINKED ISSUE #2775: combine "Create Verified Group" and "Create Group"
_before this issue is targeted, #2773 and #2774 needs to be closed._

letting the user decide whether to create a normal groups or a verified group already from the "create chat" dialog is hard. users usually won't know the differences and the advantages and disadvantages at this moment, if at all later.

to still make verified groups a  1st class element, strike the "Create Verified Group" option and let the user just create a normal group.

if it turns out on real creation that only verified members are added, ask the user whether they want to create a verified group (later on, we may skip the question):

```
All added contacts have a green verification checkmark.
Do you want to create a group that guarantees end-to-end-encryption?

[Learn more]      [Cancel] [OK]
```

(note, that we do not speak about "verified group" or "protected chats" or so - neither here nor in the rest of the app)

adbenitez:
Screenshot 2023-10-31 at 10 56 50

close #2775

### DIFF
--- a/res/values/colors.xml
+++ b/res/values/colors.xml
@@ -55,7 +55,7 @@
 
     <color name="MediaOverview_Media_selected_overlay">#88000000</color>
 
-    <color name="dummy_avatar_color">#808080</color>
+    <color name="dummy_avatar_color">#505050</color>
     <color name="recently_seen_dot">#34c759</color>
     <color name="slidearrow_color">#999999</color>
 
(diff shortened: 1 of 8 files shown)


r10s:
maybe we should make the "+ New Group" bold, to make it a bit more outstanding (before, there were two "+" options, making that area a little more visible.

Hocuri (approved):
The code looks good.

One minor possible problem I noticed is: If you create a group with only yourself (to add members later), then the text "All added contacts have a green verification checkmark" sounds a bit weird, even though technically true. But probably that's not enough of a problem to warrant the addition of a new translated string.

> maybe we should make the "+ New Group" bold, to make it a bit more outstanding

+1

r10s:
> the text "All added contacts have a green verification checkmark" sounds a bit weird

maybe say "All members have a green verification checkmark" - this is a little bit nicer in this case as it is also true if you have nothing added at all. also, we're also saying "Members" on the button, and and desktop/iOS we show sth as  "12 Members" above the list).

adbenitez:
> maybe we should make the "+ New Group" bold

done, since now they are more blackish, the gray + icons were looking bad so I changed to the same gray we are using for them in DC Desktop

PR #2785, opened 2023-10-31
add desktop strings

r10s:


res/values/strings.xml
```diff
@@ -204,6 +207,9 @@
     <string name="broadcast_lists">Broadcast Lists</string>
     <string name="new_broadcast_list">New Broadcast List</string>
     <string name="add_recipients">Add Recipients</string>
+    <string name="create_broadcast_list">Create Broadcast List</string>
```
adbenitez, line 210: if this a confirmation button for a dialog, why not just "create" ? if it is not a button then why not just reuse the "New Broadcast List"?

r10s: makes sense, also as the button easily gets too wide otherwise. i added a commit

Simon-Laux (approved):

Issue #2786, opened 2023-10-31
add "clone group"


r10s:
cloning group offers a pragmatic way to add unverified ppl to verified groups or upgrade to verified groups.

also the "birthday issue", where a 30 ppl group wants to discuss a present for one member is fixed by that.

no need to put the option too much in the way, however. the kebab menu in the group-profile will do. when tapped, the current group is left (pop 2 activities or so) and "new group" is opened with a prefilled member list and prefilled name (maybe appended by "Clone"?)

PR #2790, opened 2023-11-01
improve group creation screen

adbenitez:
show the "add member" button similar to the way it is displayed in group edition after it is created, for a more seamless UI/UX experience

also simplify the GroupCreateActivity code, removing unused logic from when the QR button was present in this page.

also fixes old known bug: the self-contact is no longer mixed with the added members but kept at the end of the list, the same as in group profile's member list 

image

image


r10s (approved):
nice cleanup!

we should remember this pr wrt SUGGESTED_CONTACT_IDS etc. when targeting https://github.com/deltachat/deltachat-android/issues/2786


layout-wise, i would remove only the thin separator lines, they looks esp. bad if there is only "Add Members" and "Me", but this better goes into a subsequent pr, i create an issue for that. EDIT: done: https://github.com/deltachat/deltachat-android/issues/2793

PR #2797, opened 2023-11-02
allow to clone groups and broadcasts

### LINKED ISSUE #2786: add "clone group"

adbenitez:
close #2786 

this allows to upgrade existing non-verified groups to verified  by drafting a new group and then getting the chance to make it verified if all contacts are verified, thanks to recent changes in the group creation workflow

### DIFF
--- a/res/menu/profile_common.xml
+++ b/res/menu/profile_common.xml
@@ -30,4 +30,8 @@
         android:id="@+id/block_contact"
         app:showAsAction="never"/>
 
+    <item android:title="@string/menu_clone"
+        android:id="@+id/menu_clone"
+        app:showAsAction="never"/>
+
 </menu>
--- a/res/menu/profile_common.xml
+++ b/res/menu/profile_common.xml
@@ -30,7 +30,7 @@
         android:id="@+id/block_contact"
         app:showAsAction="never"/>
 
-    <item android:title="@string/menu_clone"
+    <item android:title="@string/clone_chat"
         android:id="@+id/menu_clone"
         app:showAsAction="never"/>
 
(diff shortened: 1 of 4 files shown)


adbenitez:
I made it possible to clone broadcast lists as well because it can be also useful now that they appear as a new chat/channel, so it can be used for new thread/topic announcements

I tried to keep it simple with the string and just say "clone" so it fits for both groups and broadcasts, it probably is nicer to be more specific and say "clone group"/"clone broadcast list"

also the name "clone" is a bit misleading here, you are not cloning the group with name and image, but drafting a new from an existing memberlist

Hocuri:
> also the name "clone" is a bit misleading here, you are not cloning the group with name and image, but drafting a new from an existing memberlist, maybe the existing "new group"/"new broadcast" strings could be reused? starting a new group from an existing one is more in line with what is happening than "clone"

Well if we want we can also pre-fill the name and image with the ones from the previous group. Esp. setting the image could be useful since for the user it's not straightforward how to reuse the image from the old group (while they can just type in the same name if they want to reuse the name).

In any case, I think "clone" is better than "New group".

adbenitez:
> Well if we want we can also pre-fill the name and image with the ones from the previous group

done :heavy_check_mark:

r10s (approved):
nice 👌

PR #2800, opened 2023-11-02
remove "use guaranteed e2ee?" question

r10s:
this pr removes the remove "use guaranteed e2ee?" question and instead creates a verified group when all members are verified.

this Yes/No question is a hard one.
this turns out last, not least,
when trying to use self-explaining buttons for iOS without using the term "verified groups".
what would be "No"? "Best-Effort-Encryption"?, "I don't care"? :)

joking aside,
few ppl will really read the question on group creation, let alone understanding it and making and informed decision.

many ppl will just tap the rightmost button,
just to continue.

so better go with defaults here until we get
a real usecase where we would need this question.
and then consider how to solve that usecase
without reintroducing weird questions.

this pr is the outcome of longer discussions with @link2xt and @hpk42 ; in a follow-up, we'll remove the term "verified" anywhere from the ui (it may continue to exist in help)

### DIFF
--- a/src/org/thoughtcrime/securesms/GroupCreateActivity.java
+++ b/src/org/thoughtcrime/securesms/GroupCreateActivity.java
@@ -203,21 +203,7 @@ public class GroupCreateActivity extends PassphraseRequiredActionBarActivity
         if (groupChatId!=0) {
           updateGroup(groupName);
         } else {
-          if (!broadcast && allMembersVerified()) {
-            new AlertDialog.Builder(this)
-              .setMessage(R.string.create_verified_group_ask)
-              .setNeutralButton(R.string.learn_more, (d, w) -> IntentUtils.showBrowserIntent(this, "https://delta.chat/en/help#verifiedchats"))
-              .setPositiveButton(R.string.yes, (d, w) -> {
-                  verified = true;
-                  createGroup(groupName);
-              })
-              .setNegativeButton(R.string.no, (d, w) -> {
-                  createGroup(groupName);
-              })
-              .setCancelable(true)
-              .show();
-            return true;
-          }
+          verified = !broadcast && allMembersVerified();
           createGroup(groupName);
         }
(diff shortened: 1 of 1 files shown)


hpk42 (approved):
lgtm, unsurprisingly :)

adbenitez (approved):
I think this is better +1

a caveat: 
if you create the group only with yourself and postpone the decision to add members, then you will no longer be able to add your other contacts that are unverified, then user is confused why they can't add their contacts to the group if it happens that they never scanned a QR with any of them, I can't tell how often people create a group only with themselves tho, but having to select the members up-front was something that always annoyed me in WhatsApp et al.

we could think about just creating a normal group if you are the only member, but that then breaks the workflow of someone wanting to create a verified group only with themselves and then doing QR-invite to join the rest, so that is not a solution

another option: display this dialog only if the group is an empty group with yourself, then not saying "all members are verified" but just "do you want to create a verified group ..."

hpk42:
On Thu, Nov 02, 2023 at 10:29 -0700, Asiel Díaz Benítez wrote:
> @adbenitez approved this pull request.
> 
> I think this is better +1
> 
> a caveat: 
> if you create the group only with yourself and postpone the decision to add members, then you will no longer be able to add your other contacts that are unverified, then user is confused why they can't add their contacts to the group if it happens that they never scanned a QR with any of them, I can't tell how often people create a group only with themselves tho, but having to select the members up-front was something that always annoyed me in WhatsApp et al.

this "group with myself only for starters" use case appeared in some discussions, indeed. 
Even used it myself a bit sometimes.  The FAQ should say that if you want to 
be able to add "non-encrypting" users that it's best to add at least one such users from the start,
or "clone" the group and add that one member, phase out the old group. 
So it's indeed a caveat but probably not a serious one i guess
because you can learn to have groups with non-verified members if you insist on it.

PR #2804, opened 2023-11-02
remove "verified" from strings

r10s:
this pr aims to target one outcome of the recent user testing with @kermoshina , where one finding was that the "Interaction between encryption and verification is difficult to understand." and the explanations tell about end-to-end-encryption and it is unclear how this relates to "verified checkmarks" or "verified contacts" or "verified groups". 

as the term "end-to-end-encryption" is roughly known by users due to whatsapp - but the term "verified" not and is very interpretive, this pr removes this term completely. 

we do no longer talk about verification at all, we talk about "guaranteed end-to-end-encryption" that relates to "green checkmarks" - this seems easier to understand as before (iirc @Hocuri already tried to avoid too many terms and avoided to say "verified chats" or so; this was already founded by some user research)

in the help, however, we may dive deeper and talk about "verification", but there we could be a bit more verbose than in the UI, where things need to be short and on point (ppl do not read much ...)

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -29,7 +29,7 @@
     <string name="emoji">Emoji</string>
     <string name="attachment">Attachment</string>
     <string name="back">Back</string>
-    <string name="verified">Verified</string>
+    <string name="verified">Introduced</string>
     <string name="close">Close</string>
     <string name="close_window">Close Window</string>
     <string name="forward">Forward</string>
@@ -159,7 +159,6 @@
     <string name="video">Video</string>
     <string name="documents">Documents</string>
     <string name="contact">Contact</string>
-    <string name="verified_contact">Verified Contact</string>
     <string name="camera">Camera</string>
     <!-- As in "start a video recording" or "take a photo"; eg. the description of the "shutter button" in cameras -->
     <string name="capture">Capture</string>
@@ -204,7 +203,6 @@
     <string name="clone_chat">Clone Chat</string>
     <!-- Use the same wording as "Subject" to help people coming from the e-mail context; eg. "Betreff" in German -->
     <string name="new_group_or_subject">New Group or Subject</string>
-    <string name="menu_new_verified_group">New Verified Group</string>
     <!-- consider keeping the term "broadcast" as in WhatsApp or Telegram -->
     <string name="broadcast_list">Broadcast List</string>
     <string name="broadcast_lists">Broadcast Lists</string>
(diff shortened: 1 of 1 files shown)


res/values/strings.xml
```diff
@@ -922,16 +920,14 @@
     <string name="secure_join_started">%1$s invited you to join this group.\n\nWaiting for the device of %2$s to reply…</string>
     <!-- placeholder will be replaced by the name of the inviter. -->
     <string name="secure_join_replies">%1$s replied, waiting for being added to the group…</string>
-    <string name="contact_verified">%1$s verified.</string>
-    <string name="contact_not_verified">Cannot verify %1$s.</string>
-    <!-- Shown in contact profile. The placeholder will be replaced by the name of the contact that verified the contact shown. -->
-    <string name="verified_by">Verified by %1$s</string>
-    <string name="verified_by_you">Verified by me</string>
+    <string name="contact_verified">%1$s introduced.</string>
+    <string name="contact_not_verified">Cannot introduce %1$s.</string>
```
adbenitez, line 924: this gets a bit confusing, now it is not about verification but "introduction" from an user point of view the contact is "introduced" immediately after scanning and the chat appears, so "cannot introduce X" doesn't tell them anything since they don't know what was going on in the background

adbenitez (approved):

res/values/strings.xml
```diff
@@ -29,7 +29,7 @@
     <string name="emoji">Emoji</string>
     <string name="attachment">Attachment</string>
     <string name="back">Back</string>
-    <string name="verified">Verified</string>
+    <string name="verified">Introduced</string>
```
r10s, line 32: the bare `Introduced` wording that is used if we do not have introducer information is a bit weird - however, that affects only old installations, not new ones. so the issue will fade out 💀

res/values/strings.xml
r10s: suggestions welcome :)

res/values/strings.xml
adbenitez: about this one, we could consider removing/hiding the entry in such cases, then no weird word is needed, no extra translation work for legacy stuff, and the verified icon is at the top in the title bar and chat name in chat list and inside chat

res/values/strings.xml
r10s: maybe we can pick up the e2ee term here: `Cannot establish guaranteed end-to-end encryption with %1$s`

hpk42:
thanks for tackling this -- could you also link the FAQ entry/entries that are shown in some "learn more" buttons?

r10s:
> thanks for tackling this -- could you also link the FAQ entry/entries that are shown in some "learn more" buttons?

links are: 

- https://staging.delta.chat/733/en/help#verifiedchats - for the   "learn more" in green-checkmart-info and try-to-add-unverified-contact-to-verified-grtoup 

- https://staging.delta.chat/733/en/help#verificationbroken for the "learn more" in checkmark-broken info

res/values/strings.xml
r10s: i think, i will just change the existing string, so that ui do not need to updates. translators get aware of the change somehow, at least this was the case in the past

res/values/strings.xml
r10s: k, we can easily remove the line on android/ios as the icon is in the title. for desktop, things are different as there is no green checkmark in the title bar, for now, we leave the old string in this case and mark it as deprecated

PR #2805, opened 2023-11-03
don't display "new contact" in verified group contact selection

adbenitez:
the created new contact via the "new contact" button can't be added to the verified group anyways

this also fixes existing crash on master if you try to press the "new contact" button in a verified group because it was assumed all items where contacts and then it is tried to check isVerified() in a null pointer reference since the "new contact" item doesn't holds any real DcContact


r10s (approved):
lvgtm

PR #2806, opened 2023-11-03
fix bugs in GroupCreateActivity

### LINKED ISSUE #2801: "edit group" does not set existing title, "add members" missing
when using "Edit Group" from the group profile's menu, the name is no longer preset

probably a new bug by recent shiftings :)

adbenitez:
close #2801


r10s (approved):
lgtm

PR #2807, opened 2023-11-03
add new-contact activity

### LINKED ISSUE #2794: add explicit "create contact"
creating a contact from the "new chat" is currently quite implicit and hard to discover: one has to enter an email address into the search bar, then a "New Contact" button appears (the "New Contact" button is also present while typing, at the end of the list, saying "Type email address above" - it is not there, when nothing is typed in)

as an improvement, we should do what deltachat-ios or whatsapp are doing:
 
- unconditionally display a "New Contact" button atop if the list (without subtitle)
- when tapping, a new activity with the input fields "E-Mail Address" and "Contact Name"
- when an email address was entered to the search field, this should be used to prefill "E-Mail Address" field (maybe also for name otherwise)

adbenitez:
close #2794

### DIFF
--- a/res/menu/new_contact.xml
+++ b/res/menu/new_contact.xml
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<menu xmlns:android="http://schemas.android.com/apk/res/android" xmlns:app="http://schemas.android.com/apk/res-auto">
+
+    <item android:title="@string/ok"
+          android:id="@+id/menu_create_contact"
+          android:icon="?menu_accept_icon"
+          app:showAsAction="always|withText"/>
+
+</menu>
--- a/res/menu/new_contact.xml
+++ b/res/menu/new_contact.xml
@@ -2,7 +2,7 @@
 
 <menu xmlns:android="http://schemas.android.com/apk/res/android" xmlns:app="http://schemas.android.com/apk/res-auto">
 
-    <item android:title="@string/ok"
+    <item android:title="@string/create"
           android:id="@+id/menu_create_contact"
           android:icon="?menu_accept_icon"
           app:showAsAction="always|withText"/>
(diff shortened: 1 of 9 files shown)


adbenitez:
caveat: new added/created contacts are not sorted at the top of the contact list (they go to the very bottom) because they where not "seen recently" so if you create a new contact without searching first, you will not find the new contact in the list easily sadly, things that could be done to improve this:
always sort marked/selected contacts at the top, then auto-mark as selected the new created contact :thinking:

r10s:
to mitigate the sorting issue:

- in "New Chat", the flow should be "New Chat" -> "New Contact" -> enter contact name -> OK -> go directly to a one-to-one chat with that contact  
that way, we won't show the "New Chat" list after adding a contact

- for new group: hm, not sure if we should reorder the list, this makes things more complicated and may also be unexpected. i would not change that currently. 
but, maybe we can use a similar flow: ... "New Contact" -> enter contact name -> OK -> go back to profile or group-create-summary (contacts selected before would be added as well then, but i think, this is fine)  
or: do not reorder the list completely, but in exactly this case, add just added contact manually to the top (should be a simple loop searching for contact-id and moving it atop - once things are searched or re-rendered otherwise, we are back at the normal sorting

adbenitez:
> * in "New Chat", the flow should be "New Chat" -> "New Contact" -> enter contact name -> OK -> go directly to a one-to-one chat with that contact

:heavy_check_mark: done


> 1. make sure, the contact is selected, even if scrolled out. that seems fine

:heavy_check_mark: done

res/menu/new_contact.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+
+<menu xmlns:android="http://schemas.android.com/apk/res/android" xmlns:app="http://schemas.android.com/apk/res-auto">
+
+    <item android:title="@string/ok"
```
r10s, line 5:
```suggestion
    <item android:title="@string/create"
```

r10s (approved):

r10s:
i think, we can merge that in so that it gets released to 1.42

PR #2809, opened 2023-11-03
make "+ add account" bold and tweak button color in dark mode

### LINKED ISSUE #2808: Color of "+ add account" button in account switcher has too low contrast in dark mode
the old gray color could be used in dark mode

adbenitez:
close #2808


r10s (approved):
lvvgtm

PR #2821, opened 2023-11-05
contact profiles show chat's green checkmarks

r10s:
the contact profiles mix of contact and chat data for the ease of the user.

the contact profile title,
however, shall show the state of the chat (chat_is_protected()) and not the state of the contact (contact_is_verified()) to make things easier to understand and to be in sync with groups - and also with what we're saying in the faq
at https://github.com/deltachat/deltachat-pages/pull/746 :

> "contact profiles show a green checkmark
> if messaging a contact is guaranteed to be end-to-end encrypted."


r10s:
if we agree on that and merge that, we should adapt https://github.com/deltachat/deltachat-desktop/issues/3468 and file an issue for iOS

PR #2822, opened 2023-11-05
revert 'introduced by' to 'verified by'

r10s:
while it is still true that we need
to avoid the term 'verified' in the context of groups, removing 'verified' completely at #2804 was maybe a bit over the top.

for contacts, 'verified by may still be better than 'introduced' which opens the questions if not all contacts are somehow introduced.

so, to sum up, we stroke 'verified group' and 'verified checkmarks' in favor to guaranteed 'end-to-end encryption' -
but we leave 'verified contact' as is.

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -29,7 +29,7 @@
     <string name="emoji">Emoji</string>
     <string name="attachment">Attachment</string>
     <string name="back">Back</string>
-    <!-- this string is deprecated; the new wording is "introduced by ..." -->
+    <!-- deprecated, use verified_by/verified_by_you -->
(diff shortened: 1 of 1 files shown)


r10s:
hm, just read through the [new faq talking about "introduced by"](https://staging.delta.chat/746/en/help#e2eeguarantee), so, not sure about this pr ...

r10s:
argument for staying away from "verified by" is that it often just not true:

> if Alice introduces Bob, 
> this does not mean that Alice has scanned/verified Bob.
> Alice may got the key by scanning Claire.

"introduced by", however, is true in this case.

PR #2824, opened 2023-11-06
remove 40KB option from download limits

### LINKED ISSUE #2823: strike smallest auto-download option
the 40k lead to system messages as "member added" not being processed; even if there are fixes and mitigations, this 40k limit adds unneeded noise.

asking around in known low-traffic environments, the 40k limit is not needed or widely used and somehow artificial.

we should also do a pr in core to forcefully not accept too small limits (there is already a 30k limit, that should be increased to 160k)

adbenitez:
close #2823

### DIFF
--- a/res/values/arrays.xml
+++ b/res/values/arrays.xml
@@ -321,7 +321,6 @@
 
     <string-array name="pref_download_limit_entries">
         <item>@string/pref_show_emails_all</item>
-        <item>40 KiB</item>
         <item>160 KiB</item>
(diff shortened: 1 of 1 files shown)


r10s (approved):
nice, i also checked alignToMaxEntry(): if 40k was selected before, this is changed to 160k selection (but does not alter the value, but core will not allow too small values with https://github.com/deltachat/deltachat-core-rust/pull/4946)

PR #2825, opened 2023-11-06
update KeyboardAwareLinearLayout from upstream

### LINKED ISSUE #2798: title bar disappears sometimes
the title bar disappears sometimes, currently this is reproducible on the phone of @hpk42 

- send a message (virtual keyboard is still on) EDIT: no need to send just type is enough (or even just an open keyboard)
- turn screen off, wait 5+ seconds, turn screen on again

-> the title bar is gone. it reappears eg. when virtual keyboard is hidden

you may need to try this several times until it happens

adbenitez:
close #2798


r10s (approved):
lgtm, tested roughly on android7 and android5

PR #2833, opened 2023-11-07
use QR icon for QR-invite/join button

### LINKED ISSUE #2820: use "QR Code Icon"
Users expect "invite link" or QR code to be somewhere in the group profile, but not in the member list. Because of this, they open the group profile, overlook "QR Invite Code" in the member list and go search for it in 3-dot menu or click the title bar (which opens the avatar in full-screen) in an attempt to find the screen that looks more like a profile with an info section.

While it is logical for "QR Invite Code" to be considered somewhat similar to "Add Members" action, it is not where users look for it.

Maybe transforming two items ("Add Members" and "QR Invite Code") into two overlay buttons with different icons ("+" and "QR code") could help.

As a minimal change, add "QR Invite Code" into three-dot menu as well as keeping it inside the member list.

@kermoshina :
> actually may be good to add group QR in group profile - when clicking on the group name

(this is not a actually a group profile but this huge avatar screen with "edit" button looks more like a group profile than member list)

adbenitez:
close #2820


adbenitez:
image

adbenitez:
> Nice, maybe the icon could be a bit larger (like on desktop), but in any case good enough for merge IMO.

I quickly looked around for an easy way to make it bigger when I was on it, but did not find an easy way after some minutes so didn't keep trying, but should be possible

r10s:
> I quickly looked around for an easy way to make it bigger when I was on it, but did not find an easy way after some minutes so didn't keep trying, but should be possible

i added a commit that replaces raster by vector icon, this makes it bigger in that case.

i used the icon that is offered as "Clip art" build into android studio, i think, the icon even looks clearer to me, so i replaced it also on the main screen, see imanes below

@adbenitez iirc, we're using vector assets already, so that should not be an issue on older android, right?

# before:


before-invite before-main


# after:

after-invite after-main

adbenitez:
> i added a commit that replaces raster by vector icon, this makes it bigger in that case.

great!!! btw, we also need that for the "group icon placehoder" we use in the "new group" the "two persons" icon is a png of 24dp, which is too small and the avatar placeholder in "new group" screen is even bigger so it looks even worse than in the QR screen



> @adbenitez iirc, we're using vector assets already, so that should not be an issue on older android, right?

I am not sure but we are using them already in several places so it should be fine

PR #2835, opened 2023-11-08
hide toolbar onStop and show it again onResume to avoid bug with keyboard

### LINKED ISSUE #2798: title bar disappears sometimes

adbenitez:
close #2798

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -310,6 +310,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
 
     attachmentManager.onResume();
 
+    // action bar might be hidden by workaround in onStop()
     ActionBar supportActionBar = getSupportActionBar();
     if (supportActionBar != null && !supportActionBar.isShowing()) {
         supportActionBar.show();
@@ -333,6 +334,8 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
   public void onStop() {
     super.onStop();
 
+    // hack/workaround to fix https://github.com/deltachat/deltachat-android/issues/2798
+    // the real cause of the problem is still unknown but hidding the action bar here fixes it
     ActionBar supportActionBar = getSupportActionBar();
     if (supportActionBar != null && container.isKeyboardOpen()) {
       PowerManager pm = (PowerManager)getSystemService(Context.POWER_SERVICE);
--- a/src/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationActivity.java
@@ -335,12 +335,14 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     super.onStop();
 
     // hack/workaround to fix https://github.com/deltachat/deltachat-android/issues/2798
-    // the real cause of the problem is still unknown but hidding the action bar here fixes it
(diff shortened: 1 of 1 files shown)


r10s (approved):
another try! thanks a lot for holding one!

adbenitez:
> another try! thanks a lot for holding one!

hopefully this time is finally fixed, unless it re-appears after trying about 100 times in a row :unamused:

adbenitez:
more logs:
```
11-08 18:09:49.615  1466  7157 W ActivityTaskManager: Force removing ActivityRecord{7a87ff0 u0 com.b44t.messenger.beta/org.thoughtcrime.securesms.ConversationActivity t4347 f}}: app died, no saved state
11-08 18:09:49.616  1466  1695 W UsageStatsService: Unexpected activity event reported! (com.b44t.messenger.beta/org.thoughtcrime.securesms.ConversationActivity event : 23 instanceId : 223988073)
```

```
11-08 18:14:01.916  3562  3562 W WindowOnBackDispatcher: OnBackInvokedCallback is not enabled for the application.                                                                                                                          
11-08 18:14:01.916  3562  3562 W WindowOnBackDispatcher: Set 'android:enableOnBackInvokedCallback="true"' in the application manifest.
```

adbenitez:
I think the issue is that fat-debug apk is slower and there the issue happens a lot less than in release-mode apk, will have to switch to testing with release-mode compiled apks

r10s:
yeah, i also had the feeling it is related to device speed - eg. initially we could reproduce it nearly all times, but in the debugger only one of 20 times or so ...

this bug is a hard one, thanks for taking care!

adbenitez:
the issue is solved with this hack: 
* onPause, hide the bar, 
* onResume display the bar

I did this only if keyboard is open to reduce potential issues, but so far it works nice and it is not noticeable

r10s:
is it really required to hide the bar?

this sounds to me like the more dangerous part.

in any case, please add a comment that describes precisely that this is a hack and which cornercase that aims to solve

adbenitez:
> is it really required to hide the bar?

yes, if the bar is not hidden, the issue appears again, at the time onResume() if you check for the bar's visibility it says it is visible, if you anyway call bar.show() it doesn't work, it is like it is crashed in some weird state caused by the keyboard display effect



> this sounds to me like the more dangerous part.

indeed I just realized the onPause() is called also when going back to the chatlist and in its visible the bar disappears while in the transition animation, so I just improved it now this is the current behavior:

only hide the toolbar onStop() AND if keyboard is open AND screen is off

this is a constrained scenario and hiding the toolbar onStop() ensures it is hidden when the activity is no longer visible (onPause() is called when the activity is no longer in foreground but could be still visible, like in the go-to-chatlist transition)

if onStop() was called then onResume() should also be called, it can't happen that the activity was stopped (and the toolbar hidden) and then onResume() is not called, so this should be safe

adbenitez:
> in any case, please add a comment that describes precisely that this is a hack and which cornercase that aims to solve

done :heavy_check_mark:

adbenitez:
good points, I updated according to your suggestions! :+1:

r10s:
and now: fingers crossed!

PR #2836, opened 2023-11-08
use `dc_contact_profile_is_verified()` api

r10s:
this pr makes use of the [new `dc_contact_profile_is_verified()` api](https://github.com/deltachat/deltachat-core-rust/pull/4951)

the api is of limited use as group-chats, one-to-one-chats and contacts-without-chats share same profile code - and for chats, is_protected() is checked already. iOS does the same, probably desktop. so, for the very most profiles, `chat_is_protected()` is called (which is also called by `dc_contact_profile_is_verified()` in case there is a chat - but we cannot always call `dc_contact_profile_is_verified()` it does not work unconditionally for chats)

in practise, we could probably also revert the `dc_contact_profile_is_verified()` api and just use `dc_contact_is_verified()` for contacts-without-chats [as it was always in the past](https://github.com/deltachat/deltachat-android/pull/2821/files).

note that, for whatever reason, `dc_contact_profile_is_verified()` returns false if a verified-one-to-one-chat was deleted before - while `dc_contact_is_verified()` returns true - i think the latter is correct.

if we want to stay with `dc_contact_profile_is_verified()`, this pr needs a new core before merging

### DIFF
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -1913,6 +1913,12 @@ JNIEXPORT jboolean Java_com_b44t_messenger_DcContact_isVerified(JNIEnv *env, job
 }
 
 
+JNIEXPORT jboolean Java_com_b44t_messenger_DcContact_profileIsVerified(JNIEnv *env, jobject obj)
+{
+    return dc_contact_profile_is_verified(get_dc_contact(env, obj))==2;
+}
+
+
 JNIEXPORT jint Java_com_b44t_messenger_DcContact_getVerifierId(JNIEnv *env, jobject obj)
 {
     return dc_contact_get_verifier_id(get_dc_contact(env, obj));
--- a/src/com/b44t/messenger/DcContact.java
+++ b/src/com/b44t/messenger/DcContact.java
@@ -57,6 +57,7 @@ public class DcContact {
     public native boolean wasSeenRecently();
     public native boolean isBlocked      ();
     public native boolean isVerified     ();
+    public native boolean profileIsVerified();
     public native int     getVerifierId  ();
 
     // working with raw c-data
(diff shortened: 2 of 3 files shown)


r10s:
> I don't understand why the concept of "one-to-one chat profile" even exists. I thought there are two types of screens: group chat profile and contact profile. 

well, in practise, one-to-one-chat-profiles share more things with group-profiles than with contacts-without-chats-profiles - disappearing messages, sound, mute, gallery, files, apps - all that exist in group and in one-to-one-profiles but not in  contacts-without-chats-profiles

there is a single `ProfileActivity` that is used for all types - that may get a `chat_id` and/or a `contact_id` as parameter - depending on that, things are shown and hidden (it is more complicated as there is also saved-messages, device-chat, broadcast, mailinglists and maybe more)

`ProfileActivity` then has a title bar - if `chat_id` is set, chat_is_protected() is called.  
we could of course call `contact_profile_is_verified()` in case of one-to-one chats - but to me this looks overcomplex (including unneeded database calls) as if a contact has a chat (and we know this already), `contact_profile_is_verified()` calls `chat_is_protected()` anyways

`contact_profile_is_verified()`, although i was up for that initially as well, and still +/-0, looks a bit like making things flexible for a future that may or may not happen :)

r10s:
> Maybe get rid of all setTitle overloading and make it accept (possibly null) chat and contact? If contact is set, it is a contact view and chat may or may not be passed in. If no contact is set, then chat must be set because it is a chat profile rather than contact profile.

yeah, sth like that can work - and this is also how most of `ProfileActivity` work, see above.

but still, it does not change much as core also just checks

`profile_verified = if (chat_exist) chat_is_protected() else contact_is_verified()`

so, maybe just an updated documentation instead of a new api that needs some changes on all UI would have been enough until we _really_ need more complexity.

r10s:
> Currently lookup call actually returns even hidden chats as far as I see.

i also noticed that on tries, that looks like a bug to me

r10s:
> So I'm +1 on reverting the API and closing this PR.

closing this PR and create an API revert PR at https://github.com/deltachat/deltachat-core-rust/pull/4966

PR #2837, opened 2023-11-09
Revert "contact profiles show chat's green checkmarks"

r10s:
This reverts commit f6c4e189e3b35989acc30396a495d310868022ef.

after lots of discussion (#2836, https://github.com/deltachat/deltachat-core-rust/pull/4966), contact-profiles-without-chat, should show is_verified() for the green checkmarks as in the past.

it still makes sense to streamline the android code, but that is another issue.

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationTitleView.java
+++ b/src/org/thoughtcrime/securesms/ConversationTitleView.java
@@ -130,10 +130,13 @@ public class ConversationTitleView extends RelativeLayout {
     avatar.setAvatar(glideRequests, new Recipient(getContext(), contact), false);
     avatar.setSeenRecently(contact.wasSeenRecently());
 
-    title.setText(contact.getDisplayName());
+    int imgRight = 0;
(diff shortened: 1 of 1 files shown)


adbenitez (approved):
I am not 100% sure this is the less confusing thing for the users, but this mix of verified contacts and verified 1:1 chats are hard to explain and understand by the user, sadly

r10s:
> I am not 100% sure this is the less confusing thing for the users, but this mix of verified contacts and verified 1:1 chats are hard to explain and understand by the user, sadly

there is no such things as "verified 1:1 chats" that needs to be explained to the user :)

anyways, not sure what you mean here for this concrete change.

the change _only_ affects contact profiles where no one-to-one-chat exist at all yet. and for these profiles, now you see the same "green checkmark" as in member list

for contacts profiles where a one-to-one chat exist, always chat_is_protected() was used for the green checkmark, this is not changed by this pr

adbenitez:
ah I misread, thanks for explaining! :+1:

PR #2838, opened 2023-11-10
upgrade some dependencies

adbenitez:
we can merge this after 1.42

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 plugins {
-    id 'com.android.application' version '7.4.2'
+    id 'com.android.application' version '8.1.3'
 }
 
 repositories {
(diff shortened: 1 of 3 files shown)


adbenitez:
(CI failed because new gradle plugin requires java 17, I have to adapt the CI setup that is using old java)

build.gradle
```diff
@@ -86,7 +86,7 @@ dependencies {
 android {
     namespace "org.thoughtcrime.securesms"
     flavorDimensions "none"
-    compileSdkVersion 33
+    compileSdk 34
```
adbenitez, line 89: required by webkit 1.8.0

adbenitez:
I already had this merged in my fork since November, so far so good

build.gradle
```diff
@@ -216,6 +214,10 @@ android {
     lint {
         abortOnError false
     }
+    buildFeatures {
+        renderScript true
+        aidl true
+    }
```
r10s, line 220: this addition added by some automation? otherwise (or in general :) - what does it mean?

gradle.properties
```diff
@@ -1,3 +1,6 @@
+android.defaults.buildfeatures.buildconfig=true
 android.enableJetifier=true
+android.nonFinalResIds=false
+android.nonTransitiveRClass=false
```
r10s, line 4: same here: did that come by some automation?

build.gradle
adbenitez: yes, this was done automatically by AndroidStudio, it shows a dialog saying that your gradle plugin version is behind and offers to upgrade it, then it does that changes automatically, porting to the new config idioms of the new version it seems and some settings get deprecated etc

gradle.properties
adbenitez: Res IDs are non-final now in recent gradles, you probably also saw some warning in android studio where we have switch statements with branches using menu IDs etc. saying that such Res IDs are no final etc. it seems this setting here disables the dynamic Res IDs avoiding issues for our usage,

this change was done automatically by AndroidStudio

r10s (approved):

PR #2841, opened 2023-11-11
improve NewContactActivity, add "scan QR button"

adbenitez:
when creating a new contact manually, it is useful to hint the user towards the more secure "QR scan" workflow to increase the chance of getting a verified 1:1 chat


r10s:
did not look much at the code yet, but i would also swap qr code and input fields.

also, i think, the email address should be top; usually, optional fields follow, but the first one should be required. this is also the order we're using on iOS since forever.

some mockups (i did not want to mess up with non-stack-android-layouts :)





both would be fine for me atm. while the "or" divider is fancy, maybe we should go for a stupid neutral text line saying

`or enter contact data manually below:`

missytake:
In Thunderbird when you login, you have three fields:

name
email address
password

and it's veeery common that people (I have caught up by now because I configured so many accounts in thunderbird) enter their email address at "name".
So I would suggest:

email address
name (optional)

for the order and wording.

adbenitez:
> A placeholder `Name (optional)` would be nice to indicate that it is not really needed as @Simon-Laux suggested in the chat.

not entering any name is worse experience for the user, so while supported I am not sure it is worth to mention, it is better if you set a name instead of just the address, names are easier to remember etc.also this PR is about adding the QR button, we should focus on that, the dialog was like that before, so I am not sure if doing so much changes that require new translations close to the new release etc. makes sense in this PR that is about introducing the QR button

r10s:
yeah, let's keep reordering of fields and rewording of placeholder out of this issue. i also drawback my comments, sorry for unneeded noise :)

res/values/strings.xml
```diff
@@ -4,6 +4,7 @@
     <string name="app_name">Delta Chat</string>
     <string name="ok">OK</string>
     <string name="cancel">Cancel</string>
+    <string name="or_separator">or</string>
```
r10s, line 7: a hint for translators is needed here

```suggestion
   <!-- the word "or" to separate blocks in the user interface that are mutually exclusive --->
    <string name="or_separator">or</string>
```

res/layout/new_contact_activity.xml
```diff
@@ -35,16 +35,70 @@
...
+            app:layout_constraintTop_toTopOf="parent"/>
+
+        <Button
+            style="@style/ButtonSecondary"
+            android:id="@+id/qrscan_btn"
             android:layout_width="wrap_content"
-            android:layout_height="28dp" />
+            android:layout_height="wrap_content"
+            android:paddingLeft="40dp"
+            android:paddingRight="40dp"
```
r10s, line 48:
```suggestion
            android:paddingLeft="20dp"
            android:paddingRight="20dp"
```

i would use less padding here and also fewer padding atop / abottom of the "or ornament" (2/3rd or so, not totally sure which lines to tweak for that)

that way, things still look good but are are shifted a bit to the top, making it less probable, the address input filed is not covered by the keyboard on smaller devices

r10s (approved):
i am not sure if the "or ornament" is worth the effort, maybe some simpler approach would be fine as well :)

res/layout/new_contact_activity.xml
adbenitez: I reduced the padding of the divider from 50dp to 35dp

PR #2844, opened 2023-11-12
improve new-contact activity

adbenitez:
image

### DIFF
--- a/res/drawable/ic_alternate_email_24.xml
+++ b/res/drawable/ic_alternate_email_24.xml
@@ -1,0 +1,5 @@
+<vector android:height="24dp" android:tint="#FFFFFF"
+    android:viewportHeight="24" android:viewportWidth="24"
+    android:width="24dp" xmlns:android="http://schemas.android.com/apk/res/android">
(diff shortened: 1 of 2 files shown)


adbenitez:
I think with the icons changing the fields order is not necessary since having the name at the top is more natural, but if still wanted we can change that in this PR

also we could drop the "or" divider and put some text less fancy but more useful? or will we stick with the divider for now?

r10s:
the input-underline should not be below the icon - or do you know of "larger" examples where this is the different? 

it currently looks like a character one can delete

the @ for email: the letter ✉️ seems to be more common

tbh, i am not convinced about the icons at all, if they only look "nice" (a matter of taste then) or help on confusion. on ios they're less a thing than on android, however, so, yes we can give it a try

on ios, instead, we have a placeholder that makes it clear what the email address field is:

r10s:
for the divider: i would also replace that by some calmer, useful text

r10s:
sleeping it over, i am not convinced that adding icons is the right way to move forward here and solve ux issues.

- we do not have icons for "name" or so at other places in the app - why add it here and not anywhere? (answer: it will clutter UI and should not be done without reasons) - so we should have good reasons to add things and break consistency

- before adding icons, it makes more sense to think over correct placeholder texts, that could help even more

- as 90% of icons, the icons are not self-explaining, maybe they do not need to be, however, that way they will remove attention from the text and the "scan qr code button" (we're assuming ppl do not really look closely)

- finally, we do not even know that there is an issue, the dialog is not rolled out at all

all in all, i am in favour to close this pr, and rethink if there are real-world issues

adbenitez:
> the input-underline should not be below the icon - or do you know of "larger" examples where this is the different?

I just used the material design component and that is how google decided it should look, it was not intentional, I think I also would prefer it outside, but not strong opinion there



> all in all, i am in favour to close this pr, and rethink if there are real-world issues

the idea to add icons came from some devs thinking that users could get confused with the fields, in fact, I think it is not confusing at all and there is no way users can enter wrong data, so there is no real need for the icons, but IMHO it is true that having them makes things even easier for the user, but not really needed so I am fine with closing this PR :+1:

PR #2846, opened 2023-11-13
add contact: more room for name+email address

r10s:
on smaller screens, email address easily is covered by the keyboard; which will result in lots of confusion (probably much more as was discussed by field order or icon yes/no :)

this pr tweaks the or-ornament a little,
without looking much different afterwards
(less padding below, adapted colors)

before / after, on a real-world typical 5.2" device (lg x cam):

 

the keyboard has a tenkey-row, but is otherwise quite standard - other keyboards have a suggestion row and and probably of similar size

of course, there can always be smaller screens and scrolling is also in place, however, it is good to target typical 5.sth" devices, reducing the amount of devices with this issues a lot


adbenitez:
an screenshot in dark mode would be useful to compare the change in the colors of the divider tuning it down in light mode means making it stronger in dark mode :thinking:

adbenitez (approved):

r10s:
> an screenshot in dark mode would be useful to compare the change in the colors of the divider tuning it down in light mode means making it stronger in dark mode 🤔

thanks for pointing that out. indeed, i did not try in dark mode. i added a commit that makes the line lighter in light mode  and darker in dark mode

PR #2848, opened 2023-11-13
remove unused resources

r10s:
the "red trashcan" looks very familiar to me, however for avartar-removing, we use sth different? @adbenitez  can you please double check we do not delete sth. that is actually in use and the resource-checker is wrong


adbenitez:
> the "red trashcan" looks very familiar to me, however for avartar-removing, we use sth different? @adbenitez can you please double check we do not delete sth. that is actually in use and the resource-checker is wrong

in fact we used that icon in the past as the app-icon/launcher for an delete-avatar activity that appeared in the list of activities to complete action when selecting the avatar (camera app, our delete avatar activity, etc)

the delete-avatar activity was removed in https://github.com/deltachat/deltachat-android/pull/1837 when we switched to our internal AttachmentType selector, the [clear_profile_avatar image](https://github.com/deltachat/deltachat-android/commit/4c0d03772d4b5ac7d9cf964080115a4f15689640#diff-90f0a3019e1c91cfede9d4eaf5d6527ba4925d0a5ae75e97b71dfd96d365773eL262) was replaced with a [CircleColorImageView](https://github.com/deltachat/deltachat-android/commit/4c0d03772d4b5ac7d9cf964080115a4f15689640#diff-1f6d1bdc7f751d5eac8fdde297c4e9776744281f9b77012336ba388c8ac2b445R66)

TL;DR it is safe to remove :)

adbenitez (approved):

PR #2849, opened 2023-11-13
remove unused code

r10s:
ran "Run Inspection By Name > Unused Declaration" and removed some larger things (smaller things are no big win and may also make pulling things in from upstream harder)

all in all, not so much dead code - this is also good to know :)


adbenitez (approved):
:broom: niceeee

PR #2851, opened 2023-11-14
Create strings.xml

ngocanhtve:
Add Vietnamese language.

### DIFF
--- a/res/valuse-vn/strings.xml
+++ b/res/valuse-vn/strings.xml
@@ -1,0 +1,1094 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <!-- common strings without special context -->
+    <string name="app_name">Delta Chat</string>
+    <string name="ok">OK</string>
+    <string name="cancel">Hủy</string>
+   <!-- the word "or" to separate blocks in the user interface that are mutually exclusive -->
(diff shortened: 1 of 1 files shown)


r10s:
wow! thanks a lot! 🙏

i will push that with the iso-code `vi` directly to [transifex](https://delta.chat/en/contribute) (where future additions should take place then, you can log in there with your github handle, but as said, not needed just now)

@ngocanhtve the in-app language selector always shows the local name, wikipedia says the local name of the language is "tiếng Việt" - i would add that to https://github.com/deltachat/deltachat-android/blob/main/res/values/arrays.xml#L38 then

ngocanhtve:
Thank you!

r10s:
k, i did the changes at https://github.com/deltachat/deltachat-android/pull/2852

thanks again a lot for the complete (!) translation!

we have a small translators group, usually only few messages per week - in case you're using Delta Chat on your own, i can add you there (and if you do not use Delta Chat but want to, i can also help with a dedicated email address, just text me under &nbsp; _r10s &nbsp;AT&nbsp; b44t.com_ &nbsp; )

good timing btw, i was doing a release anyways just now :)

ngocanhtve:
I'm using Delta Chat. My email is ngocanh.tve@gmail.com. Thank you.

PR #2855, opened 2023-11-14
try to improve device message

adbenitez:


res/values/strings.xml
```diff
@@ -1088,7 +1088,7 @@
 
 
     <!-- device messages for updates -->
-    <string name="update_1_42_common">What\'s new:\n\n💌 one-to-one chats guarantee end-to-end encryption for introduced contacts now\n\n😌 for everyone\'s simplicity, we mark these contacts and chats with green checkmarks\n\n👥 groups are created automatically with guaranteed end-to-end encryption if possible\n\n🔄 Accept/Blocked, Archived, Pinned, Mute is synced across all your devices</string>
+    <string name="update_1_42_common">What\'s new:\n\n💌 now one-to-one chats guarantee end-to-end encryption for contacts with a green checkmark\n\n😌 for everyone\'s simplicity, we also mark these chats with green checkmarks\n\n👥 groups are created automatically with guaranteed end-to-end encryption if possible\n\n🔄 Accept/Blocked, Archived, Pinned, Mute is synced across all your devices</string>
```
r10s, line 1091: yeah, you have a favour in putting "now" to the beginning. apart for "now one..." not being fluent, i think, both are fine somehow, but note that different aspects are emphasised:
- "now" at the beginning emphasises **the timing**, the current state of things, without saying much about the future
- "now" at the end emphasises **the outcome**, that things have changed since then and suggests that this is a current and ongoing feature

```suggestion
    <string name="update_1_42_common">What\'s new:\n\n💌 one-to-one chats guarantee end-to-end encryption for contacts with a green checkmark now\n\n😌 for everyone\'s simplicity, we also mark these chats with green checkmarks\n\n👥 groups are created automatically with guaranteed end-to-end encryption if possible\n\n🔄 Accept/Blocked, Archived, Pinned, Mute is synced across all your devices</string>
```

r10s (approved):
lgtm, but maybe let's wait a little bit with merging until it is actually needed, so that we can avoid interating over another time, cc @Simon-Laux

res/values/strings.xml
adbenitez: the problem for me is that "for contacts with a green checkmark now" (or the previous "for introduced contacts now") read weird like "for contacts introduced now" like it is not clear if the now is about the 1:1 guarantee or the green check mark, but that might be me as a non-English speaker, the translation to Spanish would also be confusing if I do it like that

res/values/strings.xml
```diff
@@ -1088,7 +1088,7 @@
 
 
     <!-- device messages for updates -->
-    <string name="update_1_42_common">What\'s new:\n\n💌 one-to-one chats guarantee end-to-end encryption for introduced contacts now\n\n😌 for everyone\'s simplicity, we mark these contacts and chats with green checkmarks\n\n👥 groups are created automatically with guaranteed end-to-end encryption if possible\n\n🔄 Accept/Blocked, Archived, Pinned, Mute is synced across all your devices</string>
+    <string name="update_1_42_common">What\'s new:\n\n💌 one-to-one chats guarantee end-to-end encryption for contacts with a green checkmark now\n\n😌 for everyone\'s simplicity, we also mark these chats with green checkmarks\n\n👥 groups are created automatically with guaranteed end-to-end encryption if possible\n\n🔄 Accept/Blocked, Archived, Pinned, Mute is synced across all your devices</string>
```
adbenitez, line 1091: what about striking the "now"? we already say "what is new:" and then immediately that so you could also just say:
"What\'s new:\n\n💌 one-to-one chats guarantee end-to-end encryption for contacts with a green checkmark"

adbenitez: btw, the corrector tells me "checkmark" is not valid word

res/values/strings.xml
adbenitez: > * "now" at the end emphasises **the outcome**, that things have changed since then and suggests that this is a current and ongoing thing

I guess the correct thing then would have been:

"one-to-one chats guarantee end-to-end encryption **now** for contacts with a green checkmark"

not to put it at the very end of everything :see_no_evil:

res/values/strings.xml
adbenitez:
```suggestion
    <string name="update_1_42_common">What\'s new:\n\n💌 one-to-one chats guarantee end-to-end encryption for contacts with a green checkmark\n\n😌 for everyone\'s simplicity, we also mark these chats with green checkmarks\n\n👥 groups are created automatically with guaranteed end-to-end encryption if possible\n\n🔄 Accept/Blocked, Archived, Pinned, Mute is synced across all your devices</string>
```

adbenitez:
ok I think it is good enough now, /me done tweaking, feel free to merge or further tweak! :+1:

Simon-Laux:
much better 👍

PR #2859, opened 2023-11-14
ask for system unlock before accessing "password & account"

### LINKED ISSUE #2857: request pattern-unlock when accessing the "password and account" settings
currently the password field is hidden, but still is possible to copy it with a keyboard like AnySoftKeyboard or to modify address to a fake email server that captures the password / login attempt, it is also bad if someone manipulates your settings ex. to disable SSL/STARTTLS, strict certificate checks or SOCKS5, the whole configuration screen is too sensible and should be protected like the backup option.

then we can enable back displaying the password field so users can recover their password in case of loss

adbenitez:
make the "password and account" section more secure, ~~while at the same time more useful for that cases when you forgot your account's password, allowing to reveal your existing password, also useful for troubleshooting when you changed your account password~~

close #2857


r10s:
great! this should go after 1.42, however

r10s (approved):
code lgtm, not tested.

already approving as i am on that, merging should wait after 1.42 is released

r10s:
thinking it over, also after a discussion with  @hpk42 about situations that will cause mostly unsolvable troubles for many users: allowing to access the password is not a thing that helps us getting things sorted

> while at the same time more useful for that cases when you forgot your account's password

this will push users to a direction to "just log in" on another device, where they will get another key - writing from these two devices will cause a lot of friction, you really need to know what you're doing to solve that - and these users do not know their passwords? :)

surely, we have to deal with that "log in on multiple devices" anyways, however, we're not there. if we do not allow accessing the password, ppl will more probably find "Add Second Device", which avoids all these key issues

this is even more true for the upcoming chatmail  - and if users are using a regular account, they probably have their password ready somehow anyways (or gmx, gmail whatnot have their recovery methods)

maybe this is not a big deal, however, taking into account that it is like it is for years and not many ppl complain, i am wondering if it is really needed and a good timing to push for that flow just now - where we push towards multi device and chatmail

(asking for request lock pattern on accessing account details makes sense anyways)

adbenitez:
if it is not wanted feel free to close this PR and issue, I think it is a nice to have feature, will add it to deltalab then

r10s:
well, i am not strictly against the feature, i just want to underline, that it will have maybe unexpected side effects, creating maybe more harm than use :) 

> _"What did you do to create this mess?"_-- _"I just logged in with Delta Chat on my second device"_

surely, there are things as password forgotten or it'll be handy for us devs (i would like it!) - but apart from that:

which larger userbase we aim to target do you think would need that feature? maybe i've overseen things

r10s:
>  the lock screen is wanted, so please confirm if I should revert the "allow to show password" part

let's just leave the pr as is for now, maybe there are more comments than my 2ct. it won't be merged the next days anyways

adbenitez:
> which larger userbase we aim to target do you think would need that feature? maybe i've overseen things

in fact it is not that important, either way the impact is low, but I find it more useful than not allowing it and also know people that had this issue in the past and wanted to recover their password having it only in delta chat, etc. and currently some others are in the same situation

even for chatmail, where btw, password recovery is not possible at all so having your password under your control is not a bad idea if you have a years-old account you don't want to lose out of a sudden

also desktop allows to see the password, so maybe it should also be consistent and hide it there, but in practice even we use it for testing and it is useful

r10s:
nb: there is the idea on desktop/ios to move "Password and Account" one level up: https://github.com/deltachat/deltachat-ios/pull/2003

src/org/thoughtcrime/securesms/CreateProfileActivity.java
```diff
@@ -208,8 +218,8 @@ private void initializeResources() {
     this.statusView   = ViewUtil.findById(this, R.id.status_text);
 
     passwordAccountSettings.setOnClickListener(view -> {
-      Intent intent = new Intent(this, RegistrationActivity.class);
-      startActivity(intent);
+      boolean result = ScreenLockUtil.applyScreenLock(this, getString(R.string.pref_password_and_account_settings), getString(R.string.enter_system_secret_to_continue), ScreenLockUtil.REQUEST_CODE_CONFIRM_CREDENTIALS);
+      if (!result) { openRegistrationActivity(); }
```
r10s, line 222:
```suggestion
      if (!result) {
        openRegistrationActivity();
      }
```

although on-liners are used here and there, and sometimes, there are even reasons for it, i do not see them here.

but not a blocker, of course, feel free to leave as is :)

r10s (approved):
nice, i think, we can merge that now, 1.42 is settled enough and this is indeed a minor change :)

PR #2864, opened 2023-11-15
check for DC_DOWNLOAD_UNDECIPHERABLE download state

adbenitez:


src/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -390,7 +390,7 @@ else if (text.isEmpty()) {
     }
 
     int downloadState = messageRecord.getDownloadState();
-    if (downloadState != DcMsg.DC_DOWNLOAD_DONE) {
+    if (downloadState != DcMsg.DC_DOWNLOAD_DONE && downloadState != DC_DOWNLOAD_UNDECIPHERABLE) {
```
adbenitez, line 393: yikes, I was too fast merging, `DcMsg.` in front is missing

r10s, line 393: for me, a positive list is easier to read,  and also matches better the conditions in the branch.

which one is better in the future, is unclear, however - so feel free to ignore this suggestion :)

```suggestion
    if (downloadState == DcMsg.DC_DOWNLOAD_AVAILABLE || downloadState == DcMsg.DC_DOWNLOAD_FAILURE || downloadState == DcMsg.DC_DOWNLOAD_IN_PROGRESS) {
```

r10s (approved):

src/org/thoughtcrime/securesms/ConversationItem.java
r10s: k, too slow :) but no need to change, just forget my suggestion :)

adbenitez: I followed your suggestion in a new PR

PR #2865, opened 2023-11-15
fix bug insetBodyText()

adbenitez:


r10s (approved):
nice!

btw, this is the iOS counterpart of the logic: https://github.com/deltachat/deltachat-ios/pull/1981

weird PR title, btw

adbenitez:
> weird PR title, btw

at the beginning was about fixing caused by the previous PR (missing `DcMsg.` in front of the UNDECIPHERABLE constant) but then say your suggestion which also caused the bug to be fixed ;)

PR #2866, opened 2023-11-15
prepare 1.42.0

r10s:
k, this will be the first release candidate, makes sense to start with that now, indeed


r10s:
> Let me fix the "Member added" on the core first

k, but no hurries, i am afk now ... we'll do an new 1.42 with a fresh core tomorrow then

r10s:
sorry for being so impatient, the techno music is just too fast here ...

PR #2873, opened 2023-11-17
fix contact creation

### LINKED ISSUE #2871: After backup restore, 1:1 chat name overridden manually (by user) before changes to name of corresponding email contact in Android address book when message from respective contact is received
- Android version:
Android 12.

- Device:
Google Pixel 3a (sargo).

- Delta Chat version:
1.42.1.

- Expected behavior:
A manually (by the user) applied name for a 1:1 chat does not change after a backup is restored, and after a message from a contact who is assigned to that chat is received.

- Actual behavior:
A manually (by the user) applied name for a 1:1 chat changes after a backup is restored, and after a message from a contact who is assigned to that chat is received.

-- Run a previous version (e.g. 1.40.1) of Delta Chat.
-- Ensure that your Android address book includes a contact (e.g. `Alice C.`) with an email address that corresponds to a 1:1 chat in Delta Chat.
-- Rename the corresponding 1:1 chat in Delta Chat (e.g. from `Alice C.` to `Alice Cooper`).
-- Create a complete backup of your account, chats, and media.
-- Go to Android settings and clear the cache and storage that are applied to Delta Chat.
-- Uninstall Delta Chat.
-- Install Delta Chat version 1.42.1.
-- Restore the backup which you created before.
-- Receive a message from the contact whose 1:1 chat name you changed before. (Result: The name of the corresponding 1:1 chat has changed from `Alice Cooper` to `Alice C.`.)

- Screenshots:
N/A.

- Logs:
[deltachat-log-20231117-084832_2.txt](https://github.com/deltachat/deltachat-android/files/13390958/deltachat-log-20231117-084832_2.txt)

r10s:
this pr fixes my findings at https://github.com/deltachat/deltachat-android/issues/2871#issuecomment-1816239147 by looking up the contact-id before creating it with an empty name.

in general, it seems undefined if you call createContact() with an empty name on an existing address with a name set. as createContact() is really only for "user input", it is probably okay to reset the name (and it was wrong to use this function here, maybe a relict from the old UI; the code is also overcomplex as it passes addresses though where IDs should be good enough, but yes ...)

it need to be tested if this pr also fixes the issue @gerryfrancis reported at #2871 EDIT: closes #2871

### DIFF
--- a/src/org/thoughtcrime/securesms/NewConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/NewConversationActivity.java
@@ -107,10 +107,13 @@ public class NewConversationActivity extends ContactSelectionActivity {
                 .setCancelable(true)
                 .setNegativeButton(android.R.string.cancel, null)
                 .setPositiveButton(android.R.string.ok, (dialog, which) -> {
-                  int contactId1 = dcContext.createContact(null, addr);
+                  int contactId1 = dcContext.lookupContactIdByAddr(addr);
                   if (contactId1 == 0) {
-                    Toast.makeText(NewConversationActivity.this, R.string.bad_email_address, Toast.LENGTH_LONG).show();
-                    return;
+                    contactId1 = dcContext.createContact(null, addr);
+                    if (contactId1 == 0) {
+                      Toast.makeText(NewConversationActivity.this, R.string.bad_email_address, Toast.LENGTH_LONG).show();
+                      return;
+                    }
                   }
                   openConversation(dcContext.createChatByContactId(contactId1));
                 }).show();
--- a/src/org/thoughtcrime/securesms/NewConversationActivity.java
+++ b/src/org/thoughtcrime/securesms/NewConversationActivity.java
@@ -99,7 +99,7 @@ public class NewConversationActivity extends ContactSelectionActivity {
     else {
       int contactId = dcContext.lookupContactIdByAddr(addr);
       if (contactId!=0 && dcContext.getChatIdByContactId(contactId)!=0) {
-        openConversation(dcContext.createChatByContactId(contactId));
+        openConversation(dcContext.getChatIdByContactId(contactId));
       } else {
         String nameNAddr = contactId == 0 ? addr : dcContext.getContact(contactId).getNameNAddr();
         new AlertDialog.Builder(this)


src/org/thoughtcrime/securesms/NewConversationActivity.java
```diff
@@ -99,18 +99,21 @@ public void onContactSelected(int specialId, String addr) {
     else {
       int contactId = dcContext.lookupContactIdByAddr(addr);
       if (contactId!=0 && dcContext.getChatIdByContactId(contactId)!=0) {
-        openConversation(dcContext.createChatByContactId(contactId));
+        openConversation(dcContext.getChatIdByContactId(contactId));
```
adbenitez, line 102: why is this needed? the create-chat thing was used to unblock chats or something like that IIRC :thinking: 

in any case, what is the downside of using createChatByContactId? (just want to know)

adbenitez (approved):

gerryfrancis:
I think the APK from the PR seems to solve the (other) issue, I need to test it on the device of my mate who experienced the problem. Will keep you updated... :)

gerryfrancis:
So far so good, but I have experienced one more (related?) issue with this APK:

When I receive a message from a contact (e.g. `Sam Smith` with emails address `sam.smith@example.org`), who deleted the display name in DC, and whom I do not have registered in my Android address book (double-checked), the prior name (`Sam Smith`) is maintained. (Note: The email is sent just with `sam.smith@example.org` in the `From:` field.) I have no idea where this full name comes from, but that is in a verified 1:1 chat, does that matter?

gerryfrancis (commented):
> I think the APK from the PR seems to solve the (other) issue, I need to test it on the device of my mate who experienced the problem.

I have just tested this APK on the other device where the issue happened, and the fix solves it, thank you very much! 👍

r10s:
thanks a lot @gerryfrancis for retesting and confirmation, you're great!

src/org/thoughtcrime/securesms/NewConversationActivity.java
r10s: in a perfect world, both would be fine, however, `getChatIdByContactId()` is a read-only-operation - and therefore a precaution against bugs as the weirdly changed name

gerryfrancis:
@r10s My pleasure, thank you for the fast support, always glad to help! :)

r10s:
> So far so good, but I have experienced one more (related?) issue with this APK:
>
> When I receive a message from a contact (e.g. Sam Smith with emails address sam.smith@example.org), who deleted the display name in DC, and whom I do not have registered in my Android address book (double-checked), the prior name (Sam Smith) is maintained. (Note: The email is sent just with sam.smith@example.org in the From: field.) I have no idea where this full name comes from, but that is in a verified 1:1 chat, does that matter?

i think, this is wanted behavior. if you send  from a device with no name set, it is most probably just forgotten to be set - and better to keep the existing name

is there a change wrt 1.42?

gerryfrancis:
> is there a change wrt 1.42?

@r10s I have tested version 1.40.1 (removing the sender name), and ...

> if you send from a device with no name set, it is most probably just forgotten to be set - and better to keep the existing name

... that is correct, nothing has changed in this regard, so we are just fine - thanks! :)

PR #2891, opened 2023-11-28
Update to Rust 1.74

link2xt:


adbenitez (approved):
what is the advantage of upgrading to 1.74? is there any nice improvement that would benefit us? otherwise I would prefer to do these upgrades less frequently since it is a burden for me to download etc and slows down my work

PR #2892, opened 2023-11-28
add 'Password and Account' button to settings

r10s:
- get the button out of the hidden place in profile, however, still move it down to tune it down (it is very often not neeed)

- as we have 7+ items now, group "info" things together (the headline is not really needed, more a layout things)

if we want to go that way, the buttons needs to be filled with life (currently it is not working)

counterpart of https://github.com/deltachat/deltachat-ios/pull/2003


adbenitez (approved):
to be honest I am a bit skeptical of this change, the icon looks a bit unrelated, the place is also a bit random, after all in the account profile settings is the place where I would expect such settings, but maybe is just my old-user bias ;)

r10s:
i am also meanwhile unsure if it is really wn improvement, we did not really got complains berr, did we?

sorting non-settings to "info" otoh makey sense

r10s:
i created a more top-level discussion for that at https://github.com/deltachat/interface/issues/57 , let's figure things out there first and then just apply that on android

r10s:
currently all systems are in-sync; let's park this PR in resurrection until https://github.com/deltachat/interface/issues/57 is resolved

r10s:
moved to "advanced" on all platforms; due to "chatmail" changes this became less important

PR #2907, opened 2023-12-11
display title bar and back button in welcome screen

### LINKED ISSUE #2870: add "close" button in the upper left corner of the welcome title bar
some people click "add new account" then don't know how to go back, or close the app and forget then opening again opens in welcome screen, and they think their accounts are lost

adbenitez:
close #2870

### DIFF
--- a/src/org/thoughtcrime/securesms/WelcomeActivity.java
+++ b/src/org/thoughtcrime/securesms/WelcomeActivity.java
@@ -102,6 +102,7 @@ public class WelcomeActivity extends BaseActionBarActivity implements DcEventCen
 
         boolean canGoBack = AccountManager.getInstance().canRollbackAccountCreation(this);
         supportActionBar.setDisplayHomeAsUpEnabled(canGoBack);
+        getSupportActionBar().setTitle(canGoBack? R.string.add_account : R.string.welcome_desktop);
     }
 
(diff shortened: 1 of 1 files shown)


r10s:
i think, we should differ between the mode of "very first startup" and "adding an account  later"

- on first startup, use the title "Welcome to Delta Chat" and do not show the "back" button (it is not needed, the fewer elements the better, ppl know how to exit an app, and we also did not had that before)

- later on, use "Add Account" as the title (so, same dialog title as the invoking button reads, this is always good style :)

adbenitez:
> * and do not show the "back" button

this is already the case



> * on first startup, use the title "Welcome to Delta Chat"

done :+1:

adbenitez:
btw later on we could consider now that we have a title bar, to also put a menu there, with things like a menu option to open the offline help and maybe a link to "online support" opening the forum, but IIRC at the moment it is not possible to open the offline help from this screen

r10s:
> but IIRC at the moment it is not possible to open the offline help from this screen

yip, at some point we should streamline the two base classes to one and implement special activity redirecting for add-second-device- or welcome-activity over a flag

r10s (approved):
lgtm now, also tested a bit. thanks a lot!

Issue #2909, opened 2023-12-28
Tracking issue: Self-reporting to collect real-world statistics


Hocuri:
I'll update the issue description here whenever there is some new discussion on the topic. Below we can have an "unordered" discussion.

Goal: Collect realworld statistics on usage of verified chats and other parts of Delta Chat by allowing the user to send us some anonymized info. Not only will this be be opt-in, but also the user will be able to see what will be sent, and control when it is sent.

Timeline: Feb 2024

- [x] Implement an opt-in self-reporting mechanism for verified chat usage on Android ([UI](https://github.com/deltachat/deltachat-android/pull/2910), [Core](https://github.com/deltachat/deltachat-core-rust/pull/5129))
- [x] [Implement a data-collection bot that stores anonymized user-reports for inspection and analysis by developers.](https://github.com/deltachat/self_reporting_bot/blob/main/self_reporting_bot.py) ([Deploying PR](https://github.com/deltachat/sysadmin-pyinfra/pull/44))
  - [x] Include the bot's public key as "verified" in DC Core, so that this works with chatmail
    - [ ] Maybe we should fixate the private key on the bot-side and import it on deploy, so that even a fresh setup works and keeps the key
- [ ] Refine which metrics actually are interesting to us
- [ ] Possibly change the name from self_reporting_bot to statsbot
- [ ] Possibly send the statistics in an attachment instead of message text (see below)
- [x] Release and deploy privacy-preserving data collection bot
- [ ] Release self-reporting mechanism on Android
- [ ] List of prioritized verified chat issues obtained from field data

### Can be extracted from the db today

I checked the boxes of whatever metrics I already implemented.

- [x] Key creation date
- [x] #messages
- [x] #chats
- [x] #chats using verified, opportunistic, or no encryption
- [x] #chats with dc/non-dc users
- [ ] Android system version
- [x] Core version
- [ ] Android UI version
- [x] DB size
- [ ] Anything else that's shown at the beginning of the log already

### Additional metrics we could collect during usage

Note that before implementing these, we'll have to check whether having this information on the phone could be dangerous, e.g. maybe having created a lots of groups could make you suspicious, since you're probably some sort of leader? Probably I'm overthinking it, though.

- [ ] #failed QR Code Scans / securejoin (-> is this a realword problem?)
- [ ] #not decryptable messages (-> is this a realword problem?)
- [ ] #failed decryptions (-> is this a realword problem?)
[...]

PR #2910, opened 2023-12-28
Self-reporting, Android UI

Hocuri:
Depends on https://github.com/deltachat/deltachat-core-rust/pull/5129

Part of https://github.com/deltachat/deltachat-android/issues/2909

### DIFF
--- a/res/xml/preferences_advanced.xml
+++ b/res/xml/preferences_advanced.xml
@@ -93,6 +93,9 @@
             android:title="@string/videochat_instance"
             android:summary="@string/none"/>
 
+        <Preference android:key="pref_self_reporting"
+            android:title="Send statistics to Delta Chat's developers"/>
+
     </PreferenceCategory>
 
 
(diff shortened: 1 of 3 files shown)


res/xml/preferences_advanced.xml
```diff
@@ -93,6 +93,9 @@
             android:title="@string/videochat_instance"
             android:summary="@string/none"/>
 
+        <Preference android:key="pref_self_reporting"
+            android:title="Send statistics to Delta Chat's developers"/>
```
Hocuri, line 97: Note that this is not translated yet

src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
```diff
@@ -183,6 +186,22 @@ public void onCreate(Bundle paramBundle) {
       }
       return true;
     });
+
+    Preference selfReporting = this.findPreference("pref_self_reporting");
+    selfReporting.setOnPreferenceClickListener(((preference) -> {
```
adbenitez, line 191: so they have to manually send them, from time to time :thinking:

adbenitez (approved):

src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
Hocuri: Yes, I wanted to start with something simple. We can always make things more complicated, and since for now this will only be implemented in Android (and not the other UIs), we can easily change the core API.

r10s (approved):
good, smart thing to start with!

things as text vs. attachment etc. can be changed in core, discussions and some pros and cons are at https://github.com/deltachat/deltachat-core-rust/pull/5129

PR #2912, opened 2024-01-07
make 1to1-empty-hint working better with chatmail

r10s:
the hint already led to confusion on 37c3 testings of chatmail.

the hint that appears when a one-to-one-chat was just created manually (by entering an email address manually) is a bit misleading when using chatmail, as it says it 'does not matter if the other side uses Delta Chat'.

this is still true, however, with chatmail it is needed that messages outside the chatmail-server are encrypted,
which is arguably not so easy without delta chat and QR code scanning.

this PR therefore just removes the hint (alternative suggestions welcome);
it would still be great if the error message 'cannot encrypt' is improved, however, this is another thing.

before/after:

 

once merged, `./scripts/tx-update-changed-sources.sh` needs to be executed; ios/desktop will get the updated string automatically

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -385,7 +385,7 @@
         <item quantity="other">%d new messages</item>
     </plurals>
     <!-- The placeholder will be replaced by the name of the recipient in a one-to-one chat. -->
-    <string name="chat_new_one_to_one_hint">Send a message to %1$s; they can even receive it without Delta Chat.</string>
+    <string name="chat_new_one_to_one_hint">Send a message to %1$s.</string>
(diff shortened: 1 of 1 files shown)


adbenitez (approved):
a this point I wonder if isnt better to just don't display any hint, it doesn't say anything useful anymore

r10s:
> a this point I wonder if isnt better to just don't display any hint, it doesn't say anything useful anymore

i think, it is still better than showing an empty screen or the text "No Messages".

however, the hint fades out anyways. in many situations, the hint is not shown as the core adds an info message so that the chat is not empty.

PR #2916, opened 2024-01-07
do not show addr in guaranteed-e2ee chat titles

r10s:
with chatmail, the address is less meaningful,
it seems sufficient to show it in the profiles.

this moves focus to name and avatar and leave the email address as a "handle" that is not that meaningful and also subject to change, you check it on issues, similar to keys, similar to most other messengers.

this PR is not really a requirement for chatmail, however, tuning down the address is also reasonable by other means as the idea of changing addresses (AEAP) or even using more addresses per account

before / after:

Screenshot 2024-01-07 at 22 18 30 &nbsp; Screenshot 2024-01-07 at 22 21 02

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationTitleView.java
+++ b/src/org/thoughtcrime/securesms/ConversationTitleView.java
@@ -103,7 +103,7 @@ public class ConversationTitleView extends RelativeLayout {
       }
       else {
         DcContact dcContact = dcContext.getContact(chatContacts[0]);
-        if (profileView) {
+        if (profileView || !dcChat.isProtected()) {
           subtitleStr = dcContact.getAddr();
         }
         isOnline = dcContact.wasSeenRecently();
--- a/src/org/thoughtcrime/securesms/ConversationTitleView.java
+++ b/src/org/thoughtcrime/securesms/ConversationTitleView.java
@@ -103,12 +103,12 @@ public class ConversationTitleView extends RelativeLayout {
       }
       else {
         DcContact dcContact = dcContext.getContact(chatContacts[0]);
-        subtitleStr = dcContact.getAddr();
+        if (profileView) {
+          subtitleStr = dcContact.getAddr();
+        }
         isOnline = dcContact.wasSeenRecently();
       }
     }
-
-    subtitle.setText(subtitleStr);
 
     avatar.setAvatar(glideRequests, new Recipient(getContext(), dcChat), false);
     avatar.setSeenRecently(isOnline);


src/org/thoughtcrime/securesms/ConversationTitleView.java
```diff
@@ -103,13 +103,13 @@ else if( dcChat.isDeviceTalk() ) {
...
         DcContact dcContact = dcContext.getContact(chatContacts[0]);
-        subtitleStr = dcContact.getAddr();
+        if (profileView) {
+          subtitleStr = dcContact.getAddr();
+        }
         isOnline = dcContact.wasSeenRecently();
       }
     }
 
-    subtitle.setText(subtitleStr);
```
r10s, line 111: this was always doubled, the subtitle is set more accurate a few lines below.

adbenitez (approved):
this also makes it easier to scam people and impersonate, one idea would be to do it only for verified/protected-1:1 contacts 

another idea: if the address is not displayed anymore maybe we could put the last-seen info as subtitle

adbenitez:
another problem: if the user just sets a name, the avatar will be a letter of the address, not the name and it will be rather confusing to have something like:

(A) tim

webratte:
I absolutely agree with this.
I also use often the address to identify my buddy.
E.G. my wife uses two addresses. One time here chatmail address for chatting. And sometimes here normal mail address to forward a mail to my chatmail address.

hpk42:
@raiden-social Apart from current users, we consider future users as well -- and many of those, especially <30y aged ones, rarely if at all use e-mail conciously but regard it as a "thing of the past" that can not possibly bring any good, given Whatsapp, Signal, Matrix etc.  

Many Delta Chat folks obviously care about e-mail, its interoperability, stability, availability and diversity of the ecosystem.  Otherwise there wouldn't be a seven-year long project to build a top-notch messaging experience with it :)  Thanks for your support and contributions/feedbacks over the time, btw! 

I just checked Matrix btw: they only show the routing address in the profile but not in the "room title" but they also don't have a concept of 1:1 chats like WA/Signal do.  These technical details (matrix modeling 1:1 chats as group chats) don't really matter for most users, though (and arguably cause confusion there, a different topic). 

Please note that there is no suggestion of removing the e-mail address all together, just of relegating it to the profile.  What people care about is less the routing (email) address but rather that they talk/chat with the person they think they are chatting with.  Conversely, impersonation and fishing attacks need to be considered.  Here, e-mail addresses can serve as an additional clue for identifying, and giving away hints that the chat partner is maybe not who they claim to be.  But even with e-mail addresses, you can trick people, so it's by no means a fail-safe prevention mechanism to show e-mail addresses in many places.  And you can always go to the profile to investigate who a contact is/if you share chats with them etc. 

That all being said, there are probably three options now: 

1. Leave everything as is  

2. never show an email address in chat title (everything else stays the same, eg when choosing members for a group the email address is there with the member list) 

3. don't show email addresses for guaranteed/verified contacts in 1:1 chat titles -- chatmail users are generally lead to use QR/Invite codes which would then have the desired effect of de-emphasizing e-mail addresses  (again, not removing them from profile!) 

I don't think option 2 and 3 are in any way fatal but ultimately user-testing and feedback would be influential -- if we used either option 2 or 3 and new/current users actively dislike it/get overly confused, we'd revert.   Probably @kermoshina would also include it in upcoming user-testing sessions as a feedback item even before anything is released to the stores. While many of us have various opinions, we do take user-testing feedback pretty seriously and usually act on it quickly.

missytake:
One use case I have which I'd like to share here:

Element is currently the messenger where I most often introduce people to each other, because it is used in certain networks where people don't know each other. Very often, I send the address of someone to the person I want to introduce them to. (and it's usually very annoying because of Element's UI)

In Delta Chat this happens less often atm, but mostly because of the social graph of people using it right now.

hpk42:
On Tue, Jan 09, 2024 at 02:39 -0800, missytake wrote:
> One use case I have which I'd like to share here:
> 
> Element is currently the messenger where I most often introduce people to each other, because it is used in certain networks where people don't know each other. Very often, I send the address of someone to the person I want to introduce them to. (and it's usually very annoying because of Element's UI)
> 
> In Delta Chat this happens less often atm, but mostly because of the social graph of people using it right now.

Indeed, a use-case i can also personally relate to as i am doing introductions a lot. 

So how do i tell a contact in one chat about another contact? 
On Android you can do "attach contact" and select from a recently-seen-sorted contact list.
The display-name and e-mail address is then pasted into the input field. 
On Desktop this is much harder btw: you need to first find the "to-be-introduced" contact
somehwere, go to profile, copy-paste the e-mail address, find the chat back where
you want to paste the e-mail address etc. 

The "introduce X to Y" use case is IMO a very useful improvement area on all platforms. 
However, as such, it doesn't relate much to the "addr in chat title" issue at hand.

kermoshina:
I see that removing the email from the 1:1 chat name as a good privacy/security measure for our users in high risk areas, where the main threat models currently are not impersonalization or MiTM, but rathed device seizure or "over the shoulder attack" (sneakpeaking on someone s screen) or using surveillance cameras in public spaces to look at peoples screens (a real use case that I just heard from A., our contact from an antiwar helpline). 

I showed A. the two screenshots from @r10s example and she immediately said she finds it an excellent improvement, because of the threat model her and her partners have. She said it is good to just display the username, since Delta Chat has no central registry where a police person could search for the username (unlike telegram), and she confirmed that usually during a routine device control some screenshots or superficial examination are done, but if you are not under deep scrutiny, the adversary wont go deep into examining your device. It means that if a screenshot of a 1:1 conversation is taken with just a username, it will be harder to prove who is behind it, and also harder to further infiltrate a group. 

Another contact from a helpline, N, said that she also finds it a good feature but she asked to put the email info in the contact info screen (where currently we display common chats list). 

Other rapid feedback I got is also positive about removing email info from the chat name. 

I will definitely include this in user testing to further explore what people think. Now it is just based on a very rapid inquiry.

Simon-Laux:
Why not make this a setting, advanced users like me that use DC also as email client want to keep this. Email address is unique while display name is not.
I get the idea of a "streamer mode" that hides addresses, why not introduce sth like that, but hiding all email addresses in the app until you turn it off again?

r10s:
thanks all for the careful considerations!

i decided for the compromise of keeping the address in chat titles of classic contacts and changed the PR accordingly.

this affects chatmail contacts and delta-to-delta contacts mainly, for classic email usage, things stay the same mostly.

to be clear: the email address will stay in profile (one tap away) and there are no ideas to change that.

to sum up reasoning and discussion:
- protect against "over the shoulder surveillance", a real-world issue reported
- the address is not meaningful for chatmail. featuring chatmail is importing for future users, esp. <30yo ones considering email a "thing of the past that can't possible bring any good" :)
- act similar to other messengers, not showing the "handle" for known contacts
- focus are shift to name and avatar, not clutterung UI, making guaranteed e2ee more outstanding
- impersonation won't become worse, you could always use trustworthy looking but wrong addresses. removing the address may even improve here as other aspects (chat history etc.) may get more into focus and you need to establish guaranteed-e2ee to impersonate now
- prepare for AEAP and the idea of multilple addresses per account

hpk42:
On Tue, Jan 09, 2024 at 14:48 -0800, Simon Laux wrote:
> Why not make this a setting, advanced users like me that use DC also as email client want to keep this. Email address is unique while display name is not.

Which duplicate display names do you have in real-life? 
I think in several years i had this once or twice and then
i just set a custom name -- after all this display name also 
appears in the chatlist without any e-mail address anyway. 
Let's go with Bjoern's decision (no e-mail addr for title in guaranteed chat)
for now and call it a day. 
If it proves to be a source for common complaint we can still act. 
There are other much more interesting things (reactions, multi-account etc.) 
to go for that will interest / make a difference for far more users.

missytake:
> Which duplicate display names do you have in real-life? I think in several years i had this once or twice and then i just set a custom name -- after all this display name also appears in the chatlist without any e-mail address anyway.

Just for the record - I have this *a lot* in Signal. In one group I have *4 name conflicts*, Signal shows a warning if this appears in a group. Many people I know just use one-letter display names, or even emojis (super annoying because not searchable), for multi-identity purposes (being known with different names to different people). In Signal it makes sense because you can't easily have multi-account.

But Delta Chat does this already **a lot** better than Signal, because we can set custom display names for contacts. Sure, some people might not know that this is possible in Delta Chat; anyway I don't think we should act on it by showing the email address, I just want to show that it's a common use case.

Simon-Laux:
> Personally, I think the security aspect is greatly exaggerated.

over the shoulder attack and photographs of the phone is a real thing that can happen.

> simply not letting anyone look over your shoulder.

Not easy if you are inside of a demonstration with hundreds of people.

> I also don't think removing the email address from the title bar will lead to more young users.

I also do not think that.

adbenitez:
a practical example right now in Delta Chat are bots, instances of the same bot often will have the same name, like my two "www" bots now it is harder to know which one you are using, you still can see it in profile tho

r10s:
just a general reasoning to understand why we aim to avoid options:

> too many options makes apps worse for everyone due to overwhelming them with choices and hiding important options in a sea of non-important options.
Not to mention that every conditional behavior requires more code, more potential bugs and more testing.

... for sure only this option won't be the end of the world - but we have such discussions often. and if we would have gone the easy way of adding an optuon, we would easily have dozends more

... so, we add options only if there is really a big impact for quite some ppl. i do not see it in this case.

for the www bots example: at least the www bits i am using are not guaranteed-e2ee and therefore the adresses are shown

but even if otherwise, i doubt there are many ppl apart from devs using different instances of the same bot on the same account; and if so, tapping  profile  is a big deal and justifies an option

let's maybe reconsider when we tested for some time in the wild; i have seend this pattern of pushback on changes several times where at the end, things are just fine and even better :)

PR #2918, opened 2024-01-10
add 'Scan QR Code' to 'New Chat'

### LINKED ISSUE #2911: allow QR code scanning from "New Chat" dialog
the "New Chat" dialog is the main way we're offering to create new chats.

with chatmail becoming more into focus, scanning a QR code becomes more important and should be available from "New Chat" as well, cmp. https://github.com/deltachat/deltachat-desktop/issues/3623

r10s:
this PR add a 'Scan QR Code' button to 'New Chat' activity.

this makes setting up contacts easier when the best option is scanning a QR code (as for chatmail).

closes #2911 

EDIT: i think, best would be to combine "New Contact" and "QR Invite Code" to one item that offers QR code scanning by default and "Share Invite Link" an "Manual Setup" as option. we can also think over renaming the options. but that is another PR/discussion, the PR at hand already improves things

Screenshot 2024-01-12 at 13 22 44

### DIFF
--- a/src/org/thoughtcrime/securesms/connect/DcContactsLoader.java
+++ b/src/org/thoughtcrime/securesms/connect/DcContactsLoader.java
@@ -16,14 +16,16 @@ public class DcContactsLoader extends AsyncLoader<DcContactsLoader.Ret> {
 
     private final int     listflags;
     private final String  query;
+    private final boolean addScanQRLink;
     private final boolean addCreateGroupLinks;
     private final boolean addCreateContactLink;
     private final boolean blockedContacts;
 
-    public DcContactsLoader(Context context, int listflags, String query, boolean addCreateGroupLinks, boolean addCreateContactLink, boolean blockedContacts) {
+    public DcContactsLoader(Context context, int listflags, String query, boolean addCreateGroupLinks, boolean addCreateContactLink, boolean addScanQRLink, boolean blockedContacts) {
         super(context);
         this.listflags           = listflags;
         this.query               = (query==null||query.isEmpty())? null : query;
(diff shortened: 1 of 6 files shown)


src/org/thoughtcrime/securesms/connect/DcContactsLoader.java
```diff
@@ -46,6 +46,9 @@ DcContactsLoader.Ret loadInBackground() {
             final boolean broadcastsEnabled = Prefs.isNewBroadcastListAvailable(getContext());
             if (broadcastsEnabled) additional_items = Util.appendInt(additional_items, DcContact.DC_CONTACT_ID_NEW_BROADCAST_LIST);
         }
+        if (query == null && addCreateContactLink) {
```
adbenitez, line 49: I don't remember now if this contact list is reused in the "add member" option in group profile, but if yes then showing a contact-QR button there to add a contact doesn't make much sense maybe the group invitation QR button in that case

adbenitez (approved):

src/org/thoughtcrime/securesms/connect/DcContactsLoader.java
r10s: you're right, i will adapt the pr

r10s:
@adbenitez i also changed the "search" placeholder and moved the "QR" option to the very top

adbenitez (approved):

adbenitez:
something blocking merging?

r10s:
> something blocking merging?

there is the idea of further improvements, so combining the two "add contacts" into one, but this can also come later and is not clear yet. merging :)

PR #2922, opened 2024-01-16
register to handle i.delta.chat links

### LINKED ISSUE #2920: allow to open i.delta.chat links in-app
this makes it possible to easily launch Delta Chat from other apps or messengers ex. when sharing an invitation link to a group or contact-verification on a 3rd party channel, while sharing QR image or QR data is more cumbersome/counter-intuitive

this becomes more important when using chatmail instances, since QR scanning is mandatory to start conversation with the outside

adbenitez:
close #2920

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -138,6 +138,14 @@
                 tools:ignore="AppLinkUrlError" />
         </intent-filter>
 
+        <intent-filter android:autoVerify="true">
+            <action android:name="android.intent.action.VIEW" />
+            <category android:name="android.intent.category.DEFAULT" />
+            <category android:name="android.intent.category.BROWSABLE" />
+            <data android:scheme="https"
+                  android:host="i.delta.chat" />
+        </intent-filter>
+
         <meta-data android:name="com.sec.minimode.icon.portrait.normal"
                    android:resource="@mipmap/ic_launcher" />
         <meta-data android:name="com.sec.minimode.icon.landscape.normal"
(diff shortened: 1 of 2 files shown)


adbenitez:
in one Xiaomi phone it doesn't open in DC but in the browser, in another phone (CalyxOS) it also doesn't open in DC by default but going to system settings of DC I can see a "open supported links" option and there you have to manually enable to open i.delta.chat directly in DC, 

it says something about "zero verified links" and an info at the bottom points "An app can verify links to automatically open in the app" so it seems that to open links automatically the i.delta.chat needs to be tweaked in some way to report that DC app is allowed to handle their links instead of opening in browser, makes some sense, ex. then no fake app can open WhatsApp, or Signal links etc. but it makes all more complicated and less useful

adbenitez:
deltachat-2024-01-16-172510

adbenitez:
the link at the bottom is: https://developer.android.com/training/app-links/verify-site-associations

hpk42:
could you provide the contents of the "assetlinks.json" file? 
we can put it to i.delta.chat and then it should work, i guess? 

https://developer.android.com/training/app-links/verify-android-applinks#web-assoc

r10s (approved):
code lgtm, did not test.

for testing, it seems esp. important that existing OPENPGP4FPR links continue working (which one can easily test by open the i.delta.chat in the browser manually, then the page shows the OPENPGP4FPR links)

PR #2923, opened 2024-01-16
don't load contacts from phone address book

### LINKED ISSUE #2915: offering email-addresses from the system's address book is not helpful for chatmail
currently, the app adds all email-addresses from the system's address book to the "New Chat" dialog.

in case of chatmail, that requires encryption for foreign addresses, however, most of these addresses cannot be written to easily.

offering these addresses on case of chatmail even harms UX as it raises false expectations.

we should somehow not add these addresses. question is if we stop adding them in general or only for chatmail (however, currenlty, we do not know if self-address is chatmail or not) and/or offer a way to add them manually. are these addresses are really a thing? eg. desktop never scanned the address book, arguably, there is no common address book on desktop as for android/ios

in any way, once we found a solution here, we should do the same on ios.

adbenitez:
close #2915 

this removes the need for a "dangerous" permission that doesn't brings much value, more with chatmail where it is not possible to write unencrypted, encouraging users more in the direction of contact verification, also loading from contact address book was always a bit problematic and cause of some bugs/issues, ex. updating the contact in the phone contacts will not update it in Delta Chat until the (+) button is hit and then contacts are reloaded, 

also in multi-account this always caused me troubles, it would import all the contacts in all accounts, and for some accounts like chatmail or yggmail this didn't make any sense and cluttered the contact list

notice that, if wanted, user can still just go to the phone contacts and search contacts there and start a DC chat from there

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -13,7 +13,6 @@
     <uses-feature android:name="android.hardware.touchscreen" android:required="false"/>
 
     <!-- dangerous permissions - we need to as the user with a PermissionsRequest -->
-    <uses-permission android:name="android.permission.READ_CONTACTS" />
     <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
     <uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" android:maxSdkVersion="32" />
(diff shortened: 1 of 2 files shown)


r10s:
i think, we'll need the information if sth. is a chatmail or not anyways, at least for iOS-push. 

if we have that, maybe we can make adding address book adresses conditionally (so, no option, just added for non-chatmail)

adbenitez:
closed in favor of #2950

PR #2931, opened 2024-01-20
display last-seen information if address will not be shown

adbenitez:
with #2916 ( do not show addr in guaranteed-e2ee chat titles) now there are free space for more useful information, also without address it looks more ugly, specially for existing users that are used to the old look, this is also what is displayed in apps like Telegram

For bots I would like to show just "bot" there, there is already core API for that but just not exposed there, another situation is when user is "online" and the green dot is displayed, for that case maybe it is better to display "online" label or "last seen recently", but that needs a new string

this feature will leverage the push-back of users since not showing the address is removing a feature, while not displaying it  to display another useful information is a feature ;) in fact this makes one reconsider if always displaying this info instead the address for ALL contacts is not better than just doing so for verified contacts

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationTitleView.java
+++ b/src/org/thoughtcrime/securesms/ConversationTitleView.java
@@ -19,7 +19,11 @@ import org.thoughtcrime.securesms.components.AvatarView;
 import org.thoughtcrime.securesms.connect.DcHelper;
 import org.thoughtcrime.securesms.mms.GlideRequests;
 import org.thoughtcrime.securesms.recipients.Recipient;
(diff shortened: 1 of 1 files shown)


adbenitez:
(closing, was told it is not wanted for now in some DC chats)

r10s:
well, it is pretty inaccurate an will raise questions with that and probably also wrt privacy.

the date is an important information, but not that _outstanding_. we lived years without without adding it to a so exposed position, that gives the date too much weight.

there is already the online indicator and the profile. no need to clutter the now clean ui 

space and clean ui has its benefits as well, the challange is not to fill up all lines: the more information you show, the less important a single information will be

PR #2932, opened 2024-01-20
remove old hack to update seen-recently indicator

adbenitez:
this was a hack I added to try to compensate the lack of event for "contact is not longer seen-recently" and to update the seen-recently indicator (green dot) after that core was improved to emit contact-changed event to properly update the seen-recently status


r10s (approved):
nice! thanks for taking care! 

i assume, DC_EVENT_INCOMING_MSG and DC_EVENT_MSG_READ are still needed as otherwise you would have touched that as well :)

adbenitez:
> i assume, DC_EVENT_INCOMING_MSG and DC_EVENT_MSG_READ are still needed as otherwise you would have touched that as well :)

IIRC they are needed as there is no other event for when contact changes from not seen-recently to seen-recently, since you can already watch for the incoming message event etc. but not 100% sure, cc @link2xt

PR #2934, opened 2024-01-20
tweak reactions style

### LINKED ISSUE #2926: consistent reaction colors
desktop and iOS also show reactions meanwhile, we picked up the initial idea of reusing exiting bubble colors for self-reactions - however, even though the idea came up during android development, it is not used on android :)

so, instead of random grey tones for the reactions, use the same color as "self" and "other" bubbles:
- greenish in light mode for self-reactions
- white for others in light mode.
- in dark mode, use the existing grey tone

here are some screenshots of iOS:

https://github.com/deltachat/deltachat-ios/pull/2043

this requires a slim border around the bubbles, we can just reuse what iOS is doing here as well

adbenitez:
close #2926

### DIFF
--- a/res/values-night/colors.xml
+++ b/res/values-night/colors.xml
@@ -3,7 +3,7 @@
     <color name="reaction_pill_background">@color/black</color>
     <color name="reaction_pill_border">#80414347</color>
     <color name="reaction_pill_text_color">@color/core_dark_05</color>
-    <color name="reaction_pill_background_selected">#ff282828</color>
+    <color name="reaction_pill_background_selected">@color/gray70</color>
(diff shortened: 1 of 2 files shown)


r10s:
nice!

for darkmode's outgoing reaction, the color needs to be tweaked as as well (so the dark needs to be a little lighter), this makes also the difference a bit easier to recognize:

Screenshot 2024-01-20 at 11 45 19

adbenitez:
I modified self-reaction background in dark mode to be the same color as self-messages bubbles, same as for the light theme

r10s (approved):
looks great now! 👌

Issue #2914, opened 2024-01-07
focus on setting up contacts in initial device message


r10s:
the "device messages" chat is one of the few things that are present after "instant onboarding" (eg. via QR code scan on https://nine.testrun.org, maybe later in-app)

currently, the "welcome" messages shows some general hints, idea is to write down the most interesting thing for new users, how to get in contact with others (that are maybe sitting just next to me) - so, eg. 
- an image of two phones scanning each others QR code (needs to be changed in core iirc)
- a describing text about how to scan QR codes (cmp. [desktop/#3623](https://github.com/deltachat/deltachat-desktop/issues/3623) wrt streamlining) (can be changed on android)
- maybe show own QR code in device messages and/or allow opening the corresponding dialog via a tap/link
- to reduce noise in device chat, it is also questionable to show the "Changelog" also on initial installations (on real updates, however, it is cool)

desktop already opens the "device messages" after installations, so that should be mostly fine.
for android/ios, there is the additional idea to use space on the chatlist for additional hints, but i'd say, this is another issue, and it also depends on how good the updated device message already works (would be nice, if eg. already in the chatlist the summary "How to get in contact ... " or so is shown) 

(once this is settled here and in core, we have to check if desktop/ios needs adaptions or if just strings/core is updated)

current, a bit outdated device message:

<img width=320 src=https://github.com/deltachat/deltachat-android/assets/9800740/ddca23ae-abf0-49d7-a72c-bacd09abe69f>

nb: maybe the name "Device Messages" is also a bit cumbersome and buerocratic and sth. as "Hints" or "Notes" or just "Info" is better

adbenitez:
> * to reduce noise in device chat, it is also questionable to show the "Changelog" also on initial installations (on real updates, however, it is cool)

+100 it is rather confusing for first installations to get some message about "what is new" and it clutters the device chat, also a message about enabling disabling background optimizations will appear so the chat so too much is going on for an user that just logged in

adbenitez:
> and sth. as "Hints" or "Notes" or just "Info" is better

If it needs to be changed, from that options Info is my favorite and it fits better with the (i) icon that chat has

PR #2935, opened 2024-01-20
do not add "Changelog" for new accounts

r10s:
on new installations or accounts,
the "What's new?" messages only clutter the device messages with unimportant information (either the user is new to the app, so everyrhing is new and nothing is changed, or an existing user has created a new account, they have seen the changelog anyways)

in other words,
the changelog will be added only for updates.

when reading https://c.delta.chat/classdc__context__t.html#a1a2aad98bd23c1d21ee42374e241f389 , this was always the gist,
however, somehow forgotten or removed on the way :)

targets #2914

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -82,6 +82,7 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
   private static final String NDK_ARCH_WARNED = "ndk_arch_warned";
   public static final String CLEAR_NOTIFICATIONS = "clear_notifications";
   public static final String ACCOUNT_ID_EXTRA = "account_id";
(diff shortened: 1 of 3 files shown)


Simon-Laux:
I don't really understand the android code here, wouldn't this just be added when you restart the app? so not only on update?

r10s:
> I don't really understand the android code here, wouldn't this just be added when you restart the app? so not only on update?

no. see documentation for [`dc_add_device_msg()`](https://c.delta.chat/classdc__context__t.html#a1a2aad98bd23c1d21ee42374e241f389):  if `msg == NULL`, no message is added, but `label` is still marked as being added.

in case of the android code, `dcContext.addDeviceMsg(deviceMsgId, msg);` is called with `msg = null` if `FROM_WELCOME` is set.

on subsequent starts, when `FROM_WELCOME` is unset, `wasDeviceMsgEverAdded()` returns `true` and no device message is added

adbenitez (approved):

PR #2937, opened 2024-01-22
add string 'React'

r10s:
needed for desktop's context menu


adzialocha (approved):
Thank you!

r10s:
np, i also pushed the string to transifex now and did the german translation

PR #2938, opened 2024-01-22
update welcome message

### LINKED ISSUE #2914: focus on setting up contacts in initial device message

r10s:
this PR 

1. updates the "welcome" message, making it more useful and explaining how to get in contact.

2. the "Get in contact" summary is already visible in the chat list, before the device message is opened

3. the idea is to replace the image as well, showing two devices scanning each other, sth as ... <br><br> ... but that needs to be done in core

for now, some of the old "advertisements" (advantages of delta chat) in the second part.

 

<details>
<summary>experiment: title changed to "Info"</summary>
we can still do this, however, better in another PR; also this may require more changes in help etc., so might be not that easy as just one line change :) (as always :)

 
</details>

closes #2914


adbenitez (approved):
much more useful message, maybe also point to the get.delta.chat page like the older message, "if your friends don't have delta chat or an email account yet, point them to ..."

adbenitez:
about renaming from device messages to info, I am +/-0, maybe a bit + than -, but I guess this also requires changes FAQ/help and shouldn't it be named by core?

Simon-Laux:
We probably need some special welcome message for chatmail explaining the limitations. or at least a special message / error when an unencrypted mail is rejected by a chatmail server.

r10s:
> or at least a special message / error when an unencrypted mail is rejected by a chatmail server.

see https://github.com/deltachat/deltachat-core-rust/issues/5161 

> overlooked it [renaming "Device Messages" to "Info"] is just an experiment :)

i reverted the change, better to discuss that separately

adzialocha (changes requested):
Yes, much better!

Small english grammar finding, shouldn't it be?

> Enjoy your messenger experience over the largest decentralized network **that** ever existed:

Otherwise all good

PR #2940, opened 2024-01-24
improve status message for "Invalid Unencrypted"

r10s:
This PR refines the "Invalid Unencrypted" info messages that is added since  https://github.com/deltachat/deltachat-core-rust/pull/5198 when a message cannot be encrypted _and_ the provider does not allow unencrypted messages; this is true eg. for chatmail

In addition to stock core message:

- a "Tap to learn more" is added
- a tap shows a little dialog, allowing to perform a QR code scan or get more help about the scanning process (this seems mostly on-point to solve the user's problem at hand, and probably mostly good enough for now, we can always refine as needed)

@adbenitez this PR also removes linkifying info-messages in general - do we know of any info-messages that contain links _on purpose_? if so, we can make the linkify-removal conditional.

Screenshot 2024-01-25 at 00 39 03 &nbsp; Screenshot 2024-01-25 at 00 39 09


adbenitez (approved):
I think disabling linkifying is fine, the only links I remember are from Yandex on errors but that is not displayed in a system message but as in-message error visible in the dialog on tap or "message info"

Simon-Laux (approved):

PR #2942, opened 2024-01-28
add global unread messages badge counter

### LINKED ISSUE #2921: global badge counter in self-avatar
a badge counter in self-avatar in the main screen (chatlist) to know how many unread messages there are in other accounts

one could argue that it is not a vital feature since there is already multi-account notifications, but with the tentative of allowing to disable notifications for background accounts, a badge counter would be a less disruptive way to notice new messages

adbenitez:
close #2921

### DIFF
--- a/src/org/thoughtcrime/securesms/connect/DcEventCenter.java
+++ b/src/org/thoughtcrime/securesms/connect/DcEventCenter.java
@@ -17,10 +17,9 @@ import java.util.ArrayList;
 import java.util.Hashtable;
 
 public class DcEventCenter {
-    private @NonNull final Hashtable<Integer, ArrayList<DcEventDelegate>> allObservers = new Hashtable<>();
+    private @NonNull final Hashtable<Integer, ArrayList<DcEventDelegate>> currentObservers = new Hashtable<>();
     private @NonNull final Hashtable<Integer, ArrayList<DcEventDelegate>> multiObservers = new Hashtable<>();
     private final Object LOCK = new Object();
-    private final Object MULTI_LOCK = new Object();
     private final @NonNull ApplicationContext context;
 
     public interface DcEventDelegate {
@@ -35,20 +34,18 @@ public class DcEventCenter {
     }
 
     public void addObserver(int eventId, @NonNull DcEventDelegate observer) {
-        synchronized (LOCK) {
-            ArrayList<DcEventDelegate> idObservers = allObservers.get(eventId);
-            if (idObservers == null) {
-                allObservers.put(eventId, (idObservers = new ArrayList<>()));
-            }
-            idObservers.add(observer);
-        }
+        addObserver(currentObservers, eventId, observer);
     }
 
     public void addMultiAccountObserver(int eventId, @NonNull DcEventDelegate observer) {
-        synchronized (MULTI_LOCK) {
-            ArrayList<DcEventDelegate> idObservers = multiObservers.get(eventId);
+        addObserver(multiObservers, eventId, observer);
+    }
+
+    private void addObserver(Hashtable<Integer, ArrayList<DcEventDelegate>> observers, int eventId, @NonNull DcEventDelegate observer) {
+        synchronized (LOCK) {
+            ArrayList<DcEventDelegate> idObservers = observers.get(eventId);
             if (idObservers == null) {
-                multiObservers.put(eventId, (idObservers = new ArrayList<>()));
+                observers.put(eventId, (idObservers = new ArrayList<>()));
             }
             idObservers.add(observer);
         }
@@ -56,7 +53,7 @@ public class DcEventCenter {
 
     public void removeObserver(int eventId, DcEventDelegate observer) {
         synchronized (LOCK) {
-            ArrayList<DcEventDelegate> idObservers = allObservers.get(eventId);
+            ArrayList<DcEventDelegate> idObservers = currentObservers.get(eventId);
             if (idObservers != null) {
                 idObservers.remove(observer);
             }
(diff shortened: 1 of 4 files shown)


src/org/thoughtcrime/securesms/connect/DcEventCenter.java
```diff
@@ -18,7 +18,9 @@
 
 public class DcEventCenter {
     private @NonNull final Hashtable<Integer, ArrayList<DcEventDelegate>> allObservers = new Hashtable<>();
```
r10s, line 20: `allObservers` is misleading now, this should be `currentAccountObservers` now

src/org/thoughtcrime/securesms/connect/DcEventCenter.java
```diff
@@ -18,7 +18,9 @@
...
+    private @NonNull final Hashtable<Integer, ArrayList<DcEventDelegate>> multiObservers = new Hashtable<>();
     private final Object LOCK = new Object();
+    private final Object MULTI_LOCK = new Object();
```
r10s, line 23: i'd use the same `LOCK`, see below for `removeObservers()`

src/org/thoughtcrime/securesms/connect/DcEventCenter.java
```diff
@@ -42,6 +44,16 @@ public void addObserver(int eventId, @NonNull DcEventDelegate observer) {
...
+        synchronized (MULTI_LOCK) {
+            ArrayList<DcEventDelegate> idObservers = multiObservers.get(eventId);
+            if (idObservers == null) {
+                multiObservers.put(eventId, (idObservers = new ArrayList<>()));
+            }
+            idObservers.add(observer);
+        }
+    }
+
     public void removeObserver(int eventId, DcEventDelegate observer) {
```
r10s, line 57: it is error prone to track exactly if an observer is a multiaccount observer or not and call the correct remover. it would be better to let `removeObserver()` remove both types. otherwise, i already see bugs arising when one changes from single-observer to multi-observer for reasons - but does not change the remover-function.

this is esp. bad as it is not directly visible - but will result in slowdown over time (we had these hard-to-get-errors on iOS at some point)


same for `removeObservers()`, which is a convenience tool to remove just all observers, therefore, it should remove the multiAccountObservers as well.

two function less - and the overhead of one lookup in a most times empty hash is negligible

src/org/thoughtcrime/securesms/connect/DcEventCenter.java
```diff
@@ -62,33 +74,54 @@ public void removeObservers(DcEventDelegate observer) {
...
+            for(Integer eventId : multiObservers.keySet()) {
+                ArrayList<DcEventDelegate> idObservers = multiObservers.get(eventId);
+                if (idObservers != null) {
+                    idObservers.remove(observer);
+                }
+            }
+        }
+    }
+
+    public void sendToMultiAccountObservers(@NonNull DcEvent event) {
```
r10s, line 88:
```suggestion
    private void sendToMultiAccountObservers(@NonNull DcEvent event) {
```

src/org/thoughtcrime/securesms/connect/DcEventCenter.java
```diff
@@ -62,33 +74,54 @@ public void removeObservers(DcEventDelegate observer) {
...
+        synchronized (MULTI_LOCK) {
+            sendToObservers(event, multiObservers);
+        }
+    }
+
     public void sendToObservers(@NonNull DcEvent event) {
```
r10s, line 94:
```suggestion
    private void sendToCurrentAccountObservers(@NonNull DcEvent event) {
```

src/org/thoughtcrime/securesms/connect/DcEventCenter.java
```diff
@@ -18,7 +18,9 @@
 
 public class DcEventCenter {
     private @NonNull final Hashtable<Integer, ArrayList<DcEventDelegate>> allObservers = new Hashtable<>();
+    private @NonNull final Hashtable<Integer, ArrayList<DcEventDelegate>> multiObservers = new Hashtable<>();
```
r10s, line 21:
```suggestion
    private @NonNull final Hashtable<Integer, ArrayList<DcEventDelegate>> multiAccountObservers = new Hashtable<>();
```

r10s (commented):
if we got for the idea with the "multiAccountObserver" (which is a reasonable, more powerful alternative to [special events as on iOS](https://github.com/deltachat/deltachat-android/issues/2921#issuecomment-1912495302)), we should make the API a bit more dev-friendly and avoid future leaks and issues, see comments below

src/org/thoughtcrime/securesms/connect/DcEventCenter.java
adbenitez: yes I agree, didn't change it initially afraid of doing too much unrelated changes that might make reviewing harder, but now that I see you agree will do it :+1:

src/org/thoughtcrime/securesms/connect/DcEventCenter.java
adbenitez: indeed, in practice you rather always stop listening to all events at once

src/org/thoughtcrime/securesms/connect/DcEventCenter.java
adbenitez: sendToObservers() is also public, but in fact not used anywhere outside of DcEventCenter, maybe it is used for some tests :thinking:

r10s (approved):
nice! code lgtm, and also tested already a bit!

it is a huge improvement, making multi-account much more usable

PR #2943, opened 2024-01-29
copy invite links to clipboard

### LINKED ISSUE #2939: allow creating invite-code in-app
currently, invite-links for https://i.delta.chat need a copy+paste process.

the idea is to simplify that, maybe by offering a "share to" directly, but at least by offering copying the correct URL to the clipboard.

#2938 already mentions this option

once this is settled, we should create corresponding issues for desktop/ios

adbenitez:
close #2939

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -78,7 +78,6 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
 {
   private static final String TAG = ConversationListActivity.class.getSimpleName();
   private static final String OPENPGP4FPR = "openpgp4fpr";
-  private static final String INVITE_PAGE = "i.delta.chat";
   private static final String NDK_ARCH_WARNED = "ndk_arch_warned";
   public static final String CLEAR_NOTIFICATIONS = "clear_notifications";
   public static final String ACCOUNT_ID_EXTRA = "account_id";
@@ -375,14 +374,9 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
         return;
       }
 
-      if (uri.getScheme().equalsIgnoreCase(OPENPGP4FPR)) {
-        String uriString = uri.toString().replaceFirst(OPENPGP4FPR, OPENPGP4FPR.toUpperCase());
+      if (uri.getScheme().equalsIgnoreCase(OPENPGP4FPR) || Util.isInviteURL(uri)) {
         QrCodeHandler qrCodeHandler = new QrCodeHandler(this);
-        qrCodeHandler.handleQrData(uriString);
-      } else if (INVITE_PAGE.equals(uri.getHost()) && uri.getEncodedFragment() != null) {
-        String data = OPENPGP4FPR.toUpperCase() + ":" + uri.getEncodedFragment().replaceFirst("&", "#");
-        QrCodeHandler qrCodeHandler = new QrCodeHandler(this);
-        qrCodeHandler.handleQrData(data);
+        qrCodeHandler.handleQrData(uri.toString());
       }
     }
   }
--- a/src/org/thoughtcrime/securesms/qr/QrShowFragment.java
+++ b/src/org/thoughtcrime/securesms/qr/QrShowFragment.java
@@ -130,6 +130,18 @@ public class QrShowFragment extends Fragment implements DcEventCenter.DcEventDel
         }
     }
 
+    public void shareInviteURL() {
+        try {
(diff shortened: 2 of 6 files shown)


src/org/thoughtcrime/securesms/ConversationListActivity.java
```diff
@@ -375,14 +374,9 @@ private void handleOpenpgp4fpr() {
...
 
-      if (uri.getScheme().equalsIgnoreCase(OPENPGP4FPR)) {
-        String uriString = uri.toString().replaceFirst(OPENPGP4FPR, OPENPGP4FPR.toUpperCase());
+      if (uri.getScheme().equalsIgnoreCase(OPENPGP4FPR) || Util.isInviteURL(uri)) {
         QrCodeHandler qrCodeHandler = new QrCodeHandler(this);
-        qrCodeHandler.handleQrData(uriString);
-      } else if (INVITE_PAGE.equals(uri.getHost()) && uri.getEncodedFragment() != null) {
-        String data = OPENPGP4FPR.toUpperCase() + ":" + uri.getEncodedFragment().replaceFirst("&", "#");
-        QrCodeHandler qrCodeHandler = new QrCodeHandler(this);
-        qrCodeHandler.handleQrData(data);
```
adbenitez, line 385: no need to parse invite links and convert them to QR data, core now supports handling links directly: https://github.com/deltachat/deltachat-core-rust/pull/5219

(needs a core update in the sub-module)

src/org/thoughtcrime/securesms/qr/QrShowFragment.java
```diff
@@ -129,7 +129,8 @@ public void shareQr() {
     }
 
     public void copyQrData() {
```
adbenitez, line 131: besides self contact-qr, this is also called for groups's invite-qr

src/org/thoughtcrime/securesms/ConversationListActivity.java
r10s: i just updated core to recent `main` (as there is no release yet)

r10s (approved):
lgtm, i think, the whole stack needs quite some testinga as lots of different things are involved :)

r10s:
wait, currently the "Share Image" is still more in focus that the now more useful invite links.

i suggest at least to move the "Share Image" and "Load Image" options to the end of the overflow menu (so that the invite links options are the first options in the overflow menu) - can be done in a subsequent PR, however

src/org/thoughtcrime/securesms/util/Util.java
```diff
@@ -56,13 +57,22 @@
 
 public class Util {
   private static final String TAG = Util.class.getSimpleName();
+  private static final String INVITE_PAGE = "i.delta.chat";
```
r10s, line 60: maybe INVITE_DOMAIN fits better here

adbenitez:
I will just add the link to the share option so it is like on ios and shares the link but also + the image like in the past, I don't think the QR is now useless, in fact in some groups like on telegram, messages with links are often banned due to spam etc. so sharing as image is useful in that case, also the qr is nicer even if just as a banner for the link ;) which in theory will open directly in delta chat if the user have it installed without even opening the i.invite.delta page

r10s:
i am not sure about sending the QR-code image together with the invite-link. maybe yes, maybe no :)

for the “nice banner view”, this is true also without the QR-code image  as most apps (but Delta Chat :)  render a nice link preview, here’s eg. how the invite link is rendered by eg. Signal or WhatsApp:

   

(the subtitle could be more useful and less like an advertisement, but this is another discussion)

how does qr-code-image + link preview look like? even if nice, it may be confusing:

adding an image to the message, saying “Scan this QR code” does not really help in many cases, eg. when Delta Chat is on the same device, tapping the links just works, scanning the code may be impossible or adds further hurdles.
also, things are less clear what is needed when forwarding things etc. are both, QR code and link needed? or will only one do? if ppl forward only the image, this is usually less helpful.

it makes things esp, harder for new users that  have not installed Delta Chat yet.

wrt consistency: we have the “share QR code image” functionality only on android, not on desktop nor on iOS - this may also be confusion

wondering, what the precise usecase of the additional QR code could be, and if that justifies the drawbacks. if it is about spam, a dedicated "Share QR code image" may be better

adbenitez:
I modified the share option to share invite URL like iOS etc

r10s (approved):
lgtm, i also tested a little.

-  i added a commit that reorders the "Scan QR" menu (as primary UI works with invite links now, 'Load QR code from image' is still supported but now longer  recommended workflow and can be tuned down)

PR #2947, opened 2024-01-31
declare hasFragileUserData

### LINKED ISSUE #2941: Consider setting `hasFragileUserData`
I have noticed that some app that I recently removed (IIRC Element) offered to keep the data when I was uninstalling it.
I did it on Android 10.

Seems it is possible to trigger this with a manifest setting https://developer.android.com/reference/android/R.attr#hasFragileUserData

Should we enable it as well? This may prevent some users accidentally removing their data when they just want to switch versions or something like this and assume data is not removed. For example here is a thread from someone trying to downgrade and removing the data: https://support.delta.chat/t/how-to-recover-all-chats-after-dc-reinstallation/2686

adbenitez:
close #2941

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -50,6 +50,8 @@
                  android:largeHeap="true"
                  tools:ignore="GoogleAppIndexingWarning"
                  android:requestLegacyExternalStorage="true"
+                 tools:targetApi="TIRAMISU"
+                 android:hasFragileUserData="true"
         >
 
     <!-- android car support, see https://developer.android.com/training/auto/start/,


AndroidManifest.xml
```diff
@@ -50,6 +50,8 @@
                  android:largeHeap="true"
                  tools:ignore="GoogleAppIndexingWarning"
                  android:requestLegacyExternalStorage="true"
+                 tools:targetApi="TIRAMISU"
```
r10s, line 53: just for interest, what is the ` tools:targetApi="TIRAMISU"` needed for now?

r10s (approved):
code lgtm, and probably also makes sense if newer android offer this option.

i do not expect OS to make app-deletion much harder so that it becomes an issue for at-risk ppl, probably just a checkbox that may or may not be ticked.

AndroidManifest.xml
adbenitez: to hint the Android Studio linter, because hasFragileUserData was added on android 10 and we have minSDK 16 so the property has no effect in <10

PR #2948, opened 2024-01-31
allow to mute/unmute from chatlist

adbenitez:
this allows to mute a noisy chat you can't attend to at the moment without entering it and marking as read messages you don't intend to read at the moment and sending read receipts to the peer

also allows to mute/unmute chats in batches without having to do it one by one

### DIFF
--- a/src/org/thoughtcrime/securesms/ConversationListFragment.java
+++ b/src/org/thoughtcrime/securesms/ConversationListFragment.java
@@ -311,6 +311,28 @@ public class ConversationListFragment extends Fragment
     }
   }
 
+  private void handleMuteAllSelected() {
+    final DcContext dcContext             = DcHelper.getContext(getActivity());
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -546,6 +568,18 @@ private boolean areSomeSelectedChatsUnpinned() {
     return false;
   }
 
+  private boolean areSomeSelectedChatsUnmuted() {
```
r10s, line 571: i would have found the code easier to read if this function would be `areSomeSelectedChatsMuted()` - but i see that this double-inverting is already present with `areSomeSelectedChatsUnpinned`, so, no blocker :)

r10s (approved):
nice improvement, code lgtm and also tested a bit

PR #2950, opened 2024-01-31
add an option to 'Read System Address Book'

r10s:
with chatmail, providers exist that allow to send encrypted mails only. pulling in the system address book unconditionally for chatmail will harm UX as for most of these addresses,
no public key will be known until eg. a QR code is scanned.

we may think this over in the future.
note, that one usually also can create a Delta Chat chat from the address book, so, this works also without having Delta Chat pulled in all addresses (which has some issues, https://github.com/deltachat/deltachat-android/issues/2915 )


adbenitez (approved):
indeed, better than completely removing it :+1:

r10s:
ftr, this option was hidden for chatmail later at https://github.com/deltachat/deltachat-android/pull/3050

PR #2951, opened 2024-02-01
catch errors thrown by Uri.parse()

r10s:
catch errors thrown by Uri.parse() to avoid crashes on tapping invite links

### DIFF
--- a/src/org/thoughtcrime/securesms/util/Util.java
+++ b/src/org/thoughtcrime/securesms/util/Util.java
@@ -70,7 +70,12 @@ public class Util {
   }
 
   public static boolean isInviteURL(String url) {
-    return isInviteURL(Uri.parse(url));
+    try {
+      return isInviteURL(Uri.parse(url));
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+    return false;
   }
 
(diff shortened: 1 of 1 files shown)


adbenitez (approved):

src/org/thoughtcrime/securesms/util/Util.java
```diff
@@ -70,7 +70,12 @@ public static boolean isInviteURL(Uri uri) {
   }
 
   public static boolean isInviteURL(String url) {
-    return isInviteURL(Uri.parse(url));
+    try {
+      return isInviteURL(Uri.parse(url));
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+    return false;
```
adbenitez, line 78:
```suggestion
    return url != null && isInviteURL(Uri.parse(url));
```
according to the docs, Uri.parse() only throws NullPointerException if the given string is null, so the check could be simpler, in practice we never pass a null string here because otherwise it would not be a clickable link in message

r10s: ah, thanks! so maybe not needed as urgently as i thought :)

PR #2953, opened 2024-02-04
Update help.html

jidanni:
Actually the page technically should say "passwords" not "password" too. But maybe all this, including my change today, is just making it too complicated, so feel free to reject it.


adbenitez:
hello @jidanni, thanks for your interest, I think that offline help is actually generated from here: https://github.com/deltachat/deltachat-pages/blob/master/en/help.md

so any change should be requested there, then it will propagate to all Delta Chat clients and online website

r10s:
replaced by https://github.com/deltachat/deltachat-pages/pull/796 , thanks a lot, @jidanni

PR #2955, opened 2024-02-06
make 'Read Address Book' a per-account option

r10s:
while working on the [iOS counterpart](https://github.com/deltachat/deltachat-ios/pull/2069), i came to the conclusion that it is much better to have a per-account setting here

this fits better to chatmail:
- one can have chatmail=OFF and normal-mail=ON
- new accounts are most often chatmail, even if there is a normal-mail account, and get the setting OFF, which is better in most cases
- the option more "fades out" this way :)


adbenitez (approved):
Nice improvement!!! 🤩

PR #2958, opened 2024-02-12
add troubleshooting menu option on the scanner side

### LINKED ISSUE #2956: make 'troubleshooting' available for add-as-second-device
add a kebab-menu and the entry there, exactly the same as for the show-qr-code site at "Add Second Device"

cmp. https://github.com/deltachat/deltachat-ios/pull/2074 , https://github.com/deltachat/deltachat-desktop/issues/3676

adbenitez:
close #2956

### DIFF
--- a/src/org/thoughtcrime/securesms/qr/RegistrationQrActivity.java
+++ b/src/org/thoughtcrime/securesms/qr/RegistrationQrActivity.java
@@ -80,7 +80,6 @@ public class RegistrationQrActivity extends BaseActionBarActivity {
                 finish();
                 return true;
             case R.id.troubleshooting:
-                // once help is evolved, it may be a more visible button, may be localized, may be offline
                 WebViewActivity.openUrlInBrowser(this, "https://delta.chat/en/help#multiclient");
                 return true;
         }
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/qr/RegistrationQrActivity.java
```diff
@@ -77,6 +79,10 @@ public boolean onOptionsItemSelected(MenuItem item) {
             case android.R.id.home:
                 finish();
                 return true;
+            case R.id.troubleshooting:
+                // once help is evolved, it may be a more visible button, may be localized, may be offline
+                WebViewActivity.openUrlInBrowser(this, "https://delta.chat/en/help#multiclient");
```
r10s, line 84: these kind of comments tend to get stale :)

also, i am not sure if a more visible button is helpful, as it usually should just work - and if not, the kebab menu is mostly only other option user can tap, so they will find it. more probable is that we add "Paste" to the menu.

i agree, with the offline help, but we have that in mind anyways and won't be reminded by a commented somewhere in the code that is read randomly.


```suggestion
                WebViewActivity.openUrlInBrowser(this, "https://delta.chat/en/help#multiclient");
```

r10s (approved):
nice! thanks a lot!

src/org/thoughtcrime/securesms/qr/RegistrationQrActivity.java
adbenitez: I just copy pasted the same comment from the other place where this is called like this, so the comment doesn't get even more stale :D

r10s: :D

r10s (approved):

PR #2967, opened 2024-03-04
Update to Rust 1.76.0

link2xt:
New core will require Rust 1.74.0 so current 1.72.0 will not work.


r10s:
thanks a lot for pushing things forward! 

however, unfortunately, i am getting some "undefined symbol" errors with this PR after running `./scripts/install-toolchains.sh`:  https://gist.github.com/r10s/9bfc8c9f7ecfad134b0ddfe38d5a3a25

Issue #2969, opened 2024-03-04
undefined symbols after updating rust-toolchain


r10s:
when updating `scripts/rust-toolchain` to `1.74.0` as required by core v1.136.0, there are some undefined symbols during linking:

https://gist.github.com/r10s/fc57a140b20c14ff9a02224f8c5c8aaf

as very common things as stderr are missing, i assume, i missed sth. very basic.

what i did:
- update `scripts/rust-toolchain` to `1.74.0`
- `./scripts/install-toolchains.sh`
- `./scripts/ndk-make.sh`

r10s:
ftr, iOS upgrade to toolchain 1.74.0 and core v1.136.0 works

r10s:
maybe some caching or whatnot, will have a second look later on

PR #2970, opened 2024-03-04
update core and rust-toolchain

r10s:
does not work for me, see https://github.com/deltachat/deltachat-android/issues/2969


r10s:
completely overseen the other pr, will try over :)

PR #2976, opened 2024-03-07
Replace Signal's getFileDescriptorOwner with K-9's SafeContentResolver

### LINKED ISSUE #2975: Get rid of `FileUtils.getFileDescriptorOwner`
getFileDescriptorOwner is the last piece of native JNI code taken from Signal. Signal itself does not use it anymore as the last call to it was removed in https://github.com/signalapp/Signal-Android/commit/9bac88697b25f0ca4ae1104129bc5f1649579eb0
That commit is under GPLv3+ and not AGPL so we can take whatever changes are needed from there.

This `getFileDescriptorOwner` in `jni/utils` blocks moving JNI code to a separate package. Once it is done we can revive https://github.com/angelo-fuchs/dc-jni-playground and resolve https://github.com/deltachat-bot/echo/issues/9

Moving JNI code to a separate package will make it possible to upload prebuilds to Maven Central so Android developers and F-Droid will only need to run Gradle to build Delta Chat.

link2xt:
Closes #2975

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -62,6 +62,11 @@ dependencies {
         exclude group: 'com.google.android.gms'
     }
 
+    // Replacement for ContentResolver
+    // that protects against the Surreptitious Sharing attack.
+    // <https://github.com/cketti/SafeContentResolver>
+    implementation 'de.cketti.safecontentresolver:safe-content-resolver-v14:1.0.0'
+
     testImplementation 'junit:junit:4.13.1'
     testImplementation 'org.assertj:assertj-core:1.7.1'
     testImplementation 'org.mockito:mockito-core:1.9.5'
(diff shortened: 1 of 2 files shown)


r10s:
> This probably breaks something

maybe not. i commented a bit at https://github.com/deltachat/deltachat-android/issues/2975#issuecomment-1984675971

r10s (commented):
nice, thanks! i think, this is a good solution.

as we're currently rolling out 1.44, to avoid additional moving targets, we'd better merge this when 1.44 is really out and stable (usually, there are some minor updates needed :)

r10s (approved):

adbenitez (approved):
great! nice improvement

PR #2979, opened 2024-03-18
react with any emoji

### LINKED ISSUE #2978: react with any emoji
while android was the first supporting reactions with 👍👎❤️😂 or 🙁, [ios](https://github.com/deltachat/deltachat-ios/pull/2087) and [desktop](https://github.com/deltachat/deltachat-desktop/pull/2964) caught up, allowing to react with _any_ emoji ;) 

> on android, when it comes to emoji, always a discussion wrt the existing emoji selector pops up - replace it, rely on system selector, what about old androids etc. this all has different efforts and different drawbacks ...

to unblock this, and to bypass all discussions and side effects, a pragmatic approach might be to just use any other emoji selector for the reactions (there are comparable small ones without shipping bitmaps ...) _or_ keep the existing selector for now.

a mvp might be:

1. show "..." right of the predefined reactions
2. once selected, open the existing emoji-selector, exactly as if one hits the corresponding button in the message . the selector will be abottom of the page, maybe even the input field is visible (hidden would be better, though).  
_alternatively_, use sth. as https://developer.android.com/develop/ui/views/text-and-emoji/emoji-picker
4. once a emoji is tapped, that is sent instead of added to the input bar, then this mode is left and the selector is hidden
5. to make it possible to remove custom-self-reaction: if at 3. the emoji is equal to the the self-reaction, remove it (this is what we've done at https://github.com/deltachat/deltachat-ios/pull/2098

if we decide for another emoji-selector at some point, quite some code could be reused in any case, so it is okay to be a bit hacky at 2.  here, the result counts :)

r10s:
using the existing reaction picker seems to be a pain as it is bound very close to our keyboard - and would also need special casing if it is not used.

therefore, this PR goes for using https://developer.android.com/develop/ui/views/text-and-emoji/emoji-picker , which had some building issues (see commit messages), however, seems straight-forward.

also, only very few lines of codes are needed.

~~however, for whatever reason, emoji do not show up in the picker.  when tapping the empty space in the picker, the app crashes. might be because of kotlin, because of a missing dependency, or excluded bit, whatnot. @adbenitez can you try this as well and maybe have a closer look?~~ EDIT: thanks a lot for fixing and for taking over this PR, this is very welcome! 💯

size-wise, this PR adds about 1mb to the app, that seems okay for the given feature ;) also, we did not add dependencies to the old picker, so we can remove that at some point (or at least disable it by default).

this is the "..." button:

this is the picker:

closes #2978

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,7 @@ dependencies {
     implementation 'androidx.lifecycle:lifecycle-viewmodel:2.6.2'
     implementation 'androidx.lifecycle:lifecycle-viewmodel-ktx:2.6.2'
     implementation 'androidx.work:work-runtime:2.8.1'
+    implementation 'androidx.emoji2:emoji2-emojipicker:1.4.0'
     implementation 'com.google.android.exoplayer:exoplayer-core:2.9.6' // plays video and audio
     implementation 'com.google.android.exoplayer:exoplayer-ui:2.9.6'
     implementation 'androidx.constraintlayout:constraintlayout:2.1.4'
--- a/build.gradle
+++ b/build.gradle
@@ -32,6 +32,7 @@ dependencies {
     implementation 'androidx.lifecycle:lifecycle-viewmodel-ktx:2.6.2'
     implementation 'androidx.work:work-runtime:2.8.1'
     implementation 'androidx.emoji2:emoji2-emojipicker:1.4.0'
+    implementation 'com.google.guava:guava:29.0-android'
     implementation 'com.google.android.exoplayer:exoplayer-core:2.9.6' // plays video and audio
     implementation 'com.google.android.exoplayer:exoplayer-ui:2.9.6'
     implementation 'androidx.constraintlayout:constraintlayout:2.1.4'
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -40,6 +40,9 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission android:name="android.permission.REQUEST_INSTALL_PACKAGES" />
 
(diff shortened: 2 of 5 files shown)


build.gradle
```diff
@@ -31,6 +31,8 @@ dependencies {
     implementation 'androidx.lifecycle:lifecycle-viewmodel:2.6.2'
     implementation 'androidx.lifecycle:lifecycle-viewmodel-ktx:2.6.2'
     implementation 'androidx.work:work-runtime:2.8.1'
+    implementation 'androidx.emoji2:emoji2-emojipicker:1.4.0'
+    implementation 'com.google.guava:guava:29.0-android'
```
adbenitez, line 35: Btw, why is this new dependency needed? There is no apparent direct usage in any new line added

adbenitez (approved):

build.gradle
r10s: i do not really known. by adding `implementation 'androidx.emoji2:emoji2-emojipicker:1.4.0'` we get the error:

```
/Users/bpetersen/projects/deltachat-android/src/org/thoughtcrime/securesms/connect/FetchWorker.java:12: error: cannot access ListenableFuture
public class FetchWorker extends Worker {
       ^
  class file for com.google.common.util.concurrent.ListenableFuture not found
```

additionally adding `implementation 'com.google.guava:guava:29.0-android'` fixes that for me, i found that out my some trial and error, @adbenitez would be great if you can have an additional look what is really needed

adbenitez: yes without it it doesn't compiles, I was surprised because it is not required in the emojipicker tutorial, and the error is in a seemly unrelated area, but it seems to be related to this:
https://stackoverflow.com/questions/64290141/android-studio-class-file-for-com-google-common-util-concurrent-listenablefuture

PR #2983, opened 2024-03-22
Update to Rust 1.77

link2xt:


r10s (approved):
works for me as well locally, i could start the emulator and also general an APK.

things i've done:
```
./scripts/clean-core.sh
./scripts/install-toolchains.sh
./scripts/ndk-make.sh
```

so, from my point of view, this can be merged, cc @adbenitez

adbenitez (approved):

PR #2984, opened 2024-03-28
move pref_system_emoji pref to advanced and enable it by default

### LINKED ISSUE #2980: tune down in-app emoji selector
the in-app emoji selector is worse than what most modern keyboards offer wrt emoji, eg ...

- modern emoji
- stickers

... are available by many keyboards - but not by our in-app emoji selector.

the last update was several years ago, it seems, we cannot really maintain that effort, even if we based on existing upstream (which does not supported older android ...). also, it seems unnecessary as things are already offered by keyboards - and our limited resources are better spent elsewhere.

to move forward on this issue, get some feedback and to not annoy existing users too much, we could do the following:

1. let `pref_system_emoji` default to `true` for new installations (existing installation stay as is for now)
2. move option "Appearance / Use System Emoji" to "Advanced / Use In-App Emoji" with the subtitle "In-app emoji often has less features" (just beside "Use In-App Camera")

then, let's see what happens :)

as a successor, we could check out [EmojiCompat](https://developer.android.com/reference/androidx/emoji/text/EmojiCompat) and see if that can replace emojify() & co. but that can come later as needed and would require are larger cleanup.

@adbenitez @all or do newer androids with really worser-than-our emoji-picker exist? all i have seen recently on androids was way better, even the default stock-androids in the emulators

adbenitez:
close #2980

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -658,7 +658,9 @@
     <string name="pref_read_receipts_explain">If read receipts are disabled, you won\'t be able to see read receipts from others.</string>
     <string name="pref_manage_keys">Manage Keys</string>
     <string name="pref_use_system_emoji">Use System Emoji</string>
-    <string name="pref_use_system_emoji_explain">In-app emoji often has less features than the system\'s emoji</string>
+    <string name="pref_use_system_emoji_explain">Turn off Delta Chat\'s built-in emoji support</string>
+    <string name="pref_use_builtin_emoji">Use In-App Emoji</string>
+    <string name="pref_use_builtin_emoji_explain">In-app emoji often has less features than the system\'s emoji</string>
     <string name="pref_use_inapp_camera">Use In-App Camera</string>
     <string name="pref_use_inapp_camera_explain">In-app camera usually has less features than the system camera</string>
     <string name="pref_show_system_contacts">Read System Address Book</string>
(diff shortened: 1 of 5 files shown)


r10s:
hm, better go on pair with "in-app camera" and name the option "User In-App Emoji" as well (inverted then), see https://github.com/deltachat/deltachat-android/issues/2980 - i would also define new strings for that, reusing with changed meanings will add some noise to translators.

so, `pref_use_builtin_emoji` and `pref_use_builtin_emoji_explain`, comparable to https://github.com/deltachat/deltachat-android/pull/2674/files

adbenitez:
(notice this PR changes the default to prefer system emoji for all DC users not only new ones, I was not sure how to detect this, but also I think it should be fine to change for all, they can always enable them again in the settings)

r10s:
yeah, wording looks lice now. thanks!

> (notice this PR changes the default to prefer system emoji for all DC users not only new ones, I was not sure how to detect this

i would not try to detect it, but just continue us

> but also I think it should be fine to change for all, they can always enable them again in the settings)



while it is true that ppl can switch 
hm, but we will need to deal with requests and discussions then. i wo

r10s:
sorry for closing and adding noise /me cannot type on the phone today ...

adbenitez:
> i would not try to detect it, but just continue using the old flag.

since old flag was for "use system emoji" and it was disabled by default, there is no old key value saved, so you can't know if the key is not set because it is a new user or because it is an old user that never changed the setting and hence was using built-in emoji :thinking:

adbenitez:
closing this PR, see https://github.com/deltachat/deltachat-android/issues/2980#issuecomment-2252706584 for the reasoning

PR #2988, opened 2024-04-06
support request_internet_access again

### LINKED ISSUE #2987: Webxdc with `request_internet_access` set not working on Saved Messages
since the introduction of the FILL500 hack it seems this hidden feature was overlooked and it stopped working

adbenitez:
close #2987

### DIFF
--- a/src/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -37,6 +37,8 @@ import com.b44t.messenger.DcChat;
 import com.b44t.messenger.DcContext;
 import com.b44t.messenger.DcEvent;
 import com.b44t.messenger.DcMsg;
+import com.b44t.messenger.rpc.HttpResponse;
+import com.b44t.messenger.rpc.Rpc;
 
 import org.json.JSONObject;
 import org.thoughtcrime.securesms.connect.AccountManager;
@@ -62,6 +64,7 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
   private String baseURL;
   private String sourceCodeUrl = "";
   private boolean internetAccess = false;
+  private Rpc rpc;
 
   public static void openWebxdcActivity(Context context, DcMsg instance) {
     if (!Util.isClickedRecently()) {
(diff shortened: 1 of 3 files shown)


src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -262,7 +271,12 @@ protected WebResourceResponse interceptRequest(String rawUrl) {
         byte[] blob = this.dcAppMsg.getWebxdcBlob(path);
         if (blob == null) {
           if (internetAccess) {
-            return null; // do not intercept request
+            HttpResponse httpResponse = rpc.getHttpResponse(dcContext.getAccountId(), rawUrl);
+            String mimeType = httpResponse.getMimetype();
+            if (mimeType == null) {
+              mimeType = "application/octet-stream";
+            }
+            return new WebResourceResponse(mimeType, httpResponse.getEncoding(), new ByteArrayInputStream(httpResponse.getBlob()));
```
adbenitez, line 279: This is because otherwise normal traffic is redirected to the fake proxy introduced in #2539 but while this works with the webxdc-test app that loads the delta.chat page, the maps app doesn't work because OSM is not happy about the [wrong user agent](https://operations.osmfoundation.org/policies/tiles/) core is sending:
image

Ideally core should be fixed because otherwise the browsing will not respect the core proxy set like the "show full message" does, otherwise just not setting this fake proxy when internetAccess == true would be the simplest solution (ignoring the network leaks out of the configured proxy, but I suspect WebRTC also doesn't respect this proxy? otherwise that would have been used to disable it)

adbenitez: cc @link2xt

adbenitez: will do that, I am a bit worried about the inconsistent usage of the set sock proxy in DC, it is used only for "show full message" and it might cause false expectations we don't say anything about this traffic in Webxdc not being routed to the set proxy

adbenitez: done

Simon-Laux: There are also leaflet plugins for caching, like https://github.com/MazeMap/Leaflet.TileLayer.PouchDBCached

I think it would still make sense to use core, we could add support for defining http headers in the jsonrpc method. core respects socks5 proxy (for example tor) and on desktop we might not be able to replicate this, as dns block seem to be for the whole app there (all windows at once), or whitelisting some domains, like we currently do for mapbox.

PR #2992, opened 2024-04-16
add QR code scanning hint better fitting for desktop

r10s:
for desktop, the QR code scanning hint "Hold your camera over the QR code." is often not practical.

this PR adds the often better fitting string "Move the QR code to the camera."

see https://github.com/deltachat/deltachat-desktop/pull/3762

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -892,6 +892,7 @@
     <string name="load_qr_code_as_image">Load QR Code as Image</string>
     <string name="qrscan_title">Scan QR Code</string>
     <string name="qrscan_hint">Hold your camera over the QR code.</string>
+    <string name="qrscan_hint_desktop">Move the QR code to the camera</string>
     <string name="qrscan_failed">QR code could not be decoded</string>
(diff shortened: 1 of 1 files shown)


adbenitez (approved):

res/values/strings.xml
```diff
@@ -892,6 +892,7 @@
     <string name="load_qr_code_as_image">Load QR Code as Image</string>
     <string name="qrscan_title">Scan QR Code</string>
     <string name="qrscan_hint">Hold your camera over the QR code.</string>
+    <string name="qrscan_hint_desktop">Move the QR code to the camera.</string>
```
adzialocha, line 895: Maybe without the dot? 🙂

r10s: yeah, it's better without

i copied the old one ... that could be changed as well as some point (be we should not mix changing strings and adding strings for $reasons)

r10s:
k, the string is available on transifex and android-main-branch now - and i also removed the `.` for the old `qrscan_hint`

PR #2997, opened 2024-04-20
allow to open troubleshooting menu option directly from offline help

### LINKED ISSUE #2994: allow to open "offline help" from welcome screens
we would have needed that already for the "add as second device screen", however, it did not work (explanation attempts are at https://github.com/deltachat/deltachat-android/pull/2842#issuecomment-1810224065 ), and we went for the online help there

however, this was only meant temporary fix, using in-app-help is far better (no context switch, ppl are not always used to browsers on phones (yes, really), and, of course, works offline)

with the new onboarding, we attempt to make the FAQ accessible from the welcome screen in general, and probably also add sth about "instant accounts" there. 

one more reason to figure that out.

(probably it is needed to combing the two base classes into one and drive "special activities" by a flag)

adbenitez:
close #2994 

after this is merged, we could then easily merge `PassphraseRequiredActionBarActivity` and `BaseActionBarActivity`

### DIFF
--- a/src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
+++ b/src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
@@ -45,8 +45,12 @@ public abstract class PassphraseRequiredActionBarActivity extends BaseActionBarA
 
   protected void onCreate(Bundle savedInstanceState, boolean ready) {}
 
-  // "Locked Mode" is when the account is not configured (Welcome screen) or when sharing a backup
-  // (Add second device) and the user should not leave that screen otherwise the state would be lost.
-  // Sometimes it is fine to allow to change in these situations, like to see the logs or offline help.
+  // "Locked Mode" is when the account is not configured (Welcome screen)
+  // or when sharing a backup (Add second device).
+  // The app is "locked" to an activity if you will.
+  // In "Locked Mode" the user should not leave that activity otherwise the state would be lost -
+  // so eg. tapping app icon or notifications MUST NOT replace activity stack.
+  // However, sometimes it is fine to allow to pushing activities in these situations, 
+  // like to see the logs or offline help.
   protected boolean allowInLockedMode() { return false; }
 }
--- a/src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
+++ b/src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
@@ -44,5 +44,9 @@ public abstract class PassphraseRequiredActionBarActivity extends BaseActionBarA
   }
 
   protected void onCreate(Bundle savedInstanceState, boolean ready) {}
+
+  // "Locked Mode" is when the account is not configured (Welcome screen) or when sharing a backup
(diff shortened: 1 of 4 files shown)


src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
```diff
@@ -38,4 +44,5 @@ protected final void onCreate(Bundle savedInstanceState) {
   }
 
   protected void onCreate(Bundle savedInstanceState, boolean ready) {}
+  protected boolean allowInLockedMode() { return false; }
```
r10s, line 47: this requires a comment, what the "locked mode" is etc. and why things are not allowed by default and when one wants to change default behaviour

r10s (commented):
very nice! much better to have the help offline! and looking forward to have only one base class

some documentation is missing, when that is added, this PR is fine

src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
adbenitez: added a comment, to be honest I am not 100% sure of how this "lock mode" hack work so it is hard to explain it XD

src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
```diff
@@ -38,4 +44,9 @@ protected final void onCreate(Bundle savedInstanceState) {
   }
 
   protected void onCreate(Bundle savedInstanceState, boolean ready) {}
+
+  // "Locked Mode" is when the account is not configured (Welcome screen) or when sharing a backup
+  // (Add second device) and the user should not leave that screen otherwise the state would be lost.
+  // Sometimes it is fine to allow to change in these situations, like to see the logs or offline help.
```
r10s, line 50:
```suggestion
  // "Locked Mode" is when the account is not configured (Welcome screen)
  // or when sharing a backup (Add second device).
  // The app is "locked" to an activity if you will.
  // In "Locked Mode" the user should not leave that activity otherwise the state would be lost -
  // so eg. tapping app icon or notifications MUST NOT replace activity stack.
  // However, sometimes it is fine to allow to pushing activities in these situations, 
  // like to see the logs or offline help.
```

r10s (approved):
your comment was fine, this is also how i understood the issue. i just concretize it a bit.

what's missing and a bit vague, also to me, is how "Locked Mode" is entered - this seems to be implicitly done by using `BaseActionBarActivity` instead of `PassphraseRequiredActionBarActivity` as base (@adbenitez please double check, i am unsure here :) if we merge the two bases and add a flag to enter lock mode, this is probably even clearer

src/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
adbenitez: > what's missing and a bit vague, also to me, is how "Locked Mode" is entered 

yes, both welcome activity and backup transfer activity extend from BaseActionBarActivity, so when merging they would set the "allow in locked mode" to true which is more clearer than the currently implicit way via the parent class

PR #3000, opened 2024-04-21
add FCM PUSH notification support

### LINKED ISSUE #2998: support FCM PUSH notifications
after having some experience with [PUSH notifications on iOS](https://github.com/deltachat/deltachat-ios/pulls?q=is%3Apr+is%3Aclosed+is%3Amerged+label%3Anotify), [support of core](https://github.com/deltachat/deltachat-core-rust/pull/5286), [support of chatmail](https://github.com/deltachat/chatmail/pull/201), [having a dumb server set up](https://github.com/deltachat/notifiers/pull/17), and as android also has of issues with [IMAP IDLE not being killed](https://dontkillmyapp.com), we aim for PUSH notifications on android as well.

rough TODOs:

- create new project on [console.firebase.google.com](http://console.firebase.google.com/) and link that to our package name (this will be done by @r10s so we do not mix compliance related things too much)
-  in `build.gradle`, add implementation `com.google.firebase:firebase-messaging...`, peek at Signal to exclude analytics & co
- create class extending `FirebaseMessagingService`
- override `FirebaseMessagingService::onNewToken()`, prefix  received token by `fcm-APPLICATION_ID:` and pass it to [`dc_accounts_set_push_device_token()`](https://c.delta.chat/classdc__accounts__t.html#a8b90d31eb5914e2d620f8ee94e58fc5a) (core will take care to pass the token to accounts that support PUSH, however, notification server needs adaption for `fcm:` prefix)
- override `FirebaseMessagingService::onMessageReceived()` and add a notification-banner using existing code (that maybe needs adaption, sure)
- Register our FirebaseMessagingService class in the manifest file  as <service>, action `android:name="com.google.firebase.MESSAGING_EVENT"`
- make sure, app is still working if google is not available (not so nice would be to have different apps, esp. when it comes to other PUSH services we might support in the future ...)
- check if and which options or switches are needed in the UI's settings
- check what happens when the new APK is running on a device without FCM  or with an android sdk<19
- ...

for all steps, makes sense to peek at Signal or comparable apps.

implementing this of course requires close cooperation with the sysadmin team 👋 latest when we are at step 4, starting to pass the token to core, we should file an issue at https://github.com/deltachat/notifiers (https://github.com/deltachat/notifiers/issues/31), describing in details what we need. 

core and chatmail ideally do not require changes, as the prefix will just be passed through.

r10s:
this PR adds Firebase Clould Notifications (FCM), see #2998 for details and reasoning

the FCM token is get unconditionally currently. when the server pings us, we do nothing special - at that moment, the app should just be started (again) and process messages.

we should get this PR in to **gather feedback on a beta-version** if there are any issues of running on real devices (google and non-googles) - even if the server is not yet ready.

subsequent tasks in **another PR:** 

- init more gracefully together with POST_NOTIFICATION and other background-checks
- consider adding options and/or ask on startup
- reorganize `build.gradle`, so that `com.b44t.messenger` flavour (used for f-droid) is build without FCM - `chat.delta` and `com.b44t.messenger.beta` should stay as is

tasks done:

- [x] create firebase project linked to all used package ids
- [x] send tokens to core, depending on package ID, prefixed by `fcm-chat.delta:` (gplay and others), `fcm-com.b44t.messenger:` (fdroid), `fcm-com.b44t.messenger.beta:` (betas sent around), `fcm-chat.delta.beta:` (not used afaik)  
in general, similar to iOS it seems good if PUSH also works in betas etc., esp. as these are mainly used by devs, so if sth goes wrong it is easier noticable.  
for f-droid - idk, maybe they device to disable FCM, but for now, let's treat things the same if google-services are available on a phone anyways
- [x] show token in log
- [x] override `FirebaseMessagingService::onMessageReceived()` and add a notification-banner using existing code
- [x] check if we can run on on a device without FCM (if possible, avoid having different apps - note, it may not only be FCM in the future) - EDIT: works at least in the emulator, we need more in-the-wild-testing, however
- [x] check for analytics & co and exclude that
- [x] check if we can run on android sdk<19 - EDIT: checked - and android 4.1 - 4.3 should stays supported, [see below](https://github.com/deltachat/deltachat-android/pull/3000#issuecomment-2090656139)
- [x] check if we can build/run when using different package id - EDIT: checked, see 2nd point

useful links:

- https://console.firebase.google.com/
- https://firebase.google.com/docs/cloud-messaging
- https://github.com/deltachat/notifiers/issues/31

nb: fcm and related stuff adds about 0.5mb the the .apk size

closes #2998

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -381,6 +381,12 @@
         </intent-filter>
     </service>
 
+    <service android:name=".notifications.FcmReceiveService" android:exported="true">
+        <intent-filter>
+            <action android:name="com.google.firebase.MESSAGING_EVENT" />
+        </intent-filter>
+    </service>
+
     <receiver android:name=".notifications.MarkReadReceiver"
               android:enabled="true"
               android:exported="false">
@@ -412,6 +418,12 @@
         android:authorities="${applicationId}.attachments"
         android:grantUriPermissions="true"
         android:exported="false">
+    </provider>
+
+    <provider
+        android:name="com.google.firebase.provider.FirebaseInitProvider"
+        android:authorities="${applicationId}.firebaseinitprovider"
+        tools:node="remove">
     </provider>
 
     <receiver android:name=".service.BootReceiver"
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -40,8 +40,8 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission android:name="android.permission.REQUEST_INSTALL_PACKAGES" />
 
-    <!-- force compiling emojipicker on sdk<21; we'll add a runtime switch to not use the picker in this case -->
-    <uses-sdk tools:overrideLibrary="androidx.emoji2.emojipicker"/>
+    <!-- force compiling emojipicker on sdk<21 and firebase on sdk<19; runtime checks are required then -->
+    <uses-sdk tools:overrideLibrary="androidx.emoji2.emojipicker, com.google.firebase.messaging, com.google.android.gms.cloudmessaging"/>
 
     <application android:name=".ApplicationContext"
                  android:icon="@mipmap/ic_launcher"
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -57,6 +57,10 @@
                  android:hasFragileUserData="true"
         >
 
+    <meta-data android:name="firebase_analytics_collection_deactivated" android:value="true" />
+    <meta-data android:name="google_analytics_adid_collection_enabled" android:value="false" />
+    <meta-data android:name="firebase_messaging_auto_init_enabled" android:value="false" />
+
     <!-- android car support, see https://developer.android.com/training/auto/start/,
          as this potentially blocks releases on gplay due to extra-checks,
          we disable this during the first gplay releases -->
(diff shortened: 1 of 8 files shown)


r10s:
> check if we can run on android sdk<19 

i tried the [overrideLibrary-hack from #2979](https://github.com/deltachat/deltachat-android/pull/2979) - this does not work out as firebase seems to have another manifest that we cannot easily change, resp. i did not spend more time on it

this means, as things stand now, we would give up android 4.1-4.3 (jelly bean) - oldest supported device is then 4.4 (kitkat) - this was [subject already 4 (!) years ago](https://github.com/deltachat/deltachat-android/pull/1227)) - for now, i do not think, there are resources to dive deeper, so this should not block moving forward.

as 4 years ago, might be, we find a solution, but maybe first see if there are reasonable complains

r10s:
@link2xt the current state is not for release but for a 1.45 beta version, and is useful even if the server is not ready yet (to see if app crashes on some non-google phones etc.)

note, that the fcm-prefixes contain the APPLICATION_ID now, so the relevant ones are `fcm-chat.delta:` and `fcm-com.b44t.messenger-beta:`, cmp. https://github.com/deltachat/notifiers/issues/31#issuecomment-2091373063 (we could  also use sth as `fcm:` and `fcm-sandbox:`, however,  as functionality is bound to APPLICATION_ID, it seems practical to just add that to the token)

Issue #2995, opened 2024-04-20
add and refine instant onboarding strings


r10s:
once flow is settled in https://github.com/deltachat/deltachat-desktop/pull/3773, get the strings from desktop `_untranslated_en.json` here

this is also a good timings for refinements

@adzialocha ping me, when the time comes  :)

after that, instant-onboarding can be done for android/ios, roughly copying the pattern settled on desktop then

PR #3001, opened 2024-04-22
add strings needed for instant onboarding

r10s:
this PR extracts  new strings from [deltachat/deltachat-desktop#3773](https://github.com/deltachat/deltachat-desktop/pull/3773/files#diff-5bd8446ffbbe28eb2eacda360fd1fa0640832786a9170440fe58275a4190102b).

the following strings were not added as they are existing already in a sufficient way (desktop PR needs to be changed):

`onboarding_title` -> `welcome_desktop` (exclamation mark can be added later)
`onboarding_subtitle` -> `welcome_chat_over_email` (dot can be removed later)
`onboarding_scan_qr_code` -> `qrscan_title`
`multidevice_setup_option` -> `multidevice_receiver_title`
`import_backup_option` -> `import_backup_title` (iirc, it was renamed to "restore" on purpose, let's stay with that for now)
`instant_onboarding_confirm_label` -> `ok` (this is used in a message box, right? i would stay calm there, otherwise use the other "Let's Get Started")

for the other strings, i stayed with existing keys - they're good enough and i did not want to add additional noise.

for upper/lower case: we voted for [capitalising titles and buttons](https://capitalizemytitle.com) at some point and do not want to change this currently; i adapted the strings accordingly.

@adzialocha will add some screenshots to https://github.com/deltachat/deltachat-desktop/pull/3773 so one can get a better idea

closes https://github.com/deltachat/deltachat-android/issues/2995

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -557,8 +557,8 @@
     <string name="manual_account_setup_option">Manual Login</string>
     <!-- Instant onboarding title (there is not more to do than to set name and avatar) -->
     <string name="instant_onboarding_title">Your Profile</string>
-    <!-- The two placeholders mark beginning and end of a tappable link -->
-    <string name="instant_onboarding_agree_default">I agree to the %1$sPrivacy Policy%2$s</string>
+    <!-- The placeholder will be replaced by the tappable string "Privacy Policy" (privacy_policy) -->
+    <string name="instant_onboarding_agree_default">I agree to the %1$s</string>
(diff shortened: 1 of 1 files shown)


adzialocha:
Thank you for this @r10s, I've adjusted the strings now in the PR, they should reflect this current state

r10s:
ftr, the strings are discussed with at least @adzialocha and @hpk42

adzialocha (approved):

PR #3002, opened 2024-04-22
allow to pin, archive, etc. when searching in the chatlist

### LINKED ISSUE #2993: Offer Pin/archive/mute/delete on long-tapping chats in search results
As pre-discussed in various channels, this issue is about offering Pin/Archive/mute/delete/etc options when long-tapping a chat in the search-result chatlist.  The equivcalent right-click on search result chats exists on Desktop.

adbenitez:
close #2993 

besides the new functionality it also fixes a bug in the search results, it was not filtering with "DC_GCL_FOR_FORWARDING" when using the search while in forwarding/sharing mode so the device message and other invalid chats were proposed as search results

besides allowing to pin/delete/mute chats in search mode, this also allows to forward to several chats in the search result which is nice when you want to forward to several chats with some common key in their name, for now mixing chats and contact is not supported tho


src/org/thoughtcrime/securesms/search/SearchFragment.java
```diff
@@ -68,8 +73,8 @@ public void onCreate(@Nullable Bundle savedInstanceState) {
 
     this.locale = (Locale) getArguments().getSerializable(EXTRA_LOCALE);
 
-    viewModel = ViewModelProviders.of(this, (ViewModelProvider.Factory) new SearchViewModel.Factory(getContext())).get(SearchViewModel.class);
-    DcEventCenter eventCenter = DcHelper.getEventCenter(getContext());
+    viewModel = ViewModelProviders.of(this, (ViewModelProvider.Factory) new SearchViewModel.Factory(requireContext())).get(SearchViewModel.class);
+    DcEventCenter eventCenter = DcHelper.getEventCenter(requireContext());
```
adbenitez, line 77: all this changes from  getContext() to requireContext() were while fixing all warnings from Android Studio so I could be sure I didn't let in something wrong with the changes, ideally I should have added it in a different commit :/

r10s (approved):
tested quite a bit, all works as expected!

what an improvement! i think, this was really worth the effort - plus the fixed bugs! really good work 🤘

the code also lgtm, i compared some changes before/after and look closer to some critical parts, did not fine things to criticise :)

i think, let's merge it, giving testers a chance :) - _"if something happens, it happens out there!"_

PR #3005, opened 2024-04-24
use new experimental maps integration

adbenitez:
what is missing is checking latest core release tag in the submodule, I was not sure how it is usually done? 

counterpart of https://github.com/deltachat/deltachat-ios/pull/1912, uses https://github.com/deltachat/deltachat-core-rust/pull/5461


adbenitez:
another thing that needs to be considered, webxdc only works on android 5+ so maybe we should take the opportunity to remove the experimental option, enable maps in all platforms and disable them in android <5

r10s:
i think, at least for 1.46, we will stay as experimental, this gives us room for changes (as this one :)

i am also not really happy with the top-level UI elements: they clutter the chat always, but are needed only rarely and only in few chats. this is solved better on iOS. but then again, as long as it is experimental it's fine. i am not suggesting to change them now, it depends also on other features - eg. a "send position once" is still missing, and also a "show on map" deep link (for that, maybe wait until we have deep links on webxdc)

and yes, the map will only be available on android5+, that's a reasonable tradeoff

src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -48,40 +48,69 @@
...
+    if (msgId == 0) {
+      try {
+        InputStream inputStream = context.getResources().getAssets().open("webxdc/maps.xdc");
+        String outputFile = DcHelper.getBlobdirFile(dcContext, "maps", ".xdc");
+        Util.copy(inputStream, new FileOutputStream(outputFile));
+        dcContext.setWebxdcIntegration(outputFile);
+        msgId = dcContext.initWebxdcIntegration(chatId);
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
```
r10s, line 85: that would let the app crash, even though it should not happen in normal usable at all, maybe better be a bit more friendly and give some hints:

```suggestion
      } catch (IOException e) {
        e.printStackTrace();
      }
      if (maps == 0) {
         Toast.makeText(context, "Cannot get maps.xdc, see log for details.", Toast.LENGTH_LONG).show();
         return;
      }
```

r10s (approved):
code of integration looks good to me, thanks a lot!

question is now if the new `maps.xdc` is good enough to replace the old android map - on  iOS the decision was easier as there was just no map

what's missing:

- go to own position
- show all past tracks beyond 24h

we plan to add that, and improve even beyond, but we cannot make promises about timing. **then, we're still experimental**

whats new/nicer:

- nicer layout
- POI always show labels
- lots of POI and positions are shown much faster
- lots of POI are summed up nicely as the PNG is optimised for that
- you see the age of a position directly on the map (no longer you run to a bar 🍺 then noticing the position was from 10h ago) 
- 4 mb smaller apk

so, i would vote for merging it to move forward, even if there is some temporary worsening for some ppl

adbenitez:
> so, i would vote for merging it to move forward, even if there is some temporary worsening for some ppl

+1 also less problems bad karma with fdroid etc :)

PR #3006, opened 2024-04-24
fix bug in WebxdcActivity: don't try to get iframe in "allow internet" mode

adbenitez:
due to recent changes to fix the "allow internet" mode, this part in `handleEvent()` was forgotten

### DIFF
--- a/src/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -316,7 +316,8 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
     int eventId = event.getId();
     if ((eventId == DcContext.DC_EVENT_WEBXDC_STATUS_UPDATE && event.getData1Int() == dcAppMsg.getId())) {
       Log.i(TAG, "handleEvent");
-      webView.loadUrl("javascript:document.getElementById('frame').contentWindow.__webxdcUpdate();");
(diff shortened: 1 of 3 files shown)


adbenitez:
this needs to be merged and #3005 rebased on top

src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -316,7 +316,8 @@ public void handleEvent(@NonNull DcEvent event) {
     int eventId = event.getId();
     if ((eventId == DcContext.DC_EVENT_WEBXDC_STATUS_UPDATE && event.getData1Int() == dcAppMsg.getId())) {
       Log.i(TAG, "handleEvent");
-      webView.loadUrl("javascript:document.getElementById('frame').contentWindow.__webxdcUpdate();");
+      final String window = internetAccess? "window" : "document.getElementById('frame').contentWindow";
+      webView.loadUrl("javascript:" + window + ".__webxdcUpdate();");
```
r10s, line 320: even if slightly longer, i fine the following easier to read. 

but treat this just as a opinionated suggestion, it's up to you, i wold be fine with both, no blocker.

```suggestion
      if (internetAccess) {
        webView.loadUrl("javascript:window.__webxdcUpdate();");
      } else {
        webView.loadUrl("javascript:document.getElementById('frame').contentWindow.__webxdcUpdate();");
      }
```

r10s (approved):
thanks for the fix, lgtm.

i did not test however, would be cool if you can double check that all blocking still works as expected with test.xdc

PR #3007, opened 2024-04-25
simplify code

r10s:
this PR simplify code by following linter hints; initially i only wanted to remove the unused return value


src/org/thoughtcrime/securesms/WebViewActivity.java
```diff
@@ -25,7 +25,6 @@
 import androidx.webkit.ProxyController;
 import androidx.webkit.ProxyConfig;
 
-import com.b44t.messenger.util.concurrent.ListenableFuture;
 import com.b44t.messenger.util.concurrent.SettableFuture;
```
adbenitez, line 28: can't this import also be removed?

adbenitez (approved):

PR #3008, opened 2024-04-25
fix openOnlineUrl(), mailto/openpgp4fpr has to open externally

adbenitez:
fix bug recently introduced by #2988 

mailto/openpgp4fpr was not handled in openOnlineUrl() which caused it to be handled in interceptRequest() which only expects HTTP(S) requests

this also allows to open openpgp4fpr links (only mailto was supported in the past) as DC desktop does (what is needed by the [Public Bots Index](https://github.com/deltachat-bot/public-bots))

### DIFF
--- a/src/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -266,18 +266,16 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
     }
   }
 
+  // This is only called when internetAccess == true or for mailto/openpgp4fpr scheme,
+  // because when internetAccess == false, the page is loaded inside an iframe,
+  // and WebViewClient.shouldOverrideUrlLoading is not called for HTTP(S) links inside the iframe
   @Override
   protected boolean openOnlineUrl(String url) {
-    if (internetAccess) {
-      // internet access enabled, continue loading in the WebView
-      return false;
-    }
-    if (url.startsWith("mailto:")) {
+    Log.i(TAG, "openOnlineUrl: " + url);
+    if (url.startsWith("mailto:") || url.startsWith("openpgp4fpr:")) {
       return super.openOnlineUrl(url);
     }
-
-    Toast.makeText(this, "Please embed needed resources.", Toast.LENGTH_LONG).show();
-    return true; // returning `true` causes the WebView to abort loading
+    return false; // continue loading in the WebView
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


r10s:
nb: for testing, ex. write to <botsindex@nine.testrun.org> and try to click a bot in the list

src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -266,18 +266,16 @@ private void setScreenMode(Configuration config) {
...
-    }
-    if (url.startsWith("mailto:")) {
+    Log.i(TAG, "openOnlineUrl: " + url);
+    if (url.startsWith("mailto:") || url.startsWith("openpgp4fpr:")) {
       return super.openOnlineUrl(url);
     }
-
-    Toast.makeText(this, "Please embed needed resources.", Toast.LENGTH_LONG).show();
-    return true; // returning `true` causes the WebView to abort loading
+    return false; // continue loading in the WebView
```
r10s, line 278: i find this too dangerous and implicit. might be, that we get rid of iframes at some point (hopefully) or flow changes otherwise - it is so easy to forget this bit then.

therefore, and also as another defense line, let's stay with the `internetAccess` check:

```suggestion
    return !internetAccess; // returning `false` continues loading in WebView; returning `true` let WebView abort loading
```

removing the toast is still fine, as this is not the happy path.

src/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -266,18 +266,16 @@ private void setScreenMode(Configuration config) {
     }
   }
 
+  // This is only called when internetAccess == true or for mailto/openpgp4fpr scheme,
+  // because when internetAccess == false, the page is loaded inside an iframe,
+  // and WebViewClient.shouldOverrideUrlLoading is not called for HTTP(S) links inside the iframe
```
r10s, line 271:
```suggestion
  // This is usually only called when internetAccess == true or for mailto/openpgp4fpr scheme,
  // because when internetAccess == false, the page is loaded inside an iframe,
  // and WebViewClient.shouldOverrideUrlLoading is not called for HTTP(S) links inside the iframe
```

r10s (approved):
let's add leave a final line of defence :)

otherwise lgtm, and also testing works - thanks for fixing!

src/org/thoughtcrime/securesms/WebxdcActivity.java
adbenitez: notice that continuing loading only means that the page is loaded in place which is then blocked by CSP and if that barrier would fail somehow then it would be blocked by the fake-proxy,

but I will merge it as a 3rd layer shield :o

PR #3012, opened 2024-04-28
cleanup dead code

r10s:
some more random cleanup


src/org/thoughtcrime/securesms/TransportOption.java
```diff
@@ -11,26 +11,20 @@ public enum Type {
...
     this.composeHint         = composeHint;
   }
 
   public @NonNull Type getType() {
-    return type;
-  }
-
-  public boolean isType(Type type) {
-    return this.type == type;
+    return Type.NORMAL_MAIL;
```
r10s, line 27: whole TransportOption and TransportOptions classes are probably superfluous, however, there are some callback involved that may do things, so i stopped refactoring here for now to avoid behaviour changes

src/org/thoughtcrime/securesms/geolocation/DcLocation.java
```diff
@@ -4,11 +4,6 @@
 
 import java.util.Observable;
 
-
-/**
- * Created by cyberta on 06.03.19.
- */
-
```
r10s, line 11: in general, we avoid such boilerplates - they get outdated etc, the git history is a much better and cheaper source

adbenitez (approved):

PR #3013, opened 2024-04-29
unify the flows of DCACCOUNT: and DCLOGIN: 

r10s:
DCLOGIN: is same as DCACCOUNT: and differ only wrt confirmation message.

this PR unifies flows to avoid hickup as on (https://github.com/deltachat/deltachat-desktop/pull/3773) on the upcoming onboarding work

### DIFF
--- a/src/org/thoughtcrime/securesms/WelcomeActivity.java
+++ b/src/org/thoughtcrime/securesms/WelcomeActivity.java
@@ -451,20 +451,11 @@ public class WelcomeActivity extends BaseActionBarActivity implements DcEventCen
             String qrRaw = scanResult.getContents();
             DcLot qrParsed = dcContext.checkQr(qrRaw);
             switch (qrParsed.getState()) {
-                case DcContext.DC_QR_LOGIN:
-                    String address = qrParsed.getText1();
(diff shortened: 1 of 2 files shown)


adzialocha:
> DCLOGIN: is same as DCACCOUNT: and differ only wrt confirmation message.

Not really related to this particular PR, but as far as I understand (and I might be wrong as I'm new to the concept of `DCLOGIN`) there are other differences as well. `DCLOGIN` doesn't show the screen to set a display name etc. and just logs the user in directly after confirming while `DCACCOUNT` shows the instant onboarding screen first.

r10s:
> DCLOGIN doesn't show the screen to set a display name etc. and just logs the user in directly after confirming while DCACCOUNT shows the instant onboarding screen first

i see - but at least on android/ios, these two behave similar to DCACCOUNT wrt flow already today, so nothing would be worsend. 

we can consider another flow as needed, when real-time-usage requires that, but i would not do that in the current instant-onboarding refactoring - there it seems fine to do just the same (esp. as also core-wise  same get_qr -> set_config_from_qr)

adbenitez (approved):

PR #3015, opened 2024-04-30
adapt to new "instant onboarding" UX workflow

### LINKED ISSUE #2996: New "Instant Onboarding" UX flow to create accounts
once https://github.com/deltachat/deltachat-desktop/pull/3773 and https://github.com/deltachat/deltachat-android/issues/2995 are settled, instant onboarding can be done for android as well

in case someone is in idle, before, #2994 can be targeted already

counterpart of https://github.com/deltachat/deltachat-ios/issues/2156

adbenitez:
close #2996

- [x] use new intro image in welcome screen
- [x] change buttons in welcome screen
- [x] implement sign-in path a.k.a "I already have a login" button
- [x] implement sign-up path a.k.a "let's get started!" button for default instance
- [x] add "explore more options" and "scan QR" options to InstantOnboardingActivity
- [x] implement "scan QR" button in InstantOnboardingActivity
- [x] process "DCACCOUNT:" scheme URIs with InstantOnboardingActivity instead of WelcomeActivity 
- [x] clean-up, remove unused code in WelcomeActivity etc

Screenshot 2024-05-05 at 11 32 33 &nbsp; Screenshot 2024-05-05 at 11 35 20 &nbsp;  Screenshot 2024-05-05 at 11 35 36

counterpart of https://github.com/deltachat/deltachat-desktop/pull/3773


r10s:
did a functionality test of all but scanning, things work fine and look good. did not test QR codes yet, and also did not look at the code.

one minor comment for now:

Screenshot 2024-05-03 at 22 23 59

these linebreaks in the lower buttons do not look nice and seem superfluous. otherwise, the left/right buttons are not bad, worth a try

adbenitez:
> these linebreaks in the lower buttons do not look nice and seem superfluous. otherwise, the left/right buttons are not bad, worth a try

it doesn't do the line break in all phones it depends on screen size, that was my struggle into making that work in a way that if for some language the text is too long it would wrap around instead of pushing the second button out of view, it turned out to be harder than what I expected

r10s:
no blocker, was just wondering. maybe jt can be fixed by a simple xml flag adapting button width to content (instead of making both button 50% each)

r10s:
/me now also checked quite some QR and "Explore Other Option" flows ... all nicely so far :) and i never created so many accounts is so few time ;)

res/layout/instant_onboarding_activity.xml
```diff
@@ -0,0 +1,131 @@
...
+        android:text="@string/instant_onboarding_create"/>
+
+    <androidx.legacy.widget.Space
+        android:layout_weight="2"
+        android:layout_width="match_parent"
+        android:layout_height="0dp"/>
+
+    <androidx.constraintlayout.widget.ConstraintLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content">
```
r10s, line 84: why not also use a `LinearLayout` here?

unless there are good reasons, in general, `LinearLayout` is easier maintainable and understandable, esp. by non-pro android-devs :)

no blocker, though

res/layout/login_options_view.xml
```diff
@@ -0,0 +1,49 @@
...
+    <Button
+        style="@style/ButtonSecondary"
+        android:id="@+id/login_button"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:paddingLeft="24dp"
+        android:paddingRight="24dp"
+        android:text="@string/manual_account_setup_option"/>
+
+  </LinearLayout>
```
r10s, line 47: yeah, indeed, that looks much better than plain `AlertDialog.setItems()`; thanks for taking the effort, this is worthwhile esp as we're talking about first impressions :)

src/org/thoughtcrime/securesms/InstantOnboardingActivity.java
```diff
@@ -0,0 +1,438 @@
...
+        IntentResult scanResult = IntentIntegrator.parseActivityResult(requestCode, resultCode, data);
+        if (scanResult != null && scanResult.getFormatName() != null) {
+          setProviderFromQr(scanResult.getContents());
+        }
+        break;
+    }
+  }
+
+  private void setProviderFromQr(String rawQr) {
+        DcLot qrParsed = dcContext.checkQr(rawQr);
```
r10s, line 153: indentation in this function should be 2 spaces only

src/org/thoughtcrime/securesms/InstantOnboardingActivity.java
```diff
@@ -0,0 +1,438 @@
...
+import java.security.SecureRandom;
+import java.util.Objects;
+
+public class InstantOnboardingActivity extends BaseActionBarActivity implements DcEventCenter.DcEventDelegate {
+
+  private static final String TAG = InstantOnboardingActivity.class.getSimpleName();
+  private static final String DCACCOUNT = "dcaccount";
+  private static final String DCLOGIN = "dclogin";
+  private static final String INSTANCES_URL = "https://delta.chat/chatmail";
+  private static final String DEF_CHATMAIL_HOST = "nine.testrun.org";
```
r10s, line 63: usually, we do not abbreviate "default" to "def" (which could also be read as "define", "defer", "declare" ...)

better stay explicit and change that to:

```
  private static final String DEFAULT_CHATMAIL_HOST = "nine.testrun.org";
```

src/org/thoughtcrime/securesms/InstantOnboardingActivity.java
```diff
@@ -0,0 +1,438 @@
...
+
+  private boolean avatarChanged;
+  private boolean imageLoaded;
+  private String providerHost;
+  private String providerQrData;
+  private boolean isDcLogin;
+
+  private AttachmentManager attachmentManager;
+  private Bitmap avatarBmp = null;
+  private ProgressDialog progressDialog = null;
```
r10s, line 81: by default, Java initialises instance variables to `null`  - we we do not set them elsewhere manually, we shouln't do here either:

```suggestion
  private Bitmap avatarBmp;
  private ProgressDialog progressDialog;
```

(did not test :)

r10s (approved):
yeah, code also lgtm, just some minors, no blockers.

i think, we should get that in soon to do a beta release, to get it tested by far more ppl

res/layout/instant_onboarding_activity.xml
```diff
@@ -0,0 +1,131 @@
...
+          android:layout_height="wrap_content"
+          app:layout_constraintStart_toStartOf="parent"
+          app:layout_constraintEnd_toStartOf="@+id/separator"
+          app:layout_constraintTop_toTopOf="parent"
+          app:layout_constraintBottom_toBottomOf="parent"
+          android:gravity="start"
+          android:padding="16dp"
+          android:text="@string/instant_onboarding_show_more_instances"
+          android:textColor="@color/delta_accent"
+          android:textSize="16sp"/>
```
r10s, line 98: i suggest to tune the buttons a bit own, using the color&font of the subtitle, this is esp. nicer with the keyboard visible as well and mimics more what desktop is doing (i know this is not visible as a button at the first glance then, but that is how things are nowadays - and in this case it is okay; the more the other things stand out - and by its position it is also clear it is tappable)

```suggestion
          android:textColor="@color/gray50"/>
```

res/layout/instant_onboarding_activity.xml
```diff
@@ -0,0 +1,131 @@
...
+          android:layout_height="wrap_content"
+          app:layout_constraintStart_toEndOf="@+id/separator"
+          app:layout_constraintEnd_toEndOf="parent"
+          app:layout_constraintTop_toTopOf="parent"
+          app:layout_constraintBottom_toBottomOf="parent"
+          android:gravity="end"
+          android:padding="16dp"
+          android:text="@string/qrscan_title"
+          android:textColor="@color/delta_accent"
+          android:textSize="16sp"/>
```
r10s, line 126:
```suggestion
          android:textColor="@color/gray50"/>
```

see above for discussion

res/layout/instant_onboarding_activity.xml
adbenitez: it was LinearLayout at the beginning, but had to switch to ConstraintLayout to make the labels wrap if they are too long, in LinearLayout they just push everything out of the screen if ex. the first item is too long, but maybe similar results can be achieved using LinearLayout + weights, this was just the first working solution I found when searching for this problem

r10s: i see, and i agree that these layouts are tricky sometimes and one is happy to find sth. working :) and with the tuned down layout, even the wrapping looks okay now

so, i am fine leaving it as a `ConstraintLayout`,  might anyways be, that this parts gets changes soon

but a <Space> might be better instead of empty text, semantic-wise, screen reader etc. , see suggestion below

res/layout/instant_onboarding_activity.xml
```diff
@@ -0,0 +1,129 @@
...
+          android:layout_height="wrap_content"
+          app:layout_constraintStart_toEndOf="@+id/other_options_button"
+          app:layout_constraintEnd_toStartOf="@+id/scan_qr_button"
+          app:layout_constraintTop_toTopOf="parent"
+          app:layout_constraintBottom_toBottomOf="parent"
+          android:paddingLeft="16dp"
+          android:paddingRight="16dp"
+          android:text=" "
+          android:textSize="16sp"
+          android:textColor="@color/gray50" />
```
r10s, line 111:
```suggestion
      <androidx.legacy.widget.Space
          android:id="@+id/separator"
          android:layout_width="wrap_content"
          android:layout_height="wrap_content"
          app:layout_constraintStart_toEndOf="@+id/other_options_button"
          app:layout_constraintEnd_toStartOf="@+id/scan_qr_button"
          app:layout_constraintTop_toTopOf="parent"
          app:layout_constraintBottom_toBottomOf="parent"
          android:paddingLeft="16dp"
          android:paddingRight="16dp" />
```

this should work the same, but is semantically more correct

adbenitez: (btw, the background around why it was a TextView is because at the beginning I tried to imitate desktop and put a dot separator)

PR #3031, opened 2024-05-07
fix: allow to share attachments from one chat to another

### LINKED ISSUE #2999: not possible to share media from one chat to another
it seems after https://github.com/deltachat/deltachat-android/pull/2976 was merged, now it is not possible to share a message with media (ex. an image) from one chat to another, that is,using the share option to share to the DC app itself

adbenitez:
close #2999 

the issue was introduced in #2975 

the new dependency [SafeContentResolver ](https://github.com/cketti/SafeContentResolver?tab=readme-ov-file) says:

> Replace all occurrences of ContentResolver.openInputStream() where URIs provided by other apps are opened with SafeContentResolver.openInputStream().
>
> SafeContentResolver will refuse to open file:// URIs pointing to files belonging to your app and content:// URIs belonging to content providers of your app.
If you wish to allow access to certain content providers, add the following <meta-data> element to the appropriate <provider> entries in your manifest

so not only `file://` are rejected, also `content://` URIs are rejected if they point to content of the Delta Chat app, but this doesn't consider the situation when you want to share content from Delta Chat not to an outside app but to the app itself to share to another chat

SafeContentResolver says we can allow some content providers:

> If you wish to allow access to certain content providers, add the following <meta-data> element to the appropriate <provider> entries in your manifest:
```
<provider … >
    <meta-data
        android:name="de.cketti.safecontentresolver.ALLOW_INTERNAL_ACCESS"
        android:value="true" />
</provider>
```

### DIFF
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -409,6 +409,11 @@
         android:grantUriPermissions="true"
         android:exported="false">
 
+      <!--
+          tell SafeContentResolver to allow incoming share-file Intent pointing to files from
+          this content resolver, it could be the user sharing a file from one chat to another.
+          See: https://github.com/deltachat/deltachat-android/pull/3031
+      -->
       <meta-data
           android:name="de.cketti.safecontentresolver.ALLOW_INTERNAL_ACCESS"
           android:value="true" />
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -408,6 +408,11 @@
         android:authorities="${applicationId}.attachments"
         android:grantUriPermissions="true"
         android:exported="false">
+
+      <meta-data
+          android:name="de.cketti.safecontentresolver.ALLOW_INTERNAL_ACCESS"
+          android:value="true" />
+
     </provider>
 
     <receiver android:name=".service.BootReceiver"


AndroidManifest.xml
```diff
@@ -408,6 +408,11 @@
         android:authorities="${applicationId}.attachments"
         android:grantUriPermissions="true"
         android:exported="false">
+
+      <meta-data
```
r10s, line 412: please add some comment - here or  in the PR pr commit description.

some summary from the chat yesterday, why is that needed, what happens without. but also: who consumes this ALLOW_INTERNAL_ACCESS? android? contentresolver? safecontentresolver?

i already forgot today what exactly happens :) -- let's be kind to the maintainers next years stumbling upon that code, maybe it's our future-self :)

r10s (commented):
lgtm, but a comment seems needed

AndroidManifest.xml
adbenitez: > some summary from the chat yesterday, why is that needed, what happens without. but also: who consumes this ALLOW_INTERNAL_ACCESS? android? contentresolver? safecontentresolver?

I don't know, my guess is that this is a hint for the safecontentresolver lib to recognize which providers are whitelisted and allow them

PR #3034, opened 2024-05-08
log if we receive onDeletedMessages()

r10s:
came over that, when trying to make sense why FCM onMessageReceived() is not called.

our current strategy is to just let the app run on received PUSH, however, it is useful to log this information for debugging

this is the corresponding code, btw ...

https://github.com/firebase/firebase-android-sdk/blob/2a75605db0cfac05c1634f911a0c1f7751599035/firebase-messaging/src/main/java/com/google/firebase/messaging/FirebaseMessagingService.java#L124

... i also tried to set a breakpoint in the handleIntent(), but even that is not called

### DIFF
--- a/src/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
+++ b/src/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
@@ -82,6 +82,12 @@ public class FcmReceiveService extends FirebaseMessagingService {
   }
 
   @Override
+  public void onDeletedMessages() {
+    Log.i(TAG, "FCM push notifications dropped");
+    // nothing special to to as we're running now and notifications should be processed as usual.
+  }
+
+  @Override
   public void onNewToken(@NonNull String rawToken) {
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
```diff
@@ -81,6 +81,12 @@ public void onMessageReceived(@NonNull RemoteMessage remoteMessage) {
     // we just let the app run as long as possible.
   }
 
+  @Override
+  public void onDeletedMessages() {
```
adbenitez, line 85: no call to super?

adbenitez (approved):

src/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
r10s: this is not needed, it is just used as an event, neither Signal for the [example code](https://firebase.google.com/docs/cloud-messaging/android/receive) calls super

PR #3036, opened 2024-05-08
clear quote if current draft has no quote

### LINKED ISSUE #3032: quote reply goes to wrong chat
With 1.44 android a quote reply can land in the wrong chat like this:
- in a 1:1 chat do a quote reply of a former message, maybe type some chars 
- tap top bar to go to "shared chats" and select any shared chat with that one contact 
- the quote-reply from the 1:1 chat will now appear in the selected group chat still referencing the message from the 1:1 chat

adbenitez:
close #3032


r10s (approved):
thanks for fixing!

i was looking in the core part before you started, and tried some additional precautions there as well, https://github.com/deltachat/deltachat-core-rust/pull/5543

PR #3039, opened 2024-05-09
use last known location if it is recent enough

### LINKED ISSUE #3037: Location streaming not working
it seems `DcContext.setLocation()` is not getting called while location streaming is enabled

adbenitez:
close #3037 

this also solves this issue: even if you already had location set with dcContext.setLocation(), a chat will only get location in messages when dcContext.setLocation is called again after the location streaming was enabled for the new chat, so you could have situations where you stream location for one chat in every sent message, but in another chat where you enabled it recently it would not work

### DIFF
--- a/src/org/thoughtcrime/securesms/geolocation/DcLocation.java
+++ b/src/org/thoughtcrime/securesms/geolocation/DcLocation.java
@@ -1,12 +1,16 @@
 package org.thoughtcrime.securesms.geolocation;
 
 import android.location.Location;
+import android.util.Log;
 
 import java.util.Observable;
 
 public class DcLocation extends Observable {
+    private static final String TAG = DcLocation.class.getSimpleName();
     private Location lastLocation;
     private static DcLocation instance;
+    private static final int TIMEOUT = 1000 * 15;
+    private static final int EARTH_RADIUS = 6371;
 
     private DcLocation() {
         lastLocation = getDefault();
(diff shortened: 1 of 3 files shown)


r10s (approved):
lgtm, did not test, though

r10s:
> this also solves this issue: even if you already had location set with dcContext.setLocation(), a chat will only get location in messages when dcContext.setLocation is called again after the location streaming was enabled for the new chat, so you could have situations where you stream location for one chat in every sent message, but in another chat where you enabled it recently it would not work

i do not fully understand that paragraph - if `dc_set_location()` does not send locations to all chats waiting for it - isn't that a core bug?

nb: `dc_set_location()` should be called for all accounts, and only if no account wants another location, location streaming should be stopped. we can also consider a `dc_accounts_set_self_location()`.

well, well, it is still experimental for reasons :)

adbenitez:
> i do not fully understand that paragraph - if `dc_set_location()` does not send locations to all chats waiting for it - isn't that a core bug?

it does the next time dc_set_location() is called AFTER location was enabled for the chat, what is happening is that if you are already streaming location for other chat, core will not reuse the already known location for the new chat, and will only consider the new calls to dc_set_location() after you enabled location streaming for that chat, yeah core could have been smarter about this and do what the UI is doing and use the last known location if it is recent enough, but you still would want to provide this to core anyway because the system one could be more updated



> nb: `dc_set_location()` should be called for all accounts, and only if no account wants another location, location streaming should be stopped. we can also consider a `dc_accounts_set_self_location()`.

indeed it makes sense this is a global thing for all accounts +1

r10s:
> we currently only support location streaming for the selected account!?

yes. and switching accounts probably even disables streaming for the previous account. but that is no new bug, it was always like that

but a fix should be simple, even without help of core:

- dc_is_sending_locations_to_chat(0) should be done in a loop for all accounts, if one is sending, we need streaming

- dc_set_location() should be called should be done in a loop for all accounts, if needs continuation, continue streaming

PR #3043, opened 2024-05-10
display vcard view type and allow to import/export contacts as vcard

### LINKED ISSUE #3004: Allow to share contacts as vCards
Tracking issue for sending a contact via Delta Chat, see: https://github.com/deltachat/deltachat-core-rust/issues/5422

when the core API is ready, it should be used by the client instead of just sending a raw address

adbenitez:
close #3004 

- [x] add RPC API for parsing vcard (using https://github.com/deltachat/deltachat-core-rust/pull/5536)
- [x] display in-chat
- [x] display in quotes/replies
- [x] display in draft/staging
- [x] set new core stock string "Contact" 
- [x] allow to share contact as vcard (using https://github.com/deltachat/deltachat-core-rust/pull/5571)
- [x] allow to import vcard (using https://github.com/deltachat/deltachat-core-rust/pull/5582)

preview: 
image


adbenitez:
this is already functional for displaying and also at least starting a chat with the email address in the vcard, it is already an improvement and could be merged and then the remaining items that need the core API for import/export can be added in another PR

adbenitez:
I incorporated https://github.com/deltachat/deltachat-core-rust/pull/5571 now when selecting a contact with the "attach contact" button, they are shared as vcard!

r10s (approved):
nice, lgtm and tested a bit - however, if sth happens, it happens out there, so let's get this in asap

PR #3044, opened 2024-05-11
add strings for new 'please wait' securejoin flow

r10s:
this PR adds the strings for the new 'please wait' securejoin flow.

background: when https://github.com/deltachat/deltachat-core-rust/pull/5550 gets merged, a one-to-one chat is read only the first seconds:

 - **either** until the protocol finishes. if both devices are line and connection is fine already after 1 or 2 seconds, this should be how things should usually go:<br>
 
 - **or**, on bad network or if devices are offline, we start with the same "please wait" message (we do not know yet that things will last) - after 15 seconds sending is allowed, but ppl will get a warning and can message already - if devices get online again or network gets better, also then in the "guaranteed e2ee" is shown. this looks as follows:

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -960,7 +960,7 @@
     <!-- placeholder will be replaced by the name of the inviter. -->
     <string name="secure_join_replies">%1$s replied, waiting for being added to the group…</string>
     <string name="secure_join_wait">Establishing guaranteed end-to-end encryption, please wait…</string>
-    <string name="secure_join_wait_timeout">Could not yet establish guaranteed end-to-end encryption, but you may already send a message</string>
+    <string name="secure_join_wait_timeout">Could not yet establish guaranteed end-to-end encryption, but you may already send a message.</string>
     <string name="contact_verified">%1$s introduced.</string>
     <string name="contact_not_verified">Cannot establish guaranteed end-to-end encryption with %1$s.</string>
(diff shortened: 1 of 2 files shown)


r10s:
once this gets merged, we need to set the strings also for iOS/desktop or file issues for that

res/values/strings.xml
```diff
@@ -959,6 +959,8 @@
     <string name="secure_join_started">%1$s invited you to join this group.\n\nWaiting for the device of %2$s to reply…</string>
     <!-- placeholder will be replaced by the name of the inviter. -->
     <string name="secure_join_replies">%1$s replied, waiting for being added to the group…</string>
+    <string name="secure_join_wait">Establishing guaranteed end-to-end encryption, please wait…</string>
+    <string name="secure_join_wait_timeout">Could not yet establish guaranteed end-to-end encryption, but you may already send a message</string>
```
iequidoo, line 963:
```suggestion
    <string name="secure_join_wait_timeout">Could not yet establish guaranteed end-to-end encryption, but you may already send a message.</string>
```

iequidoo (approved):

hpk42 (approved):
lgtm, eager to try :)

res/values/strings.xml
r10s: yeah, makes sense to stay with what we have for inner consistency :)

PR #3050, opened 2024-05-14
hide IMAP folder options on chatmail

r10s:
this PR makes use of the `is_chatmail` config introduced by https://github.com/deltachat/deltachat-core-rust/pull/5567 and hides the IMAP folder options.

~~for whatever reason, `is_chatmail` is currently always `0` (also in "View Log" which is generated by core directly),  might be things are not rolled out yet.~~ EDIT: works now, see below

this is what it will look like:

Screenshot 2024-05-14 at 23 19 00 &nbsp; Screenshot 2024-05-15 at 00 54 22 <br><i>left: chatmail=0, right: chatmail=1</i>

targets https://github.com/deltachat/deltachat-core-rust/issues/5380

when this PR gets merged, we need to file issues/PR for the same change on desktop/ios

### DIFF
--- a/src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
+++ b/src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
@@ -207,6 +207,13 @@ public class AdvancedPreferenceFragment extends ListSummaryPreferenceFragment
       }
       return true;
     }));
+
+    if (dcContext.isChatmail()) {
+      sentboxWatchCheckbox.setVisible(false);
+      bccSelfCheckbox.setVisible(false);
+      mvboxMoveCheckbox.setVisible(false);
+      onlyFetchMvboxCheckbox.setVisible(false);
+    }
   }
 
   @Override
--- a/src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
+++ b/src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
@@ -209,6 +209,8 @@ public class AdvancedPreferenceFragment extends ListSummaryPreferenceFragment
     }));
 
     if (dcContext.isChatmail()) {
+      preferE2eeCheckbox.setVisible(false);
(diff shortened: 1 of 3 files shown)


r10s:
great! works perfectly fine

adbenitez:
probably "prefer encryption" toggle should be locked/disabled in ON mode, to avoid people accidentally disabling encryption in chatmail servers

more importantly, the "import contacts from address book" should also be hidden

src/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
```diff
@@ -207,6 +207,13 @@ public void onCreate(Bundle paramBundle) {
       }
       return true;
     }));
+
+    if (dcContext.isChatmail()) {
+      sentboxWatchCheckbox.setVisible(false);
+      bccSelfCheckbox.setVisible(false);
```
adbenitez, line 213: if someone that is already using chatmail disabled this toggle, now they will not be able to enable it anymore, would be nice to do some migration checking for this property and maybe also the other ones, and set/force the correct default values

r10s: imu, whatever is chosen for these four options, it does not make a difference for chatmail. but not sure - otherwise, i would prefer some migration in the core, or that the settings are ignored, at least when they're harmful, i think, we do want to have less footguns for chatmail :) cc @link2xt

adbenitez: bcc self makes a difference without it multidevice sync is broken, or do you mean that internally core already ignores their value if chatmail flag is set?

r10s:
"prefer e2ee" and "read system address book" are hidden now as well, thanks for the pointer!

if the latter was set before, it is now ignored on chatmail - the other options are reset by https://github.com/deltachat/deltachat-core-rust/pull/5568

adbenitez (approved):

PR #3051, opened 2024-05-15
allow to show password on "account & password" editing

adbenitez:


r10s:
ftr, this PR aims to target issues where users once set up eg. gmail - but then forgot the password and failing to recover it otherwise, eg. recently at https://support.delta.chat/t/how-to-recover-gmail-password-from-phone/3067

in the past ([eg](https://github.com/deltachat/deltachat-android/pull/2859#issuecomment-1814876203)), we hesitated to show the password to avoid users shooting theirselves in the foot:

- some chatmail users will overlook "Add Second Device" - or stop searching as the pattern of address+password seems sufficient
- log in on a second device with address+password
- bam 💥 nothings works correctly as the  two devices have different private keys

as chatmail users will probably become the larger part and the issue is _really_ bad and hard to understand/fix for the user, this is really worth thinking about (nb: randomly noticed that [protonmail does not allow other private keys by clients for similar reasons](https://news.ycombinator.com/item?id=36639530))

to mitigate the footgun, we could add another warning (that no one reads :) or check isChatmail() before enabling the button.

however, maybe moving "accounts & password" to the lower part of "advanced", makes this footgun already unlikely enough :)

r10s (approved):
well, just had a little talk with @hpk42 let's accept the remaining footgun - if it turns out that ppl run into problems because of that, we can iterate

PR #3052, opened 2024-05-15
allow to share log instead of copying to clipboard

### LINKED ISSUE #3046: Support directly "sharing" a log to some chat
posting a copy-paste is a giant message in DC, and saving, and refinding is bothersome. 
If there were a way to directly share a log it'd be very helpful especially for the push-notification debugging now.

adbenitez:
close #3046

### DIFF
--- a/src/org/thoughtcrime/securesms/LogViewActivity.java
+++ b/src/org/thoughtcrime/securesms/LogViewActivity.java
@@ -1,18 +1,26 @@
 package org.thoughtcrime.securesms;
 
 import android.Manifest;
+import android.content.Intent;
+import android.net.Uri;
 import android.os.Bundle;
+import android.util.Log;
+import android.view.Menu;
+import android.view.MenuInflater;
+import android.view.MenuItem;
 
 import androidx.annotation.NonNull;
 import androidx.appcompat.app.AlertDialog;
 import androidx.fragment.app.FragmentTransaction;
-import android.view.Menu;
-import android.view.MenuInflater;
-import android.view.MenuItem;
-import android.widget.Toast;
 
 import org.thoughtcrime.securesms.permissions.Permissions;
-import org.thoughtcrime.securesms.util.Util;
+import org.thoughtcrime.securesms.util.FileProviderUtil;
+
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileWriter;
+import java.text.SimpleDateFormat;
+import java.util.Date;
 
 public class LogViewActivity extends BaseActionBarActivity {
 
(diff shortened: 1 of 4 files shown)


r10s:
does "share" on android guarantee to allow "copy to clipboard"? otherwise the latter may still be useful (i mean we offer both also elsewhere)

adbenitez:
my idea is to discourage sending a big text message in delta chat, which is extremely hard to read, it is better to share as text file, then you can always tap it and view the file in full screen in a more readable way

hpk42:
On Wed, May 15, 2024 at 05:20 -0700, Asiel Díaz Benítez wrote:
> my idea is to discourage sending a big text message in delta chat, which is extremely hard to read, it is better to share as text file, then you can always tap it and view the file in full screen in a more readable way

that's a separately useful consideration IMO.  

A share-button in the debug-log view is meant to support our real-world field testers.

Current: on android in debug log view: 

1. copy-paste 
2. go back
3. go back 
4. go back 
5. select chat
6. tap input field
7. long click and click "paste" 
8. send (but then you didn't provide any context) 

Suggested: 

1. hit "share"
2. select chat (or other app) 
3. be in draft message mode with log as attachment, add context like "slow connectivity in crowded situation" 
4. send

src/org/thoughtcrime/securesms/LogViewActivity.java
```diff
@@ -90,6 +97,28 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+  public void shareLog() {
+    try {
+      String timestamp = new SimpleDateFormat("yyyy-MM-dd-HH-mm").format(new Date());
+      File file = new File(getExternalCacheDir(), "log-" + timestamp + ".txt");
+      file.createNewFile();
+      file.setReadable(true, false);
+      FileWriter logFileWriter = new FileWriter(file, false);
+      BufferedWriter logFileBufferWriter = new BufferedWriter(logFileWriter);
+      logFileBufferWriter.write(logViewFragment.getLogText());
+      logFileBufferWriter.close();
```
r10s, line 109: there is already LogViewFragment::saveLogFile() doing similar things, let's see to use the same code, to get also the same filenames etc.

r10s (commented):
it does not work for me in the emulator, android13:


https://github.com/deltachat/deltachat-android/assets/9800740/f50e8ae2-f3be-4da1-be55-dc23b5ba2a68

adbenitez:
> it does not work for me in the emulator, android13

the "prevent own URI attack" thing strikes back again, we have to also allow this files provider similar to what we did for sharing attachments in https://github.com/deltachat/deltachat-android/pull/3031

r10s (commented):
nice, lgtm and works :)

PR #3053, opened 2024-05-15
change welcome headline

r10s:
just that this is not forgotten, cc @hpk42


hpk42 (approved):
matches current https://delta.chat and it's better to stay  on message. Why not just change the string directly if it's anyway a draft?

adbenitez (approved):

PR #3056, opened 2024-05-17
show overridden sender name also for self messages if present

adbenitez:
only after https://github.com/deltachat/deltachat-core-rust/pull/5585 is merged will this be visible, but this already can be merged without waiting  for it since it is just an internal change in DcMsg.getOverrideSenderName() that was not setting the overridden sender name for outgoing messages


r10s (approved):
nice, lgtm, thanks for one-to-one explanations, @adbenitez

PR #3059, opened 2024-05-17
change "three dots" option in "add reaction" view to ellipsis

### LINKED ISSUE #3047: "three dots" in reaction selector might wrap in small screens
image

adbenitez:
close #3047


r10s (approved):
hm, can we have sth. in between the bold and the normal?
and please a little more space between the point, which, btw, should be more squarish
and a little more grayish.

code otherwise lgtm, so feel free to do the requested changes in a subsequent PR

PR #3060, opened 2024-05-17
move 'manual log in' to 'create profile'

r10s:
after some internal discussion with @link2xt @hpk42 @adbenitez, "Manual Log In" still creates a client-side profile on the client, and therefore it makes some sense to move it to the first "Create New Profile" button.

this makes the "i already have a log in" pretty bulletproof, also, it seems less probable ppl will try to somehow manually login there. idea is also that ppl will less often "just so manually log in" - as the "create new profile" is sth else than "log in". but well, user testing will tell.

moreover, this PR takes the chance to move the "scan qr code" button to "other options" as well.

"explorer other options" is a bit more visible as before (where the text was smaller and gray), however, by its position still  pretty much tuned down; this is on purpose as the user really needs to know what to do or it is sth where we do not want to push users to.

however, we can also easily make the "explorer other options" a regular secondary button if we get feedback that it is too hard to discover.

in a subsequent PR we may refine the menu or so, but it seems already a good start if we still think it is a good idea tomorrow :)

Screenshot 2024-05-18 at 00 22 55
Screenshot 2024-05-18 at 13 26 46
Screenshot 2024-05-18 at 13 56 41

cc @kermoshina

### DIFF
--- a/res/layout/instant_onboarding_activity.xml
+++ b/res/layout/instant_onboarding_activity.xml
@@ -101,24 +101,12 @@
           android:layout_width="wrap_content"
           android:layout_height="wrap_content"
           app:layout_constraintStart_toEndOf="@+id/other_options_button"
-          app:layout_constraintEnd_toStartOf="@+id/scan_qr_button"
+          app:layout_constraintEnd_toEndOf="parent"
           app:layout_constraintTop_toTopOf="parent"
           app:layout_constraintBottom_toBottomOf="parent"
           android:paddingLeft="16dp"
           android:paddingRight="16dp" />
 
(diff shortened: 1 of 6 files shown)


r10s:
sleeping over it, i'd try at least how a regular secondary button called **"Use Other Server"** would look like. it would open the following options then:

- **List in Browser**
- **Classic Login**
- **Scan Invite Code**

r10s:
reworded some things after discussion with @hpk42 and @kermoshina ; screenshots updates

adbenitez (approved):

PR #3067, opened 2024-05-19
save avatar and name onPause() in InstantOnboardingActivity

### LINKED ISSUE #3063: name/avatar settings do not survive leaving/re-entering InstantOnboardingActivity
when settings a name + avatar and then selecting another server, the name is lost.

in addition to saving on hitting "create" button, we should set name/avatar unconditionally when leaving InstantOnboardingActivity and restore it on re-entering

adbenitez:
close #3063

### DIFF
--- a/src/org/thoughtcrime/securesms/InstantOnboardingActivity.java
+++ b/src/org/thoughtcrime/securesms/InstantOnboardingActivity.java
@@ -114,7 +114,7 @@ public class InstantOnboardingActivity extends BaseActionBarActivity implements 
     avatarChanged = false;
     registerForEvents();
     initializeResources();
-    initializeProfileAvatar();
+    initializeProfile();
     handleIntent();
     updateProvider();
   }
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/InstantOnboardingActivity.java
```diff
@@ -199,6 +199,24 @@ public void onStart() {
...
+
+    if (avatarChanged) {
+      try {
+        AvatarHelper.setSelfAvatar(InstantOnboardingActivity.this, avatarBmp);
+        Prefs.setProfileAvatarId(InstantOnboardingActivity.this, new SecureRandom().nextInt());
+        avatarChanged = false;
+      } catch (IOException e) {
+        Log.e(TAG, "Failed to save avatar", e);
+      }
+    }
```
r10s, line 217: one could factor out these lines as they're similar to createProfile(), however, due to different threads and additional ongoings in createProfile()  it is probably not worth the effort

r10s (approved):
nice! thanks a lot!

PR #3073, opened 2024-05-20
improve saved messages

### LINKED ISSUE #3573: adapt to new saved-messages api
since https://github.com/deltachat/deltachat-core-rust/pull/5606 we have a new saved-messages-api; to benefit from that on android, the following steps are needed:

- [x] make [`dc_save_msgs()`](https://c.delta.chat/classdc__context__t.html#ac8075b06088ddbe39cd8cc3802141102), [`dc_msg_get_original_msg_id()`](https://c.delta.chat/classdc__msg__t.html#a3d3fdca5164b38d93e1f42f55ca67f72) and [`dc_msg_get_saved_msg_id()`](https://c.delta.chat/classdc__msg__t.html#ab13b251a52cf6b36627dff46f3c0be3e) available in java

- [x] when long tapping a regular message (not info [^1]), **show save/unsave menu option**. in case of android, this should be a ★ in the action bar, directly accessible (not the overflow menu). in self-chat, the "unsave" button should be shown unconditionally, so also for message saved otherwise. an unsave-confirmation is only needed if there is no original message

- [x] for every message shown in a chat, check if it was saved. if so, show a little ★ beside the date ([screenshot](https://github.com/deltachat/deltachat-ios/pull/2527))

- [x] in self-chat, for incoming messages, **show group avatar and sender name**, this is a simple one-line-change as drafted already at https://github.com/deltachat/deltachat-android/pull/3073/commits/75e9984959004e632737cd1e0635ff0593789dae

- [x] add a button to go to the original beside each saved message in the self-chat ([screenshot](https://github.com/deltachat/deltachat-ios/pull/2527))

[^1]: at least for the first iteration, we do not want to show saved info messages and thelike in the "Saved Messages" chat; this turned out to be confusing and might also require adaptions on systems other than android (remember: saved messages are synced), cmp. https://github.com/deltachat/deltachat-ios/issues/2567

r10s:
close #3573

to make place for the "Save" icon, the "Info" icon, which is anyway more a debug view currently, is moved to the overflow menu.

currently, only one selected message can be saved per action, mutli-select-mode can be done if needed in a subsequent PR

what is missing and should go to a subsequent PR, is a button to go  from the "saved message" to the "original message"; my android-layout-constraints skills are too unused to target that in reasonable time :)


r10s:
closing, this can serve as a template, acutal prototyping was done on iOS at https://github.com/deltachat/deltachat-ios/pull/2527

r10s:
/me adapted to new api, updated pr description

adbenitez:
I am testing the PR a bit, some comments at a first glance:

* appending the star as text to the footer label is not good, localization-wise but also it looks weird for outgoing messages, it should probably be at the beginning for outgoing messages, not mixed between data label and delivery checkmark
* normally sent messages in "saved messages" chat don't have the star, that is ok, but then when selecting the message the star button is not hidden for those, instead it is available and if you click it you get an option to remove the message which is basically the same duplicated action of the trash icon besides, I think it is better to just hide the option for messages directly sent in saved messages
* the previous point also applies to messages that its "parent"/original message was deleted, I think it is better to just show the delete trash can there and not show the "un save" button, because it is a bit confusing, if saving/unsaving sometimes show the same dialog than the trash can I can already see the problems of people accidentally deleting messages with the trash can when wanting to un-star etc
* I find the "save/unsave" terms a bit confusing wrt meaning, saving sounds more like the "export attachments" option, probably "bookmark" or "star" would have fit better, but maybe it is too late for renamings

adbenitez:
For other testers: if you are confused why saved messages doesn't have a padlock, it was a bug in core, it was fixed and it will show a padlock when we upgrade to core 1.155.2 or superior

r10s:
- star/bookmark was considered (among tons of other things) but there the expectation is more, that star/bookmark gets deleter when original is deleted. what we'de doing js actually saving. but, yes, that discussion is too late :)

- i agree that a dedicated view is better than a concatenated string, feel free to do a subsequent pr, that viewstuff is above my paygrade :)

- for hiding unsave for messages added directly: on iOS, we had it like that before, but it was confusing. but the menu js different there (delete a tap more far away) - so, i am fine with doing it differently on android. consider, however, ppl do not really know how the ine kr the other message was saved, and do not undersatnd why "unsave" is sometimes there and sometimes not.  we can also consider to remove "unsave" completely in "saved messages" and let ppl just hit "delete" - that would skip that uncertainity.
but i would leave that change also for a later PR

adbenitez (approved):

PR #3079, opened 2024-05-21
tweak menu options in QR show/scanning activities

### LINKED ISSUE #3076: too many menu options in group's invite code
since #3064, the group's invite code shows the options "add contact manually", "paste", "load as image", which do not make sense there.

when on that, in the normal scanner, we should remove  "copy" and "deactivate" again, there was already a hint in some chat.

the other options are mostly fine and okay: share button is a bit overall - and "paste, "manuall", "load as image" still makes sense in the overall context of "Add Contact"

adbenitez:
close #3076

### DIFF
--- a/src/org/thoughtcrime/securesms/qr/QrActivity.java
+++ b/src/org/thoughtcrime/securesms/qr/QrActivity.java
@@ -113,6 +113,13 @@ public class QrActivity extends BaseActionBarActivity {
     public boolean onPrepareOptionsMenu(Menu menu) {
       menu.clear();
       getMenuInflater().inflate(R.menu.qr_show, menu);
+      if(tabLayout.getSelectedTabPosition() == TAB_SHOW) {
+        menu.findItem(R.id.withdraw).setVisible(true);
+        menu.findItem(R.id.copy).setVisible(true);
+      } else {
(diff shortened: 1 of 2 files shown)


src/org/thoughtcrime/securesms/qr/QrActivity.java
```diff
@@ -113,6 +113,13 @@ private void checkPermissions(int position, ProfilePagerAdapter adapter, ViewPag
     public boolean onPrepareOptionsMenu(Menu menu) {
       menu.clear();
       getMenuInflater().inflate(R.menu.qr_show, menu);
+      if(tabLayout.getSelectedTabPosition() == TAB_SHOW) {
+        menu.findItem(R.id.withdraw).setVisible(true);
+        menu.findItem(R.id.copy).setVisible(true);
+      } else {
```
r10s, line 119: with `inflate()` we just added all items visible; therefore, we only need to hide them, not to make them visible again.

```suggestion
      if(tabLayout.getSelectedTabPosition() == TAB_SCAN) {
```

r10s (approved):
a minor optimisation, otherwise lgtm.

thanks a lot!

PR #3084, opened 2024-05-22
allow to delete the selected account

r10s:
not allowing to delete the selected account forces users to switch to another account after reviewing that an account can be deleted - then, in the list, they have to pick the correct account to delete, which can be challenging and is error-prone.

it is better to just allow deletion of the selected account (i think it was introduced for technical reasons)

todo:

- [x] show the delete account button unconditionally
- [ ] handle last account deletion
- [ ] update chatlist after account deletion

no priority :)


r10s:
closing stale PR for now

PR #3097, opened 2024-05-24
auto-generated RPC bindings

adbenitez:
- [x] add new auto-generated `chat.delta` package with RPC bindings
- [x] remove old manually generated rpc and util package
- [x] use the new auto-generated Rpc API
- [x] remove gson library from dependencies


r10s:
thanks for making this possible! this will make things much easier in the future!

however, for this release, this is a too large and massive change for 1.46.0 to 1.46.1+ - even if most things are auto-generated, i do not want to risk getting issues by that or dive into this topic completely, which is needed for review.

therefore, for the two missing jsonrpc for the peer-to-peer, we should add them by hand as before, i think, they do not even have parameters, so that should be easy.

when 1.46 is finally in all stores and things are settled so far we can finally go for this PR

adbenitez:
> Where is the generator itself?

it is closed source with a paid license subscription

adbenitez:
> Where is the generator itself?

published it here: https://github.com/deltachat/dcrpcgen

PR #3107, opened 2024-05-28
feat: add stock strings for realtime

Septias:
Needed for https://github.com/deltachat/deltachat-desktop/pull/3741

### DIFF
--- a/res/values/strings.xml
+++ b/res/values/strings.xml
@@ -430,6 +430,9 @@
     <string name="send_file_to">Send \"%1$s\" to…</string>
     <!-- title shown above a list contacts where one should be selected (eg. when a webxdc attempts to send a message to a chat) -->
     <string name="send_message_to">Send Message to…</string>
+    <!-- Options in config menu -->
+    <string name="enable_realtime">Enable Realtime APIs</string>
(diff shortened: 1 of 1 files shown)


res/values/strings.xml
```diff
@@ -430,6 +430,9 @@
     <string name="send_file_to">Send \"%1$s\" to…</string>
     <!-- title shown above a list contacts where one should be selected (eg. when a webxdc attempts to send a message to a chat) -->
     <string name="send_message_to">Send Message to…</string>
+    <!-- option in config menu -->
+    <string name="enable_realtime">Enable Realtime APIs</string>
+    <string name="enable_realtime_explain"> Enable Webxdc realtime APIs which create direct connection between devices.</string>
```
r10s, line 435: - remove leading space
- remove misleading comment about "config menu" which is not true even for desktop
- remove "enable" from name - this is clear by the user of a switch/checkbox and also not used elsewhere
- name the option more from the view of the user  (mentioning one time APIs is fine)

```suggestion
    <string name="enable_realtime">Realtime Webxdc Channels</string>
    <string name="enable_realtime_explain">Enable Webxdc realtime APIs to create direct connections between devices.</string>
```

r10s (approved):

PR #3110, opened 2024-05-31
call setConfigFromQR() from background thread

r10s:
setConfigFromQR() involves network and should not block the UI.

(/me in deutsche bahn :)

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/InstantOnboardingActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/InstantOnboardingActivity.java
@@ -50,6 +50,7 @@ import org.thoughtcrime.securesms.profiles.ProfileMediaConstraints;
 import org.thoughtcrime.securesms.qr.RegistrationQrActivity;
 import org.thoughtcrime.securesms.scribbles.ScribbleActivity;
 import org.thoughtcrime.securesms.util.Prefs;
+import org.thoughtcrime.securesms.util.Util;
 import org.thoughtcrime.securesms.util.views.ProgressDialog;
 
 import java.io.File;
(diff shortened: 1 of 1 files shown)


src/org/thoughtcrime/securesms/InstantOnboardingActivity.java
```diff
@@ -467,12 +468,16 @@ private void startQrAccountCreation(String qrCode)
 
     DcHelper.getEventCenter(this).captureNextError();
 
-    if (!dcContext.setConfigFromQr(qrCode)) {
-      progressError(dcContext.getLastError());
-      return;
-    }
-    DcHelper.getAccounts(this).stopIo();
-    dcContext.configure();
+    new Thread(() -> {
```
adbenitez, line 471: I think it is better to use `Util.runOnBackground()` or  `Util.runOnAnyBackgroundThread()`

r10s: i am always unsure about these options. we're using the java-nly approach also ar other places.

which one is better? and why?

adbenitez: I am not sure, but since this was created in utils and used in other places in a similar way (runOnBackground + inside runOnMainThread() at some point) I guess is better to use the same approach everywhere, in any case what runOnBackground does can then be changed according to what is "better" unlike manually creating the thread and calling start()

adbenitez (approved):

src/org/thoughtcrime/securesms/InstantOnboardingActivity.java
r10s: ftr, `runOnBackground()` uses `AsyncTask.THREAD_POOL_EXECUTOR` which is marked as deprecated ("Using a single thread pool for a general purpose results in suboptimal behavior for different tasks [...] Use or create an Executor configured for your use case" - i think, real differences do only pop up when really using threads massively, in our case it probably just does not matter.

PR #3116, opened 2024-06-02
tweak asking to disable battery optimisations

### LINKED ISSUE #3114: on second chatmail account, question about "disable battery optimisations" appears
when using only chatmail accounts, the question about "disable battery optimisations" is not needed.

however, sometimes it pops up even then.

one way to reproduce:

- install new Delta Chat
- onboard with default nine.testrun-profile
- "Add Profile"
- again  onboard with default nine.testrun-profile

-> question about "disable battery optimisations" is shown.

it is important to have a fresh install so that the system/deltachat does not block the question by other reasons.

r10s:
only ask to disable battery optimisations,
if we're actually connected to the server that reports non-chatmail - otherwise the information is not yet there.

needs to be tested on a real device.

closes #3114

### DIFF
--- a/src/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
+++ b/src/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
@@ -110,18 +110,24 @@ public class DozeReminder {
     }
   }
 
-  private static boolean isAllChatmail() {
+  private static boolean isPushAvailableAndSufficient() {
+    if (FcmReceiveService.getToken() == null) {
+      return false;
+    }
+
     for (int accountId : ApplicationContext.dcAccounts.getAll()) {
       DcContext context = ApplicationContext.dcAccounts.getAccount(accountId);
       if (!context.isChatmail()) {
-        return false;
+        // only report "not sufficient" if we're actually connected -
+        // otherwise we'll recheck next time the chatlist is opened
+        int connectivity = context.getConnectivity();
+        if (connectivity >= DcContext.DC_CONNECTIVITY_WORKING) {
+          return false;
+        }
       }
     }
-    return true;
-  }
 
-  private static boolean isPushAvailableAndSufficient() {
-    return isAllChatmail() && FcmReceiveService.getToken() != null;
+    return true;
   }
 
   public static void maybeAskDirectly(Context context) {


src/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
```diff
@@ -110,18 +110,24 @@ public static void dozeReminderTapped(Context context) {
...
+      return false;
+    }
+
     for (int accountId : ApplicationContext.dcAccounts.getAll()) {
       DcContext context = ApplicationContext.dcAccounts.getAccount(accountId);
       if (!context.isChatmail()) {
-        return false;
+        // only report "not sufficient" if we're actually connected -
+        // otherwise we'll recheck next time the chatlist is opened
+        int connectivity = context.getConnectivity();
```
iequidoo, line 123: So, this `isChatmail()` API is really unclear. Maybe the core should just return `true` until it really knows that? Btw, `isChatmail()` affects not only disabling battery optimisation, but also hiding some settings like "move to mvbox". Maybe these settings also should be hidden until the core connects to the server?

iequidoo: Yes, but `Context::configure()` doesn't call `inbox_fetch_idle()`. So there's no guarantee that when the user enters Settings, we already know if we a chatmail client. User can do that being offline. Maybe better return `true` until we know?

iequidoo: Then `Config::IsChatmail` needs to be set in this block: https://github.com/deltachat/deltachat-core-rust/blob/90c30879b182824f1d823db73bc76e54dc0c4f91/src/configure.rs#L461

iequidoo: I'm just trying to understand why `Config::IsChatmail` is set in `inbox_fetch_idle()`, but other chatmail-related config keys here: https://github.com/deltachat/deltachat-core-rust/blob/90c30879b182824f1d823db73bc76e54dc0c4f91/src/configure.rs#L461

iequidoo: So, it's guaranteed that `inbox_fetch_idle()` is called by the time `Context::configure()`, and `Config::IsChatmail` is set properly? Then i don't understand why the subject fix needs to look into the connectivity state

iequidoo: Ok, not sure my understanding is correct, but anyway i've created a PR for this: https://github.com/deltachat/deltachat-core-rust/pull/5671

r10s:
so, this pr can be closed as core provides the information now soon enough, right?

iequidoo:
> so, this pr can be closed as core provides the information now soon enough, right?

I think that yes. Now the core should always report the actual value of `Config::IsChatmail` for configured accounts

r10s:
> It's not in the released core yet

np, it is fine if it gets to the next core. the issue is a minor one and will be fixed when we do a 1.46.x then

closing this PR then, thanks a lot for fixing it in core! <3

PR #3122, opened 2024-06-10
include firebase only in gplay flavour

### LINKED ISSUE #3030: adapt build files so that it skips FCM for unsupported flows
FCM needs help of server and firebase-console to work; for default/unknown flavours, this is not the case - if possible, we should change the build files so that it links against FCM only if application_id is known and makes use of FCM

this would be `chat.delta` (gplay etc) and `com.b44t.messenger.beta` (beta we send around) currently

esp. for `com.b44t.messenger`, as used by f-droid, we should avoid getting FCM in, but also for other unknown application_ids as deltalab

(i personally do not know much about build files on android, so if someone already knows _something_, step ahead :)

adbenitez:
- [x] only include firebase dep in gplay flavor
- [x] include firebase-related settings in AndroidManifest only for gplay flavor
- [x] create fake do-nothing FcmReceiveService implementation for fat flavor
- [x] disable push notifications toggle in setting if feature is not available in the built apk flavor

close #3030


.gitignore
```diff
@@ -5,7 +5,6 @@ project.properties
 .settings
 bin/
 gen/
-gplay/
 .idea/
```
r10s, line 8: excluding the top-level gplay directory is still needed once one is doing a release, this directory is created and get filled with artefacts

however, it should be `/gplay/` to really only exclude the top-level directory and not `/src/gplay/`

```suggestion
/gplay/
.idea/
```

r10s (approved):
yeah, very nice!

also that we get a bit more standard wrt project structure 👌

my experiments yesterday in #3123 came to roughly the same code, that's a good sign - but your's more complete wrt run-timeflag, disabled option etc. (wondering btw why the AndroidManifest.xml worked for me without changes - maybe some caching)

one small comment wrt .gitignore, otherwise lgtm

PR #3124, opened 2024-06-12
cancel muting should not exit select-mode

r10s:
exiting 'select mode' is unexpected when hitting 'cancel' - the user may still want to do other things with the selection.

this makes the 'mute' behavior consisten with other cases showing a dialog, eg. 'delete'


adbenitez (approved):
good improvement, didn't test

r10s:
yeah, came over that while searching for a "Mark All As Read" :)

let's wait for merging until #3122 is merged ... rebasing this PR seems easier :)

adbenitez:
it seems no rebasing is needed since the file was just moved but there is no content conflict :thinking:

r10s:
github said there's a merge conflict, rebase then did not need interaction .... some say so and some say so :)

PR #3127, opened 2024-06-12
organize project structure

adbenitez:
- [x] move res and assets folders to src/main/
- [x] update scripts and transifex config to use the new paths


.tx/config
```diff
@@ -3,8 +3,8 @@ host     = https://www.transifex.com
 lang_map = id: in, ja_JP: ja, nl_NL: nl, pt_BR: pt-rBR, zh_CN: zh-rCN, zh_TW: zh-rTW
 
 [o:delta-chat:p:delta-chat-app:r:stringsxml]
-file_filter = res/values-<lang>/strings.xml
-source_file = res/values/strings.xml
+file_filter = src/main/res/values-<lang>/strings.xml
+source_file = src/main/res/values/strings.xml
```
adbenitez, line 7: I am not sure if this will make the tx program to output to src/main/res instead of ./res, this needs to be tested

r10s:
these are the defaults for android projects, right?

.tx/config
r10s: should be fine, we will find out easily sooner or later :)

r10s:
but is it really default to put `res` into `src` dir? if so, then it's that, just wondering.

apart from begin default, what is the advantage of the now longer paths? that we can more easily adapt for flavours in the already know way? are there other advantages?

r10s (approved):
some questions, but overall probably just fine!

thanks for the chore!

r10s:
> just because of interest, iirc, it took me some effort years ago to go for an alternative app-icon that includes the word "DEV" ...

ftr, that seems to use the same approach as now everywhere, see documentation in `src/debug/REAME.md` -  i think, we can then delete that file alltogether, this is now obvious and standard

adbenitez:
> but is it really default to put `res` into `src` dir? if so, then it's that, just wondering.

yes, in fact that is what the debug folder is already doing to override the app icon, it might sound confusing at first because you might thing of src as code, but it is more like all your app sources and content, and it is more clear when you see the flavor folders inside, then each flavor can have its own res, assets, AndroidManifest and code

you can see a common project structure creating a new project in Android Studio, the common thing is to wrap the src folder inside an app folder but that seemed to add more complexity, since you then need to have a top level build.gradle plus another one inside app folder etc

adbenitez:
> apart from begin default, what is the advantage of the now longer paths? that we can more easily adapt for flavours in the already know way? are there other advantages?

not much, other than being in the standard place were other android developers would expect, and in the same level as alternative res/assets for other flavors 
also to reduce the clutter in the root of the project, less folders and files from the app source spread together with config files etc

PR #3130, opened 2024-06-14
allow to mark as read selected chats

### LINKED ISSUE #3128: add "Mark All As Read"
we have a "Mark All As Read" for the archive, but not for the normal chatlist. would be nice to have, seems easy to add (also in maintenance mode ...), and that was also requested several times.

while marking all chats are read is definitely an advanced feature where it'd be okay if it is discovered only when searching for it, the three-dot menu atop of the chatlist seems a natural place for that, i'd just add it there (alternative might be "long tab -> select all -> mark as read"- but that seems too complicated)

additional notes:

- archived chats shall probably being marked as read at the same time - when the user uses the option, they are a bit in a "leave me alone" mood, not taking care in details. archived chats are shown as being muted, so it would be weird to exclude them from the action and force the user to go to the archive and hit "mark all as read" there again. also it makes sense if all badge counters disappear

- we're going for maintenance mode, this issue has low priority

- when this issue is closed, we can file corresponding issues for desktop (right tap on profile icon) and iOS (not sure, maybe long tap on "Chats" icon)

nb: we can consider removing "Switch Account" from the menu at the same time, it is there mainly to make the move to the account switcher via the avatar easier (which is the same on all platforms) and comes from a time where we did not have the avatar there at all.

adbenitez:
close #3128 

user can select certain chats they want to mark as seen, they still can mark as sen all by using "select all" and then use "mark all as read"

### DIFF
--- a/src/main/res/menu/conversation_list.xml
+++ b/src/main/res/menu/conversation_list.xml
@@ -21,6 +21,10 @@
           android:id="@+id/menu_mute_selected"
           app:showAsAction="never"/>
 
+    <item android:title="@string/mark_all_as_read"
+          android:id="@+id/menu_marknoticed_selected"
+          app:showAsAction="never"/>
+
     <item android:title="@string/menu_select_all"
           android:id="@+id/menu_select_all"
           app:showAsAction="never"/>
--- a/src/main/res/menu/conversation_list.xml
+++ b/src/main/res/menu/conversation_list.xml
@@ -21,7 +21,7 @@
           android:id="@+id/menu_mute_selected"
           app:showAsAction="never"/>
 
-    <item android:title="@string/mark_all_as_read"
+    <item android:title="@string/mark_as_read"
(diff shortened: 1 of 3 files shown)


src/main/res/menu/conversation_list.xml
```diff
@@ -21,6 +21,10 @@
           android:id="@+id/menu_mute_selected"
           app:showAsAction="never"/>
 
+    <item android:title="@string/mark_all_as_read"
```
adbenitez, line 24: I reused "mark all as read" string, because it fits for "mark all (selected chats) as read" but we could also add a new "mark as read" string

r10s (approved):
let's go for a dedicated string, otherwise lgtm, and also tested

src/main/res/menu/conversation_list.xml
adbenitez: done

PR #3131, opened 2024-06-14
only show debug entries if developer mode is enabled

### LINKED ISSUE #3106: remove DEBUG entries from log
with the now [colored debug](https://github.com/deltachat/deltachat-android/pull/3087) log, the _green_ DEBUG entries are easily recognizable - and i do not think, there is any worth in it for out typical debugging of network problems, this level is not even emitted by core/us but only by other libraries.

it is probably just noise in most, if not all cases. and it may result in INFO/WARNING/ERROR being added less for older entries.

i do not think, we need a switch, i would consider a switch only if we really have an issue where that would be helpful.

adbenitez:
close #3106

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/LogViewFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/LogViewFragment.java
@@ -127,21 +127,27 @@ public class LogViewFragment extends Fragment {
     return logFile;
   }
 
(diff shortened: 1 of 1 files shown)


r10s (approved):
lgtm. nice, pragmatic  way to get alone with no additional regex calls

PR #3132, opened 2024-06-14
rename fat flavor to foss

adbenitez:


r10s (approved):
lgtm, if sth. is missing, we can do that in subsequent pr

PR #3135, opened 2024-06-15
don't go back to welcome screen when pressing back after instant onboarding

### LINKED ISSUE #3119: "back" button behaviour after creating/adding profiles
when a profile was added, one ends up in the chatlist, which is fine.

however, when then hitting the "back" button, the app should just exit (as it does otherwise as well)

this is esp. bad for the very first profile created, on very first app start ppl land again in "add profile" unexpectedly.

not a blocker or so, but should still be fixed soon

tested on android7 - might be different in other android versions, this stacking of activities is tricky sometimes when supporting a broad range of android versions :)

adbenitez:
close #3119


r10s (approved):
perfect!

what a nice PR!

r10s:
(but i also tested a bit ;)

PR #3143, opened 2024-07-03
allow to enable/disable notifications per-account

### LINKED ISSUE #3126: make “Notifications on/off” a per-profile setting
similar to desktop, it makes sense also on android to disable notifications for some accounts. we got real-world user request for that several times; and in the team this was also considered useful :)

UI-wise, no new options are needed. as a minimum, make “Settings / Notifications / On/Off” a per-profile switch.

optionally on first iteration, the state can be shown in the “profile switcher” as well (similar to desktop). however, if we do so, we should also offer a way to change the setting directly there, eg. on long tap (for a permanent button cluttering UI, the option is usually not changed frequently enough)

as this is only a minor change, this should align well with the "maintenance" approach. the issue has low priority, however.

sidenotes:

- we should regard this option similar to “Mute Chat”

- similar to other “Mute chat” settings, “Mute profile” can be synced across devices in the future (if one really wants to have one device silent and the other not, it is better to mute on device level) to prepare this future, it makes sense to introduce a core-config-setting (that then should be used by desktop/iOS as well)

- this issue is _not_ about disable syncing; also for disabled notifications we should update in background

- for other notification settings (sounds, vibrate, led. etc): in the first iteration, that can be unchanged and stay a global setting. it would be nice to have that as per-profile settings as well (esp. as we even have per-chat sounds etc.), however, it should not be synced across devices (but can be stored in ui.*-config so that it goes to the backup)

- this is also in-line with most other settings that are only exceptionally considered global - at some point we can also double check if more settings can become per-profile and go to ui.*-config

- when this is done, we should file issues for iOS and desktop (to use the core-config then)

adbenitez:
close #3126 

image

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -287,7 +287,10 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
     int skipId = dcAccounts.getSelectedAccount().getAccountId();
     for (int accountId : dcAccounts.getAll()) {
       if (accountId != skipId) {
-        unreadCount += dcAccounts.getAccount(accountId).getFreshMsgs().length;
+        DcContext dcContext = dcAccounts.getAccount(accountId);
+        if (!dcContext.isMuted()) {
+          unreadCount += dcContext.getFreshMsgs().length;
+        }
       }
     }
 
--- a/src/main/java/org/thoughtcrime/securesms/LogViewFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/LogViewFragment.java
@@ -218,7 +218,7 @@ public class LogViewFragment extends Fragment {
     PowerManager powerManager = (PowerManager)context.getSystemService(Context.POWER_SERVICE);
     final PackageManager pm      = context.getPackageManager();
     final StringBuilder  builder = new StringBuilder();
-
+    final DcContext dcContext = DcHelper.getContext(context);
 
     builder.append("device=")
            .append(Build.MANUFACTURER).append(" ")
@@ -251,8 +251,7 @@ public class LogViewFragment extends Fragment {
         builder.append("ignoreBatteryOptimizations=").append(
             powerManager.isIgnoringBatteryOptimizations(context.getPackageName())).append("\n");
       }
-      builder.append("notifications=").append(
-              Prefs.isNotificationsEnabled(context)).append("\n");
+      builder.append("notifications=").append(!dcContext.isMuted()).append("\n");
       builder.append("reliableService=").append(
               Prefs.reliableService(context)).append("\n");
 
@@ -271,7 +270,6 @@ public class LogViewFragment extends Fragment {
     }
 
     builder.append("\n");
-    DcContext dcContext = DcHelper.getContext(context);
     builder.append(dcContext.getInfo());
 
     return builder.toString();
--- a/src/main/java/org/thoughtcrime/securesms/connect/KeepAliveService.java
+++ b/src/main/java/org/thoughtcrime/securesms/connect/KeepAliveService.java
@@ -29,7 +29,7 @@ public class KeepAliveService extends Service {
         // note, that unfortunately, the check for isIgnoringBatteryOptimizations() is not sufficient,
         // this checks only stock-android settings, several os have additional "optimizers" that ignore this setting.
         // therefore, the most reliable way to not get killed is a permanent-foreground-notification.
-        if (Prefs.isNotificationsEnabled(context) && Prefs.reliableService(context))  {
+        if (Prefs.reliableService(context))  {
             startSelf(context);
         }
     }
--- a/src/main/java/com/b44t/messenger/DcContext.java
+++ b/src/main/java/com/b44t/messenger/DcContext.java
@@ -236,6 +236,14 @@ public class DcContext {
       return getConfigInt("is_chatmail") == 1;
     }
 
+    public boolean isMuted() {
+      return getConfigInt("ui.muted") == 1;
+    }
+
+    public void setMuted(boolean muted) {
+      setConfigInt("ui.muted", muted? 1 : 0);
+    }
+
     /**
      * @return true if at least one chat has location streaming enabled
      */
--- a/src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListAdapter.java
+++ b/src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListAdapter.java
@@ -39,7 +39,7 @@ public class AccountSelectionListAdapter extends RecyclerView.Adapter
       super(itemView);
     }
 
-    public abstract void bind(@NonNull GlideRequests glideRequests, int accountId, DcContact self, String name, String addr, int unreadCount, boolean selected);
+    public abstract void bind(@NonNull GlideRequests glideRequests, int accountId, DcContact self, String name, String addr, int unreadCount, boolean selected, boolean isMuted);
     public abstract void unbind(@NonNull GlideRequests glideRequests);
   }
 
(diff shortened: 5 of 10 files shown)


adbenitez:
currently ui.muted config is used, we need a proper core config for this

r10s (approved):
android part lgtm. **but let's wait for core:**

i agree, we should use a core settings instead of `ui.muted` - we then have the option to do $things on the option, eg. not doing FCM (which may or may not be wanted, no need to decide that now).  
also, it is in sync with muted-chats, which are also handled in core - that way, we have the option to not fire fresh-events and do the notification-logic more in core (again, this may or may not be wanted, no need to decide that now, it is just about the option)

r10s:
i created a PR offering `is_muted` at https://github.com/deltachat/deltachat-core-rust/pull/5759 

btw, i think we should migrate the old setting, so that we do not start adding notifications the user has disabled before. we could do that similar to "migrating chat backgrounds" (which can be removed then)

src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
```diff
@@ -287,7 +287,10 @@ public void refreshUnreadIndicator() {
     int skipId = dcAccounts.getSelectedAccount().getAccountId();
     for (int accountId : dcAccounts.getAll()) {
       if (accountId != skipId) {
-        unreadCount += dcAccounts.getAccount(accountId).getFreshMsgs().length;
+        DcContext dcContext = dcAccounts.getAccount(accountId);
+        if (!dcContext.isMuted()) {
+          unreadCount += dcContext.getFreshMsgs().length;
+        }
```
r10s, line 293: this is about the badge counter in the upper left corner right?

i think, it is not correct to leave muted contexts out, see comments in the core-pr.

we also do not hide these counters for muted-chats.

instead, we should grey-out the counter if there are _only_ muted messages.

for the global badge-counter or unread-dot on the app-icon, i am unsure, however :)

Simon-Laux: > for the global badge-counter or unread-dot on the app-icon

I vote for exclude, because muted chats are also excluded. (at least on desktop and in every messaging app I know that supports muting chats).

I understand and agree it makes some sense to have an indicator of "there are new messages in muted" accounts in the app, though.

adbenitez: > we also do not hide these counters for muted-chats.

what??? I think we already do, it is done automatically by core, muted chats don't count in the unread messages count of an account 

similarly it feels natural that if you mute an account you don't want it to keep making noise in your badge counter and then you go check and it turned out it was just a muted account, if you want them to appear there then just don't mute the account

adbenitez: > instead, we should grey-out the counter if there are _only_ muted messages.

mmmh it seems there is a misunderstanding, the account badge is already shown in gray for muted accounts, what these lines of code do is gather all the badge counts of all accounts and aum it but muted accounts are left out, trying to be smart here and show the badge counter in gray if all the messages are from muted accounts it is tricky, since you can have mixed counters from muted and not muted accounts and then it is confusing

r10s: if it is muted+unmuted, i'd use normal color

r10s: k, after internal discussions, we leave the muted accounts out of the sum-up badge counter in the upper left corner.

one can argue for both, and then, it is easier to not introduce another color/state there :) also, this makes it easier wrt global badge counter. and we can always iterate :)

src/main/java/org/thoughtcrime/securesms/LogViewFragment.java
```diff
@@ -251,8 +251,7 @@ private static String buildDescription(LogViewFragment fragment) {
         builder.append("ignoreBatteryOptimizations=").append(
             powerManager.isIgnoringBatteryOptimizations(context.getPackageName())).append("\n");
       }
-      builder.append("notifications=").append(
-              Prefs.isNotificationsEnabled(context)).append("\n");
+      builder.append("notifications=").append(!dcContext.isMuted()).append("\n");
```
r10s, line 254: this should be removed; the state is added by core now

```suggestion
```

src/main/java/org/thoughtcrime/securesms/connect/KeepAliveService.java
```diff
@@ -29,7 +29,7 @@ public static void maybeStartSelf(Context context) {
         // note, that unfortunately, the check for isIgnoringBatteryOptimizations() is not sufficient,
         // this checks only stock-android settings, several os have additional "optimizers" that ignore this setting.
         // therefore, the most reliable way to not get killed is a permanent-foreground-notification.
-        if (Prefs.isNotificationsEnabled(context) && Prefs.reliableService(context))  {
+        if (Prefs.reliableService(context))  {
```
r10s, line 32: k, so we do not check for notification state any longer (we could check if notifications are enabled in some account) but start the `KeepAliveService` unconditionally.

probably fine, as also the option is now independent, also that makes code and states simpler in general.

src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -131,15 +131,16 @@ public void onCreate() {
...
+    // migrating global notifications pref. to per-account config, added  10/July/24
+    final String NOTIFICATION_PREF = "pref_key_enable_notifications";
+    boolean isMuted = !Prefs.getBooleanPreference(this, NOTIFICATION_PREF, true);
+    if (isMuted) {
       for (int accId : dcAccounts.getAll()) {
-        Prefs.setBackgroundImagePath(this, accId, backgroundImagePath);
+        dcAccounts.getAccount(accId).setMuted(true);
       }
-      Prefs.setStringPreference(this, "pref_chat_background", "");
+      Prefs.removePreference(this, NOTIFICATION_PREF);
```
r10s, line 141: as `SwitchPreferenceCompat` (resp. its base classes)  persist changes in "shared preferences" under the name `pref_key_enable_notifications` as well, the migration is executed whenever the user muted notifications for any account.

on the next app start, all accounts get muted then.

i do not know if we can prevent `SwitchPreferenceCompat` from persisting things, an easy fix would be to use a different name as `pref_enable_notifications` (strike the "key")

r10s (commented):
the migration seems buggy, i could also confirm that in a test.

apart from that and the log, things lgtm.

when these two things are  fixed, this PR should **maybe better be squashed** - i assume this will re-create a commit that not confuses git, sth. is broken in the git history still on some system

src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
adbenitez: good catch! I renamed it as per your suggestion

src/main/java/org/thoughtcrime/securesms/connect/KeepAliveService.java
adbenitez: yeah, also simpler for the user, if they enable the reliable background service/permanent notification, it is harder to understand if it is only enable only under certain circumstances like "at least one account have notifications enabled" and since it is a global option it is easier to enable/disable it directly instead of relaying in disabling notifications for all accounts

src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
r10s: the references need to be changed as well (i'd do a grep over .xml and .java and rename everything but the migration)

adbenitez: I fixed it, I wonder if it is possible to move such keys to some xml file and then don't have this problem, it is easy to overlook and the keys are manually typed in both sides

r10s: for me, there are still dependencies in `preferences_notifications.xml` as `android:dependency="pref_key_enable_notifications"` , this leads to crashes on opening the page.

but maybe a commit is missing again ...

adbenitez: sorry, fixed it and compiled locally this time before pushing it (was thinking it was a quick simple single-line renaming fix and didn't test last times)

PR #3147, opened 2024-07-05
save chatmail diskspace by tweaking autodelete-from-server

r10s:
this PR aims to help saving disk space on chatmail servers by deleting messages _at once after download_ for new profiles (new profiles are usually not multi device, and multi device is the only reason to keep messages longer).

when we later get a hint that the user _might_ set up a second device, we go back to _automatic/never_ - which was the default up to now.

we assume that we will save 90% or more space by that as many user never set up a second device.

this PR also makes the available options a bit clearer (only two are left for chatmail, normal server still have opions as week/month/year) and adds a warning when an option may break multi device:

  

moreover, for chatmail the the alarming confirmation dialog is removed - there is anyways an auto-deletion and  hints as "Do not use this function if you want to keep data on the server" and "Do not use this function if you are using other e-mail clients besides Delta Chat" are not matching and not doable for chatmail.

### DIFF
--- a/src/main/java/com/b44t/messenger/DcContext.java
+++ b/src/main/java/com/b44t/messenger/DcContext.java
@@ -236,6 +236,19 @@ public class DcContext {
       return getConfigInt("is_chatmail") == 1;
     }
 
+    // Called for new profiles on chatmail servers that are "single device" initially;
+    // to save server disk space, we make use of that delete all messages immediately after download.
+    public void assumeSingleDevice() {
+      setConfigInt("delete_server_after", 1 /*at once*/);
+    }
+
+    // Called when we get a hint that another device may be set up.
+    public void assumeMultiDevice() {
+      if (getConfigInt("delete_server_after") == 1 /*at once*/) {
+        setConfigInt("delete_server_after", 0 /*never/automatic*/);
+      }
+    }
+
     /**
      * @return true if at least one chat has location streaming enabled
      */
--- a/src/main/java/com/b44t/messenger/DcContext.java
+++ b/src/main/java/com/b44t/messenger/DcContext.java
@@ -239,12 +239,14 @@ public class DcContext {
     // Called for new profiles on chatmail servers that are "single device" initially;
     // to save server disk space, we make use of that delete all messages immediately after download.
     public void assumeSingleDevice() {
-      setConfigInt("delete_server_after", 1 /*at once*/);
+      if (isChatmail()) {
+        setConfigInt("delete_server_after", 1 /*at once*/);
+      }
     }
 
     // Called when we get a hint that another device may be set up.
     public void assumeMultiDevice() {
-      if (getConfigInt("delete_server_after") == 1 /*at once*/) {
+      if (isChatmail() && getConfigInt("delete_server_after") == 1 /*at once*/) {
         setConfigInt("delete_server_after", 0 /*never/automatic*/);
       }
     }
--- a/src/main/java/org/thoughtcrime/securesms/WelcomeActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/WelcomeActivity.java
@@ -369,6 +369,7 @@ public class WelcomeActivity extends BaseActionBarActivity implements DcEventCen
             else if (progress==1000/*done*/) {
                 DcHelper.getAccounts(this).startIo();
                 progressSuccess();
+                dcContext.assumeMultiDevice();
                 notificationController.close();
                 cleanupTempBackupFile();
             }
--- a/src/main/java/org/thoughtcrime/securesms/preferences/ChatsPreferenceFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/preferences/ChatsPreferenceFragment.java
@@ -68,10 +68,12 @@ public class ChatsPreferenceFragment extends ListSummaryPreferenceFragment {
 
     autoDelServer = findPreference("autodel_server");
     autoDelServer.setOnPreferenceChangeListener(new AutodelChangeListener("delete_server_after"));
-    if (!dcContext.isChatmail()) {
-      CharSequence[] entries = autoDelServer.getEntries();
-      entries[0] = getString(R.string.never);
-      autoDelServer.setEntries(entries);
+    if (dcContext.isChatmail()) {
+      autoDelServer.setEntries(new CharSequence[]{getString(R.string.automatic), getString(R.string.autodel_at_once)});
+      autoDelServer.setEntryValues(new CharSequence[]{"0", "1"});
+      if (dcContext.getConfigInt("delete_server_after") > 1) {
+        dcContext.setConfigInt("delete_server_after", 0 /*never/automatic*/);
+      }
     }
   }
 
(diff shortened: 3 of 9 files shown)


iequidoo:
Btw, this option (DeleteServerAfter) isn't synced across devices in Core, seems this was forgotten.

iequidoo:
Also i checked the core code, for `DeleteServerAfter` we count time from the sort timestamp which is derived from the "sent" timestamp, so we delete messages earlier than "X hours after fetch", but as there's the word "least" maybe it's not that important.

r10s:
the PR does a little more than changing only the wording now, i adapted the PR description accordingly

hpk42:
Great stuff.  The warning message is too implicit though. It's better to assume the reader has already forgotten what they tapped a second ago.  So "turn on at-once deletion" as title and something like "if you enable at-once deletion you can not use multiple devices on this profile anymore"

r10s:
yeah, your text suggestion is much better. i added a commit and also changed the screenshot in the PR description.

src/main/java/org/thoughtcrime/securesms/WelcomeActivity.java
```diff
@@ -369,6 +369,7 @@ else if (progress<1000/*progress in permille*/) {
             else if (progress==1000/*done*/) {
                 DcHelper.getAccounts(this).startIo();
                 progressSuccess();
+                dcContext.assumeMultiDevice();
```
adbenitez, line 372: this might reset deletion config for backups that are not from chatmail accounts :thinking:

src/main/java/org/thoughtcrime/securesms/preferences/ChatsPreferenceFragment.java
```diff
@@ -68,6 +68,13 @@ public void onCreate(Bundle paramBundle) {
 
     autoDelServer = findPreference("autodel_server");
     autoDelServer.setOnPreferenceChangeListener(new AutodelChangeListener("delete_server_after"));
+    if (dcContext.isChatmail()) {
+      autoDelServer.setEntries(new CharSequence[]{getString(R.string.automatic), getString(R.string.autodel_at_once)});
```
adbenitez, line 72: it seems with this you only have the option to delete immediately or after 20 days as per chatmail server config, but you lost other options like 1 week or 1 day that might be better to avoid full inbox while still supporting multidevice

src/main/res/values/strings.xml
```diff
@@ -803,16 +803,19 @@
...
     <!-- shown below enabled autodel_server-option, should be a summary of autodel_server_ask and remind about the impact -->
     <string name="autodel_server_enabled_hint">This includes e-mails, media and \"Saved messages\" in all server folders. Do not use this function if you want to keep data on the server or if you are using other e-mail clients besides Delta Chat.</string>
+    <string name="autodel_sever_warn_multi_device_title">Turn on at-once deletion</string>
+    <string name="autodel_sever_warn_multi_device">If you enable at-once deletion you cannot use multiple devices on this profile anymore.</string>
     <string name="autodel_confirm">I understand, delete all these messages</string>
     <!-- "At once" in the meaning of "Immediately", without any intervening time. -->
-    <string name="autodel_at_once">At once</string>
+    <string name="autodel_at_once">At once after Download</string>
```
adbenitez, line 810: "Immediately after Download" sounds better also wrt translations

adbenitez (approved):

src/main/java/org/thoughtcrime/securesms/preferences/ChatsPreferenceFragment.java
r10s: the thing is that the 20 days are not guaranteed anyways, so all the other options are vague as well

src/main/res/values/strings.xml
r10s: iirc, we had lots of discussions wrt that wording that time, wnd ended in "at once"; i would keep that

adbenitez:
I think this is missing resetting the delete at-once when backup transfer is finished in BackupProviderFragment

also if user does a normal backup, the device will keep deleting immediately despite the other device (that will be the one with deletion reset) 

or is this "delete_server_after" synchronized via chatmail server special support?

src/main/java/org/thoughtcrime/securesms/preferences/ChatsPreferenceFragment.java
r10s: and: we can always ipen for more as needed, having  now two cleqr options sounds good to me; also as it does not make other changes harder.

we can do whatever we want in the server as is "automatic" :)

src/main/java/org/thoughtcrime/securesms/WelcomeActivity.java
r10s: that's true, but it would change only "at once" to "never/automatic", maybe okay and may improve things also for normal accounts.

but i agree, it would be better to do this for chatmail only, where there are no other options that can go wrong (one more reason for no days/weeks/months here, btw)

@link2xt when we check for `is_chatmail` - is that set accordingly even if there is no network after import? so, is it cached in the database?

r10s:
> also if user does a normal backup

idea is that the "never/automatic" is set on export as well as on import

> or is this "delete_server_after" synchronized via chatmail server special support?

no

src/main/java/com/b44t/messenger/DcContext.java
```diff
@@ -236,6 +236,19 @@ public boolean isChatmail() {
...
 
+    // Called for new profiles on chatmail servers that are "single device" initially;
+    // to save server disk space, we make use of that delete all messages immediately after download.
+    public void assumeSingleDevice() {
+      setConfigInt("delete_server_after", 1 /*at once*/);
+    }
+
+    // Called when we get a hint that another device may be set up.
+    public void assumeMultiDevice() {
+      if (getConfigInt("delete_server_after") == 1 /*at once*/) {
```
r10s, line 247:
```suggestion
      if (isChatmail() && getConfigInt("delete_server_after") == 1 /*at once*/) {
```

that might be better - but requires `is_chatmail` to be set accordingly after backup import or on export when no network is available - afaik, this is the case, but not 100% sure

src/main/java/com/b44t/messenger/DcContext.java
```diff
@@ -236,6 +236,19 @@ public boolean isChatmail() {
       return getConfigInt("is_chatmail") == 1;
     }
 
+    // Called for new profiles on chatmail servers that are "single device" initially;
+    // to save server disk space, we make use of that delete all messages immediately after download.
+    public void assumeSingleDevice() {
```
adbenitez, line 241: I am wondering if this isn't better handled automatically by core, similar to how the defaults are selected for some providers via provider-db, core could already have the correct default value for chatmail instances

I think this will set wrong defaults for mailcow deployments with the QR-invites

r10s: well, we want to have an update soon, and the issue is mainly about android, as user onboard there mostly. if things work out, we can think over doing that in core.

src/main/java/org/thoughtcrime/securesms/WelcomeActivity.java
r10s: i added a commit that resets to "never/automatic" only for chatmail profiles, assuming `is_chatmail` is persisted (it is, otherwise, we would also have some settings availablility flickering :)

src/main/java/com/b44t/messenger/DcContext.java
r10s: did an explict commit chat checks for is_chatmail on both, assumeSingleDevice() as well as assumeMultiDevice()

src/main/java/com/b44t/messenger/DcContext.java
r10s: > @adbenitez i am afk - can you check what would be a good place to call assumeSingleDevice()?

nm, i shifted the call and check for isChatmail() before changing the autodel settings.

still, makes sense if you have an additional look if the last commit makes sense to you :)

PR #3154, opened 2024-07-10
expand advanced settings if they are modified

adbenitez:
this helps avoid issues like user changing email address and password and pressing "configure" button without realizing they have custom imap, smpt, etc. settings

this already happened to a friend when I was trying to help them to migrate to a chatmail address, it was really confusing because address was changed and configuration succeed but it took me a while realize that the client was actually connecting to a different IMAP/SMTP server without even using the new set address and password because it was using the manual configuration set in advanced


r10s (approved):
definitely makes sense functionality-wise.

code-wise, one could consider to create one value for each element and build `expandAdvanced` at the end - however, not sure if that is better at the end wrt maintainability as one have to adapt `expandAdvanced` explicitly when a new value is added. in the current approach it would just happen by copy+paste.

so, i'd say it is also fine code-wise.

Issue #3164, opened 2024-07-13
The invite links are hard to discover


adbenitez:
when user goes to the QR screen it is hard for them to discover that they can also share an invite link instead of the QR

in fact for some reason they take pictures/screenshot of the QR code instead of trying the share button at the top (maybe it is not obvious for everybody that that buttons mean "share")

there is enough space in that screen to show the QR button and a "copy invite link" (or so, probably with a link/chain icon link in Telegram at al) button bellow that would be easier to discover

PR #3166, opened 2024-07-13
hide "add contact manually" for chatmail accounts

### LINKED ISSUE #3163: "create contact manually" option is confusing for chatmail
many chatmail users for some reason (if it can go wrong it will go wrong) end up selecting the option to add contact manually in the QR scanning screen instead of sharing invitation link, adding contact manually is not recommended and will not work at all if user adds an address from another server instance

adbenitez:
close #3163

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/qr/QrActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/qr/QrActivity.java
@@ -31,6 +31,7 @@ import com.google.zxing.common.HybridBinarizer;
 
 import org.thoughtcrime.securesms.BaseActionBarActivity;
 import org.thoughtcrime.securesms.R;
+import org.thoughtcrime.securesms.connect.DcHelper;
 import org.thoughtcrime.securesms.contacts.NewContactActivity;
 import org.thoughtcrime.securesms.mms.AttachmentManager;
 import org.thoughtcrime.securesms.permissions.Permissions;
@@ -114,6 +115,7 @@ public class QrActivity extends BaseActionBarActivity {
     public boolean onPrepareOptionsMenu(Menu menu) {
       menu.clear();
       getMenuInflater().inflate(R.menu.qr_show, menu);
+      menu.findItem(R.id.new_classic_contact).setVisible(!DcHelper.getContext(this).isChatmail());
       if(tabLayout.getSelectedTabPosition() == TAB_SCAN) {
         menu.findItem(R.id.withdraw).setVisible(false);
         menu.findItem(R.id.copy).setVisible(false);


hpk42:
i guess this shouldn't be merged/released without also caring for https://github.com/deltachat/deltachat-android/issues/3164  ?

adbenitez:
https://github.com/deltachat/deltachat-android/issues/3164 is not a blocker for this PR, although together they improve the "get in contact situation", but already not going the path to add a contact manually will lead users to scan QR or use the share button and discover the invite link instead of falling in the "message can be delivered unencrypted" trap after adding manually

r10s:
removing this option would probably already push users to "share", as that is the only viable option left. 

downside of the PR is that it removes the option to contact users from the same instance using just the email address - but then again, if you share the address, you can probably also share the link.

but yes, that all should be regarded in the "larger picture" of "make inviting friends" easier, but already alone, it removes one pitfall

adbenitez:
> downside of the PR is that it removes the option to contact users from the same instance using just the email address -

sending unencrypted messages on the same instance should not be a use case we should support and in fact it will be better for chatmail operators legally-wise if such possibility is removed

iequidoo:
The problem maybe is that you can't readd a deleted contact then. Now it works because the peerstate isn't deleted, so you can send encrypted messages to the readded contact and the chat is verified from the beginning. I checked this on my Android.

r10s:
> The problem maybe is that you can't readd a deleted contact then. Now it works because the peerstate isn't deleted, so you can send encrypted messages to the readded contact and the chat is verified from the beginning. I checked this on my Android.

i do not really understand. iirc, ["contacts in use" anyways are not really deleted](https://github.com/deltachat/deltachat-core-rust/pull/3751) (as they are referenced) but their origin is decreased so that they do not appear in add-contact-UI (but eg. still as group members).

anyways, if there is an issue with deletion/recreation, that should be fixed in a way we do not need "add contact manually"

iequidoo:
> i do not really understand. iirc, ["contacts in use" anyways are not really deleted](https://github.com/deltachat/deltachat-core-rust/pull/3751) (as they are referenced) but their origin is decreased so that they do not appear in add-contact-UI (but eg. still as group members).
> 
> anyways, if there is an issue with deletion/recreation, that should be fixed in a way we do not need "add contact manually"

True, contacts in use can't be really deleted, but anyway currently the only way to look up such a contact is "Add Contact Manually". So despite you still have the contact's peerstate, you can't chat with them again. I don't think this really prevents merging the PR, this isn't a frequent scenario, but probably even for chatmail it makes sense to have some fallback option to add a contact by an address.

r10s:
> but probably even for chatmail it makes sense to have some fallback option to add a contact by an address.

if doing the same as for initially adding a contact works again after deletion, that seems enough, totally fine and reasonable.

concretely, as long as tapping or scanning the contact's invite code works also for once deleted contacts, things are fine. if this is not working for re-adding a contact, that's a bug.

iequidoo:
> if doing the same as for initially adding a contact works again after deletion, that seems enough, totally fine and reasonable.
> 
> concretely, as long as tapping or scanning the contact's invite code works also for once deleted contacts, things are fine. if this is not working for re-adding a contact, that's a bug.

If the user has the invite code, i guess it should work. But this is not always the case, e.g. if the user shared their own invite code instead. Ok, i think it's not that important, now it works only because peerstates are never cleaned up.

r10s (approved):

PR #3170, opened 2024-07-14
share vcards from contact profile

### LINKED ISSUE #3169: forward/share from contact-profile
we currently allow to "attach" a contact's vcard to some chat, requiring to select first the destination, then the object to share.

sometimes, the other way round is more practical: when having the object to share already open (so a  contact's profile), just select the destination.

we are going that way already eg. when "forwarding" messages (the object) to other chats. also the "system-wide" sharing works more that way. so, the "components" seems already be there mostly and it should be an overall not too complicated enhancement, making it a bit easier to get ppl into contact.

after internal discussion, it is about "forwarding" (internal naming, we still may want to call it "share" in the UI), so completely inside Delta Chat. forwarding a vcard to other apps is of less use for improving onboarding.

concrete suggestion:

1. add "Forward"  item to the contact profile's three-dot-menu (it can be at the position of the "Copy to clipboard" item which is quite unimportant in the light of chatmail) 

2. when tapping, open a chatlist, allowing the user to select a chat (`RelayUtil` probably needs some adaption, otherwise, we should just do the same as for existing forwarding/sharing)

3. when a chat is selected, stage the contact's vcard as usual

 
open question: what to do with  "Copy to clipboard"? maybe hide it is a "more options" submenu, maybe together with other options? or offer  "Copy to clipboard" only when long-tapping the address itself? is it needed at all?

r10s:
closes #3169 , see there for details and reasoning


adbenitez (approved):
nice, I wonder if it doesn't make sense also to hide the copy-to-clipboard option for chatmail instances?

r10s:
> nice, I wonder if it doesn't make sense also to hide the copy-to-clipboard option for chatmail instances?

yeah, i also thought about that, but then focused more on adding "share", we can refine on another PR.

we could also think about copying vcard to clipboard, copying the address as text seems to be mainly useful for debugging purposes in case of chatmail.

r10s:
> nice, I wonder if it doesn't make sense also to hide the copy-to-clipboard option for chatmail instances?

i created a PR for that: https://github.com/deltachat/deltachat-android/pull/3175

PR #3172, opened 2024-07-15
update README screenshots

r10s:
just use the ones from the store

preview: https://github.com/deltachat/deltachat-android/blob/aa2cd03bbbb696dda08528d2ddd1748d567c3321/README.md

before / after:


adbenitez (approved):
great this save us work in the future

PR #3175, opened 2024-07-17
tune down copying email address

r10s:
copying email addresses to the clipboard
is not proper way to get in contact in the light of chatmail - even if the account is non-chatmail, the receiver may use chatmail. or the copied contact.

it is better to point ppl to use 'Share' or 'Attach Contact' therfore and not offer failing/corner-case things in the primary menu.

therefore, we do basically the same what desktop is doing all the years: no special function, however, if you select text (desktop) or long tap it (android now), you can copy it.

nb: for whatever reason, we're not using context menu before, so it makes sense to double check if things are working as expected

  

_left: long-tap the address text, right: the cleaned up "main options"_

(i was also considering [removing copying the address alltogether](https://github.com/deltachat/deltachat-android/pull/3170), however, it may be of some use here and there. if problems stay we can reconsider)

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationTitleView.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationTitleView.java
@@ -1,5 +1,6 @@
 package org.thoughtcrime.securesms;
 
+import android.app.Activity;
 import android.content.Context;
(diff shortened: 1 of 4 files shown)


adbenitez (approved):
really nice

PR #3176, opened 2024-07-18
add "copy invite link" button to QrShowFragment

### LINKED ISSUE #3164: The invite links are hard to discover

adbenitez:
image

close #3164

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/qr/QrShowFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/qr/QrShowFragment.java
@@ -86,6 +86,8 @@ public class QrShowFragment extends Fragment implements DcEventCenter.DcEventDel
             e.printStackTrace();
         }
 
+        view.findViewById(R.id.copy_link_button).setOnClickListener((v) -> copyQrData());
+
         return view;
     }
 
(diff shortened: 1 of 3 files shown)


r10s:
the function is already in the menu, i assume the gist is to make an option more visible.

that in mind - wouldn't "Share" make much more sense here? esp. as it includes "Copy to Clipboard" functionality on newer androids.

while this would be an improvement, we'll see if it is sufficient or if we need some larger rework of the screen, shifting focus further.

adbenitez:
> that in mind - wouldn't "Share" make much more sense here?

I am fine with changing the functionality but it should be "Share invite link" or so then, the main point is mentioning "invite link" so people know it is not about sharing the QR image

r10s (approved):
let's go for "Share" for now, otherwise lgtm. it is an improvement.

there are ideas for some larger improvements of the whole dialog, but that's not settled yet and this PR won't stand in the way

r10s (approved):

PR #3177, opened 2024-07-19
prevent creating contact without encryption in chatmail account

### LINKED ISSUE #3162: Don't offer to chat when clicking an email address
when clicking an email address inside and/or outside delta chat, a chat with that person is created, for chatmail accounts this will not work since sending any message without encryption will fail, so if the contact doesn't exist yet, instead of the dialog "chat with Foo?" a dialog explaining that it is not possible to write to the address directly without encryption and pointing to QR scanning/link sharing (and maybe a link to some FAQ entry if it exist) should be shown

adbenitez:
close #3162

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -81,7 +81,7 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
   public static final String CLEAR_NOTIFICATIONS = "clear_notifications";
   public static final String ACCOUNT_ID_EXTRA = "account_id";
   public static final String FROM_WELCOME   = "from_welcome";
-  public static final String WARN_ADDR_EXTRA = "warn_addr";
+  public static final String WARN_CANNOT_ENCRYPT = "warn_cannot_encrypt";
 
   private ConversationListFragment conversationListFragment;
   public TextView                  title;
@@ -239,7 +239,7 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
       AccountManager.getInstance().switchAccountAndStartActivity(this, accountId);
     }
 
-    String warnAddr = getIntent().getStringExtra(WARN_ADDR_EXTRA);
+    String warnAddr = getIntent().getStringExtra(WARN_CANNOT_ENCRYPT);
     if (!TextUtils.isEmpty(warnAddr)) {
       DcHelper.showEncryptionRequiredDialog(this, warnAddr);
     }
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -81,6 +81,7 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
   public static final String CLEAR_NOTIFICATIONS = "clear_notifications";
   public static final String ACCOUNT_ID_EXTRA = "account_id";
   public static final String FROM_WELCOME   = "from_welcome";
+  public static final String WARN_ADDR_EXTRA = "warn_addr";
 
   private ConversationListFragment conversationListFragment;
   public TextView                  title;
@@ -236,6 +237,11 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
     }
     if (accountId != dcContext.getAccountId()) {
       AccountManager.getInstance().switchAccountAndStartActivity(this, accountId);
+    }
+
+    String warnAddr = getIntent().getStringExtra(WARN_ADDR_EXTRA);
+    if (!TextUtils.isEmpty(warnAddr)) {
+      DcHelper.showEncryptionRequiredDialog(this, warnAddr);
     }
 
     refreshAvatar();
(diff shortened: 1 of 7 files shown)


src/main/res/values/strings.xml
```diff
@@ -917,6 +917,7 @@
     <string name="chat_protection_broken_explanation">End-to-end encryption cannot be guaranteed anymore, likely because %1$s reinstalled Delta Chat or sent a message from another device.\n\nYou may meet them in person and scan their QR code again to reestablish guaranteed end-to-end encryption.</string>
     <string name="invalid_unencrypted_tap_to_learn_more">⚠️ %1$s requires end-to-end encryption which is not setup for this chat yet. Tap to learn more.</string>
     <string name="invalid_unencrypted_explanation">To establish end-to-end-encryption, you may meet contacts in person and scan their QR Code to introduce them.</string>
+    <string name="encryption_required_for_new_contact">End-to-end encryption is required but it is not setup for %1$s yet.\n\nYou may share your invite link with them or meet in person and scan their QR Code.</string>
```
adbenitez, line 920: fill free to tweak the message

r10s: s/fill free/feel free/ - tweaked :)

r10s (commented):
did a  quick test by adding sth. as `send foo@bar.de a mail` to "Saved messages" on chatmail-profile - the message is tappable, but then just nothing happens. for non-chatmail-profile, the old "chat with ...?" is shown (emulator, android14)

apart from that, the gist of the PR makes sense to me :)

adbenitez:
> did a quick test by adding sth. as `send foo@bar.de a mail` to "Saved messages" on chatmail-profile - the message is tappable, but then just nothing happens. for non-chatmail-profile, the old "chat with ...?" is shown (emulator, android14)

fixed it

src/main/java/org/thoughtcrime/securesms/qr/QrCodeHandler.java
```diff
@@ -55,8 +55,13 @@ public void handleQrData(String rawString) {
                 showFingerPrintError(builder, nameAndAddress);
                 break;
 
-            case DcContext.DC_QR_FPR_OK:
             case DcContext.DC_QR_ADDR:
+                if (dcContext.isChatmail()) {
+                    dcContext.deleteContact(qrParsed.getId());
+                    DcHelper.showEncryptionRequiredDialog(activity, nameAndAddress);
+                    return;
+                }
```
adbenitez, line 63: this is a bit aggressive, the contact might exist and have encryption, but we can't check if the contact exists since the QR parsing from core always creates a contact as side effect just by the action of scanning the QR which is a bit dangerous and questionable, probably should be fixed in core 

cc @link2xt, @iequidoo

adbenitez: other option would be for now to allow to create chats with email address in QR it is not that common and the other changes catch the more common situations

iequidoo: There are many ways to get contacts w/o keys for encryption anyway, e.g. contacts are auto-created from incoming messages' From/To/Cc. What happens now if you try to start a chat with such a contact? So maybe we should allow creating them, but handle this later? Though probably we should warn a user after scanning such a QR if we see that we don't have the contact's public key

iequidoo: The contact may still be useful, you can copy its address and use your another account to send a message f.e.

adbenitez: how can I check if a contact has encryption key or not?

iequidoo: You can use jsonrpc's `CommandApi::get_contact_encryption_info()`. But the problem is that currently it returns the stock string "No encryption." instead of an empty string or other meaningful value for contacts w/o public keys. Also we have `ContactObejct::is_verified` flag, but don't have anything for the encryption state. Do i correctly understand that adding it to `ContactObject` fits, so the app can show a contact info after scanning a QR + a warning if we don't have the contact's key. I'm not sure it should be done this way, users sometimes don't read warnings, but not creating a contact at all may be worse.

Another way is not create a contact, but show an error with the scanned address so that the user can copy it at least

adbenitez: so there is no way to know if a contact has encryption available, cheking for CommandApi::get_contact_encryption_info() string that is meant for display is no good, we need an api to check this, meanwhile we can just leave qr scanning mailto links as-is without the warning since we can't effectively take action with what core provides today

src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
```diff
@@ -81,6 +81,7 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
   public static final String CLEAR_NOTIFICATIONS = "clear_notifications";
   public static final String ACCOUNT_ID_EXTRA = "account_id";
   public static final String FROM_WELCOME   = "from_welcome";
+  public static final String WARN_ADDR_EXTRA = "warn_addr";
```
r10s, line 84: let's call it `WARN_CHATMAIL_ADDR_EXTRA` or `WARN_CANNOT_ENCRYPT` or otherwise be more explicit.

r10s (approved):

src/main/java/org/thoughtcrime/securesms/qr/QrCodeHandler.java
iequidoo: > so there is no way to know if a contact has encryption available

Created a PR for this: https://github.com/deltachat/deltachat-core-rust/pull/5824, please check if it is helpful

PR #3182, opened 2024-07-26
cleanup menu

r10s:
this PR removes less often used and dangerous options of a chat to a new "more" menu - the old menu was just a little big.

- "Archive" is not considered less often used, it is often used as a common pattern to handle chats
- "Add Attachment" could also be removed completely, however, better done in another PR/discussion (it is needed only for the rare usecase where ppl write first - while Delta  Chat is praised for never forgetting attachments by forcing adding them first if one really writes first accidentally, one could fix that by copy+paste)
- ~~would be nice to have the "dangerous" items red, but that can be figured out in another PR (at least there is no obvious option)~~ EDIT: found an easy-enough solution

nb: this should also fix/mitigate the - hopefully rare - issue of accidentally chat deletion: some ppl remember by "finger memory" that "last option in menu is used to switch account" - but being in the chat and not in the chatlist deletes the chat then (seen that two times now... ppl really do not read, though there is a warning saying "do you want to delete the chat" they read "do you want to switch account" :)

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
@@ -440,10 +440,9 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
 
   @Override
   public boolean onPrepareOptionsMenu(Menu menu) {
-    MenuInflater inflater = this.getMenuInflater();
     menu.clear();
 
-    inflater.inflate(R.menu.conversation, menu);
+    getMenuInflater().inflate(R.menu.conversation, menu);
 
     if (dcChat.isSelfTalk() || dcChat.isBroadcast()) {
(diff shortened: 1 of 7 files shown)


iequidoo:
Maybe also move "Add Attachment" under "Delete Chat" as an additional protection? Or add "Less Options ◀️"

r10s:
i think it is fine, let's see. we can always iterate. in general, it is usual to have "delete" last, mixing too wild would open other issues :)

adbenitez:
how does the red options look in the dark theme? I am afraid it might be hard to read with the gray background of the menu, so we might need different red tones for dark theme

adbenitez (approved):
actually it looks fine in dark mode:
image


great!

adbenitez (approved):

r10s:
yeah, it is one of the "reds" that do not "bleed" to white nor black

PR #3183, opened 2024-07-27
show '(un)block contact' in red as well

r10s:
one can argue that 'unblock' is more or even less dangerous than 'block', however, the reason to leave it in 'red' is that it makes it easier to find the option again if the color does not change by state

successor of https://github.com/deltachat/deltachat-android/pull/3182


adbenitez (approved):

r10s:
now, it probably makes sense to tint the "trash icons" and dangerous buttons in "red" as well :) but that can go to another PR

PR #3188, opened 2024-07-30
remove switch account from main menu

### LINKED ISSUE #3185: remove "Switch Profile" from main menu
default way of switching profiles is the avatar in the upper left corner - this is also consistent with iOS and also with Desktop's profile bar. and it is the way we described on the help.

for historic reasons, we also have a "Switch Account" item in the chatlist's three-dot-menu (account switching is older than showing avatars in the title).

at a first glance one would think, that the repetition of the account switching option in the main menu cannot hurt. however, this turns out to be wrong :)

- ppl that discover the "Switch Account" chatlist's do never discover that there is an easier way (with one tap less)
- as switching accounts is a task that is done _very often_, that totally goes to the "finger memory" - tap last item of menu, do sth. in the dialog, no need to read anything, this is a fraction of a second
- however, if only replays this "finger memory" while being in the chat ... one deletes the chat quickly

maybe unbelievable  - but that's what it is. i got direct reports more than one time about that :) ("Delta Chat deleted my chat!" "What did you do?" "Nothing!, just gone!" ... after some research on the second time, i discovered the issue).

one can argue that this is the user's failure, for sure.

however, we thought about removing "Switch Account" from the chatlist's three-dot-menu already before, so this is just another reason beside:

- multi-profile is not the main things new, unexperienced users should try out. it may be even good to have that hidden a bit more
- it is a bit illogical having "Switch account" in the context menu of the chatlist
- "Switch account" is also confusing if you have only one account

we decided for keeping it that time for some period of time, but that time seems to be long enough now :)

disadvantages may be:

- users already using multi-profile may need to rediscover the option. but it seems good to  assume most already have, and for the few others, it is good to force them to that. there are no so many things to tap one. latest if you get a message, it the account switch gets a badge and comes to focus
- new users,  not knowing about multi-profile will need more time to discover. however, this may also be a feature, as other things get into focus by  that, esp. on the first steps when getting into contact, multi-profile is not helpful and may be stand in the way even

we also considers hints, a "+" in the settings beside profile, and whatnot. but i think, it is fine to just go for it.

adbenitez:
close #3185

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -392,9 +392,6 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
       case R.id.menu_global_map:
         WebxdcActivity.openMaps(this, 0);
         return true;
-      case R.id.menu_switch_account:
-        AccountManager.getInstance().showSwitchAccountMenu(this);
-        return true;
       case android.R.id.home:
         onBackPressed();
         return true;
--- a/src/main/res/menu/text_secure_normal.xml
+++ b/src/main/res/menu/text_secure_normal.xml
@@ -21,7 +21,4 @@
     <item android:title="@string/menu_settings"
           android:id="@+id/menu_settings" />
 
-    <item android:title="@string/switch_account"
-        android:id="@+id/menu_switch_account" />
-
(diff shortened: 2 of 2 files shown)


r10s:
we need to check that there's an accessibility hint, otherwise lgtm

r10s (approved):

PR #3191, opened 2024-07-31
replace trash icon in profile switcher with context menu

### LINKED ISSUE #3190: remove trash can from profile switcher
instead allow to delete by long-pressing the account and opening a context menu

this protects more users from having such a dangerous action always directly accessible and making noise in the profiles list since it is not an action user wants to use constantly but more occasional

adbenitez:
close #3190


adbenitez:
I followed this guide:
https://developer.android.com/develop/ui/views/components/menus#context-menu

r10s (approved):

r10s (approved):

PR #3192, opened 2024-08-01
remove push notifications toggle

### LINKED ISSUE #3142: Make it possible to opt out of obtaining FCM token before creating a profile
Currently application checks if push notification preference is enabled at the application start:
https://github.com/deltachat/deltachat-android/blob/7e7257bf4180060495c86cb10cd09121f9e69d77/src/main/java/org/thoughtcrime/securesms/ApplicationContext.java#L230-L239

If preference is enabled, FCM token is requested and once it is obtained it is uploaded to the server.

User has no chance to change the setting before creating a profile and even then FCM token is already obtained.

adbenitez:
close #3142

### DIFF
--- a/src/foss/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
+++ b/src/foss/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
@@ -10,6 +10,5 @@ import androidx.annotation.Nullable;
 public class FcmReceiveService {
   public static void register(Context context) {}
   public static void waitForRegisterFinished() {}
-  public static void deleteToken() {}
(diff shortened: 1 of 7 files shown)


r10s (approved):
`deleteToken()` can be removed as well and the now unused strings should be marked at least as deprecated.

i was also thinking if we should show the "got push notification" success, however, we have that in the log and it is probably not needed to show that in the dialog. the "in Push Service is unavailable" is probably still good enough for most users.

for the help, i created another  PR, https://github.com/deltachat/deltachat-pages/pull/934

otherwise lgtm

r10s (approved):

PR #3196, opened 2024-08-02
add "Invite Friends" to main menu

### LINKED ISSUE #3167: allow to "invite friends"
most apps like Telegram, Signal, etc. have an easy to discover "invite friends" option in the main menu

such option is central when users install the app and don't have any contact to chat with, more in Delta Chat where we don't use the address-book/phone-numbers

in the past we used to have a nice "invite friends" option, but at that time it was a bit useless because it just sent a "join me on delta chat" message, but now we have the invite links that are an important way to get in contact for chatmail users since they can't get in contact typing the address manually and scanning a QR code is easy only if meeting in person

such option could also include some text like "Contact me in Delta Chat: https://i.delta.chat/..."

this enhancement is in a similar direction than https://github.com/deltachat/deltachat-android/issues/3164

adbenitez:
close #3167

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -386,6 +386,9 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
     super.onOptionsItemSelected(item);
 
     switch (item.getItemId()) {
+      case R.id.menu_new_chat:
+        createChat();
+        return true;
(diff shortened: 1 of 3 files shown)


r10s:
- shouldn't it be "Contact me on ..."?
- not sure, if removing "New Chat" is really good idea, the "+" may also be easily overseen (iirc, this was a reason to double that)
- "Settings" should be the last item, this is a usual order, so maybe leave  "New Chat" and add "Invite Friends" below

r10s (approved):

PR #3212, opened 2024-08-05
tint trash red

### LINKED ISSUE #3186: tint dangerous icons red
at #3182 and #3183 we tint dangerous menu texts with a red color.

however, we're not doing that for menu icons yet, this is mainly "delete", so the trash bin. 

maybe code changes are not even needed and we can just make the "trash bin image"  use the color `0xffff0c16` (typical "destructive red" that works nicely in light and dark mode)

r10s:
the first two commits unify the usage of the trashcan icon. in case of "enlarge media", trashing images should also not be a primary action; it is not what is done in many cases.

the last commits then replace the old "white trashcan" by a red one, which looks like the following:

<details>
<summary>color details</summary>

in fact, the red goes to some orange, giving contrast by also making the menu background slightly darker in light mode.

the folling images show the first attempt with more red and unchanged background, however, this resulted in more "bleeding":

</details>

closes #3186


r10s:
ah, CI buils again :)

PR #3215, opened 2024-08-08
stop using jcenter

### LINKED ISSUE #1791: JCenter is closing on August 15th
https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

Something should be done about this line in `build.gradle`:
https://github.com/deltachat/deltachat-android/blob/f03c8a5062e23ac980e7cfb5e508ab551215a1a5/build.gradle#L9

adbenitez:
close #1791

## Migration plan:

- [x] upgrade `com.google.android.exoplayer:exoplayer-core/ui` to recent version
- [x] upgrade `com.journeyapps:zxing-android-embedded` to more recent version **⚠️REQUIRES MinSdk 19+**
- [x] remove unused dependency `'mobi.upod:time-duration-picker`
- [x] get [com.amulyakhare:com.amulyakhare.textdrawable](https://github.com/amulyakhare/TextDrawable) from JitPack.io instead of jcenter
- [x] include [com.codewaves.stickyheadergrid:stickyheadergrid](https://github.com/Codewaves/Sticky-Header-Grid) in our code (upstream is unmantained)
- [x] declare recent version of  dep `relinker`  to avoid `de.cketti.safecontentresolver:safe-content-resolver-v14` depending on old relinker version


Hocuri:
So, this means that, since we're including the binary blob now, we just can't publish any more F-Droid versions until either:
- We manage to get the source code of the old dependency versions, include it in the repo, and get gradle to compile them together with our own source code
- Or we switch to newer versions that are available via Maven Central, dropping support for some older Android versions?

adbenitez:
good point, some of the jars have the .java files but others are compiled .class, so this is not the way to go to solve the problem with fdroid!?

we will have to go the migration route :(

Hocuri:
- For Exoplayer, there is a migration guide [https://developer.android.com/media/media3/exoplayer/migration-guide](https://github.com/deltachat/deltachat-android/pull/3215#) in case this helps
- For textdrawable, simply replacing `implementation 'com.amulyakhare:com.amulyakhare.textdrawable:1.0.1'` with `implementation 'com.github.amulyakhare:TextDrawable:558677ea31'` seems to work (i.e. getting it from JitPack.io instead of jcenter)
- JitPack unfortunately [can't build](https://www.jitpack.io/com/github/Codewaves/Sticky-Header-Grid/e3bd61623c/build.log) StickyHeaderGrid because it [depends on other jcenter dependencies](https://github.com/Codewaves/Sticky-Header-Grid/blob/master/build.gradle). Maybe it's enough to fork Sticky-Header-Grid and add `mavenCentral()` to build.gradle, idk.

adbenitez:
> * with `implementation 'com.github.amulyakhare:TextDrawable:558677ea31'` seems to work (i.e. getting it from JitPack.io instead of jcenter)

this is a rather fragile solution, tried forking the repo and currently jitpack.io doesn't manage to build the repo anymore (see https://jitpack.io/#deltachat/TextDrawable) but for now we can go with that one until it breaks again, anyway I will keep the fork around in the org just in case it gets deleted etc.

build.gradle
```diff
@@ -35,7 +33,7 @@ android {
         applicationId "com.b44t.messenger"
         multiDexEnabled true
 
-        minSdkVersion 16
+        minSdkVersion 21
```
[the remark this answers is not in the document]
adbenitez: yes

src/main/java/org/thoughtcrime/securesms/qr/QrScanFragment.java
```diff
@@ -106,7 +106,7 @@ public MyCaptureManager(Activity activity, DecoratedBarcodeView barcodeView) {
         // which makes _showing_ the QR-code impossible if scanning goes wrong.
         // therefore, we only show a non-disturbing error here.
         @Override
-        protected void displayFrameworkBugMessageAndExit() {
+        protected void displayFrameworkBugMessageAndExit(String message) {
             Toast.makeText(myActivity, R.string.zxing_msg_camera_framework_bug, Toast.LENGTH_SHORT).show();
```
r10s, line 110: it probably makes sense to make use of the given `message`. looking at the source code, i found eg. [this](https://github.com/journeyapps/zxing-android-embedded/blob/master/zxing-android-embedded/src/com/journeyapps/barcodescanner/CaptureManager.java#L406) which does the following:

```suggestion
            if (message.isEmpty()) {
                message = activity.getString(R.string.zxing_msg_camera_framework_bug);
            }
            Toast.makeText(myActivity, message, Toast.LENGTH_SHORT).show();
```

r10s (approved):
chapeau! thanks a lot for this PR! code lgtm, one minor comment. i also played around with videos in the emulator a bit. we anyways need some testing ...

... mile stones and concrete stones - whatever it takes, you know how to hold a hammer! 🔨

src/main/java/org/thoughtcrime/securesms/qr/QrScanFragment.java
adbenitez: just to be safe I will check if message is not null

src/main/java/org/thoughtcrime/securesms/qr/QrScanFragment.java
```diff
@@ -106,8 +106,11 @@ public MyCaptureManager(Activity activity, DecoratedBarcodeView barcodeView) {
         // which makes _showing_ the QR-code impossible if scanning goes wrong.
         // therefore, we only show a non-disturbing error here.
         @Override
-        protected void displayFrameworkBugMessageAndExit() {
-            Toast.makeText(myActivity, R.string.zxing_msg_camera_framework_bug, Toast.LENGTH_SHORT).show();
+        protected void displayFrameworkBugMessageAndExit(String message) {
+            if (message == null || message.isEmpty()) {
```
r10s, line 110: good catch with the null pointer. the following is shorter, but less explicit, a matter of taste i'd say :)

```suggestion
            if (TextUtils.isEmpty(message)) {
```

adbenitez:
@r10s did something to lower minSdk to 19, that is the minimum supported by `com.journeyapps:zxing-android-embedded`

cc @link2xt

r10s (approved):

src/main/java/com/codewaves/stickyheadergrid/StickyHeaderGridAdapter.java
```diff
@@ -0,0 +1,612 @@
...
+
+import androidx.annotation.NonNull;
+import androidx.recyclerview.widget.RecyclerView;
+
+import java.security.InvalidParameterException;
+import java.util.ArrayList;
+
+/**
+ * Created by Sergej Kravcenko on 4/24/2017.
+ * Copyright (c) 2017 Sergej Kravcenko
```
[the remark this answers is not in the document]
r10s: targeted at https://github.com/deltachat/deltachat-android/pull/3231, thanks a lot  for the hint!

PR #3220, opened 2024-08-09
mark bots in chat title and profile as such

r10s:
- in chat titles, show "Bot" instead of the email address of an empty line

- in profiles, show "Bot" instead of "Contact" in the corresponding tab

there are several ideas  for further iterations, but that seems to be a good start, that can also be done on other platforms very easily

Screenshot 2024-08-09 at 16 43 49
Screenshot 2024-08-09 at 16 44 23


adbenitez (approved):

r10s:
ftr: the additional idea to mark bots in chat messages as such might be regarded over the top when the bot is a bridge. then, messages actually written by ppl would be marked as generated as bots. so, maybe the current approch is just a good compromise and good enough (outcome of a discussion with @adbenitez )

Issue #3213, opened 2024-08-06
Changing the language direction of Luri Bakhtiari


hosseinabaspanah:
Hello, please set the language for Luri Bakhtiari like English from left to right, because I have translated this language in Luritian script...

- Android version:14
- Device:Galaxy A23
- Delta Chat version:1.46.8
- Expected behavior:
- Actual behavior:
- Steps to reproduce the problem:
- Screenshots:
- Logs:

r10s:
> because I have translated this language in Luritian script

this sounds like if there are several alphabets/scrips possible?

i am wondering if the used code `bqi` is the correct one for "Luritian script" still.

because, as far as i know, "right to left", "left to right" and other settings are not really supposed to be under control of the app but are defined in the used locale.

hosseinabaspanah:
@r10s 
Yes, there is an Arabic script for this language normally, but translation is allowed for this language and other Luri languages ​​with a special script for Luri, which is taken from Latin, and its name is Luritin...

If you want to check a program to see the status of this language and the source code of the program, look at the link below...

https://github.com/TeamAmaze/AmazeFileManager

Except for one case in which they mistakenly attributed the Luxembourg region to Luri Bakhtiari's language in this program, there is no problem with the display and arrangement from left to right for Luri Bakhtiari's language...

r10s:
thanks for the pointer, then it is what i meant, the language code is wrong. my assumptions:

- Delta Chat uses `bqi`, the system expects Arabic script and right-to-left for that
- Amaze uses `bqi-rLU`, the system expects Latin script and left-to-right for that

i will do a PR that changes the identifier for android, you should be able to download a testing file from there and can give feedback if it fixes the issue

r10s:
i created a PR that uses - the wrong, but same as Amaze - `bqi-rLU` code.

you can try that out by **setting Delta Chat's language to "Default"** - and then select "Luri Bakhtiari" in the global system settings.

selecting the language that way in Delta Chat, does not work (and would probably also not work for Amaze if there would be such an option) as it would set the locale to "Luri Bakhtiari" using region "Luxembourg" which seems to be just wrong and the latter is ignored, resulting in only setting "Luri Bakhtiari" which again is right-to-left.

otherwise, i am a bit lost here - "Luri Bakhtiari" exists in transifex in the following variations:

[...]

PR #3228, opened 2024-08-12
attempt to make 'Luri Baktharari' left-to-right

r10s:
this seems to use the wrong region code, see https://github.com/deltachat/deltachat-android/issues/3213#issuecomment-2283903100 for discussion.


r10s:
this was just a test, adding the region "Luxemburg" seems to add errors and problems, not solve them

PR #3229, opened 2024-08-12
build: update NDK to r27

link2xt:
If we are dropping Android 4 support anyway,
can as well upgrade to the current LTS NDK
which requires API level 21 (Android 5.0).


adbenitez:
I just lowered the support to 19 (android 4.4)

r10s:
as nightlies also get stuck because of this PR not being merged, we should get that in, when 1.48 is out and considered stable (in a few days :)

it is needed also that devs make sure, the new NDK is running on their machines. tbh, i also totally forgot how to install / update / select NDK, will have a look at that these days as well :)

PR #3234, opened 2024-08-14
handle now unsupported oauth2

### LINKED ISSUE #3233: remove gmail's oauth2
oauth2 for gmail stopped working, at least partially.

we decided that we do not want to spend additional development resources and extra money for that (we would need a yearly  "CASA Tier 2/3 sth. security assessment" only for that - [despite we have much better security reviews already passed](https://delta.chat/en/help#security-audits), that alone would require quite some bureaucracy effort, which would result in other things not being done).

gmails oauth2 was annoying all the time, it seems a waste of resources for a provider that is not even playing well together. also, there are comparable few users using gmail oauth2 with Delta Chat. finally, with 2FA and app passwords, user can still use gmail.

- for existing gmail oauth2 users, add a device message how they can "exit"

- do not offer oauth2 for new gmail users, instead, use the normal hint, "With 2FA and an app password set for Delta Chat, logging in works fine." (gmail dc user)

this may require adapting provider database

r10s:
this PR adds a device message for existing gmail-oauth2 users, telling them how to fix log in problems.

new gmail profiles won't be created using gmail-oauth2

closes #3233 

~~need to update the hints in the provider-db now~~ EDIT: done: https://github.com/deltachat/provider-db/pull/299

### DIFF
--- a/src/main/java/com/b44t/messenger/DcContext.java
+++ b/src/main/java/com/b44t/messenger/DcContext.java
@@ -260,6 +260,15 @@ public class DcContext {
       }
     }
 
+    public boolean isGmailOauth2Addr(String addr) {
+      final String oauth2url = getOauth2Url(addr, "chat.delta:/foo");
+      return isGmailOauth2Url(oauth2url);
+    }
+
+    public boolean isGmailOauth2Url(String oauth2url) {
+      return oauth2url.startsWith("https://accounts.google.com/");
+    }
+
     /**
      * @return true if at least one chat has location streaming enabled
      */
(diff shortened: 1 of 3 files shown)


iequidoo:
Maybe also add this link: https://myaccount.google.com/apppasswords. It pretty hard to find from the Google account settings. I even had to google it :)

r10s:
> Maybe also add this link: https://myaccount.google.com/apppasswords. It pretty hard to find from the Google account settings. I even had to google it :)

i followed your device and added that to https://github.com/deltachat/provider-db/pull/299, which is available at "More Info" then; the in-app-box itself does not allow links beside "More Info"

adbenitez (approved):

PR #3240, opened 2024-08-16
stop 'encrypted database' experiment

r10s:
this PR removes options to create encrypted database. existing encrypted databases will continue working for now.

the 'encrypted database' experiment started some years ago was never finished and already stopped working partly, eg. importing backups as encrypted accounts stopped working some years ago for newer android.

in the light of Signal's "encryption key flaw",
but also recent issues with worsending core things in Delta Chat, it seems wise to concentrate on the core things ("you have one job!"), and not on side topics comparable few persons are interested in.

also, it will become harder to maintain that path on upcoming api changes, leaving the database open as opening is slow is needed but will become harder, this will make notifications tricy,
may need additional effort to run in background etc.

all that is doable,
however, it will remove resources from far more important parts, dangerously up to a level that breaks the app - a focus is limited.

also, no one really complained as it is already partly not-working, it was _never_ working on desktop
(where it would make much more sense),
removal on iOS was flawless, see https://github.com/deltachat/deltachat-ios/pull/2011

tl;dr: better leave the protection up to the operating system and concentrate on other things.

### DIFF
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -282,12 +282,6 @@ JNIEXPORT jint Java_com_b44t_messenger_DcAccounts_addAccount(JNIEnv *env, jobjec
 }
 
 
-JNIEXPORT jint Java_com_b44t_messenger_DcAccounts_addClosedAccount(JNIEnv *env, jobject obj)
-{
-    return dc_accounts_add_closed_account(get_dc_accounts(env, obj));
-}
-
-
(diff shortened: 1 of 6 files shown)


adbenitez (approved):
nice reduction of code, good for karma ;)

r10s:
counterpart for adapting FAQ: https://github.com/deltachat/deltachat-pages/pull/946

adbenitez:
> This does not seem to remove much code and I don't understand what does it make easier.

just remove the "encrypt database" checkbox from advanced account creation

r10s:
it removes the need to think about, deal with and maintain a whole set of states and bugs. esp. on api updates and changes in related areas

PR #3241, opened 2024-08-16
use system language, support per-app-language-selection

r10s:
this PR let's the app always just use the system language,  removing the in-app language selection.

if the user wants to have Delta Chat in a different language, they can change the system language. since android13, it is even possible to do that per app. however, as "on the wire" things are neutral now and important strings are rendered in the receivers language, there is not much need to use sth. else than the system language.

sure, there may be cornercases, however, they're outweighed by the advantages:

- fewer, less complex code
- no more bugs in language selection code - passing around locale objects was error-prone - and, during creation of this PR, it turns out, also sometimes forgotten
- we had bugs when creating the xml file, that is also no longer possible now
- things as https://github.com/deltachat/deltachat-android/issues/3213 are easier to target, as there is one state less


src/main/java/org/thoughtcrime/securesms/util/Util.java
```diff
@@ -357,4 +357,6 @@ public static boolean isTouchExplorationEnabled(Context context) {
       return false;
     }
   }
+
+
```
adbenitez, line 361: looks like leftover?
```suggestion
```

r10s:
it needs to be tested if using a per-app language on android13/android14 is working

adbenitez:
the information about per-app language is here:
https://developer.android.com/guide/topics/resources/app-languages

instructions on how to add support for it to the apps is here:
https://developer.android.com/guide/topics/resources/app-languages#auto-localeconfig

r10s:
thanks for the links, very helpful. i added a commit, in-app-language-selection works for be in the emulator with that :)

adbenitez:
indeed, just tested in my device, now when I go to "App info" (clicking the app icon from the "recent used apps" / tasks manager) it shows the language selector of the app now integrated in the system!

image

adbenitez (approved):
really nice clean-up work! while using the recommended system API, elegant!

r10s:
ftr: another way to select the language is "System Settings / System / Languages / App languages":

Screenshot 2024-08-16 at 23 34 50

AndroidX also shows a way to again use an in-app language selector, while probably more straight-forward as before, it is still some stuff, so i do not think it is worth the effort, esp. as the user base that can use the new method is growing

PR #3242, opened 2024-08-16
tweak colors of the status bar

adbenitez:
use same color as the app bar to give a more "modern" feeling to the app, other apps like Telegram, Signal, etc. use the same color for the app bar and status bar these days


r10s (approved):
lgtm, i think `delta_primary_dark` can be removed completely.

nb: bbq is on with the korn

adbenitez:
> lgtm, i think `delta_primary_dark` can be removed completely.

it is used for `pref_icon_tint` and an apparently unused `contact_selection_header_text` but indeed it could be replaced with the literal value or use the primary color for that, but for now I will leave it as-is in this PR to avoid side effects, a "cleanup" PR make sense at a later point

PR #3243, opened 2024-08-16
refresh title bar color 

r10s:
after all the years, the blueish green
could need a freshup.

the new color is close enough,
that it does not break layout of other apps,
existing screenshots and so on.
that can be updated without stress as needed.

the new color was inspired by "Rinmans Green",
which has the additional safetly feature
to be [known as an antibacterial color!](https://www.monumente-online.de/de/ausgaben/2015/3/was-auf-der-hohen-kante-lag.php)

 &nbsp; &nbsp;


r10s:
after some discussion with @adbenitez we decided that the old color is a better default, also contrast wise to the default avatars and colors

PR #3247, opened 2024-08-19
limit number of lines to fetch from logcat

### LINKED ISSUE #3194: crash when opening log if log lines are too many
recently @lkfasnrqrl reported:

> When I try to get the logs after some time when the high battery usage is reported, the log page is frozen and throws the following exception after a minute

 see original post here: https://github.com/deltachat/deltachat-android/issues/3184#issuecomment-2262759006

this seems related to the log string being several megabytes

adbenitez:
for performance and to avoid crashes with potential +250MB in logs when there are unexpected errors resulting in the same lines logging over and over several times per second

close #3194


r10s (approved):
very nice!

PR #3252, opened 2024-08-19
clean content of obj and build folders

r10s:
explicitly list the content to be deleted,
to have some more general protection in case sth. goes wrong accidentally ...


adbenitez (approved):

r10s:
also tested when building https://github.com/deltachat/deltachat-android/pull/3256

PR #3254, opened 2024-08-22
improve nickname dialog

r10s:
this PR improves the "Edit Name" dialog.

inspired by [this forum post](https://support.delta.chat/t/ux-for-setting-alias-name-for-contacts-is-not-obvious/3224) - indeed, there was just nothing. thanks a lot for the hints!

after some internal discussions, we went for placeholder+description. using radio buttons looks to complex for a comparable easy task, also in relation to comparable dialogs in-app and in other apps.

moreover, the PR uses the string "Edit Name" instead of "Edit" in the menu. i was thinking about renaming that even to "Nickname", however, that may be confusing on other levels,  we can always iterate.

finally, the dialog is no longer closed by one wrong tap, avoiding potential loss of typed names.

once this is merged, desktop and iOS can pick up these strings.

_before / after_

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -268,7 +268,7 @@
     <string name="select_more">Select more</string>
     <string name="menu_edit_name">Edit Name</string>
     <!-- The placeholder will be replaced by the name the contact gave themself (if any) or by an e-mail address. -->
-    <string name="edit_name_explain">Set a local nickname to replace \"%1$s\". Leave empty to use the contact\'s chosen name.</string>
+    <string name="edit_name_explain">Set a nickname that will be displayed for you instead of \"%1$s\". Leave empty to use the contact\'s chosen name.</string>
     <!-- The placeholder will be replaced by the name the contact gave themself (if any) or by an e-mail address. -->
     <string name="edit_name_placeholder">Nickname for \"%1$s\"</string>
     <string name="menu_settings">Settings</string>
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -268,7 +268,7 @@
     <string name="select_more">Select more</string>
     <string name="menu_edit_name">Edit Name</string>
     <!-- The placeholder will be replaced by the name the contact gave themself (if any) or by an e-mail address. -->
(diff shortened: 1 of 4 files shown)


src/main/res/values/strings.xml
```diff
@@ -267,6 +267,10 @@
     <string name="select_chat">Select Chat</string>
     <string name="select_more">Select more</string>
     <string name="menu_edit_name">Edit Name</string>
+    <!-- The placeholder will be replaced by the name the contact gave themself (if any) or by an e-mail address. -->
+    <string name="edit_name_explain">Set a local nickname to replace \"%1$s\". Leave empty to use the contact\'s chosen name.</string>
```
adbenitez, line 271: I think saying just "local" is not necessarily straight forward the implications for most marginal users, so emphasizing "only you will see this nickname" or something like that in some part of the text would be nice

adbenitez (approved):

src/main/res/values/strings.xml
r10s: so maybe the first attempt was better:

"Set a nickname that will be displayed for you instead of \"%1$s\". Leave empty to use the user\'s chosen name."

adbenitez: "Set a nickname that only you will see instead of "%1$s". Leave empty to use the user's chosen name."

or

"Set a nickname that will be shown to you instead of "%1$s". Leave empty to use the user's chosen name."

would be my picks but that one is fine as well, the reason I don't like it too much is that "that will be displayed for you" could be misunderstood as if the nick is for you / your nick and not the other person's

r10s: further feedback:

> "Set a nickname that only you will see instead of "%1$s". Leave empty to use the user's chosen name."
> 
> or
> 
> "Set a nickname that will be shown to you instead of "%1$s". Leave empty to use the user's chosen name."
>
> would be my picks but that one is fine as well, the reason I don't like it too much is that "that will be displayed for you" could be misunderstood as if the nick is for you / your nick and not the other person's

i went for 

> "Set a nickname that will be shown to you instead of "%1$s". Leave empty to use the contact's chosen name."

PR #3262, opened 2024-08-24
Increase account switcher hitbox

Hocuri:
The hitbox for the account switcher was only the own avatar in and the title, so when you tapped a bit too low or too high, nothing happened.

This PR changes it to be the whole title bar.

The area where you can now tap to open the account switcher but you couldn't before:

Screenshot_20240824-184814_1_1

(the slim area above and below the menu items on the very right is a bit unnecessary, but I don't know an easy way to avoid it. Before this PR, nothing happened when you tapped there, so I think it's fine)

Anyway, I find it quite a bit easier to open the account switcher now, just try it for yourself if you want :)

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -182,7 +182,7 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
     TooltipCompat.setTooltipText(searchAction, getText(R.string.search_explain));
 
     TooltipCompat.setTooltipText(selfAvatar, getText(R.string.switch_account));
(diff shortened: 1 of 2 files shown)


adbenitez (approved):
btw, would be nice if between the search icon and the rest of the toolbar there were some neutral area/margin so when trying to tap the search icon you don't accidentally open the account switcher

Hocuri:
That does make the PR quite a bit more complex, but it's possible :)

adbenitez:
an extra LinearLayout wrapping two elements should be fine, this solution now looks more correct without any of the more potential user-visible bugs/annoyances, and in general onClick on the whole Toolbar sounds more bug-prune future-wise, so I think it is worth to go with this solution, thanks a lot! :+1:

PR #3263, opened 2024-08-24
Visual feedback when tapping the action button on messages

Hocuri:
When tapping the action button on a message (i.e. "Show  Full Message", "Start" for Webxdcs, and "Download" for big messages"), there was no immediate visual feedback so far.

Since the Full Message / Webxdc activity takes some time to open, I was always a bit unsure whether I had misclicked or whether it's loading.

So, I find it quite a bit more comfortable like this.

**Note that this also changes how the emoji and camera buttons in the message input field look when tapped.**

**I explained the reasoning for the changes in the commit messages.**

If necessary, of course it's easily possible to tweak the colors.

### DIFF
--- a/src/main/res/drawable-v21/touch_highlight_background.xml
+++ b/src/main/res/drawable-v21/touch_highlight_background.xml
@@ -1,7 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
 <ripple xmlns:android="http://schemas.android.com/apk/res/android"
-    android:color="@color/touch_highlight">
-    <item
-        android:id="@android:id/mask"
-        android:drawable="@android:color/white" />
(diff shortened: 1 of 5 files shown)


adbenitez (approved):

r10s (commented):
well, as already mentioned, there are side effects. however, i think, the change is too weird/bad for the the existing elements, esp. as these are the most important elements in the app:

highlighting before / after:

 &nbsp; 

there is not much contrast left.

the highlight for the button itself is nice.

i suggest to fix the issue on-point, really only for the buttons in questions. the other ones have already nice working animations and feedback

Hocuri:
I'll create a follow-up.

PR #3271, opened 2024-08-29
fix wrong translations of videochat_instance_example add doc comment

### LINKED ISSUE #3270: Wrongly translated video instance example link
I just realized that some people have translated the string `videochat_instance_example` (`"Example: https://your-server.org/$ROOM"`) translating the `$ROOM` parameter which should not be changed to something like:

* `Príklady: https://meet.jit.si/$MIESTNOST [...]`
* `Voorbeeld: https://meet.jit.si/$KAMER`

maybe a comment should be put on top of the line of this string clarifying that $ROOM should not be translated and maybe also check that $ROOM was not translated in some of the translations scripts we have

adbenitez:
close #3270

### DIFF
--- a/src/main/res/values-nl/strings.xml
+++ b/src/main/res/values-nl/strings.xml
@@ -331,7 +331,7 @@
     <string name="videochat_tap_to_open">Druk om te openen</string>
     <string name="videochat_instance">Videogespreksinstantie</string>
     <string name="videochat_instance_placeholder">Mijn videogespreksinstantie</string>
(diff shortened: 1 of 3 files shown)


Hocuri (approved):

gerryfrancis:
FYI, please consider https://github.com/deltachat/deltachat-desktop/issues/3366 for translations, too (cc @r10s).

r10s (approved):

r10s:
i also updated the translations on transifex; for https://github.com/deltachat/deltachat-desktop/issues/3366 - i think, this is fine here, the original string does not refer to jitsi, but uses "your-server.org"

PR #3273, opened 2024-08-30
improve profile deletion dialog

adbenitez:
image


adbenitez:
mmh I think the padding in the account item doesn't look good, also the account size is missing (desktop display that info) so maybe better create the item from scratch instead of trying to reuse the ContactSelectionListItem layout

Hocuri (approved):
Looks good! I do think we should wait with merging until the last hotfix release is about a week ago. Then we will have plenty of time left before the next release, and we don't have to do a stable-branch-dance if we need another hotfix release.

PR #3277, opened 2024-09-05
Add separate view for proxy settings

### LINKED ISSUE #3276: Add separate view for proxy settings
Proxy (currently SOCKS5 only) settings are independent of configuration and can be changed without rerunning configuration.

Delta Chat can also have an icon leading to this screen when proxy is enabled, similar to what Telegram does:
telegramproxy

If proxy screen is separate from Password and Account screen, then it can also be made accessible for unconfigured accounts even during chatmail onboarding so users can configure local proxy before creating a chatmail account.

I thought about adding proxy information to connectivity view as a part of https://github.com/deltachat/deltachat-core-rust/pull/5926 but connectivity view is cluttered already so a separate native screen for proxy settings can solve this already.

Adding an icon is outside of the scope for this issue, first these settings need to be moved outside of "advanced settings" for server configuration.

adbenitez:
close #3276 

- [x] add ProxySettingsActivity
- [x] allow to open proxy settings from InstantOnboardingActivity's menu
- [x] update RegistrationActivity (classic email login screen) to open proxy settings activity
- [x] add proxy settings outside in Advanced settings 
- [x] add proxy settings to main menu if it is enabled (in the future when we have a proxy list that should be the determining factor)

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -249,6 +249,11 @@
         </intent-filter>
     </activity>
 
+    <activity android:name=".ProxySettingsActivity"
+              android:label="@string/proxy_settings"
+              android:windowSoftInputMode="stateHidden"
+              android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize"/>
+
     <activity android:name=".LogViewActivity"
               android:label="@string/pref_log_header"
               android:windowSoftInputMode="stateHidden"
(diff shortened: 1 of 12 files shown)


adbenitez:
one known caveat atm: if you enable proxy in an already configured/connected account, IMAP will not reconnect automatically, so it needs to be done automatically by core or otherwise the UI will have to restart IO after the proxy toggle is changed

adbenitez:
merging as is, if we want to re-order or even move up the "server" section we can do it in a next PR

PR #3285, opened 2024-09-11
Fix weblate problem

Hocuri:
This PR will probably fix the problem, if there is one (and if we merge it via "Create a merge commit"), but to see if there is a problem in the first place, I think we should first try to click "Update" at https://hosted.weblate.org/projects/deltachat/android-metadata/#repository


metadata/it/changelogs/6864.txt
adbenitez, line 1: btw we removed this changelog files, we will need to remove them again afterwards

adbenitez (approved):

PR #3292, opened 2024-09-16
use new proxy_url list API

### LINKED ISSUE #3284: New UI for proxy list
https://github.com/deltachat/deltachat-core-rust/pull/5955 introduces new config values `proxy_enabled` and `proxy_url` that are going to replace SOCKS 5 settings. `proxy_url` is actually a list of URLs separated by newlines. The first one is the URL that is used if proxy is enabled. URLs can have a user-visible name which is URL-encoded into anchor part of the URL - ideally this should be displayed separately so spaces are displayed as spaces and not as `%20`.

With core release 1.143.0 old `socks5_*` settings are deprecated and `proxy_*` should be used instead.

This should be done when core is updated from 1.142.x.

adbenitez:
close #3284

missing:
- [ ] improve layout of ProxySettingsActivity
- [ ] create proper layout for the ProxyUrl items
- [ ] allow to long-press proxy-url items to select and delete or share
- [ ] if the current selected proxy is deleted, disable "use proxy" toggle
- [ ] if a proxy is clicked/selected/added, enable "use proxy" toggle


adbenitez:
> Would be nice to have a way to show proxy as a QR code for easy sharing as well.

that is planned for a subsequent PR as part of #3316 



> When I have a proxy that is "Connected" and turn off "Use Proxy", it is strange that proxy shows "Not connected" status while actually I am connected

you are "not connected" to the proxy ;) the idea is not to show the connectivity status in general but the connectivity to the proxy and it is only "connected", "connecting" or "not connected", I think it is clear to the user what this mean, it might look confusing to you because you know that internally this is just using the connectivity status of the account ;)

PR #3296, opened 2024-09-17
simplify 'Proxy Settings' to just 'Proxy'

r10s:
- shorter, less cluttering
- we usually do not say 'Settings' inside 'Settings', also having several 'Settings' menu entries in the main menu us not so nice
- same as telegram is doing

once merged, this needs a `./scripts/tx-push-source.sh`

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -604,7 +604,8 @@
     <string name="login_smtp_port">SMTP Port</string>
     <string name="login_smtp_security">SMTP Security</string>
     <string name="login_auth_method">Authorization Method</string>
-    <string name="proxy_settings">Proxy Settings</string>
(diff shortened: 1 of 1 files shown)


adbenitez:
> * same as telegram is doing

where?

image

I agree with the menu, but in the proxy screen just "Proxy" doesn't look good IMHO, not 100% against this change tho

adbenitez:
also notice that the idea is to use an icon for the menu so the string will not be visible, only when long-pressing where it is more clear if it says "proxy settings"

r10s:
hm, on telegram iOS it is only "proxy", everywhere. i still find the long version superfluous, maybe in the dialog headline , but esp there it is clear that it is about  "settings" :)

+1 for an icon, but also then "Proxy" is fine

adbenitez (approved):
personally prefer a bit "Proxy Settings" for the settings screen, but just "Proxy" is also fine :+1:

PR #3297, opened 2024-09-17
'Mark Read' instead 'Dismiss' notifications

r10s:
'Dismiss' is a bit unclear - even though introduced at #2083 to make things clearer, i do not think this has happened :) -
i also needed to dive into the source code to see what exactly 'Dismiss' is doing.

meanwhile, we're anyway using 'Mark Read' to remove unread counter from chats, and this is basically the same. good to be consistent. finally, this is also the term WhatsApp/Telegram are using.

closes https://support.delta.chat/t/what-does-dismiss-button-on-notifications-do/3261/3

nb: we also change the english 'short' term from 'Read' to 'Mark Read' - this is sill shorter than many translations
and removes the ambiguousness 'Mark Read' vs 'Open for Reading" - esp. in the notification without much context, this is better, but also on iOS, where the term was introduced for, it is good to be explict.

once merged, `./scripts/tx-update-changed-sources.sh` should be called

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/notifications/MarkReadReceiver.java
+++ b/src/main/java/org/thoughtcrime/securesms/notifications/MarkReadReceiver.java
@@ -8,6 +8,8 @@ import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
 
+import com.b44t.messenger.DcContext;
+
 import org.thoughtcrime.securesms.connect.DcHelper;
 import org.thoughtcrime.securesms.util.Util;
 
@@ -16,6 +18,7 @@ public class MarkReadReceiver extends BroadcastReceiver {
   public static final  String CANCEL_ACTION         = "org.thoughtcrime.securesms.notifications.CANCEL";
   public static final  String ACCOUNT_ID_EXTRA      = "account_id";
   public static final  String CHAT_ID_EXTRA         = "chat_id";
+  public static final  String MSG_ID_EXTRA          = "msg_id";
 
   @Override
   public void onReceive(final Context context, Intent intent) {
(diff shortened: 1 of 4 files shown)


adbenitez:
this reintroduces a problem that was fixed by the rename, I agree with going back to "mark as read" but then the proper way to fix is not simply to re-introduce a fixed issue but to change the behavior of the button from "dismiss" to actually marking as read, as the tentative new name

WofWca:
> the proper way to fix is not simply to re-introduce a fixed issue but to change the behavior of the button from "dismiss" to actually marking as read, as the tentative new name

Maybe removing the button completely is better then.

adbenitez:
> Maybe removing the button completely is better then.

why? marking as read to let the other person to know you read the message has been the wanted and expected behavior by other reports in the past (and the reason we changed the name to "dismiss" because people said the message was not marked as read :heavy_check_mark: :heavy_check_mark: )

r10s:
so marking the chat as noticed and marking the explicit message as being read - this sounds good to me, as the message is arguably read, as least as good as scrolling through a chat and skimming over messages

r10s:
> and the reason we changed the name to "dismiss" because people said the message was not marked as read ✔️ ✔️

this is only visible on the other (recipients) device and onl when read receipts are enabled (which is the default), but as said, i do not see a reason why not really mark the message as reas

WofWca:
> why?

Sorry, I mean until the original bug, that caused https://github.com/deltachat/deltachat-android/pull/2083, is fixed.

r10s:
k, i added the following to the PR:

- really mark the message as seen when tapping "Mark Read"
- when replying, also mark the message as seen (this is also what happens when replying in-app)

as a bonus, as we now anyway have the message-id in place, the reply will get a reference (a quote for the receiver), so that eg. a "yes" will have context if while answering "do you join us?" another message "or not?" arrives ...

adbenitez (approved):
nice improvements! :1st_place_medal:

PR #3298, opened 2024-09-17
marked chat replied to as being noticed

r10s:
when replying from within a notification,
the corresponding chat should be marked as being noticed as well.

otherwise, it looks wrong that one has replied to a chat, but still the chat is shown as being "unread".

the 'reply' from a notification should leave the chat in the same state as 'dismiss' resp. 'mark read' - plus the added message.


adbenitez (approved):

WofWca (approved):
Didn't check the code besides the surrounding 10 lines, but the idea makes sense.

PR #3306, opened 2024-09-19
log notifications permission status

### LINKED ISSUE #3305: Add info about notification permission to the log
Here we should have some info about notification permission:
https://github.com/deltachat/deltachat-android/blob/e6f9eda78727a8f06af6df0994910cb6bca44179/src/main/java/org/thoughtcrime/securesms/LogViewFragment.java#L244-L249

We ask for the permission here:
https://github.com/deltachat/deltachat-android/blob/e6f9eda78727a8f06af6df0994910cb6bca44179/src/main/java/org/thoughtcrime/securesms/ConversationListFragment.java#L200-L223

Otherwise when user reports that notification are not working on Android >=13 we don't know if it is just because they denied notification permission on the first start.

adbenitez:
close #3305

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/LogViewFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/LogViewFragment.java
@@ -259,7 +259,7 @@ public class LogViewFragment extends Fragment {
 
       final String token = FcmReceiveService.getToken();
       boolean notifPermGranted = PermissionChecker.checkSelfPermission(context, Manifest.permission.POST_NOTIFICATIONS) == PermissionChecker.PERMISSION_GRANTED;
-      builder.append("notifications-permission=").append(notifPermGranted).append("\n");
+      builder.append("post-notifications-granted=").append(notifPermGranted).append("\n");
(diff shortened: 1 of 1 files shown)


src/main/java/org/thoughtcrime/securesms/LogViewFragment.java
```diff
@@ -256,6 +258,8 @@ private static String buildDescription(LogViewFragment fragment) {
       }
 
       final String token = FcmReceiveService.getToken();
+      boolean notifPermGranted = PermissionChecker.checkSelfPermission(context, Manifest.permission.POST_NOTIFICATIONS) == PermissionChecker.PERMISSION_GRANTED;
+      builder.append("notifications-permission=").append(notifPermGranted).append("\n");
```
r10s, line 262: i would be more explicit and on point here, using the same wording. things are complex enough :)

```suggestion
      builder.append("post-notifications-granted=").append(notifPermGranted).append("\n");
```

r10s (approved):
suggested a rename, otherwise lgtm.

PR #3310, opened 2024-09-20
cleanup account switcher

r10s:
do not show 'Add Profile' in bold.

this looks less cluttered
and makes the "bold" of the selected account more outstanding:

before, there were two types of "boldness",
the add profile button (using bold only)
and the selected profile (using bold and a background), this results in a bit cluttered, unsteady look,
esp. when there are few accounts and more bold than normal is shown.

also, this makes "bold" more outstanding,
as this attribute is given to one line only.
there is no need for the "add profile" to be that outstanding (in contrast to eg. the "add chat" dialog)
("Add Profile" is clearly less often used that an temporarily unselected profile and should stand back also from that point of view)

the "Add Profile" button with its "+" is still recognisable enough and is not really mixed with an existing profile. otherwise, we could add some more whitespace between the accounts and "add profile". btw: wondering if "Add Profile" shouln't be the last item

it is a minor, sure :)

before - a bit noisy and looking as if two things are selected. 

after - less agitated:

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListItem.java
+++ b/src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListItem.java
@@ -91,7 +91,7 @@ public class AccountSelectionListItem extends LinearLayout {
       nameView.setTypeface(null, Typeface.BOLD);
     } else {
       addrView.setTypeface(null, Typeface.NORMAL);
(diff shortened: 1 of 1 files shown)


adbenitez (approved):
to be honest I liked it more bold, but it is an +-0 change, it is kind of OK both ways

Issue #3281, opened 2024-09-08
Start data sync foreground service task when FCM notification is received


link2xt:
Currently FCM notifications are only used to start the application if it is in background so it can download the messages:
https://github.com/deltachat/deltachat-android/blob/56717d3f445a315e9a1623d1614464e6b8575359/src/gplay/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java#L95-L101

Notification is shown only once core downloads a new message from IMAP and decides that notification should be shown. It is done this way to prevent showing notifications for read receipts as the server does not know for encrypted messages if the message is read receipt or not.

Users however report that on some Android phones application fails to connect immediately after receiving FCM notification and only downloads the messages when user opens the application. Because of this FCM notifications effectively do nothing, application starts in the background but cannot download messages. This seems to happen at least on some stock Samsung phones. See <https://support.delta.chat/t/push-notifications-not-working-with-chatmail-server/3245/5>. It should also be possible to replicate on vanilla Android by going to App info -> Mobile data & Wi-Fi and switching off "Background data" or configuring non-existing proxy.

In this case it would be nice to be able to show a local notification saying something like "Failed to connect". This can be done by checking connectivity status when FCM notification is received and if it is `DC_CONNECTIVITY_NOT_CONNECTED` show a notification "Failed to connect". Normally this should not happen because if we can connect to FCM, we should at least be in the "connecting" state in Delta Chat unless OS blocks internet connectivity.

As this may be annoying to receive these notifications constantly on OS that don't allow fixing this behavior, it should be possible to disable the option and it should suggest disable this in the notification if it is enabled by default.

iequidoo:
I have WhatsApp on my phone and it seems that it does the same. At least i see this "You may have new messages" when a bus with wi-fi passes  by me. But i don't remember that it shows any failures. A user may have a poor connection quite a long time, so i'm not sure we should scare users with errors.

r10s:
shouldn't we also for  [`dc_accounts_background_fetch()`](https://c.delta.chat/classdc__accounts__t.html#a4b735a1e51222a06aa992773e42d06b0) [as on iOS](https://github.com/deltachat/deltachat-ios/pull/2071), instead of just hoping that starting the app is good enough? cc @link2xt 

iirc, `dc_accounts_background_fetch()` somehow does maybe_network and whatever is needed to perform a fetch - and can be called unconditionally. on iOS, things seem to work fine with that so far.

[...]

PR #3311, opened 2024-09-20
refine post-notifications-granted logging

r10s:
saying post-notifications-granted=false on API that do not need this grant is misleading as it looks as some error or if the user has rejected sth.

just stumbled upon that and was irritated when trying out things wrt https://github.com/deltachat/deltachat-android/issues/3281 on android7


adbenitez (approved):
good point!

PR #3312, opened 2024-09-20
use dataSync service on receiving FCM notifications

### LINKED ISSUE #3281: Start data sync foreground service task when FCM notification is received

r10s:
this PR marks the code running after receiving an FCM push notification as a "foreground service", aiming not being killed. the "foreground service" requires a visible notification  (nb: while we call this "background fetch", from the view of android, this is a foreground service, see https://developer.android.com/develop/background-work/services , most typical "background" things are "foreground services" nowadays)

when we're done fetching, "foreground service" is stopped, this also removes the notification. 

in most cases, this is only a second or less and the notification should not really be visible

we are using these kind of "foreground service" already for a bunch of things - backup, key export, setup, therefore, i was first trying to use existing `GenericForegroundService`. however, `GenericForegroundService` has the additional side effect to change the flow of activities (one does not want to eg. open the chatlist, when eg. a setup is not yet done), i did not want to add extra states to this already critical area.

(also `GenericForegroundService` might be more complicated as it needs to be - the new `FetchForegroundService` looks so much more straight forward :)

closes #3281

### DIFF
--- a/src/gplay/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
+++ b/src/gplay/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
@@ -102,7 +102,7 @@ public class FcmReceiveService extends FirebaseMessagingService {
   public void onMessageReceived(@NonNull RemoteMessage remoteMessage) {
     Log.i(TAG, "FCM push notification received");
     synchronized (NOTIFICATION_CONTROLLER_LOCK) {
-      notificationController = GenericForegroundService.startForegroundTask(this, getString(R.string.connectivity_updating));
+      notificationController = GenericForegroundService.startForegroundTask(this, getString(R.string.connectivity_updating), false);
       if (!ApplicationContext.dcAccounts.backgroundFetch(19)) { // we should complete within 20 seconds
         notificationController.close();
         notificationController = null;
--- a/src/gplay/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
+++ b/src/gplay/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
@@ -17,20 +17,15 @@ import com.google.firebase.messaging.RemoteMessage;
 
 import org.thoughtcrime.securesms.ApplicationContext;
 import org.thoughtcrime.securesms.BuildConfig;
-import org.thoughtcrime.securesms.R;
-import org.thoughtcrime.securesms.service.GenericForegroundService;
-import org.thoughtcrime.securesms.service.NotificationController;
+import org.thoughtcrime.securesms.service.FetchForegroundService;
 import org.thoughtcrime.securesms.util.Util;
 
 public class FcmReceiveService extends FirebaseMessagingService {
   private static final String TAG = FcmReceiveService.class.getSimpleName();
   private static final Object INIT_LOCK = new Object();
-  private static final Object NOTIFICATION_CONTROLLER_LOCK = new Object();
-
   private static boolean initialized;
   private static volatile boolean triedRegistering;
   private static volatile String prefixedToken;
-  private static NotificationController notificationController;
 
   public static void register(Context context) {
     if (Build.VERSION.SDK_INT < 19) {
@@ -101,21 +96,11 @@ public class FcmReceiveService extends FirebaseMessagingService {
   @Override
   public void onMessageReceived(@NonNull RemoteMessage remoteMessage) {
     Log.i(TAG, "FCM push notification received");
-    synchronized (NOTIFICATION_CONTROLLER_LOCK) {
-      notificationController = GenericForegroundService.startForegroundTask(this, getString(R.string.connectivity_updating));
-      if (!ApplicationContext.dcAccounts.backgroundFetch(19)) { // we should complete within 20 seconds
-        notificationController.close();
-        notificationController = null;
-      }
+    FetchForegroundService.start(this);
+    if (!ApplicationContext.dcAccounts.backgroundFetch(19)) { // we should complete within 20 seconds
+      FetchForegroundService.stop(this);
     }
     Log.i(TAG, "background fetch done");
-  }
-
-  public static void backgroundFetchDone() {
-    synchronized (NOTIFICATION_CONTROLLER_LOCK) {
-      notificationController.close();
-      notificationController = null;
-    }
   }
 
   @Override
--- a/src/gplay/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
+++ b/src/gplay/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
@@ -97,14 +97,6 @@ public class FcmReceiveService extends FirebaseMessagingService {
   public void onMessageReceived(@NonNull RemoteMessage remoteMessage) {
     Log.i(TAG, "FCM push notification received");
     FetchForegroundService.start(this);
-
-    // we should complete within 20 seconds,
-    // see https://firebase.google.com/docs/cloud-messaging/android/receive
-    if (!ApplicationContext.dcAccounts.backgroundFetch(19)) {
-      FetchForegroundService.stop(this);
-    }
-
-    Log.i(TAG, "background fetch done");
   }
 
   @Override
--- a/src/gplay/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
+++ b/src/gplay/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
@@ -95,9 +95,8 @@ public class FcmReceiveService extends FirebaseMessagingService {
   @Override
   public void onMessageReceived(@NonNull RemoteMessage remoteMessage) {
     Log.i(TAG, "FCM push notification received");
-    // the app is running (again) now and fetching and notifications should be processed as usual.
-    // to support accounts that do not send PUSH notifications and for simplicity,
-    // we just let the app run as long as possible.
+    ApplicationContext.dcAccounts.backgroundFetch(120);
+    Log.i(TAG, "background fetch done");
   }
 
   @Override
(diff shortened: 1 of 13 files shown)


adbenitez (approved):

src/gplay/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
```diff
@@ -92,12 +92,15 @@ public static String getToken() {
...
 
+  @WorkerThread
   @Override
   public void onMessageReceived(@NonNull RemoteMessage remoteMessage) {
     Log.i(TAG, "FCM push notification received");
-    // the app is running (again) now and fetching and notifications should be processed as usual.
-    // to support accounts that do not send PUSH notifications and for simplicity,
-    // we just let the app run as long as possible.
+    FetchForegroundService.start(this);
+    if (!ApplicationContext.dcAccounts.backgroundFetch(19)) { // we should complete within 20 seconds
```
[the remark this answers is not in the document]
r10s: i stumbled upon that when jumping into the source code of FirebaseMessagingService:

Screenshot 2024-09-22 at 11 37 05

but it is also mentioned at https://firebase.google.com/docs/cloud-messaging/android/receive (re-reading that, it is crucial that only `Data` is set in the notification)

src/main/java/org/thoughtcrime/securesms/connect/DcEventCenter.java
```diff
@@ -177,6 +178,10 @@ public long handleEvent(@NonNull DcEvent event) {
         DcHelper.getNotificationCenter(context).removeNotifications(accountId, event.getData1Int());
         break;
 
+      case DcContext.DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE:
+        FetchForegroundService.stop(context);
```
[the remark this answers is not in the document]
r10s: imu, same as on iOS, we need to wait for `DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE` to make sure all events are processed - otherwise, the app may be terminated just after `backgroundFetch()`, leaving `DC_EVENT_INCOMING_MSG` unprocessed, resulting in no notification being shown

src/gplay/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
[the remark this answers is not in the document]
r10s: i abandoned the idea initially, as it requires an additional worker thread in `FetchForegroundService.onCreate()` - whereas `FcmReceiveService.onMessageReceived()` already arrives from a worker thread. and for "just running" code, that seems fine.

> If starting FetchForegroundService takes some time or somehow fails then we fetch the messages without the service actually starting and may not have network access yet.

not having internet or being killed too soon indeed is a strong argument, so thanks for the pointer, i added a commit.

it is indeed cleaner and more logical that way - and also when it comes to other push handlers, it is good to have things more concentrated in once class

src/main/java/org/thoughtcrime/securesms/service/FetchForegroundService.java
```diff
@@ -0,0 +1,79 @@
...
+      .setSmallIcon(R.drawable.notification_permanent)
+      .build();
+
+    startForeground(NotificationCenter.ID_FETCH, notification);
+
+    // Start explicit fetch only after we marked ourselves as requiring foreground;
+    // this may help we on getting network and time adequately
+    // Fetch is started in background to not block the UI.
+    // We then run not longer than the max. of 20 seconds,
+    // see https://firebase.google.com/docs/cloud-messaging/android/receive .
```
Hocuri, line 60: I understood https://firebase.google.com/docs/cloud-messaging/android/receive differently:
- FirebaseMessagingService "should handle any message within 20 seconds of receipt", i.e. onMessageReceived should return within 20s
- "The time window for handling a message may be shorter than 20 seconds"
- Therefore, in the code example, they try to need only 10 seconds to be on the safe side:

```
        if (/* Check if data needs to be processed by long running job */ true) {
            // For long-running tasks (10 seconds or more) use WorkManager.
            scheduleJob();
        } else {
            // Handle message within 10 seconds
            handleNow();
        }
```

- As soon as we are in a foreground service, we may need as long as we need to (within some bounds, IIRC on newer Androids you told me it's 6 hours per day or until the user swipes away the notification, anyway, it's way more than 20s)

Conclusion: IIUC, we can set the timeout to more than 20s, and I personally think that we should because on a slow network or so it may be nice to have a minute or even more.

src/main/java/org/thoughtcrime/securesms/service/FetchForegroundService.java
```diff
@@ -0,0 +1,79 @@
...
+    Util.runOnAnyBackgroundThread(() -> {
+      Log.i(TAG, "Starting fetch");
+      if (!ApplicationContext.dcAccounts.backgroundFetch(19)) {
+        FetchForegroundService.stop(this);
+      }
```
Hocuri, line 65: Do I understand correctly that we only stop the service if backgroundFetch() returns false, i.e. `dc_accounts_background_fetch()` returns 0, i.e. `accounts.is_null() || timeout_in_seconds <= 2`?

If so, why? Shouldn't we stop it in any case?

r10s: see [dc_accounts_background_fetch()](https://c.delta.chat/classdc__accounts__t.html#a4b735a1e51222a06aa992773e42d06b0):
- usually,  `dc_accounts_background_fetch()` emits the event  `DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE` and returns true; `FetchForegroundService.stop(this)` os called as a reaction to this event 
- only if the event is not emitted (errors or timeout probably), `false` is returned. in this case, we stop the service here

src/main/java/org/thoughtcrime/securesms/service/FetchForegroundService.java
r10s: > As soon as we are in a foreground service, we may need as long as we need to

indeed, that makes sense to me as well. once there is a foreground service notification, one is usually allowed to run long times - things are visible to the user. only without a foreground service notification one has 20 seconds - and maybe we got hit by that. though, still a bit confusing in the documentation ... 

what would be a reasonable timeout for slow networks then? 60 seconds? the service is not started a second time anyways in case another push arrives and we're not completed.

adbenitez: > Yes, I also wrote in the issue that datasync foreground task is allowed to run for 6 hours or so.

IIRC  it is 6 hours for the whole day, so if you use it for several minutes in every push notification you might run out of "quota"

src/main/java/org/thoughtcrime/securesms/service/FetchForegroundService.java
Hocuri: Ooooh ok. Maybe put a comment similar to: "Else, we need to wait for `DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE` to make sure all events are processed before calling `FetchForegroundService.stop(this)` - otherwise, the app may be terminated just after `backgroundFetch()`, leaving `DC_EVENT_INCOMING_MSG` unprocessed, resulting in no notification being shown."

FTR, slightly off-topic:

The only reason for `dc_accounts_background_fetch()` to return 0 is `accounts.is_null() || timeout_in_seconds <= 2`:

https://github.com/deltachat/deltachat-core-rust/blob/774307241163a959505f20725e486b2df6abaeb9/deltachat-ffi/src/lib.rs#L4880-L4898

In case of other errors or a timeout, the event is emitted and 1 is returned.

src/main/java/org/thoughtcrime/securesms/service/FetchForegroundService.java
r10s: i changed the timeout to 300 seconds in the last commit and adapted the comment

src/main/java/org/thoughtcrime/securesms/service/FetchForegroundService.java
r10s: i added a little pointer in the comment

src/main/java/org/thoughtcrime/securesms/service/FetchForegroundService.java
r10s: > The "ideal" solution that only requires a foreground service (with "Updating..." notification) when necessary

yeah, that may or may not work, probably hard to say without quite some testing across different android versions, manufactors etc.

might be, systems require foreground service and grant network access only there. at least, this was the reasoning to move backgroundFetch() inside FetchForegroundService above.

so, i think, the current approach is kind of more bullet proof as it has less state and is pretty straight forward. let's see if the additional notification is really annoying and visible in case there is good connectivity

PR #3317, opened 2024-09-23
Also start the FetchForegroundService in onDeletedMessages()

Hocuri:
See https://firebase.google.com/docs/cloud-messaging/android/receive#override-ondeletedmessages

Shouldn't happen very often, but if it does, it's good to connect, too.

Follow-up for https://github.com/deltachat/deltachat-android/issues/3281


r10s (approved):
makes sense! thanks a lot!

PR #3320, opened 2024-09-26
start foreground-service for background-fetch only if really in background

### LINKED ISSUE #3319: do not call FetchForegroundService.start() if the app is in foreground
instead, call maybe_network.

this will prevent a possibly flashing permanent notification and will prevent network connection from  being cut and re-established

r10s:
closes #3319 

however, closing this pr as #3319 seems not to be needed:
- the permanent notification is usually not annoying and shortly displayed only
- the "maybe network" optimisation will be done in core, at https://github.com/deltachat/deltachat-core-rust/issues/5989


r10s:
we can reopen if it turns out the permanent notification is annoying

PR #3329, opened 2024-10-04
show display name in title bar when user has multiple profiles

### LINKED ISSUE #3307: Show self-name in title bar
with instant onboarding and chatmail servers multi-profile usage is more common, users that have multiple profiles often get confused or find it hard to know which profile they are using

in ArcaneChat the user's display name is displayed in the top bar, which fits well since the avatar is displayed there, in official DC the avatar is displayed but then the text "Delta Chat" which looks a bit weird/odd more now that clicking it shows the account switcher

adbenitez:
close #3307

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -298,7 +298,13 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
       }
       getSupportActionBar().setDisplayHomeAsUpEnabled(true);
     } else {
-      title.setText(DcHelper.getConnectivitySummary(this, R.string.app_name));
+      String titleText;
+      if (DcHelper.getAccounts(this).getAll().length > 1) {
+        titleText = DcHelper.getConnectivitySummary(this, DcHelper.getContext(this).getName());
+      } else {
+        titleText = DcHelper.getConnectivitySummary(this, R.string.app_name);
+      }
+      title.setText(titleText);
       getSupportActionBar().setDisplayHomeAsUpEnabled(false);
     }
   }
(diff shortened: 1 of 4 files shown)


src/main/java/org/thoughtcrime/securesms/connect/DcHelper.java
```diff
@@ -451,6 +452,10 @@ public static String getConnectivitySummary(Context context, int connectedString
       }
   }
 
+  public static String getConnectivitySummary(Context context, int connectedString) {
+    return getConnectivitySummary(context, context.getString(connectedString));
+  }
+
```
r10s, line 458:
```suggestion
```

src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
```diff
@@ -298,7 +298,13 @@ public void refreshTitle() {
       }
       getSupportActionBar().setDisplayHomeAsUpEnabled(true);
     } else {
-      title.setText(DcHelper.getConnectivitySummary(this, R.string.app_name));
+      String titleText;
+      if (DcHelper.getAccounts(this).getAll().length > 1) {
+        titleText = DcHelper.getConnectivitySummary(this, DcHelper.getContext(this).getName());
+      } else {
+        titleText = DcHelper.getConnectivitySummary(this, R.string.app_name);
```
r10s, line 305:
```suggestion
        titleText = DcHelper.getConnectivitySummary(this, getString(R.string.app_name));
```

i think, it is easier to move `getString()` here than maintaining a second signature of `getConnectivitySummary()` - i needed a moment to understand what is going on otherwise :)

r10s (approved):
some minor suggestion for the additional signature of the `getConnectivitySummary()`, but in general, nice!

src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
r10s: ah, i see, that's a reason for two signatures then, even if minor :)

let's go with the PR as is then

r10s (approved):

adbenitez:
> some minor suggestion for the additional signature of the `getConnectivitySummary()`, but in general, nice!

I updated the code and removed the `getConnectivitySummary(context, int)` it was only used in two other places

PR #3334, opened 2024-10-05
allow to set proxies from QRs

### LINKED ISSUE #3316: Process proxy URLs in QR scanner
if QR type is "proxy", offer the user to use/set the proxy

adbenitez:
close #3316

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/qr/QrCodeHandler.java
+++ b/src/main/java/org/thoughtcrime/securesms/qr/QrCodeHandler.java
@@ -89,6 +89,18 @@ public class QrCodeHandler {
                 BackupTransferActivity.appendSSID(activity, alertDialog.findViewById(android.R.id.message));
                 return;
 
+            case DcContext.DC_QR_PROXY:
+                builder.setTitle(R.string.proxy_use_proxy);
+                builder.setMessage(activity.getString(R.string.proxy_use_proxy_confirm, qrParsed.getText1()));
+                builder.setPositiveButton(R.string.proxy_use_proxy, (dlg, btn) -> {
+                    dcContext.setConfigFromQr(rawString);
+                    dcContext.stopIo();
+                    dcContext.startIo();
+                });
+                builder.setNegativeButton(R.string.cancel, null);
+                builder.setCancelable(false);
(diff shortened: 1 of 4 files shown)


src/main/res/values/strings.xml
```diff
@@ -615,6 +615,7 @@
     <string name="proxy_list_header">Saved Proxies</string>
     <string name="proxy_delete">Delete Proxy</string>
     <string name="proxy_delete_explain">Are you sure you want to delete \"%1$s\"?</string>
+    <string name="proxy_use_proxy_confirm">Do you want to use \"%1$s\" proxy?</string>
```
r10s, line 618:
```suggestion
    <string name="proxy_use_proxy_confirm">Do you want to use proxy \"%1$s\"?</string>
```

src/main/java/org/thoughtcrime/securesms/qr/QrCodeHandler.java
```diff
@@ -89,6 +89,18 @@ public void handleQrData(String rawString) {
                 BackupTransferActivity.appendSSID(activity, alertDialog.findViewById(android.R.id.message));
                 return;
 
+            case DcContext.DC_QR_PROXY:
+                builder.setTitle(R.string.proxy_use_proxy);
+                builder.setMessage(activity.getString(R.string.proxy_use_proxy_confirm, qrParsed.getText1()));
+                builder.setPositiveButton(R.string.proxy_use_proxy, (dlg, btn) -> {
+                    dcContext.setConfigFromQr(rawString);
+                    dcContext.stopIo();
+                    dcContext.startIo();
```
r10s, line 98: this sets implicitly the default proxy and enables proxy settings, right?

and for that to take place, stop/start IO is needed?

r10s (approved):
lgtm so far, did not test

src/main/java/org/thoughtcrime/securesms/qr/QrCodeHandler.java
adbenitez: yes, it sets the configs but current IMAP connection is not reconnected via the proxy automatically, so restarting IO is necessary

PR #3340, opened 2024-10-06
improve proxy settings UI

adbenitez:
improve buttons' tap area and don't show confusing "not connected" if account is not configured or proxy is disabled


r10s (approved):
lgtm, however, i am failing to connect to any proxy in the emulator, so i could not really test

(would be cool if you can add a working proxy somewhere to the PR description, i always have a hard time to find sth) (and do not know if it is the emulator at the end)

r10s (approved):

Issue #3342, opened 2024-10-08
Account name/address in title bar not changed back to "Delta Chat" after all but current account are deleted


gerryfrancis:
- Android version:
Android 14.

- Device:
Google Pixel 8.

- Delta Chat version:
1.46.15 (nightly built 2024-10-06).

- Expected behavior:
After all but the current Delta Chat account have been deleted, the name/address displayed in the title bar of the only account left is changed back to "Delta Chat".

- Actual behavior:
After all but the current Delta Chat account have been deleted, the name/address displayed in the title bar of the only account left is not changed back to "Delta Chat".

- Steps to reproduce the problem:
-- Create at least two Delta Chat accounts.
-- Remove all but the current account. (Result: The name/address of the only account left is still displayed in the title bar.)

- Screenshots:
N/A.

- Logs:
N/A.

adbenitez:
it is back to delta chat eventually, ex. when you open a chat and go back again

r10s:
>  Remove all but one account

for reproducing: it is important to delete the _other_ accounts, not the selected account.

gerryfrancis:
> for reproducing: it is important to delete the other accounts, not the selected account.

@r10s Many thanks, I updated the description accordingly.

adbenitez:
closed by https://github.com/deltachat/deltachat-android/pull/3343

gerryfrancis:
Works fine now, many thanks to everyone involved! :)

r10s:
thanks for checking and for coming back!

PR #3343, opened 2024-10-08
update chatlist title on profile deletion

r10s:
this is needed to show 'Delta Chat' in case only one profile is left

i think, this is simple enough to go for closing #3342 accordingly


adbenitez (approved):
nice, it was good enough but now it is PERFECT :sparkles:

PR #3351, opened 2024-10-10
attach multiple images

r10s:
this PR allows to select and send multiple images directly from within the app's gallery.

after confirmation, the images are send as separated messages then - similar to images selected from the system gallery.

sending single images is done as usual and unchanged.

this should make the usecase of "sharing quite some holiday images" much easier - even though more advanced things as per-image comment or mini-galleries are not there, let alone sending the images in a single mail. all that would be _much_ more effort and is out of scope currently

https://github.com/user-attachments/assets/874e237e-1cdf-41ff-aec2-db0800a82cc2

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
@@ -678,7 +678,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
   }
 
   private void askSendingFiles(ArrayList<Uri> uriList, Runnable onConfirm) {
-    String message = String.format(getString(R.string.ask_send_files_to_selected_chat), uriList.size());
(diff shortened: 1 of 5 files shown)


adbenitez (approved):
really nice improvement!

one step forward to some kind of multi-attachment message, I was thinking that even if it is not practical to send all attachments in a single email, maybe the UI could still group all the images if there is some header/indication of "this images belong to the same message and should be grouped"

r10s:
yes, that is a a more user-visible step and probably more interesting step than really sending on one email. but already that raises quite some more advanced challenges: download-on-demand, handling different aspect ratios, rotations (sounds easy, but it was quite some effort to settle that for one image :) - probably more.

i'd wait at least until we have https://github.com/deltachat/deltachat-core-rust/issues/5888 settled

PR #3359, opened 2024-10-14
[do not merge] test a three letter language

### LINKED ISSUE #3318: make sure, 3-letter-language-codes that are supported by the system are supported by Delta Chat as well
After using the ability to select the program language separately from the system language; Local languages ​​that are not available in the phone cannot be used in Delta Chat...

If you don't understand what I mean, tell me so I can explain more...

If there is a way to enter local languages ​​in Android and Google has provided a platform for Android translation to solve the said problem. Please let me know...

r10s:
targets https://github.com/deltachat/deltachat-android/issues/3318

i tried to get a three [letter-BCP47-language-code](https://github.com/deltachat/deltachat-android/issues/3318#issuecomment-2394451433) of a language that is easier to recognize for me runnin, i chosed "Hawaii" for that.

and: the language is available in the system settings of the Delta Chat language selector, and is working as expected:

so, _in general_ this is the way to use letter-BCP47-language-code.

so, to close #3318, we should follow the same naming convention for `cbk` and `bqi`

might still be, the language is not supported by OS, but that is out of our control then, and one should better advocate on the OS developers then.

### DIFF
--- a/src/main/res/values-b+haw/strings.xml
+++ b/src/main/res/values-b+haw/strings.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources xmlns:tools="http://schemas.android.com/tools" tools:ignore="MissingTranslation">
+    <string name="app_name">Delta Hawaii</string>
+    <string name="invite_friends">Freunde Hawaii</string>
+</resources>


r10s:
tried to rename the folder to the simpler `values-haw` (as signal is [doing](https://github.com/deltachat/deltachat-android/issues/3318#issuecomment-23944514339) - and this is still working. so, the way we're adding three-letter-languages-is okay, it is really just that `ckb` and `bqi` are not supported by most androids

PR #3362, opened 2024-10-14
auto-focus group name field

### LINKED ISSUE #3355: `Group Name` field isn't focused / keyboard isn't shown when creating a New Group
So to type the group name, the user has to do one tap more which is especially inconvenient because group name is mandatory. On Desktop it's auto-focused, so would be good to unify things here. Telegram also auto-focuses the group name field and brings the keyboard on screen.

adbenitez:
close #3355


Hocuri (approved):
LGTM, didn't test

PR #3363, opened 2024-10-14
Do not remember "New Contact" screen's tab state

### LINKED ISSUE #3357: Do not remember "New Contact" tab state
Currently "New Contact" screen remembers the last open tab: either "QR INVITE CODE" or "SCAN QR CODE".
This makes it difficult to give instructions or write guides because the screen that actually opens when you click the qr code button in the top panel or "new contact" (item with the QR code in the contact list) you may get one of the two screens depending on the last action that you did.

Instead of remembering the state, QR code button at the top panel should always open the scanner (so you can use Delta Chat as a QR code scanner, this should be easy to access the scanner with one click) and "New contact" option should always open the QR invite code. "New contact" is not a frequent action anyway as this is only used by new users and most users get their contact from groups afterwards. Worst case users have to switch to another tab from there.

Reworking the "QR INVITE CODE" screen itself is outside of the scope of this issue.

adbenitez:
close #3357

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/qr/QrActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/qr/QrActivity.java
@@ -75,8 +75,7 @@ public class QrActivity extends BaseActionBarActivity {
         getSupportActionBar().setTitle(R.string.menu_new_contact);
         getSupportActionBar().setDisplayHomeAsUpEnabled(true);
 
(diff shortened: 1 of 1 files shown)


adbenitez:
> What does this do, always opens the left tab with a QR code?

yes

Hocuri (approved):

PR #3365, opened 2024-10-14
revert #3177

### LINKED ISSUE #3361: don't prevent creating contacts without keys
#3177 introduced a dialogue to prevent chatmail-profiles  from creating chats when there is no key. 
However, chatmail servers have "passthrough_recipients"  and there is a pending PR to allow to message users on an XMPP bridge (not encrypted), see https://github.com/deltachat/chatmail/pull/408 

Effectively, even if it's a chatmail profile, you may still succeed sending out an unencrypted message but the dialogue prevents even trying it, arguably a bug in dc/chatmail interaction.  

To keep things simple, it seems best to drop the #3177 "a-priori" dialogue.  There already is a "tap more" system message when a message fails to get sent because of missing encryption, which shows the dialogue, so we don't need to per-emptively
do it.  

The more complex alternative is to implement some protocol between chatmail-server and core to tell about "unencrypted" addresses and to query that from the UI to determine if to show the dialogue, but that seems hardly worth it.

adbenitez:
close #3361

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -86,7 +86,6 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
   public static final String CLEAR_NOTIFICATIONS = "clear_notifications";
   public static final String ACCOUNT_ID_EXTRA = "account_id";
   public static final String FROM_WELCOME   = "from_welcome";
-  public static final String WARN_CANNOT_ENCRYPT = "warn_cannot_encrypt";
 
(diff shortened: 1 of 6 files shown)


r10s:
did not look in detail about the impact (will do tomorrow) but in any case, we should not remove the strings but deprecate them, otherwise other os will get in trouble, we shoudl give them time to adapt

r10s (approved):

adbenitez:
closing stale PR

adbenitez:
ping @r10s, @hocuri, it was already approved in the past, anything speaking against merge?

Hocuri (approved):

PR #3366, opened 2024-10-14
Visual feedback when tapping the action button on messages, attempt 2

Hocuri:
Follow-up for https://github.com/deltachat/deltachat-android/pull/3263.

This does still slightly change the highlight color for the whole app, since I do think it's nice to have it the same throughout the app. The previous highlight color was from the old Signal code was seemingly forgotten when changing from Signal look-and-feel to Delta look-and-feel.

Before/after: ("before" the visual feedback is barely visible except for when you pointed onto it with a mouse):

Screenshot_20241014-194207
Screenshot_20241014-202253

_But_, it would be simple to revert the last two commits and instead directly write `#5C4CB8DB` into `touch_highlight_background_strong.xml`, then nothing will change in the app except for the background of the action button.


adbenitez:
TBH I still like the gray highlight more and would have preferred just changing the highlight for the "button links" in the message bubble, in any case we need to test that this doesn't worsen things in some other areas that could have colors in background

PR #3368, opened 2024-10-16
reflect status of the proxy (enable/disabled) in the proxy icon

### LINKED ISSUE #3364: Proxy icon doesn't reflect enable/disable status of the proxy
the new proxy icon in the main screen's menu doesn't reflect the state of enabled/disabled of the proxy toggle, so user can't realize they have the proxy enabled/disabled easily and troubleshoot connection problems

adbenitez:
the icon is also shown in the onboarding screen to remind the user that they have the proxy enabled to help troubleshooting

I think the icons could probably be replaced at some point but this is the best I could find and it is pretty similar to the one used by Telegram, in some way you are protecting the connection against blocking etc. and if you are using shadowsocks it is even encrypted

close #3364

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -19,6 +19,7 @@ package org.thoughtcrime.securesms;
 import static org.thoughtcrime.securesms.ConversationActivity.CHAT_ID_EXTRA;
 import static org.thoughtcrime.securesms.ConversationActivity.STARTING_POSITION_EXTRA;
 import static org.thoughtcrime.securesms.connect.DcHelper.CONFIG_ADDRESS;
+import static org.thoughtcrime.securesms.connect.DcHelper.CONFIG_PROXY_ENABLED;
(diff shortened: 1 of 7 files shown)


r10s (approved):
nice improvement,  also showing the state during onboarding makes lots of sense.

code lgtm, and i am also fine with the "shield", esp. as it is known by users. UX is also about historic experiences :)
also, the shield could be read as "defend blockings". 

for the concrete design of the esp. filled shield: this does neither look good nor really fitting (the other icons have another "weight"), but that is no blocker. maybe the outlined-shield as "on" and a stroked-outline-shield for "off"?

adbenitez:
when looking for shield this is what I found in android studio, I like the shield with the padlock, but it is fitting only for VPN shadowsocks proxies that are the ones with encrypted connection, my favorite would be like the icon telegram uses: a shield with a checkmark in the middle, I tried to modify the android vector asset manually but it is too difficult to modify such vector paths that are just a sequence of numbers hard to understand so I gave up eventually
  
image

r10s:
> I tried to modify the android vector asset manually but it is too difficult to modify such vector paths that are just a sequence of numbers hard to understand so I gave up eventually

iirc, one can convert these .xml vectors to/from .svg - and then modify them using inkscape

Issue #3369, opened 2024-10-19
QR code scanner for "Add as Second Device" should have an option to paste from clipboard


link2xt:
Currently 3-dot menu only has "Troubleshooting".

Same for "Scan Invitation Code", it should have a way to paste from clipboard.

See https://support.delta.chat/t/how-do-i-add-a-second-device-that-doesnt-have-a-camera/3289

r10s:
we had a bit hidden fallback by the option to use the scanner available from the chatlist - however, that was forgotten to adapt by the move to DC_QR_BACKUP2, i did a fix at #3370 

but yes, we should probably have a paste option also directly at "add second device scanner"

PR #3370, opened 2024-10-20
support BACKUP2 in chatlist-qr-code scanner, 

r10s:
this was forgotten by the move to BACKUP2 in #3173

targets #3369, however, indeed "paste" should be added to "add second device scanner" as well

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/qr/QrCodeHandler.java
+++ b/src/main/java/org/thoughtcrime/securesms/qr/QrCodeHandler.java
@@ -76,6 +76,7 @@ public class QrCodeHandler {
                 break;
 
             case DcContext.DC_QR_BACKUP:
+            case DcContext.DC_QR_BACKUP2:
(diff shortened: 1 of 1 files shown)


adbenitez (approved):
I wonder why did we even need a new type for this QR result? I was expecting it to always be DC_BACKUP

cc @link2xt

adbenitez:
> Because it has a different format, in new iroh there are separate node address and auth token

but that is something internal, right? the UI doesn't use anything of that and this PR literally used the exact same code to handle both events :thinking:

iequidoo:
> > Because it has a different format, in new iroh there are separate node address and auth token
> 
> but that is something internal, right? the UI doesn't use anything of that and this PR literally used the exact same code to handle both events 🤔

You're right, this originates in `deltachat-ffi/src/lot.rs`: `Qr::Backup2 { .. } => LotState::QrBackup2,`. Probably `QrBackup` can be used, it's not passed to the core anyway. I have no idea why the `QrBackup2` variant was added

PR #3372, opened 2024-10-21
allow to paste from clipboard in RegistrationQrActivity

### LINKED ISSUE #3369: QR code scanner for "Add as Second Device" should have an option to paste from clipboard

adbenitez:
close #3369


r10s (approved):
code lgtm, and general pasting also tested in emulator.

@adbenitez did you test on a real device that the original _scanning_ of QR codes still work as expected?

adbenitez:
> @adbenitez did you test on a real device that the original _scanning_ of QR codes still work as expected?

yes, it works fine :partying_face:

PR #3373, opened 2024-10-21
allow to set tags on profiles

adbenitez:
see https://support.delta.chat/t/chat-folders-to-collect-chats-by-context/2201

image

image

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
@@ -36,6 +36,7 @@ import org.thoughtcrime.securesms.util.Util;
 import org.thoughtcrime.securesms.util.ViewUtil;
 
 import static com.b44t.messenger.DcContact.DC_CONTACT_ID_ADD_ACCOUNT;
+import static org.thoughtcrime.securesms.connect.DcHelper.CONFIG_PRIVATE_TAG;
 
 public class AccountSelectionListFragment extends DialogFragment
 {
@@ -129,14 +130,15 @@ public class AccountSelectionListFragment extends DialogFragment
     View view = View.inflate(activity, R.layout.single_line_input, null);
     EditText inputField = view.findViewById(R.id.input_field);
     inputField.setHint(R.string.profile_tag_hint);
-    inputField.setText(dcContext.getTag());
+    inputField.setText(dcContext.getConfig(CONFIG_PRIVATE_TAG));
 
     new AlertDialog.Builder(activity)
       .setTitle(R.string.profile_tag)
       .setMessage(R.string.profile_tag_explain)
       .setView(view)
       .setPositiveButton(android.R.string.ok, (d, b) -> {
-        dcContext.setTag(inputField.getText().toString().trim());
+        String newTag = inputField.getText().toString().trim();
+        dcContext.setConfig(CONFIG_PRIVATE_TAG, newTag);
         AccountManager.getInstance().showSwitchAccountMenu(activity);
       })
       .setNegativeButton(R.string.cancel, (d, b) -> AccountManager.getInstance().showSwitchAccountMenu(activity))
--- a/src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
@@ -128,7 +128,7 @@ public class AccountSelectionListFragment extends DialogFragment
     DcContext dcContext = DcHelper.getAccounts(activity).getAccount(accountId);
     View view = View.inflate(activity, R.layout.single_line_input, null);
     EditText inputField = view.findViewById(R.id.input_field);
-    inputField.setHint(R.string.enter_tag);
+    inputField.setHint(R.string.tag_field_hint);
     inputField.setText(dcContext.getTag());
 
     new AlertDialog.Builder(activity)
--- a/src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
@@ -128,7 +128,7 @@ public class AccountSelectionListFragment extends DialogFragment
     DcContext dcContext = DcHelper.getAccounts(activity).getAccount(accountId);
     View view = View.inflate(activity, R.layout.single_line_input, null);
     EditText inputField = view.findViewById(R.id.input_field);
-    inputField.setHint(R.string.tag_field_hint);
+    inputField.setHint(R.string.profile_tag_hint);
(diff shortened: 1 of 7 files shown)


adbenitez:
this uses `ui.tag` it make sense to add a core setting for this eventually but should not be a blocker if we don't want to invest time into that for now

r10s:
nice! 

indeed i also drafted sth. these days, see https://github.com/deltachat/deltachat-core-rust/pull/6088 and https://github.com/deltachat/deltachat-android/compare/main...r10s/internal-profile-names

while the android PR is far behind yours (maybe only some inspiration for wordings), we should consider to add an "official" config for core - i think, it makes sense to sync them at some point and also that is better wrt compatibility and synchronisation. might also be, core will use the labels for sth at some point, name of backups or so

src/main/res/values/strings.xml
```diff
@@ -647,6 +647,10 @@
     <string name="accept_invalid_certificates">Accept invalid certificates</string>
     <string name="switch_account">Switch Profile</string>
     <string name="add_account">Add Profile</string>
+    <string name="profile_tag">Profile Tag</string>
+    <string name="enter_tag">Enter Tag</string>
+    <string name="set_tag">Set Tag</string>
```
r10s, line 652: "set" seems redundant here, usually, we do not use that verb. the menu entry would better just read "Profile Tag" (or whatever we chose here)

src/main/res/values/strings.xml
```diff
@@ -647,6 +647,10 @@
     <string name="accept_invalid_certificates">Accept invalid certificates</string>
     <string name="switch_account">Switch Profile</string>
     <string name="add_account">Add Profile</string>
+    <string name="profile_tag">Profile Tag</string>
+    <string name="enter_tag">Enter Tag</string>
```
r10s, line 651: for the placeholder, maybe write down some examples?

`eg. Work, Family, ...`

src/main/res/layout/account_selection_list_item.xml
```diff
@@ -60,16 +60,17 @@
...
+              android:id="@+id/addr"
+              android:layout_width="wrap_content"
+              android:layout_height="wrap_content"
+              android:textDirection="ltr"
+              android:singleLine="true"
+              android:ellipsize="marquee"
+              android:textAppearance="?android:attr/textAppearanceSmall"
+              android:textSize="14sp"
+              android:fontFamily="sans-serif-light"
+              tools:text="user@example.com" />
```
r10s, line 73: ah, this is the old "address" :)

maybe we should rename it to sth. more fitting.

also, if no label is set, the alignment is not centered vertically:

r10s (commented):
very nice!

src/main/res/values/strings.xml
adbenitez: the examples are in the explanation text :thinking:

src/main/res/layout/account_selection_list_item.xml
adbenitez: fixed

src/main/res/values/strings.xml
adbenitez: I took inspiration from your branch, kept "Tag" instead of label, but if you prefer Label we can change it

src/main/res/values/strings.xml
```diff
@@ -647,6 +647,9 @@
     <string name="accept_invalid_certificates">Accept invalid certificates</string>
     <string name="switch_account">Switch Profile</string>
     <string name="add_account">Add Profile</string>
+    <string name="profile_tag">Private Tag</string>
+    <string name="tag_field_hint">eg. Work, Family, ...</string>
```
r10s, line 651:
```suggestion
    <string name="profile_tag_hint">eg. Work, Family, ...</string>
```

src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
```diff
@@ -113,9 +114,35 @@ private void onContextItemSelected(MenuItem item, int accountId) {
...
 
+  private void onSetTag(int accountId) {
+    Activity activity = getActivity();
+    if (activity == null) return;
+    AccountSelectionListFragment.this.dismiss();
+
+    DcContext dcContext = DcHelper.getAccounts(activity).getAccount(accountId);
+    View view = View.inflate(activity, R.layout.single_line_input, null);
+    EditText inputField = view.findViewById(R.id.input_field);
+    inputField.setHint(R.string.tag_field_hint);
```
r10s, line 131: i thought somehow, setting placeholders is also possible directly in`.xml`, but maybe i'm wrong :)

adbenitez: the `single_line_input` layout is used for set contact profile name and add proxy dialogs, it is generic for any dialog wanting to add a single line input field, so the hint can't be set in the xml

src/main/res/values/strings.xml
r10s: tag is fine. and i also think "private" is an improvement over my "personal"

adbenitez:
updated to use the new `private_tag` setting now that it is part of main branch in core, we need a new tagged core before this can be merged

r10s:
updated the core on main; this PR needs a rebase

src/main/res/values/strings.xml
```diff
@@ -652,6 +652,9 @@
     <string name="accept_invalid_certificates">Accept invalid certificates</string>
     <string name="switch_account">Switch Profile</string>
     <string name="add_account">Add Profile</string>
+    <string name="profile_tag">Private Tag</string>
+    <string name="profile_tag_hint">eg. Work, Family, ...</string>
```
r10s, line 656: i think, either `eg.` or `...` is enough:

```suggestion
    <string name="profile_tag_hint">eg. Work, Family</string>
```

or 

```suggestion
    <string name="profile_tag_hint">Work, Family, …</string>
```

in any case, `...` which should be an ellipsis, so `…` and not `...`

r10s (approved):
wondering, if it makes sense to show the tag additionally in the subtitle above the chatlist. but that can go to another PR, if we think it makes sense. it is already very useful as it is now

src/main/res/layout/account_selection_list_item.xml
```diff
@@ -60,16 +60,17 @@
...
-                      android:ellipsize="marquee"
-                      android:textAppearance="?android:attr/textAppearanceSmall"
-                      android:textSize="14sp"
-                      android:fontFamily="sans-serif-light"
-                      tools:text="user@example.com" />
+          <org.thoughtcrime.securesms.components.emoji.EmojiTextView
+              android:id="@+id/addr_or_tag"
+              android:layout_width="wrap_content"
+              android:layout_height="wrap_content"
+              android:textDirection="ltr"
```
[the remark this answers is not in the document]
r10s: i think, this comes maybe copy+paster from old signal code, where this was a _phone number_ ...

PR #3377, opened 2024-10-23
notify reactions added

r10s:
this PR shows notifications for reactions to one's own messages.

the notifications are added unconditionally, assuming, this is also known from other messengers. also, only reaction to _own_ messages are notified  - compared to that otherwise _all_ messages of others are notified, this usually won't add much  - either one cares or not (and in the latter case, you have muted the chat). we can always iterate :)

for the "mark read": there is nothing to mark as read for reactions, however, i still think, it makes sense to keep the button as it marks the chat as noticed - and it is maybe more confusion if it is sometimes not there.

counterpart of https://github.com/deltachat/deltachat-ios/pull/2331

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/connect/DcEventCenter.java
+++ b/src/main/java/org/thoughtcrime/securesms/connect/DcEventCenter.java
@@ -171,7 +171,11 @@ public class DcEventCenter {
 
     switch (id) {
       case DcContext.DC_EVENT_INCOMING_MSG:
-        DcHelper.getNotificationCenter(context).addNotification(accountId, event.getData1Int(), event.getData2Int());
+        DcHelper.getNotificationCenter(context).notifyMessage(accountId, event.getData1Int(), event.getData2Int());
+        break;
+
+      case DcContext.DC_EVENT_INCOMING_REACTION:
+        DcHelper.getNotificationCenter(context).notifyReaction(accountId, event.getData1Int(), event.getData2Int(), event.getData2Str());
         break;
 
       case DcContext.DC_EVENT_MSGS_NOTICED:
--- a/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
+++ b/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
@@ -20,12 +20,14 @@ import android.util.Log;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.WorkerThread;
 import androidx.core.app.NotificationCompat;
 import androidx.core.app.NotificationManagerCompat;
 import androidx.core.app.RemoteInput;
 import androidx.core.app.TaskStackBuilder;
 
 import com.b44t.messenger.DcChat;
+import com.b44t.messenger.DcContact;
 import com.b44t.messenger.DcContext;
 import com.b44t.messenger.DcMsg;
 import com.bumptech.glide.load.engine.DiskCacheStrategy;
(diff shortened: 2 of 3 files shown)


src/main/java/org/thoughtcrime/securesms/connect/DcEventCenter.java
```diff
@@ -171,7 +171,11 @@ public long handleEvent(@NonNull DcEvent event) {
 
     switch (id) {
       case DcContext.DC_EVENT_INCOMING_MSG:
-        DcHelper.getNotificationCenter(context).addNotification(accountId, event.getData1Int(), event.getData2Int());
+        DcHelper.getNotificationCenter(context).notifyMessage(accountId, event.getData1Int(), event.getData2Int());
+        break;
+
+      case DcContext.DC_EVENT_INCOMING_REACTION:
+        DcHelper.getNotificationCenter(context).notifyReaction(accountId, event.getData1Int(), event.getData2Int(), event.getData2Str());
```
adbenitez, line 178: one problem with this is that if user "reads"/see  the reaction in one chat, will it be marked as read and the notification disappear in other devices??? or will the user need to handle duplicated notifications in every device?

r10s: good question!

r10s:
we decided to keep that out of 1.48, also because of the [open question wrt multi-device above](https://github.com/deltachat/deltachat-android/pull/3377#pullrequestreview-2389859336)

Hocuri (approved):
Nice :tada: :heart:!

src/main/java/org/thoughtcrime/securesms/connect/DcEventCenter.java
Hocuri: The `MsgsNoticed` contains a chat id (not a message id), so we just need to make sure that core sends it not only when messages are noticed but also when reactions are noticed on another device

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -494,7 +518,7 @@ public void addNotification(int accountId, int chatId, int msgId) {
                             lines = new ArrayList<>();
                             accountInbox.put(chatId, lines);
                         }
-                        lines.add(line);
+                        lines.add(tickerLine);
```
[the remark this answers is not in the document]
Hocuri: I think there is a misunderstanding, I wanted to suggest changing `lines.add(tickerLine);` to `lines.add(shortLine);` in order not to have the redundant `Bob: ` prefix

Hocuri, line 521:
```suggestion
                        lines.add(shortLine);
```

r10s: see discussion above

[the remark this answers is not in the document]
Hocuri: > but, most importantly for this PR: this is offtopic, as far as i see, things are unchanged here, at least this was the gist :)

As far as I understood the code, `line` was renamed to `shortLine` and `tickerLine` is still called `tickerLine` throughout the code? So, to get the behavior of the old code, `line` should be renamed to `shortLine` also here.

However, I just see that there was a bug in the old code:

```java
            // if privacy allows, for better accessibility,
            // prepend the sender in the ticker also for one-to-one chats (for group-chats, this is already done)
            String tickerLine = line;
            if (!dcChat.isMultiUser() && privacy.isDisplayContact()) {
                line = dcMsg.getSenderName(dcContext.getContact(dcMsg.getFromId()), false) + ": " + line;
            }
```
here, `line` is changed instead of `tickerLine`, even though the comment says that the sender should be prepended to the ticker. 

This bug is gone now, and it looks like the old behavior is produced again by replacing `line` with `tickerLine` here.

r10s: maybe better discuss on point to the code here:

- it is true, that this was the variable `line` but not `tickerLine` before
- however, in the old code, `line` was  converted to a _ticker-line-including-sender_, see L406 in the old code - and `tickerLine` was never a _ticker-line-including-sender_, even though that was probably the idea

this is indeed confusing, took me a while to re-understand this after these weeks of writing the PR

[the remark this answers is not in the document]
r10s: let's maybe continue below on code-level, not sure if i could explain it better there :) (or if i am wrong at the end :)

adbenitez:
is there a core issue for the "mark as read reactions" problem? I don't see it linked to this PR

r10s:
afaik, there is no core issue.

imu, the issue is: 
- if a chat _without_ unread messages is opened, DC_EVENT_MSGS_NOTICED is _not_ triggered on other devices (only locally).
- and therefore, the reaction-notifications on the other devices are not removed.

imu, already now, reaction-notifications are not triggered when they're not fresh for whatever reason.

so, i am wondering, how bad that is in practise - and if ppl really care about notifications, long scrolled away, hanging around a bit longer - sound etc. already has happened.

so, tbh, the impact is not totally clear to me. cc @adbenitez 

surely, it also depends on how hard the fix in core would be, maybe it is comparable easy to do and it is not even worth about thinking about impact and possible alternatives :) cc @iequidoo

adbenitez (approved):

PR #3380, opened 2024-10-24
Use tag or display name for notifications instead address

### LINKED ISSUE #3378: remove account address from notifications
i just realised there is quite some noise in the notifiation that show the email address of the account, see screenshots at https://github.com/deltachat/deltachat-android/pull/3377

i suggest to:
- if there is only one account, show nothing thelike
- if there are multiple accounts, show the "private tag", if set. otherwise show the account's displayname

adbenitez:
close #3378 

before - quite some noise in the notification:

Screenshot 2024-10-24 at 11 20 29

after - on a default installation with one account we do not need to show anything:

Screenshot 2024-10-24 at 11 34 55

after - using multiple accounts and label set (if not label is set, the displayname is used):

Screenshot 2024-10-24 at 11 24 06

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
+++ b/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
@@ -396,7 +396,9 @@ public class NotificationCenter {
 
             if (privacy.isDisplayContact()) {
                 builder.setContentTitle(dcChat.getName());
-                builder.setSubText(accountTag);
+                if (!TextUtils.isEmpty(accountTag)) {
+                    builder.setSubText(accountTag);
+                }
             }
 
             // if privacy allows, for better accessibility,
@@ -540,7 +542,7 @@ public class NotificationCenter {
                     .setContentTitle("Delta Chat") // content title would only be used on SDK <24
                     .setContentText("New messages") // content text would only be used on SDK <24
                     .setContentIntent(getOpenChatlistIntent(accountId));
-                  if (privacy.isDisplayContact()) {
+                  if (privacy.isDisplayContact() && !TextUtils.isEmpty(accountTag)) {
                     summary.setSubText(accountTag);
                   }
                   notificationManager.notify(String.valueOf(accountId), ID_MSG_SUMMARY, summary.build());
--- a/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
+++ b/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
@@ -1,5 +1,7 @@
 package org.thoughtcrime.securesms.notifications;
 
+import static org.thoughtcrime.securesms.connect.DcHelper.CONFIG_PRIVATE_TAG;
+
 import android.app.Notification;
 import android.app.NotificationChannel;
 import android.app.NotificationChannelGroup;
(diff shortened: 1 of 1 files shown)


r10s (approved):
very nice!

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -387,10 +389,14 @@ public void addNotification(int accountId, int chatId, int msgId) {
...
-            String accountAddr = dcContext.getConfig("addr");
+            String accountTag = dcContext.getConfig(CONFIG_PRIVATE_TAG);
+            if (accountTag.isEmpty() && context.dcAccounts.getAll().length > 1) {
+                accountTag = dcContext.getName();
+            }
+
             if (privacy.isDisplayContact()) {
                 builder.setContentTitle(dcChat.getName());
-                builder.setSubText(accountAddr);
+                builder.setSubText(accountTag);
```
r10s, line 399: otherwise, a 3rd, empty line may be added, see NotificationCompat.java source.

it's also better style to just not set in case nothing is to set.

```suggestion
                if (!TextUtils.isEmpty(accountTag)) {
                    builder.setSubText(accountTag);
                }
```

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -535,7 +541,7 @@ public void addNotification(int accountId, int chatId, int msgId) {
                     .setContentText("New messages") // content text would only be used on SDK <24
                     .setContentIntent(getOpenChatlistIntent(accountId));
                   if (privacy.isDisplayContact()) {
-                    summary.setSubText(accountAddr);
+                    summary.setSubText(accountTag);
```
r10s, line 544: same here:

```suggestion
                if (!TextUtils.isEmpty(accountTag)) {
                    builder.setSubText(accountTag);
                }
```

(could be combined with the outer `if`, this seems not possible on github)

r10s (commented):
we need to check for empty, however

r10s:
oh, the submodule sneaked in ... np, i was only wondering what happens on main :)

needed to update to ~~1.148.3~~ 1.148.4 anyways

adbenitez:
> oh, the submodule sneaked in ... np, i was only wondering what happens on main :)

oops!!! sorry, it seems it was when the usual git-control tool I use inside emacs broke after an upgrade and somehow I messed it up while doing the commit on the command line, accidentally used  a cli shortcut to "commit all" without manually adding the desired file/changes :/

r10s:
"commit all" is considered harmful ... ftr, since $known_accident, i disabled `git -a` in `.zshrc` as follows:

```
# git -a is considered harmful                                                  
git() {                                                                         
  for arg                                                                       
  do                                                                            
    if [[ $arg == -a* || $arg == -[^-]*a* ]]                                    
    then                                                                        
      echo "\033[0;35mdo not use \"git -a\" to avoid accidental check ins! use eg. \"git -p\" instead\033[0m"
      return 1                                                                  
    fi                                                                          
  done                                                                          
  command git "$@"                                                              
} 
``` 

a bit clumsy, it catches a bit too much, but it is good enough for me :)

PR #3384, opened 2024-10-24
do not start foreground service in foreground

r10s:
the foreground service was introduced unconditionally in #3312, however turns out to be annoying and flickering
if the app is actually in foreground.

the service _might_ be needed in this case only
if the app is about going to background during fetch, but in this case, we assume,
that we have some 10 seconds time still.
@Hocuri iirc, you wanted to double check that :)

in any case,
this is still an improvement to the released apps, that _never_ show a foreground service.

(for the naming, it could also be maybeStart() or so, but this was also the case before as it checks if already started, so i do not mind. moving the check outside will add code, and makes tweaking harder. also, as the service will potentially be used also from other places, eg. when it comes to huawei, so i'd be pragmatic here)

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/service/FetchForegroundService.java
+++ b/src/main/java/org/thoughtcrime/securesms/service/FetchForegroundService.java
@@ -13,6 +13,7 @@ import androidx.core.content.ContextCompat;
 
 import org.thoughtcrime.securesms.ApplicationContext;
 import org.thoughtcrime.securesms.R;
+import org.thoughtcrime.securesms.connect.ForegroundDetector;
 import org.thoughtcrime.securesms.notifications.FcmReceiveService;
 import org.thoughtcrime.securesms.notifications.NotificationCenter;
 import org.thoughtcrime.securesms.util.Util;
@@ -23,6 +24,11 @@ public final class FetchForegroundService extends Service {
   private static Intent service;
 
   public static void start(Context context) {
+    ForegroundDetector foregroundDetector = ForegroundDetector.getInstance();
+    if (foregroundDetector != null && foregroundDetector.isForeground()) {
+      return;
+    }
+
     GenericForegroundService.createFgNotificationChannel(context);
     synchronized (SERVICE_LOCK) {
       if (service == null) {


src/main/java/org/thoughtcrime/securesms/service/FetchForegroundService.java
```diff
@@ -13,6 +13,7 @@
 
 import org.thoughtcrime.securesms.ApplicationContext;
 import org.thoughtcrime.securesms.R;
+import org.thoughtcrime.securesms.connect.ForegroundDetector;
```
adbenitez, line 16: nice! we had a foreground detector!

adbenitez (approved):

Hocuri:
> the service _might_ be needed in this case only
> if the app is about going to background during fetch, but in this case, we assume,
> that we have some 10 seconds time still.
> @Hocuri iirc, you wanted to double check that :)

No, unfortunately not - in order to have the up to 20 seconds (so, 10 seconds to be on the very safe side), we would need to block in `onMessageReceived()`, which we don't currently.

Full explanation (or, the way I understood the documentation, but I'm confident that it's correct) is [here](https://github.com/deltachat/deltachat-android/pull/3312#discussion_r1770647114) and [here](https://github.com/deltachat/deltachat-android/pull/3312#discussion_r1771274548), summary:
- By blocking in `onMessageReceived()` (by calling a blocking core function or by calling `wait()` on an Object that is `notify()`ed when an event arrives), we would be allowed to take up to 20 seconds to process a message. Since there can be delays everywhere, the documentation says that one should only take up to 10 seconds.
- If we want to avoid being killed as soon as returning from `onMessageReceived()`, we need a foreground service.

Hocuri:
OTOH, the PR here is simple enough and solves an actual problem - one would need to be quite unlucky that the app goes into background and is destroyed immediately right when a message arrives.

I do assume though that there will be some, very very rare missed notifications on very slow networks.

What about starting the foreground service in ForegroundDetector.onPause() if the connectivity is Working? That would be a lot simpler that [what I proposed in the other PR](https://github.com/deltachat/deltachat-android/pull/3312#discussion_r1771274548) and should fix most race conditions. (it can happen that the app is killed without onStop() being called first, so, it needs to be onPause(), not onStop()).

r10s:
the 20/10 seconds are about when onMessageReceived() is called in background, or? do we have hints that apps are killed immediately when switching from foreground to background? https://developer.android.com/about/versions/oreo/background says _" When an app goes into the background, it has a window of several minutes in which it is still allowed to create and use services"_ - assuming the reports of missing notifications are when the app is in background for some time already, this is probably not the issue at hand.

> What about starting the foreground service in ForegroundDetector.onPause() if the connectivity is Working? That would be a lot simpler that https://github.com/deltachat/deltachat-android/pull/3312#discussion_r1771274548 and should fix most race conditions. (it can happen that the app is killed without onStop() being called first, so, it needs to be onPause(), not onStop()).

the "end condition" needs to be different then: DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE is only added when fetch was initiated by `dc_accounts_background_fetch()`. we would need to wait for state=no-longer-working or so.

doable, and maybe makes sense - but let's see if things aren't already fine :)

Hocuri:
> the 20/10 seconds are about when onMessageReceived() is called in background, or?

Yes, exactly.

> do we have hints that apps are killed immediately when switching from foreground to background?

No.

> but let's see if things aren't already fine :)

:crossed_fingers:

PR #3391, opened 2024-10-25
improve logging: always log events for accId=0

adbenitez:
That "global"(accid=0) account manager events where never logged at all was a bug

This PR also log other accounts' events if the developer mode is enabled, his can be useful to debug multi-account problems like backgroundFetch()

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/connect/DcEventCenter.java
+++ b/src/main/java/org/thoughtcrime/securesms/connect/DcEventCenter.java
@@ -12,7 +12,6 @@ import com.b44t.messenger.DcEvent;
 import org.thoughtcrime.securesms.ApplicationContext;
 import org.thoughtcrime.securesms.R;
 import org.thoughtcrime.securesms.service.FetchForegroundService;
-import org.thoughtcrime.securesms.util.Prefs;
 import org.thoughtcrime.securesms.util.Util;
 
 import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


src/main/java/org/thoughtcrime/securesms/connect/DcEventCenter.java
```diff
@@ -187,7 +188,26 @@ public long handleEvent(@NonNull DcEvent event) {
         return 0;
     }
 
+    boolean devMode = Prefs.isDeveloperModeEnabled(context);
+    String logPrefix = devMode? "[accId="+accountId + "] " : "";
+
     if (accountId != context.dcContext.getAccountId()) {
```
r10s, line 194: `devMode` and `logPrefix` are needed only inside the branch and could be final:

```suggestion
    if (accountId != context.dcContext.getAccountId()) {
        final boolean devMode = Prefs.isDeveloperModeEnabled(context);
        final String logPrefix = devMode? "[accId="+accountId + "] " : "";
```

r10s (approved):
lgtm.

ftr, on iOS we're always logging all account unconditionally (with accountId), that might be an option for android as well - eg. when hunting bugs, it may be important to know that another account is blocking or so. and if we get a log from random internet, we would not know that a second account even exist.
but i am fine with both, this PR is definitely an improvement

PR #3398, opened 2024-10-29
change realtime preference position

adbenitez:


src/main/res/values/strings.xml
```diff
@@ -447,7 +447,7 @@
     <!-- title shown above a list contacts where one should be selected (eg. when a webxdc attempts to send a message to a chat) -->
     <string name="send_message_to">Send Message to…</string>
     <string name="enable_realtime">Realtime Webxdc Channels</string>
-    <string name="enable_realtime_explain">Enable Webxdc realtime APIs to create direct connections between devices.</string>
+    <string name="enable_realtime_explain">Enable Webxdc realtime APIs to create direct connections between devices. If this is enabled, your chat partners may be able to see your IP address.</string>
```
r10s, line 450: this is a bit shorter - and also removes the advice "enable ... to ...", which is a bit off as it will be enabled by default in the future.

```suggestion
    <string name="enable_realtime_explain">Webxdc realtime APIs allow direct connections between devices. If enabled, chat partners may be able to see your IP address.</string>
```

nb: we're otherwise not talking about much "webxdc" inside Delta Chat, but more about "apps" - but that is a bit forth and back anyways :) so, as being advanced, probably okay to call the option "Realtime Webxdc Channels" - alternative may be "Realtime App Channels", however, i do not really like that and also  do not have a really good other suggestion here

r10s (approved):
i would adapt the string as the advice "enable to ..." is misleading as it will be already enabled with https://github.com/deltachat/deltachat-core-rust/pull/6125

otherwise, lgtm

adbenitez:
feel free to suggest new strings for the explanation, I would like it to transmit the idea "some apps might not work properly if you disable this" since it is more damage than helpful if user disable it just because it seems like an useless feature that just "make apps faster" or "reveals my IP why would I want that?"

src/main/res/values/strings.xml
adbenitez: > nb: we're otherwise not talking about much "webxdc" inside Delta Chat,

I kept it because the title is also saying "webxdc" but otherwise I am up for renaming it, just saying "apps" is a bit broad/unclear tho

adbenitez:
see https://github.com/deltachat/deltachat-core-rust/pull/6125 for more context

src/main/res/values/strings.xml
```diff
@@ -447,7 +447,7 @@
     <!-- title shown above a list contacts where one should be selected (eg. when a webxdc attempts to send a message to a chat) -->
     <string name="send_message_to">Send Message to…</string>
     <string name="enable_realtime">Realtime Webxdc Channels</string>
-    <string name="enable_realtime_explain">Enable Webxdc realtime APIs to create direct connections between devices.</string>
+    <string name="enable_realtime_explain">Webxdc realtime APIs allow direct connections between devices. If enabled, chat partners may be able to see your IP address.</string>
```
Hocuri, line 450:
```suggestion
    <string name="enable_realtime_explain">This allows webxdc apps to create direct connections between devices. If enabled, chat partners may be able to see your IP address.</string>
```

Better not use the technical term "API here". 

I'm wondering if we should just add "If disabled, some apps may not work."

hpk42: i'd say it's clear that if you disable this option that realtime apps may not function.  
here is my suggestion (avoiding "webxdc realtime" which really does not tell users much if they don't know about it yet). 

```suggestion
    <string name="enable_realtime_explain">Enable real-time support for apps shared in chats. If enabled, chat partners may be able to discover your IP address after you started an app.</string>
```

r10s: note, that this string is only the _subtitle_, the _option title_ itself still has "webxdc" and "realtime" in, i think, we need to change both (it is only github, that does not allow to adapt unchanged lines)

so, if we take the suggestion from @hpk42 , maybe change the title from `Realtime Webxdc Channels` to `Support Real-Time Apps`?

r10s (approved):

Hocuri (approved):

adbenitez:
I added a dialog if real-time is disabled and the app needs it, to give a chance user can quickly fix problems if they disabled the setting without properly understanding scared by the "leaks IP" 

@r10s, @Hocuri improvements to the dialog message welcome

image

Hocuri (commented):
I just did a quick user test on this (we really ought to do those more often! though I was lucky to be around some non-technical users).

Both my users thought that "real-time support" is a chatbot that supports you in real-time😅

src/main/res/values/strings.xml
```diff
@@ -446,8 +447,9 @@
     <string name="send_file_to">Send \"%1$s\" to…</string>
     <!-- title shown above a list contacts where one should be selected (eg. when a webxdc attempts to send a message to a chat) -->
     <string name="send_message_to">Send Message to…</string>
-    <string name="enable_realtime">Realtime Webxdc Channels</string>
-    <string name="enable_realtime_explain">Enable Webxdc realtime APIs to create direct connections between devices.</string>
+    <string name="enable_realtime">Support Real-Time Apps</string>
+    <string name="enable_realtime_explain">Enable real-time support for apps shared in chats. If enabled, chat partners may be able to discover your IP address when you start an app.</string>
+    <string name="realtime_disabled_warning">Real-time support is disabled, this app may not work properly. Do you want to enable it? Chat partners may be able to discover your IP address</string>
```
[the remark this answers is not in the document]
adbenitez: this needs to be escaped

r10s (approved):

r10s:
ftr, i we want to come back to a dialog later on, we should remember this commit: https://github.com/deltachat/deltachat-android/pull/3398/commits/d986e83cc15fae458ba52e385d4ff449ca300624
(it was taken out of this PR as it may interfere with apps dealing nicely with realtime not available. and also may annoy users by nudging into re-enabling, which we maybe do not want. esp. for apps that work without realtime as written in the spec. we can reconsider later, if we have more experience and there are really non-testing-only realtime apps out)

adbenitez:
also notice that apps can make the real-time optional in their settings and then control it inside the app without causing the dialog to prompt every time since it is only shown if the app tries to use realtime,

alternatively, the API could be improved so apps realize realtime is disabled and don't try to use it, currently apps are totally blind trying to use realtime without realizing it is not available, it is good enough for apps that don't need it but that is not all apps

r10s:
i am not against a dialog, there are a lot of _"may"_ in my _"ftr"_ :)

PR #3400, opened 2024-10-29
improve msg_action_button

### LINKED ISSUE #3395: mini-apps' "start" button is hard to tap
it would be better to show as primary button like on desktop

adbenitez:
close #3395

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationItem.java
@@ -111,6 +111,7 @@ public class ConversationItem extends BaseConversationItem
   protected ViewGroup              contactPhotoHolder;
   private   ViewGroup              container;
   private   Button                 msgActionButton;
+  private   Button                 showFullButton;
 
   private @NonNull  Stub<ConversationItemThumbnail> mediaThumbnailStub;
   private @NonNull  Stub<AudioView>                 audioViewStub;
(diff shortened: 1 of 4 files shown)


r10s:
fine for that "start..." button. 

and also for the other buttons ("show full message", "download", "downloading") enlarging the click area sin a good idea, however, apart from that, we should not change the layout there - they are not primary buttons, and would get too much focus by that. eg. the "show full message" button is very often superfluous as we do not exactly know if the is needed or not, and in doubt, add the button. so putting additional focus on that is not needed, that is more a "more ..." button and should stay as is, layout wise

Hocuri:
Looking forward to a better start button on webxdcs!

I agree with @r10s that for "show full message" it would be nicer to keep the layout as-is, though for me it's not that important. But the "Download" button _is_ the primary button on this message and I do think that it would be nice to emphasize it similarly to the "Start" button. Not sure about the "Downloading..." and "Download failed" button because they are not primary buttons, but it's nicer to make them similar to the "Download" button in order to reduce flickering when the button changes from "Download" to "Downloading".

r10s:
yes, i agree that "Download" is a primary button for these bubbles bubble. but i may still look a bit weird, as there is just no other content in, same with having a primary button change title result in more flickering. therefore, as we anyway need two styles, still my suggestion to change the issue #3395 on point and only change the "start" button as proposed in the issue, not unrelated buttons.

Hocuri:
+1 for "Show full message".

But I don't think that this looks weird:

Screenshot_20241030-164122

and, in fact, I think it looks a lot better than it did before.

r10s:
i agree, it is fine :)

adbenitez:
I will need to add a extra button then for show-full-message and hide one and show the other in dependency, I knew this was coming, just tried to see if I could get away with this without the extra work :sweat_smile:

r10s:
yeah, it is probably easier to add a second button to the xml than to try to change the style programatically

adbenitez:
this is the results, nicer indeed:

adbenitez:
oops! I shared screenshot but forgot to push! please review now

Hocuri (approved):

PR #3405, opened 2024-10-31
add "realtime is disabled" dialog

adbenitez:
also add `window.webxdc.isRealtimeSupported()` so apps that don't require the feature to work can avoid the dialog pop up and just keep working without real-time, to mitigate annoyances when apps work without realtime and use it only to add extra features opportunistically 

`window.webxdc.isRealtimeSupported()`  makes sense in general as some webxdc implementations besides allowing to disable it, might just not support realtime channels feature, as it is more work to implement than the simple basic API, giving the client a way to communicate that the feature is missing without silently doing nothing, so apps can realize the lack of support and adapt or show a warning if they can't work without realtime

I guess this requires the Webxdc spec to be updated as well in some our repos

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -449,7 +449,7 @@
     <string name="send_message_to">Send Message to…</string>
     <string name="enable_realtime">Real-Time Apps</string>
     <string name="enable_realtime_explain">Enable real-time connections for apps shared in chats. If enabled, chat partners may be able to discover your IP address when you start an app.</string>
-    <string name="realtime_disabled_warning">\"Real-time\" feature is disabled, this app may not work properly. Do you want to enable it? Chat partners may be able to discover your IP address</string>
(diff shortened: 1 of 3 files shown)


src/main/res/values/strings.xml
```diff
@@ -448,6 +449,7 @@
     <string name="send_message_to">Send Message to…</string>
     <string name="enable_realtime">Real-Time Apps</string>
     <string name="enable_realtime_explain">Enable real-time connections for apps shared in chats. If enabled, chat partners may be able to discover your IP address when you start an app.</string>
+    <string name="realtime_disabled_warning">\"Real-time\" feature is disabled, this app may not work properly. Do you want to enable it? Chat partners may be able to discover your IP address</string>
```
r10s, line 452: call the feature here as it is called in the settings. and do not forget the final, serious dot.

```suggestion
    <string name="realtime_disabled_warning">\"Real-Time Apps\" feature is disabled, this app may not work properly. Do you want to enable it? Chat partners may be able to discover your IP address.</string>
```

adbenitez:
not priority for now

Issue #3392, opened 2024-10-26
"Add (As) Second Device" does not show progress


r10s:
when adding an android as a second device of iOS (iOS sends, android receives), the progress is shown on iOS, but not on android.
instead, android stays at "0%" all the time - still, the transfer succeeded.

tested with ios1.148.1 and android 1.48.1, both core 1.148.4. the progress used to be working nicely before.

![IMG_0045](https://github.com/user-attachments/assets/5b284dc1-0fc8-41ad-867b-112b81c5a56f)

~~did not test other combinations, it is also unclear if that is an android or a core issue.~~ EDIT: i also tested the other way round, iOS as receiver shows the progress then, android as sender not, so **it seems to be an android issue**:

![IMG_0046](https://github.com/user-attachments/assets/ac730ed6-c417-4c45-bb88-6e954c4c48a9)

adbenitez:
I could not reproduce this, tried: 
- from Desktop `core-v1.142.12` to DC Android `core-v1.148.4` and it works fine
- from ArcaneChat `core-v1.148.4` to DC Android `core-v1.148.4` and it also works as expected

![image](https://github.com/user-attachments/assets/507ba46a-ba9f-41f5-8406-da91e693cc6b)

r10s:
thanks for testing, good to know it is not a general thing, so, this is no blocker anymore :)

i tried with an older android7 phone, maybe he progress bar is broken only there. 

still i am sure, it was displaying correctly there before as this is my "only"  working test phone - and add-second-device does not work in the emulator for me.

Hocuri:
As a quick mitigation that would be nice to have anyway, we should simply show "Transferring..." instead of "Transferring... 0%" as long as the progress is 0%, similar to what I started at https://github.com/deltachat/deltachat-android/pull/3337/

Simon-Laux:
transfer progress 0 is connecting devices, we should tell users to look at trouble shooting if this takes long, see https://github.com/deltachat/deltachat-desktop/issues/4420

r10s:
android  already now shows unconditionally a hint about checking wifi, so that's fine so far, and not the issue at hand.

![image_2025-01-11_14-18-29](https://github.com/user-attachments/assets/4a5f1600-ec85-4e07-bff3-09825841ab49)

for the issue at hand: i just retried, that is working now (1.48.3 using core 1.148.6), maybe a temporary hickup, idk .)

PR #3409, opened 2024-11-04
Don't show "0%" at the beginning when being added as second device

Hocuri:
Mitigation for https://github.com/deltachat/deltachat-android/issues/3392, follow-up for https://github.com/deltachat/deltachat-android/pull/3337.

When creating a chatmail account, when logging in to an existing email account, and when importing a backup from the Welcome Activity, we're still showing "0%" even after this PR. So, an alternative to this PR would be to make `percent > 0 ? String.format(Util.getLocale(), " %d%%", percent) : "";` into a function and call it from all these places, too, and never show "0%" anywhere ever again.

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/qr/BackupReceiverFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/qr/BackupReceiverFragment.java
@@ -84,7 +84,8 @@ public class BackupReceiverFragment extends Fragment implements DcEventCenter.Dc
             } else if (permille < 1000) {
                 percent = permille/10;
                 percentMax = 100;
-                statusLineText = getString(R.string.transferring) + String.format(Util.getLocale(), " %d%%", percent);
+                String formattedPercent = percent > 0 ? String.format(Util.getLocale(), " %d%%", percent) : "";
+                statusLineText = getString(R.string.transferring) + formattedPercent;
                 hideSameNetworkHint = true;
             } else if (permille == 1000) {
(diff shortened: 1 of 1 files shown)


r10s (approved):

src/main/java/org/thoughtcrime/securesms/qr/BackupReceiverFragment.java
```diff
@@ -84,7 +84,8 @@ public void handleEvent(@NonNull DcEvent event) {
             } else if (permille < 1000) {
                 percent = permille/10;
                 percentMax = 100;
-                statusLineText = getString(R.string.transferring) + String.format(Util.getLocale(), " %d%%", percent);
+                String formattedPercent = percent > 0 ? String.format(Util.getLocale(), " %d%%", percent) : "";
+                statusLineText = getString(R.string.transferring) + formattedPercent;
```
adbenitez, line 88: btw, not related to this PR but this string concatenation is not good, for RTL languages it will not be mirrored properly :thinking:

adbenitez (approved):

src/main/java/org/thoughtcrime/securesms/qr/BackupReceiverFragment.java
Hocuri: The good news is that it's not a sentence, so that it's not that bad if the order is wrong for RTL messages. I.e. "10% Transferring..." and "Transferring... 10%" is both right.

Hocuri:
Merging since it's a mitigation for a bug that's possibly a regression

PR #3411, opened 2024-11-04
remove old emojis

### LINKED ISSUE #1993: Marked single emoji not exchanged by pasted clipboard text in message input field
- Android version:
Android 11.

- Device:
Fairphone FP2 (Lineage OS 18.1 build RQ3A.210705.001 from 2021-07-23).

- Delta Chat version:
1.20.5 (nightly build).

- Expected behavior:
When marked, a single emoji character in the message input field shall be overwritten when text from the clipboard is pasted.

- Actual behavior:
The marked emoji in the message input field is not overwritten by text from the clipboard when pasted. When the user taps the send icon afterwards to send the message, text from the clipboard is sent instead of the emoji character.

-- Open the chat "Saved Messages".
-- Enter "Test" in the text input field and tap the send icon to send the message.
-- Mark the message that has been sent and tap the "Copy to clipboard" icon from the top bar to copy the text to the clipboard.
-- Enter any emoji character in the text input field and long tap the emoji character to mark it.
-- Select "Paste" to exchange the emoji character with the text from the clipboard. (You will see that nothing has changed.)
-- Tap the send icon to send the message. (Result: The text from the clipboard has been sent instead of the emoji character.)

- Screenshots:
grafik

- Logs:
N/A.

### LINKED ISSUE #3399: problems with reactions and old built-in emojis
the old built-in emojis cause some reactions to be displayed 3 times for some user:

image

I tried to test it myself but reacting with 😮‍💨 causes this other problematic reaction for me:

image

the user confirmed that after disabling built-in emojis, the triple-emoji problem disappeared for them

adbenitez:
use modern `emoji2-emojipicker` instead (already in use by reactions picker)

close #3399
close #1993

To Do:
- [x] remove old emoji images assets
- [x] remove "use system emojis" preference
- [x] use emoji2-emojipicker instead of old emoji picker
- [x] implement on-click event in new emoji picker
- [x] fix emoji text views
- [x] delete old emoji code
- [x] fix emoji color for reaction pills
- [x] enlarge font size when text is only a few emojis

image


adbenitez:
image

adbenitez:
caveats: 

- after a lot of searching for solutions I didn't find any proper&easy way to detect emojis in text in java for the font size scaling effect, so if you send a single emoji the message will look the same as any other message, hopefully core could help with this in the future
- the emoji picker doesn't have a delete button, this could be added eventually when we add  a row to switch between emojis and sticker picker, that is how it is in whatsapp

adbenitez:
ok, I added the emoji scaling feature using a regex based on data from:
https://util.unicode.org/UnicodeJsps/list-unicodeset.jsp?a=%5B%3AEmoji%3DYes%3A%5D&g=&i=

src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
```diff
@@ -0,0 +1,82 @@
...
+    float scale = text != null ? getTextScale(text.toString()) : 1;
+    super.setTextSize(TypedValue.COMPLEX_UNIT_PX, originalFontSize * scale);
+    super.setText(text, type);
+  }
+
+    
+  private float getTextScale(String text) {
+    if (text.length() > 16) {
+      return 1;
+    }
```
Hocuri, line 47: Just wanted to point out: Very long emojis like :scotland: won't be shown big because of this. I do think that the code is fine as-is, because, generally only rather rare emojis are very long, and for those it's not important to be shown big. Also, it's better for performance to return early whenever possible. If someone complains that some emojis are not shown big, we can adapt.

Hocuri (commented):
Minus 2.6MB, plus fixing an issue :tada:!

image

Partial review.

src/main/java/org/thoughtcrime/securesms/components/InputPanel.java
```diff
@@ -304,25 +292,30 @@ public void onSuccess(Void result) {
...
+  @Override
+  public void onHidden() {
+    emojiToggle.setToMedia();
   }
 
-  public void releaseRecordingLock() {
-    microphoneRecorderView.unlockAction();
+  @Override
+  public void onEmojiPicked(String emoji) {
+    composeText.append(emoji);
```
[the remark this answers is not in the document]
adbenitez: fixed it, good catch! I was thinking append would append at the current cursor position :rofl:

src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
```diff
@@ -0,0 +1,82 @@
...
...👦👧🧑👱👨🧔👩🧓👴👵🙍🙎🙅🙆💁🙋🧏🙇🤦🤷👮🕵💂🥷👷🫅🤴👸👳👲🧕🤵👰🤰🫃🫄🤱👼🎅🤶🦸🦹🧙-🧟🧌💆💇🚶🧍🧎🏃💃🕺🕴👯🧖🧗🤺🏇⛷🏂🏌🏄🚣🏊⛹🏋🚴🚵🤸🤼-🤾🤹🧘🛀🛌👭👫👬💏💑🗣👤👥🫂👪👣🦰🦱🦳🦲🐵🐒🦍🦧🐶🐕🦮🐩🐺🦊🦝🐱🐈🦁🐯🐅🐆🐴🫎🫏🐎🦄🦓🦌🦬🐮🐂-🐄🐷🐖🐗🐽🐏🐑🐐🐪🐫🦙🦒🐘🦣🦏🦛🐭🐁🐀🐹🐰🐇🐿🦫🦔🦇🐻🐨🐼🦥🦦🦨🦘🦡🐾🦃🐔🐓🐣-🐧🕊🦅🦆🦢🦉🦤🪶🦩🦚🦜🪽🪿🐸🐊🐢🦎🐍🐲🐉🦕🦖🐳🐋🐬🦭🐟-🐡🦈🐙🐚🪸🪼🐌🦋🐛-🐝🪲🐞🦗🪳🕷🕸🦂🦟🪰🪱🦠💐🌸💮🪷🏵🌹🥀🌺-🌼🌷🪻🌱🪴🌲-🌵🌾🌿☘🍀-🍃🪹🪺🍄🍇-🍍🥭🍎-🍓🫐🥝🍅🫒🥥🥑🍆🥔🥕🌽🌶🫑🥒🥬🥦🧄🧅🥜🫘🌰🫚🫛🍞🥐🥖🫓🥨🥯🥞🧇🧀🍖🍗🥩🥓🍔🍟🍕🌭🥪🌮🌯🫔🥙🧆🥚🍳🥘🍲🫕🥣🥗🍿🧈🧂🥫🍱🍘-🍝🍠🍢-🍥🥮🍡🥟-🥡🦀🦞🦐🦑🦪🍦-🍪🎂🍰🧁🥧🍫-🍯🍼🥛☕🫖🍵🍶🍾🍷-🍻🥂🥃🫗🥤🧋🧃🧉🧊🥢🍽🍴🥄🔪🫙🏺🌍-🌐🗺🗾🧭🏔⛰🌋🗻🏕🏖🏜-🏟🏛🏗🧱🪨🪵🛖🏘🏚🏠-🏦🏨-🏭🏯🏰💒🗼🗽⛪🕌🛕🕍⛩🕋⛲⛺🌁🌃🏙🌄-🌇🌉♨🎠🛝🎡🎢💈🎪🚂-🚊🚝🚞🚋-🚎🚐-🚙🛻🚚-🚜🏎🏍🛵🦽🦼🛺🚲🛴🛹🛼🚏🛣🛤🛢⛽🛞🚨🚥🚦🛑🚧⚓🛟⛵🛶🚤🛳⛴🛥🚢✈🛩🛫🛬🪂💺🚁🚟-🚡🛰🚀🛸🛎🧳⌛⏳⌚⏰-⏲🕰🕛🕧🕐🕜🕑🕝🕒🕞🕓🕟🕔🕠🕕🕡🕖🕢🕗🕣🕘🕤🕙🕥🕚🕦🌑-🌜🌡☀🌝🌞🪐⭐🌟🌠🌌☁⛅⛈🌤-🌬🌀🌈🌂☂☔⛱⚡❄☃⛄☄🔥💧🌊🎃🎄🎆🎇🧨✨🎈-🎋🎍-🎑🧧🎀🎁🎗🎟🎫🎖🏆🏅🥇-🥉⚽⚾🥎🏀🏐🏈🏉🎾🥏🎳🏏🏑🏒🥍🏓🏸🥊🥋🥅⛳⛸🎣🤿🎽🎿🛷🥌🎯🪀🪁🔫🎱🔮🪄🎮🕹🎰🎲🧩🧸🪅🪩🪆♠♥♦♣♟🃏🀄🎴🎭🖼🎨🧵🪡🧶🪢👓🕶🥽🥼🦺👔-👖🧣-🧦👗👘🥻🩱-🩳👙👚🪭👛-👝🛍🎒🩴👞👟🥾🥿👠👡🩰👢🪮👑👒🎩🎓🧢🪖⛑📿💄💍💎🔇-🔊📢📣📯🔔🔕🎼🎵🎶🎙-🎛🎤🎧📻🎷🪗🎸-🎻🪕🥁🪘🪇🪈📱📲☎📞-📠🔋🪫🔌💻🖥🖨⌨🖱🖲💽-📀🧮🎥🎞📽🎬📺📷-📹📼🔍🔎🕯💡🔦🏮🪔📔-📚📓📒📃📜📄📰🗞📑🔖🏷💰🪙💴-💸💳🧾💹✉📧-📩📤-📦📫📪📬-📮🗳✏✒🖋🖊🖌🖍📝💼📁📂🗂📅📆🗒🗓📇-📎🖇📏📐✂🗃🗄🗑🔒🔓🔏-🔑🗝🔨🪓⛏⚒🛠🗡⚔💣🪃🏹🛡🪚🔧🪛🔩⚙🗜⚖🦯🔗⛓🪝🧰🧲🪜⚗🧪-🧬🔬🔭📡💉🩸💊🩹🩼🩺🩻🚪🛗🪞🪟🛏🛋🪑🚽🪠🚿🛁🪤🪒🧴🧷🧹-🧻🪣🧼🫧🪥🧽🧯🛒🚬⚰🪦⚱🧿🪬🗿🪧🪪🏧🚮🚰♿🚹-🚼🚾🛂-🛅⚠🚸⛔🚫🚳🚭🚯🚱🚷📵🔞☢☣⬆↗➡↘⬇↙⬅↖↕↔↩↪⤴⤵🔃🔄🔙-🔝🛐⚛🕉✡☸☯✝☦☪☮🕎🔯🪯♈-♓⛎🔀-🔂▶⏩⏭⏯◀⏪⏮🔼⏫🔽⏬⏸-⏺⏏🎦🔅🔆📶🛜📳📴♀♂⚧✖➕-➗🟰♾‼⁉❓-❕❗〰💱💲⚕♻⚜🔱📛🔰⭕✅☑✔❌❎➰➿〽✳✴❇©®™🔟-🔤🅰🆎🅱🆑-🆓ℹ🆔Ⓜ🆕🆖🅾🆗🅿🆘-🆚🈁🈂🈷🈶🈯🉐🈹🈚🈲🉑🈸🈴🈳㊗㊙🈺🈵🔴🟠-🟢🔵🟣🟤⚫⚪🟥🟧-🟩🟦🟪🟫⬛⬜◼◻◾◽▪▫🔶-🔻💠🔘🔳🔲🏁🚩🎌🏴🏳🇦-🇿\uD83E\uDD89\uD83E\uDD8F\uD83E\uDDBE\uD83E\uDDC6\uD83E\uddcd\uD83E\udddf\uD83E\ude99]|#️⃣|\\*️⃣|0️⃣|1️⃣|2️⃣|3️⃣|4️⃣|5️⃣|6️⃣|7️⃣|8️⃣|9️⃣)+");
```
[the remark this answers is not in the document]
adbenitez: added :+1:

src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
adbenitez: wow an emoji with a length of 7, still the reason it is not shown big is because the regex doesn't support it, but it is true that more than two 🏴󠁧󠁢󠁳󠁣󠁴󠁿 would not be shown big even if supported by the regex,

I tried to optimize, a bit scared about performance impact of counting the graphemes iterating over the characters and then the regex, but maybe increasing to 21 makes sense, 

in general I think the emojification rule is a bit over the board, I would simplify it to making 1-3 emojis big and the rest normal, the size difference is really subtle for 4-8 and doesn't make much sense

r10s:
> Minus 2.6MB, plus fixing an issue 🎉!

yeah, i was also curious at the same time as you :) - can confirm the 2.6MB smaller APK, excellent!

r10s:
ftr, this is how it looks like on and old android7, where the system emoji were still in "blob style" - one can see quite nicely there, how the gaps are filled by he new-style-emoji - what a nice improvement, compared to having just nothing before:

src/main/res/layout/reactions_pill.xml
```diff
@@ -12,11 +12,13 @@
     android:paddingEnd="7dp"
     android:gravity="center">
 
-  <org.thoughtcrime.securesms.components.emoji.EmojiTextView
+  <!-- to avoid default transparency in text color that makes emoji transparent -->
```
r10s, line 15: took me a moment to get, what the comment is about - and that even though the changed line was hilited on github :)

i suggest to either remove the comment or say:

```suggestion
  <!-- android:textColor is set to avoid default transparency making emoji transparent -->
```

r10s (approved):
what a cleanup!

i have not much additional to add, apart from saying that it works quite nicely on first test, great job!

with removing the switch "use system emojis", we also removed the option to get rid of the emoji button from the keyboard. i am fine with that, if ppl complain in masses, we can iterate (_some_ will complain for sure :)

and, if issues pop up later on (i would assume so :), it is also a good timing as 1.48 is just released

src/main/res/layout/reactions_pill.xml
adbenitez: oops! yes, the reason was that the comment was initially right above the android:textColor line but that caused a compilation error due to not allowed in xml to put comments inside a tag declaration and I moved it outside to avoid the error and keep testing and forgot to update the comment, thanks a lot for pointing out!

src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
adbenitez: (tried only increasing 1-3 and have to admit that then 4-6 feels too small, so I left the behavior unchanged :)

Hocuri: > in general I think the emojification rule is a bit over the board, I would simplify it to making 1-3 emojis big and the rest normal, the size difference is really subtle for 4-8 and doesn't make much sense

+1

PR #3410, opened 2024-11-04
fix: Accept the deprecated 3 as an alias for 2

Hocuri:
fix #3408 by automatically changing a 3 to 2.

3 is a [deprecated alias for 2](https://github.com/deltachat/deltachat-core-rust/blob/37ca9d7319dabcf1cd3cf0bb75c2b4c062aa3bf0/src/login_param.rs#L18-L41), so all UIs now need to handle both values. I think it wasn't a good idea to deprecate 3 in the first place, but it's not worth it going back now.

I tested this PR like this:
- Without this PR:
  - Set certificate checks to "Accept invalid Certificates" in the settings -> `imap_certificate_checks` is 2
  - Import a backup from Desktop with "Accept invalid Certificates"
  - Open "Password and Account" in the settings -> the app crashes
- With this PR:
  - Open "Password and Account" in the settings -> the app doesn't crash anymore
  - -> `imap_certificate_checks` is 3
  - Save the settings
  -  -> `imap_certificate_checks` is 2

PR #3412, opened 2024-11-05
check spinner bounds

r10s:
_can be merged to #3410 or later to main_

Spinner is very picky about bad selections,
this is an accident waiting to happen -
eg. if new values are introduced
or someone thinks, the "deprecated" case can be removed.

better do not trust incoming data.

nb: first i did a "clip to bounds", so that larger values get the "maximum", however, although anything is better than _always_ crashing,  i think it is better to for for "default" in this case. however, this requires the "deprecated" check introduced by #3410 for the next time at least (iOS and desktop always write a `3`, probably also DeltaTouch)

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/util/ViewUtil.java
+++ b/src/main/java/org/thoughtcrime/securesms/util/ViewUtil.java
@@ -31,6 +31,8 @@ import androidx.annotation.Nullable;
 import androidx.core.view.ViewCompat;
 import androidx.core.view.WindowInsetsCompat;
 import androidx.interpolator.view.animation.FastOutSlowInInterpolator;
+
+import android.util.Log;
 import android.view.Gravity;
 import android.view.LayoutInflater;
 import android.view.View;
@@ -47,6 +49,8 @@ import com.b44t.messenger.util.concurrent.SettableFuture;
 import org.thoughtcrime.securesms.util.views.Stub;
 
 public class ViewUtil {
+  private final static String TAG = ViewUtil.class.getSimpleName();
+
   @SuppressWarnings("deprecation")
   public static void setBackground(final @NonNull View v, final @Nullable Drawable drawable) {
     v.setBackground(drawable);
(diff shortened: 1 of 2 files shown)


src/main/java/org/thoughtcrime/securesms/util/ViewUtil.java
```diff
@@ -267,4 +268,14 @@ public static int getStatusBarHeight(@NonNull View view) {
       return result;
     }
   }
+
+  // Checks if a selection is valid for a given Spinner view.
+  // Returns given selection if valid.
+  // Otherwise, to avoid ArrayIndexOutOfBoundsException, 0 is returned, assuming to refer to a good default.
+  public static int checkBounds(int selection, AbsSpinner view) {
+    if (selection < 0 || selection >= view.getCount()) {
+      return 0;
```
Hocuri, line 277: Should print a stack trace to the log here since something went wrong, maybe even show a toast

Hocuri (approved):

src/main/java/org/thoughtcrime/securesms/util/ViewUtil.java
r10s: not idea how to print a stack trace without having an exception object.

but we could log a warning, that is probably good enough

r10s: i added a commit, the shown warning looks like the following now:

```
index -9 out of bounds of androidx.appcompat.widget.AppCompatSpinner{a8b85c4 VFED..CL. ......I. 0,0-0,0 #7f0a009e app:id/cert_check}
```

Hocuri: > not idea how to print a stack trace without having an exception object.

you can simply do
```
new RuntimeException("Selection " + selection + " is not available for this spinner").printStackTrace();
```

> but we could log a warning, that is probably good enough

yes, also good enough

adbenitez: btw, printing stacktrace instead of using the logging API is discourage IIRC

adbenitez (approved):

adbenitez:
should we merge this?

PR #3413, opened 2024-11-05
Opt out of webview metrics collection

link2xt:


r10s (approved):
good catch! even though data seems to be uploaded only if the user consents, this is  good to have that disabled as a general precaution

r10s:
i  think, it is this one (seen on android7, motog at "System Settings / Google"):

adbenitez (approved):

PR #3415, opened 2024-11-07
feat: Add a very simple about:config

Hocuri:
Add a very simple version of about:config where you need to write "about:config" in the "Saved Messages" chat and then tap on it.

Screenshot_20241107-151947
Screenshot_20241107-152020
Screenshot_20241107-152039

Then, a message is added to the device chat:

Screenshot_20241107-152045

I first thought about including a feature in core of writing `/set config value` into the Saved Messages chat, but that seemed too "dangerous" since an attacker (or a troll) could trick a user into sending e.g. `/set proxy_url attacker_controlled_url` into the Saved Messages chat.

I can't think of any way how this PR could be misused.


adbenitez:
what is the purpose of this? I guess only for development to test something, so I would only activate this feature if developer mode is enabled in the app settings

would be nice to gather a list of the configs (there is a core value that list all keys) and then show a list (or a two columns table with key and value?) to allow easily looking at values and selecting the config to change, but as-is it is good enough already

I am skeptical in general of the utility of this feature (for developers only) vs chances to mess up things by some user

r10s:
> I am skeptical in general of the utility of this feature (for developers only) vs chances to mess up things by some user

i share this scepticism. 

we had sth. like that in early android versions as well, and removed it for the exact reason: that user do not mess sth. up - one can easily imagine some "tipps" for "power users" etc., esp. if this door is open and we devs start adding questionable things ...

and seems not to be needed. devs can find other ways.

so, i prefer to not get sth. like that in.

Hocuri:
I changed it to be possible only in developer builds (i.e. only in the versions build locally - normal users won't be able to enable it).

I would find it useful, e.g. when testing https://github.com/deltachat/deltachat-core-rust/pull/6122 it would have been useful to be able to change notify_about_wrong_pw. But yes, of course it's always possible to put what I want into the code and recompile, so, it's fine not to merge, too.

r10s:
> But yes, of course it's always possible to put what I want into the code and recompile, so, it's fine not to merge, too.


i would definitely prefer that, it is easy enough.

pretty sure, we get some issues about that at some point and will it disable it a second time :)

adbenitez:
as @r10s I also think it is better not to do this

since @Hocuri also said it is fine not to merge, lets close this for now, while testing one is building the apk anyways and recompiling with some change in android code is cheap

PR #3416, opened 2024-11-07
Rename "Back up Chats to External Storage" to "Export Backup"

Hocuri:
A backup contains more than just the chats, and it's not true that the backup will be exported to "external storage". (the backup will be exported to the Downloads folder, which may well be on the internal storage rather than the SD card).

Old/New:

image

image

### DIFF
--- a/src/main/res/xml/preferences_chats.xml
+++ b/src/main/res/xml/preferences_chats.xml
@@ -48,7 +48,7 @@
 
     <PreferenceCategory android:title="@string/pref_backup">
         <Preference android:key="pref_backup"
-            android:title="@string/pref_backup_explain"/>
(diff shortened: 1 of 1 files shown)


adbenitez:
some user already reported in the past that they didn't realize this was the option to export the whole profile

wondering if there is a better naming that could transmit the idea "this backups your whole profile" or maybe take the opportunity to add something like that as explaination subtitle!

r10s:
> some user already reported in the past that they didn't realize this was the option to export the whole profile

let's see after things change i'd say. 

the new wording (that is also used on desktop/iOS) does not limit things to "chats", so, a simple "export" indeed might work much better than "export chats"

adbenitez (approved):

PR #3418, opened 2024-11-09
tweak phased rollout instructions

r10s:
this PR adapts the "phased rollout instructions", as a result of internal discussions


adbenitez (approved):

r10s:
let's on the next releases, see how much "real rollout" we get after the two-days-fifty-percent - maybe we can streamline this further and go for 99% immediately and then to 100% after 4 days ...

PR #3419, opened 2024-11-12
don't increase reaction selector's font size according to system

### LINKED ISSUE #3291: Bigger Android font and display sizes cause truncated reaction bar
- Android version:
Android 13 & 14.

- Devices:
Google Pixel 8 & Samsung Galaxy A71.

- Delta Chat version:
1.46.13.

- Expected behavior:
The reaction bar is not truncated when the sizes for font and display in Android are set to high levels.

- Actual behavior:
The reaction bar is truncated when the sizes for font and display in Android are set to high levels.

-- Set the sizes for font and display in Android to high levels.
-- Open a chat and long-tap a message you would like to react to. (Result: The reaction bar is truncated, the three-dot icon on the right is invisible and cannot be selected.)

- Screenshot:
grafik

- Logs:
N/A.

adbenitez:
also avoid moving the selector horizontally out of screen 

close #3291

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/reactions/AddReactionView.java
+++ b/src/main/java/org/thoughtcrime/securesms/reactions/AddReactionView.java
@@ -105,7 +105,7 @@ public class AddReactionView extends LinearLayout {
         } else {
             x += offset;
         }
-        ViewUtil.setLeftMargin(this, x);
+        ViewUtil.setLeftMargin(this, Math.max(x, 0));
 
         int y = Math.max((int)parentView.getY() - offset, offset/2);
         ViewUtil.setTopMargin(this, y);
--- a/src/main/res/values/text_styles.xml
+++ b/src/main/res/values/text_styles.xml
@@ -32,9 +32,9 @@
     </style>
 
     <style name="AddReaction" parent="Base.TextAppearance.AppCompat.Body1">
-        <item name="android:paddingLeft">8sp</item>
-        <item name="android:paddingRight">8sp</item>
-        <item name="android:textSize">28sp</item>
+        <item name="android:paddingLeft">8dp</item>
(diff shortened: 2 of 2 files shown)


r10s (approved):
good and easy fix 👍

Hocuri (approved):
It's not working when the screen is actually small or when the display size is increased in the settings (only when the font size is increased in the settings). Nonetheless, it's an improvement to how it is today.
Screenshot_20241112-142150

adbenitez:
> It's not working when the screen is actually small or when the display size is increased in the settings (only when the font size is increased in the settings). Nonetheless, it's an improvement to how it is today.

mmmh weird, it should span to the full screen size but it is smaller for you, I tested with both, font and screen amplification and it looked fine for me :(

Hocuri:
> could you try with system emojis?

I tried with current main (which has both this PR and https://github.com/deltachat/deltachat-android/pull/3411/ merged), still too big

r10s:
seem the screen of @Hocuri is even smaller than the one in the original post - see the whitespace between "1" and "reply".

i fear, we need to be a little more adaptive here. maybe the following is good enough:

```
if screenWidth < N {
  set laugh-emoji and sad-emoji to View.GONE - maybe even heart-emoji
}
```

we were considering to hide sad-emoji alltogether, but that may not fix the issue accodingly, also for consistency with desktop (and also iOS has these emojis prefilled in the selector) that would not be a very nice solution

adbenitez:
notice that the selector looks fine in hocuri is phone, it is about the case when in such small screen you also use the screen scaling/augmentation which is not what  someone with such small screen would do,

a more generic solution could be to use a scrollbar inside the bubble to catch such extreme cases at least without errors

PR #3420, opened 2024-11-12
calculate account size in background

### LINKED ISSUE #3308: Account deletion confirmation dialog takes a while to appear in big accounts
this is because the computation of the account size takes a few seconds, maybe display "..." or "--" there and make the core API call in a thread to load dynamically? what will happen if user click "delete" button and the API call didn't returned yet?? (cc @link2xt) 

other option would be not to display the account size, but it is nicer to know

r10s:
calculation may take a moment,
do this in background.

it is totally fine to just display nothing before, that avoids flickering (WHAT was this? i cannot look so fast!) - and in most cases, it is close to instant.

closes #3308

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
@@ -168,11 +168,16 @@ public class AccountSelectionListFragment extends DialogFragment
     avatar.setAvatar(GlideApp.with(activity), recipient, false);
     nameView.setText(name);
     addrView.setText(contact.getAddr());
-    try {
-      sizeView.setText(Util.getPrettyFileSize(rpc.getAccountFileSize(accountId)));
-    } catch (RpcException e) {
-      e.printStackTrace();
-    }
+    Util.runOnAnyBackgroundThread(() -> {
+      try {
+        final int sizeBytes = rpc.getAccountFileSize(accountId);
+        Util.runOnMain(() -> {
+          sizeView.setText(Util.getPrettyFileSize(sizeBytes));
+        });
+      } catch (RpcException e) {
+        e.printStackTrace();
+      }
+    });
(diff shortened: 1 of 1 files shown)


src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
```diff
@@ -168,11 +168,16 @@ private void onDeleteAccount(int accountId) {
...
-    try {
-      sizeView.setText(Util.getPrettyFileSize(rpc.getAccountFileSize(accountId)));
-    } catch (RpcException e) {
-      e.printStackTrace();
-    }
+    Util.runOnAnyBackgroundThread(() -> {
+      try {
+        final int sizeBytes = rpc.getAccountFileSize(accountId);
+        Util.runOnMain(() -> {
+          sizeView.setText(Util.getPrettyFileSize(sizeBytes));
```
adbenitez, line 175: if you cancel the dialog in a GBs account that takes several seconds, would be got to know if this causes some background exception because the dialog was already destroyed

adbenitez: but my guess is that this should be fine, since the view is created separately from the dialog

adbenitez (approved):

src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
r10s: i think, there is a strong reference to the object as defined in the outer block, if at all, this could irritate the garbage collection, but even that should be fine as when the callback is finally left, the object can be freed

i tried that in practise by adding a `final int sizeBytes = rpc.getAccountFileSize(accountId);Util.sleep(10000);`  and closing all dialogs before the 10 seconds, everything seems fine

PR #3422, opened 2024-11-12
change proxy protocol label position

### LINKED ISSUE #3421: always show proxy-type label at the beginning
currently the selected proxy has connectivity message at the beginning  and this causes the type label to shift to the right which is not nice

adbenitez:
close #3421

### DIFF
--- a/src/main/res/layout/proxy_list_item.xml
+++ b/src/main/res/layout/proxy_list_item.xml
@@ -38,6 +38,22 @@
         android:gravity="center_vertical"
         >
 
+        <TextView
+            android:id="@+id/protocol"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_marginEnd="5dp"
+            android:paddingStart="5dp"
+            android:paddingEnd="5dp"
+            android:paddingTop="1dp"
+            android:paddingBottom="1dp"
+            android:textAllCaps="true"
+            android:textColor="?attr/conversation_list_item_date_color"
+            android:background="@drawable/archived_indicator_background"
+            style="@style/Signal.Text.Caption"
+            tools:text="socks5"
+            />
+
       <ImageView
           android:id="@+id/checkmark"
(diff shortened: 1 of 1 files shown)


r10s (approved):
much better!

when on that, i'd use the same padding for checkmark and status as well:

- the checkmark does not belong to status nor protocol, it is an symbol on its own (using radio buttons might be better here, or move it left of the "share icon" - but it is also not bad as it is now, i am not suggesting to change this PR further)
- don't make me think wrt RTL layout (not saying it was wrong before)

```
      <ImageView
          android:id="@+id/checkmark"
          android:layout_gravity="center_vertical|end"
          android:paddingStart="5dp"
          android:paddingEnd="5dp"
          android:contentDescription="@null"
          android:src="@drawable/ic_delivery_status_sent"
          app:tint="?attr/conversation_list_item_date_color"
          />

      <TextView
          android:id="@+id/status"
          android:singleLine="true"
          android:ellipsize="marquee"
          android:paddingStart="5dp"
          android:paddingEnd="5dp"
          android:textAppearance="?android:attr/textAppearanceSmall"
          android:textSize="14sp"
          android:fontFamily="sans-serif-light"
          tools:text="@string/connectivity_connected"
          />
```

adbenitez:
> when on that, i'd use the same padding for checkmark and status as well:
> 
>     * the checkmark does not belong to status nor protocol, it is an symbol on its own

the padding is used because the label has the line background, the actual separation between the items is the same with `android:layout_marginEnd="5dp"`

adbenitez:
> * the checkmark does not belong to status nor protocol, it is an symbol on its own

that is why it looked better at the start, yeah we will likely need to further improve this layout, ideally the protocol label would go up to the server title where it fits more together, and the subtitle would be the availability of the proxy or ping latency (that is how it is in Telegram) but since we don't have that, it was looking too empty with smaller items for not selected proxy and bigger item for selected proxy, that is why I put the label in the subtitle

PR #3424, opened 2024-11-13
remove internal font scaling preference

### LINKED ISSUE #3423: internal setting for text scaling not working
with the removal of the old emoji textview, it seems it was there where the internal text size scale setting was handled, now changing the text size in appearance settings in-app does nothing

we have two options here, implementing old behavior or the simpler and removing maintenance efforts (and feature requests like scaling ALL text inside the app): just remove the setting and let user handle this using the much better system settings that allow scaling  of text and also of the whole UI

adbenitez:
close #3423


src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
```diff
@@ -45,6 +45,16 @@ public void setText(@Nullable CharSequence text, BufferType type) {
...
+  @Override
+  public void setTextSize(float size) {
+    setTextSize(TypedValue.COMPLEX_UNIT_SP, size);
+  }
+
+  @Override
+  public void setTextSize(int unit, float size) {
+    originalFontSize = TypedValue.applyDimension(unit, size, getResources().getDisplayMetrics());
+    super.setTextSize(unit, size);
+  }
```
r10s, line 57: why is that needed now?

adbenitez: re-review, I just fixed it instead of removing it, for some user with Samsung 12 the system setting was not taking effect in the text size and I suspect it is because this is necessary anyways to refresh the font size, since the saved value is in PX, not SP

r10s:
sth seems wrong with the PR on github, i do not see any preference option removed in the codes

adbenitez:
> sth seems wrong with the PR on github, i do not see any preference option removed in the codes

I updated the title, it is not about removing the preference but fixing the issue with the font size changes not taking effect

adbenitez:
or do you want to take the opportunity to remove the preference altogether?

r10s (approved):

PR #3425, opened 2024-11-13
perf: Only set text size if the scale isn't 1

Hocuri:
According to the Android Studio profiler, this is an easy win of 1.5ms when loading the chat activity.

I tested that when one or multiple of the latest message are large, and then I scroll up to text messages, these text messages are at scale 1. (because sometimes with the RecylcerView it happens that when you change something, it isn't reset to the default value when the item is recycled).

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
+++ b/src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
@@ -41,7 +41,9 @@ public class AutoScaledEmojiTextView extends AppCompatTextView {
   @Override
   public void setText(@Nullable CharSequence text, BufferType type) {
     float scale = text != null ? getTextScale(text.toString()) : 1;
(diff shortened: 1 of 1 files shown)


adbenitez:
> I tested that when one or multiple of the latest message are large, and then I scroll up to text messages, these text messages are at scale 1. (because sometimes with the RecylcerView it happens that when you change something, it isn't reset to the default value when the item is recycled).

this is because in ConversationItem we always set/reset the font size (due to the custom font size setting we have in appearance which we contemplated removing recently,

I think the proper way to fix this is to check `getTextSize()` and compare with the target font size `originalFontSize * scale` and if they are different then call `setTextSize()`

adbenitez:
this PR becomes invalid with #3424

adbenitez:
https://github.com/deltachat/deltachat-android/pull/3424 got merged, text size is not set anymore outside in ConversationItem so it needs to be set inside AutoScaledEmojiTextView, so the issue of setting the size twice that this PR was targeting is solved! :tada:

PR #3427, opened 2024-11-13
fix: Show composite emojis enlarged, too

Hocuri:
There are a lot of composite emojis like the flags and the family emojis. Here, the grapheme starts with an emoji, but then come some other characters that modify the first emoji (e.g. all flags start with the flag emoji).

This PR shows those as big, too, by checking whether a grapheme _starts_ with an emoji to determine whether it's an emoji.


src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
```diff
@@ -20,7 +20,7 @@ public class AutoScaledEmojiTextView extends AppCompatTextView {
...
...🧑👱👨🧔👩🧓👴👵🙍🙎🙅🙆💁🙋🧏🙇🤦🤷👮🕵💂🥷👷🫅🤴👸👳👲🧕🤵👰🤰🫃🫄🤱👼🎅🤶🦸🦹🧙-🧟🧌💆💇🚶🧍🧎🏃💃🕺🕴👯🧖🧗🤺🏇⛷🏂🏌🏄🚣🏊⛹🏋🚴🚵🤸🤼-🤾🤹🧘🛀🛌👭👫👬💏💑🗣👤👥🫂👪👣🦰🦱🦳🦲🐵🐒🦍🦧🐶🐕🦮🐩🐺🦊🦝🐱🐈🦁🐯🐅🐆🐴🫎🫏🐎🦄🦓🦌🦬🐮🐂-🐄🐷🐖🐗🐽🐏🐑🐐🐪🐫🦙🦒🐘🦣🦏🦛🐭🐁🐀🐹🐰🐇🐿🦫🦔🦇🐻🐨🐼🦥🦦🦨🦘🦡🐾🦃🐔🐓🐣-🐧🕊🦅🦆🦢🦉🦤🪶🦩🦚🦜🪽🪿🐸🐊🐢🦎🐍🐲🐉🦕🦖🐳🐋🐬🦭🐟-🐡🦈🐙🐚🪸🪼🐌🦋🐛-🐝🪲🐞🦗🪳🕷🕸🦂🦟🪰🪱🦠💐🌸💮🪷🏵🌹🥀🌺-🌼🌷🪻🌱🪴🌲-🌵🌾🌿☘🍀-🍃🪹🪺🍄🍇-🍍🥭🍎-🍓🫐🥝🍅🫒🥥🥑🍆🥔🥕🌽🌶🫑🥒🥬🥦🧄🧅🥜🫘🌰🫚🫛🍞🥐🥖🫓🥨🥯🥞🧇🧀🍖🍗🥩🥓🍔🍟🍕🌭🥪🌮🌯🫔🥙🧆🥚🍳🥘🍲🫕🥣🥗🍿🧈🧂🥫🍱🍘-🍝🍠🍢-🍥🥮🍡🥟-🥡🦀🦞🦐🦑🦪🍦-🍪🎂🍰🧁🥧🍫-🍯🍼🥛☕🫖🍵🍶🍾🍷-🍻🥂🥃🫗🥤🧋🧃🧉🧊🥢🍽🍴🥄🔪🫙🏺🌍-🌐🗺🗾🧭🏔⛰🌋🗻🏕🏖🏜-🏟🏛🏗🧱🪨🪵🛖🏘🏚🏠-🏦🏨-🏭🏯🏰💒🗼🗽⛪🕌🛕🕍⛩🕋⛲⛺🌁🌃🏙🌄-🌇🌉♨🎠🛝🎡🎢💈🎪🚂-🚊🚝🚞🚋-🚎🚐-🚙🛻🚚-🚜🏎🏍🛵🦽🦼🛺🚲🛴🛹🛼🚏🛣🛤🛢⛽🛞🚨🚥🚦🛑🚧⚓🛟⛵🛶🚤🛳⛴🛥🚢✈🛩🛫🛬🪂💺🚁🚟-🚡🛰🚀🛸🛎🧳⌛⏳⌚⏰-⏲🕰🕛🕧🕐🕜🕑🕝🕒🕞🕓🕟🕔🕠🕕🕡🕖🕢🕗🕣🕘🕤🕙🕥🕚🕦🌑-🌜🌡☀🌝🌞🪐⭐🌟🌠🌌☁⛅⛈🌤-🌬🌀🌈🌂☂☔⛱⚡❄☃⛄☄🔥💧🌊🎃🎄🎆🎇🧨✨🎈-🎋🎍-🎑🧧🎀🎁🎗🎟🎫🎖🏆🏅🥇-🥉⚽⚾🥎🏀🏐🏈🏉🎾🥏🎳🏏🏑🏒🥍🏓🏸🥊🥋🥅⛳⛸🎣🤿🎽🎿🛷🥌🎯🪀🪁🔫🎱🔮🪄🎮🕹🎰🎲🧩🧸🪅🪩🪆♠♥♦♣♟🃏🀄🎴🎭🖼🎨🧵🪡🧶🪢👓🕶🥽🥼🦺👔-👖🧣-🧦👗👘🥻🩱-🩳👙👚🪭👛-👝🛍🎒🩴👞👟🥾🥿👠👡🩰👢🪮👑👒🎩🎓🧢🪖⛑📿💄💍💎🔇-🔊📢📣📯🔔🔕🎼🎵🎶🎙-🎛🎤🎧📻🎷🪗🎸-🎻🪕🥁🪘🪇🪈📱📲☎📞-📠🔋🪫🔌💻🖥🖨⌨🖱🖲💽-📀🧮🎥🎞📽🎬📺📷-📹📼🔍🔎🕯💡🔦🏮🪔📔-📚📓📒📃📜📄📰🗞📑🔖🏷💰🪙💴-💸💳🧾💹✉📧-📩📤-📦📫📪📬-📮🗳✏✒🖋🖊🖌🖍📝💼📁📂🗂📅📆🗒🗓📇-📎🖇📏📐✂🗃🗄🗑🔒🔓🔏-🔑🗝🔨🪓⛏⚒🛠🗡⚔💣🪃🏹🛡🪚🔧🪛🔩⚙🗜⚖🦯🔗⛓🪝🧰🧲🪜⚗🧪-🧬🔬🔭📡💉🩸💊🩹🩼🩺🩻🚪🛗🪞🪟🛏🛋🪑🚽🪠🚿🛁🪤🪒🧴🧷🧹-🧻🪣🧼🫧🪥🧽🧯🛒🚬⚰🪦⚱🧿🪬🗿🪧🪪🏧🚮🚰♿🚹-🚼🚾🛂-🛅⚠🚸⛔🚫🚳🚭🚯🚱🚷📵🔞☢☣⬆↗➡↘⬇↙⬅↖↕↔↩↪⤴⤵🔃🔄🔙-🔝🛐⚛🕉✡☸☯✝☦☪☮🕎🔯🪯♈-♓⛎🔀-🔂▶⏩⏭⏯◀⏪⏮🔼⏫🔽⏬⏸-⏺⏏🎦🔅🔆📶🛜📳📴♀♂⚧✖➕-➗🟰♾‼⁉❓-❕❗〰💱💲⚕♻⚜🔱📛🔰⭕✅☑✔❌❎➰➿〽✳✴❇©®™🔟-🔤🅰🆎🅱🆑-🆓ℹ🆔Ⓜ🆕🆖🅾🆗🅿🆘-🆚🈁🈂🈷🈶🈯🉐🈹🈚🈲🉑🈸🈴🈳㊗㊙🈺🈵🔴🟠-🟢🔵🟣🟤⚫⚪🟥🟧-🟩🟦🟪🟫⬛⬜◼◻◾◽▪▫🔶-🔻💠🔘🔳🔲🏁🚩🎌🏴🏳🇦-🇿\uD83E\uDD89\uD83E\uDD8F\uD83E\uDDBE\uD83E\uDDC6\uD83E\uddcd\uD83E\udddf\uD83E\ude99]|#️⃣|\\*️⃣|0️⃣|1️⃣|2️⃣|3️⃣|4️⃣|5️⃣|6️⃣|7️⃣|8️⃣|9️⃣)+.*");
```
Hocuri, line 23: In this line, what I changed is add the `.*` at the end so that the pattern matches everything that starts with an emoji.

adbenitez: this seems invalid, if it match anything starting with an emoji it will match `👋hi`

src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
```diff
@@ -50,36 +50,34 @@ private float getTextScale(String text) {
...
-    // Initialize count
     int graphemeCount = 0;
 
     // Iterate over the text and count graphemes
     int start = graphemeIterator.first();
     for (int end = graphemeIterator.next(); end != BreakIterator.DONE; start = end, end = graphemeIterator.next()) {
-      if (++graphemeCount > max) break;
+      String grapheme = text.substring(start, end);
+      if (!emojiRegex.matcher(grapheme).matches()) return 0;
+      if (++graphemeCount > max) return 0;
```
adbenitez, line 80: now the returned value is never `> max`  so the `if` condition checking `emojiCount > 8` can be removed, the function doc-string needs also to indicate this. I would prefer returning `-1` in this case then we can check for `emojiCount > 0`

adbenitez: did you test the performance impact of regex over the whole text vs doing it for each grapheme? in theory it should be cheap since we do it at most in 8 graphemes and break the loop otherwise

since now this is more robust and we don't loop over whole text, maybe makes sense to increase the length restriction from 21 to 56

there is another PR modifying this, maybe first merge that one and rebase

src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
```diff
@@ -50,36 +50,34 @@ private float getTextScale(String text) {
...
```
adbenitez, line 80:
```suggestion
      if (++graphemeCount > max) return -1;
```

src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
```diff
@@ -50,36 +50,34 @@ private float getTextScale(String text) {
     if (text.length() > 21 || text.isEmpty() || Character.isLetter(text.charAt(0))) {
       return 1;
     }
-    int emojiCount = countGraphemes(text, 8);
-    if (emojiCount > 8) {
+    int emojiCount = countEmojis(text, 8);
+    if (emojiCount > 8 || emojiCount == 0) {
```
adbenitez, line 54:
```suggestion
    if (emojiCount <= 0) {
```

src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
```diff
@@ -50,36 +50,34 @@ private float getTextScale(String text) {
     if (text.length() > 21 || text.isEmpty() || Character.isLetter(text.charAt(0))) {
```
adbenitez, line 50:
```suggestion
    if (text.length() > 56 || text.isEmpty() || Character.isLetter(text.charAt(0))) {
```

adbenitez (approved):

src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
adbenitez: without this, 4-8 emojis of some flags would not be properly scaled

src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
Hocuri: I finally ran some performance tests (4 times opening a chat with a lot of emojis, 2 times opening a chat with just 2 emoji messages). The performance varies a lot, and all in all, it's inconclusive

getTextScale()

Chat with a lot of emojis, new:
16ms, of which 7.4ms in getCharacterInstance, 3.5ms in setText, 4.3ms in next, 1.15ms in matches
3.6ms, of which 1ms in getCharacterInstance, 1ms in setText, 1ms in matches
9ms, of which 5.6ms in getCharacterInstance, 1.5ms in next, 1ms in matcher, 1ms in matches
8.3ms, of which 4.3 in getCharacterInstance, 1.7 in next, 1.4 in setText, 1ms in memcmp (weird...)

Chat with just two messages with emojis, new:
0.8ms, all in setText
1ms, all in next

Chat with a lot of emojis, old:
9ms, of which 5ms in getCharacterInstance, 2ms in next, 1ms in setText, 1ms in matches
1ms, all in matches
5.7ms, of which 4.7 in getCharacterInstance, 1ms in matches

Chat with just two messages with emojis, old:
not present at all
1ms, all in next

src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
Hocuri: See comment above, there is no evidence that this PR here improves performance, so it's no reason to increase the limit

src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
```diff
@@ -73,36 +73,34 @@ private float getTextScale(String text) {
...
+    if (emojiCount <= 6) scale += 0.25f;
+    if (emojiCount <= 4) scale += 0.25f;
+    if (emojiCount <= 2) scale += 0.25f;
+    return scale;
   }
 
-  public static int countGraphemes(String text, int max) {
-    // Create a BreakIterator for grapheme boundaries
+  /**
+   * Returns the number of emojis if there are only emojis, or 0 otherwise.
```
Hocuri, line 89:
```suggestion
   * Returns the number of emojis if there are only emojis, or -1 otherwise.
```

src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
adbenitez: the reason to increase the limit is not for performance reasons but for what this PR is targeting that is showing composite emojis enlarged, if I got it correctly this PR was never about performance :D

src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
Hocuri: I do find it a bit surprising that this PR doesn't improve performance; I'd be fine with not merging if we say that performance is more important than enlarging all emojis here

adbenitez (approved):

src/main/java/org/thoughtcrime/securesms/components/emoji/AutoScaledEmojiTextView.java
adbenitez: oth, showing big only up to 3 of such emojis is good enough probably

PR #3432, opened 2024-11-16
fix: check draft.isOk() instead of comparing to null

link2xt:
dcContext.getDraft() never returns null.

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
@@ -717,7 +717,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     DcMsg draft = dcContext.getDraft(chatId);
     final String sharedText = RelayUtil.getSharedText(this);
 
-    if (draft == null) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -717,7 +717,7 @@ private ListenableFuture<Boolean> initializeDraft() {
     DcMsg draft = dcContext.getDraft(chatId);
     final String sharedText = RelayUtil.getSharedText(this);
 
-    if (draft == null) {
+    if (!draft.isOk()) {
```
adbenitez, line 720: wouldn't this cause a crash if draft is null??? I think what you wanted is:
```suggestion
    if (draft == null || !draft.isOk()) {
```

adbenitez: wow so this was never entering that path

adbenitez (approved):

adbenitez:
maybe the correct fix would be to return null in getDraft, having to ask for isOk() is harder to remember than a more visible NullPointerException

PR #3435, opened 2024-11-20
allow to open in browser http links that look like proxies

adbenitez:
see discussion at https://support.delta.chat/t/in-the-new-android-client-it-is-not-possible-to-click-to-open-urls-with-port-numbers/3327


Hocuri (approved):
Looks good, didn't test

PR #3438, opened 2024-11-23
allow to open deep-link to webxdc

adbenitez:
- [x] extend WebxdcActivity to allow passing deep-link when calling `WebxdcActivity.openWebxdcActivity()`
- [x] add DcMsg.getWebxdcHref()
- [x] use new core API to get deep-link from info-message when user click it and directly open webxdc?

### DIFF
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -1565,6 +1565,15 @@ JNIEXPORT jstring Java_com_b44t_messenger_DcMsg_getWebxdcInfoJson(JNIEnv *env, j
 }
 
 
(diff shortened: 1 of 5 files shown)


adbenitez:
when user click a webxdc info-message, we want to open directly the webxdc if it has a link, I find it a bit unexpected if some info messages open the webxdc while other doesn't so I think it is better to always open the apps when you click in an info message

adbenitez:
I think opening deep-links from the new webxdc notifications is not part of the task at hand right?

r10s (approved):

adbenitez:
this partially solves #3439

PR #3440, opened 2024-11-23
Reverse order of the tiny inbox in notifications

Hocuri:
When someone sends you 4 messages "1", "2", "3", "4" (in this order), this now looks like this (unexpanded/expanded):
Screenshot_20241123-200045
Screenshot_20241123-200048

Before, the unexpanded notification looked the same, the expanded notification had the other direction.

This PR here makes DC show the same behavior as Signal and WhatsApp.


r10s (approved):

adbenitez (approved):
really nice!!!

PR #3441, opened 2024-11-23
Don't always show the sender in notifications

Hocuri:
In 1:1 chats, don't prepend the sender name to every line. The exception is when the display name is not equal to the chat name (i.e. there is an OverwriteSenderDisplayname), in this case we still prepend it.

This is how it looks when Bob2 sends 4 messages, and then I get 1 message from Hocuri4 that is assigned to the same chat:

Screenshot_20241123-202426

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
+++ b/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
@@ -349,7 +349,14 @@ public class NotificationCenter {
         }
         String tickerLine = shortLine;
         if (!dcChat.isMultiUser() && privacy.isDisplayContact()) {
-          tickerLine = dcMsg.getSenderName(dcContext.getContact(dcMsg.getFromId()), false) + ": " + tickerLine;
+          DcContact contact = dcContext.getContact(dcMsg.getFromId());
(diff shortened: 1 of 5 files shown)


src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -347,7 +347,14 @@ public void notifyMessage(int accountId, int chatId, int msgId) {
         }
         String tickerLine = shortLine;
         if (!dcChat.isMultiUser() && privacy.isDisplayContact()) {
-          tickerLine = dcMsg.getSenderName(dcContext.getContact(dcMsg.getFromId()), false) + ": " + tickerLine;
+          DcContact contact = dcContext.getContact(dcMsg.getFromId());
+          tickerLine = dcMsg.getSenderName(contact, false) + ": " + tickerLine;
+
+          if (!Util.equals(dcChat.getName(), contact.getDisplayName())) {
```
adbenitez, line 353: this sounds hacky, I would directly check:
`IF dcChat is multi-user OR dcMsg has overriden sender name THEN show it ELSE don't show it

Hocuri:
This PR now contains the commits of https://github.com/deltachat/deltachat-android/pull/3442, actually the PR here is only about the first commit

PR #3443, opened 2024-11-23
notify DC_EVENT_INCOMING_WEBXDC_NOTIFY

### LINKED ISSUE #3439: add webxdc selfAddr/notify/href APIs (M2/M3)

adbenitez:
close #3439

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
+++ b/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
@@ -368,7 +368,7 @@ public class NotificationCenter {
 
         DcContact sender = dcContext.getContact(contactId);
         String shortLine = context.getString(R.string.reaction_by_other, sender.getDisplayName(), reaction, dcMsg.getSummarytext(2000));
-        maybeAddNotification(accountId, dcContext.getChat(dcMsg.getChatId()), msgId, shortLine, shortLine, true);
+        maybeAddNotification(accountId, dcContext.getChat(dcMsg.getChatId()), msgId, shortLine, shortLine, false);
       });
     }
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -366,7 +368,28 @@ public void notifyReaction(int accountId, int contactId, int msgId, String react
 
         DcContact sender = dcContext.getContact(contactId);
         String shortLine = context.getString(R.string.reaction_by_other, sender.getDisplayName(), reaction, dcMsg.getSummarytext(2000));
-        maybeAddNotification(accountId, dcContext.getChat(dcMsg.getChatId()), msgId, shortLine, shortLine, false);
+        maybeAddNotification(accountId, dcContext.getChat(dcMsg.getChatId()), msgId, shortLine, shortLine, true);
```
r10s: the "play in chat sound" needs to be false here and for webxdc-info.

at some point, we can factor that out , but well ...

r10s (commented):
otherwise lgtm

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -370,6 +372,27 @@ public void notifyReaction(int accountId, int contactId, int msgId, String react
...
+          return; // showing "New Message" is wrong, just do nothing.
+        }
+
+        DcContext dcContext = context.dcAccounts.getAccount(accountId);
+        DcMsg dcMsg = dcContext.getMsg(msgId);
+        JSONObject info;
+        if(dcMsg.getType() == DcMsg.DC_MSG_WEBXDC) {
+          info = dcMsg.getWebxdcInfo();
+        } else { // info message, get from parent xdc
+          info = dcMsg.getParent().getWebxdcInfo();
```
r10s, line 388: getParent() can be null, so maybe be more resilient here, a try/catch around things or so

r10s (approved):

adbenitez:
there are some problems I see: since if the notification comes from a summary change, there is no actual message id that is different per notification, several notifications like that for the same message would probably override each other since they will have the same id

we would need to generate some new notifications ID, and handle them differently also to be able to discard notifications of a webxdc when you open it (similar to how we discard notifications of a chat when you open it)

not notifying if the notification is from the currently opened webxdc is more easy to do

r10s (approved):

PR #3453, opened 2024-11-27
prepare 1.49.0

adbenitez:


CHANGELOG.md
```diff
@@ -1,5 +1,22 @@
...
+* Add "Show in Chat" to the menu of opened Webxdc apps
+* Reverse order of messages in the notification group
+* Notify reactions to own messages
+* Improve the button to start Webxdc apps
+* Make account deletion confirmation dialog faster
+* Rename "Back up Chats to External Storage" to "Export Backup"
+* Improve compatibility with classic email clients in the outgoing messages
+* Removed internal font scaling setting in favor of the better system settings
+* several bug fixes
+* update to core 1.151.2
```
r10s, line 18: seems entries for core 1.148.7 - 1.151.1 are missing, worth mentioning:

Use privacy-preserving webxdc addresses
Use Rustls for connections with strict TLS 
fix: Trim whitespace from scanned QR codes.
fix quotes: Line-before-quote may be up to 120 character long instead of 80
fix: Prevent accidental wrong-password-notifications 
fix: Remove footers from "Show Full Message..."
fix: Only add "member added/removed" messages if they actually do that 
fix: Update state of message when fully downloading it
fix: send message: Do not fail if the message does not exist anymore
fix: Do not percent-encode dot when passing to autoconfig server (so, fix handling of some servers)
fix displaynames not being updated when intially scanned by a QR code

r10s (approved):

r10s:
merging, as 1.49 is out

PR #3456, opened 2024-11-29
notify webxdc events, replies and reactions to own messages even if chat is muted

adbenitez:
even if group is muted:

* notify quote-reply to own messages
* notify reactions to own messages
* notify webxdc notifications

what is still unsolved:

if webxdc sends notification while the chat is open, there is no notification, if there is no info message there is no hint that there was a notification at all, and if there is info message that doesn't warranties that the notification was the same

### DIFF
--- a/src/main/java/com/b44t/messenger/DcContext.java
+++ b/src/main/java/com/b44t/messenger/DcContext.java
@@ -85,7 +85,7 @@ public class DcContext {
     public final static int DC_CONNECTIVITY_WORKING = 3000;
     public final static int DC_CONNECTIVITY_CONNECTED = 4000;
 
-    private static final String CONFIG_MENTION_NOTIF_ENABLED = "ui.notify_mentions";
+    private static final String CONFIG_NOTIFY_MENTIONS_IF_MUTED = "ui.notify_mentions_if_muted";
 
     // when using DcAccounts, use DcAccounts.addAccount() instead
     public DcContext(String osName, String dbfile) {
@@ -136,7 +136,8 @@ public class DcContext {
     public void                setConfigInt         (String key, int value) { setConfig(key, Integer.toString(value)); }
     public native boolean      setConfigFromQr      (String qr);
     public native String       getConfig            (String key);
-    public int                 getConfigInt         (String key) { try{return Integer.parseInt(getConfig(key));} catch(Exception e) {} return 0; }
+    public int                 getConfigInt         (String key) { return getConfigInt(key, 0); }
+    public int                 getConfigInt         (String key, int defaultValue) { try {return Integer.parseInt(getConfig(key));} catch(Exception e) {} return defaultValue; }
     public native String       getInfo              ();
     public native int          getConnectivity      ();
     public native String       getConnectivityHtml  ();
@@ -214,11 +215,11 @@ public class DcContext {
     public DcProvider          getProviderFromEmailWithDns (String email) { long cptr = getProviderFromEmailWithDnsCPtr(email); return cptr!=0 ? new DcProvider(cptr) : null; }
 
     public boolean isMentionsEnabled() {
-      return !"0".equals(getConfig(CONFIG_MENTION_NOTIF_ENABLED));
+      return getConfigInt(CONFIG_NOTIFY_MENTIONS_IF_MUTED, 1) == 1;
     }
 
     public void setMentionsEnabled(boolean enabled) {
-      setConfigInt(CONFIG_MENTION_NOTIF_ENABLED, enabled? 1 : 0);
+      setConfigInt(CONFIG_NOTIFY_MENTIONS_IF_MUTED, enabled? 1 : 0);
     }
 
     public String getName() {
--- a/src/main/java/com/b44t/messenger/DcContext.java
+++ b/src/main/java/com/b44t/messenger/DcContext.java
@@ -85,7 +85,7 @@ public class DcContext {
     public final static int DC_CONNECTIVITY_WORKING = 3000;
     public final static int DC_CONNECTIVITY_CONNECTED = 4000;
 
-    private static final String CONFIG_NOTIFY_MENTIONS_IF_MUTED = "ui.notify_mentions_if_muted";
+    private static final String CONFIG_MUTE_MENTIONS_IF_MUTED = "ui.mute_mentions_if_muted";
 
     // when using DcAccounts, use DcAccounts.addAccount() instead
     public DcContext(String osName, String dbfile) {
@@ -136,8 +136,7 @@ public class DcContext {
     public void                setConfigInt         (String key, int value) { setConfig(key, Integer.toString(value)); }
     public native boolean      setConfigFromQr      (String qr);
     public native String       getConfig            (String key);
-    public int                 getConfigInt         (String key) { return getConfigInt(key, 0); }
-    public int                 getConfigInt         (String key, int defaultValue) { try {return Integer.parseInt(getConfig(key));} catch(Exception e) {} return defaultValue; }
+    public int                 getConfigInt         (String key) { try{return Integer.parseInt(getConfig(key));} catch(Exception e) {} return 0; }
     public native String       getInfo              ();
     public native int          getConnectivity      ();
     public native String       getConnectivityHtml  ();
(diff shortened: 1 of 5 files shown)


r10s:
> [remove unnecessary thread](https://github.com/deltachat/deltachat-android/pull/3456/commits/05a5e4317ac7625c1602dc5f3d1efe5fc3b48432)

just wanted to comment on that :)

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -387,26 +393,33 @@ public void notifyWebxdc(int accountId, int contactId, int msgId, String text) {
...
+        if (Util.equals(visibleWebxdc, new Pair<>(accountId, parentMsg.getId()))) {
+            return; // do not notify if the app is already open
         }
+
+        JSONObject info = parentMsg.getWebxdcInfo();
         final String name = JsonUtils.optString(info, "name");
         String shortLine = name.isEmpty()? text : (name + ": " + text);
-        maybeAddNotification(accountId, dcContext.getChat(dcMsg.getChatId()), msgId, shortLine, shortLine, false);
+        DcChat dcChat = dcContext.getChat(dcMsg.getChatId());
+        maybeAddNotification(accountId, dcChat, msgId, shortLine, shortLine, false, dcChat.isMultiUser());
```
r10s, line 411: how is that "mentioning" supposed to work?

if i understand the code correctly, that would result in _any_ webxdc-notification to  be treated as "mention" and bypass group-muting

adbenitez: that was the initial idea, but that was before "*" came in, oth, notifying all even if group is muted is still nice, and is similar to what happen in a muted group when someone uses the `@all` mention

Hocuri: I think that for webxdcs, `isMention` should either always be `false` or `wasDirectlyNotified` (i.e. not only with `*`). Because I think it would be quite annoying to get notifications `XY added an event to the calendar` all the time even though you muted the group, and look like a bug.

adbenitez: what we could do to mitigate potential annoyances is to have a setting in notifications to disable notification of "mentions" and webxdc in muted chats, like I did for ArcaneChat:

image

adbenitez: 2024? he is moving forward since the 80s

r10s: > Because I think it would be quite annoying to get notifications `XY added an event to the calendar all the time` even though you muted the group, and look like a bug.

agree, we should not do that.

in general, not sure to introduce these "mentions" last minute that the user cannot get rid of.

an option might be nice, but we should have that on all UI then, this is nothing for today.

adbenitez: > in general, not sure to introduce these "mentions" last minute that the user cannot get rid of.

the webxdc notifications in muted groups was intended since the very beginning or that was hpk told me, I am fine with not doing it, then notifications for webxdc will be lost since there is no way to read them later unlike with messages

r10s:
as adding "mentions" last-minute without an option to really mute a chat are questionable - and also need help of core to do correctly (which is also nothing helping getting releases out), i again, suggest to split the PR (in general, it is better to do PR that are doing one thing)

- let's merge the is-foreground-thingie on
- postpone mentions

adbenitez:
created: https://github.com/deltachat/deltachat-android/pull/3460

adbenitez:
> to do better, might also need help of core

I am not sure it is better, notify ALL is also considered a mention in other platforms, and would break the workflow of the Calendar example I described above

adbenitez:
I updated this PR, now this one is about the extra notifications in "mentions" situations in muted chats

r10s:
> I am not sure it is better, notify ALL is also considered a mention in other platforms, and would break the workflow of the Calendar example I described above

well, from that point of view every message could be regarded mention 🤷‍♂️ :)

question is, if a webxdc notification to "*" is more like "Hello" or "@all Hello".  so, a normal message that should be notified as any other message or a explicit mention

adbenitez:
> well, from that point of view every message could be regarded mention 🤷‍♂️ :)
> 
> question is, if a webxdc notify to "*" is more like `Hello` or `@all Hello` - so, a normal message that should be notified as any other message or a explicit mention

the property is called NOTIFY, so yes, it is not like any normal chat message, which we might add the possibility to send to webxdc as already said here and there

r10s:
at a first glance, the option is disabled by default, or?

as from another chat ...

> the chat is muted user will lose webxdc notifications without way to realize

... so that would still be true by default.

src/main/java/com/b44t/messenger/DcContext.java
```diff
@@ -211,6 +213,14 @@ protected void finalize() throws Throwable {
     public native boolean      isSendingLocationsToChat(int chat_id);
     public DcProvider          getProviderFromEmailWithDns (String email) { long cptr = getProviderFromEmailWithDnsCPtr(email); return cptr!=0 ? new DcProvider(cptr) : null; }
 
+    public boolean isMentionsEnabled() {
+      return !"0".equals(getConfig(CONFIG_MENTION_NOTIF_ENABLED));
```
r10s, line 217: while working, it is a bit hard to understand, every string by "0" means "enabled" then, so also "" (which is the default). getConfigInt() would not work as well, although used for setting.

suggestion: call the thing `CONFIG_NOTIIFY_MENTIONS_IF_MUTED` / `ui.notify_mentions_if_muted`, then the default of "" or 0 matches and we can use get/setConfigInt(), and that is also in sync with the other options

src/main/res/values/strings.xml
```diff
@@ -702,6 +702,8 @@
     <string name="pref_screen_security_explain">Request to block screenshots in the recents list and inside the app</string>
     <string name="pref_screen_security_please_restart_hint">To apply the screen security setting please restart the app.</string>
     <string name="pref_notifications">Notifications</string>
+    <string name="pref_mention_notifications">Notify mentions</string>
+    <string name="pref_mention_notifications_explain">Notify messages in muted chats if they are directed to you, like replies or reactions to your messages</string>
```
r10s, line 706: this is 3+ lines explanations, maybe we can get shorter;

```suggestion
    <string name="pref_mention_notifications">Mentions</string>
    <string name="pref_mention_notifications_explain">In muted groups, notify messages directed to you, like replies or reactions</string>
```

"notify" in the title seems redundant.

not sure about the group/chat

r10s (approved):
some minor comments, i am fine with that in general, having a switch we can also experiment more :)

src/main/java/com/b44t/messenger/DcContext.java
adbenitez: > suggestion: call the thing `CONFIG_NOTIIFY_MENTIONS_IF_MUTED` / `ui.notify_mentions_if_muted`, then the default of "" or 0 matches and we can use get/setConfigInt(), and that is also in sync with the other options

renaming would be ok, but I don't see how it changes the problem, "" (unset) still needs to mean "1" (enabled) since it is the default, you want to "ui.notify_mentions_if_muted" to be on by default :thinking:

adbenitez: I renamed it :+1: and I think I improved the code to be easier to understand now

r10s: sorry, my fault 🤦‍♂️ 

i meant to inverse the meaning of the var, so `CONFIG_MUTE_MENTIONS_IF_MUTED`; that way our default matches the default core will return and we do not need to deal with that.

in fact, we had the default-passing in old cores, but got rid of that as it turned out to be tricky and error-prone in detail. tbh, i would prefer to not introduce a getConfig(key, default) again, at least not for a comparable side-quested thing

r10s: i also think, it just does not work. getConfig("ui.unkown_sth") returns an empty string, not null.

the try/catch is only to catch Integer-parsing exceptions

adbenitez: I renamed it again

r10s (approved):
nice, with the option i think it is fine to get it in, even in 1.50.x

PR #3457, opened 2024-11-30
fix WebxdcActivity.callJavaScriptFunction()

adbenitez:
now even if internet access is enabled, the index.html is wrapped in an iframe


r10s:
can you please also say in a short line what the issue/impact was, esp as flagged as a blocker? that makes review focus easier :)

adbenitez:
> can you please also say in a short line what the issue/impact was and why the change is needed? esp as flagged as a blocker, that makes review focus easier :)

the function was executing code based in `index.html` or wrapper being loaded, since now `index.html` is never loaded directly, in apps with internet access this call would fail as you don't want to run the javascript in the parent window but in the iframe

r10s (approved):
lgtm, thanks for clarifications.

i did not test, however, affected xdc would only be the maps.xdc, right?

PR #3460, opened 2024-12-02
do not notify if the webxdc is open

adbenitez:


r10s (approved):
great, thanks for splitting off.

after merging #3456 should be rebased

PR #3462, opened 2024-12-03
prepare 1.50.0

adbenitez:


src/main/res/values/strings.xml
```diff
@@ -1181,7 +1181,7 @@
     <string name="perm_enable_bg_already_done">You already allowed Delta Chat to receive messages in the background.\n\nIf messages still do not arrive in background, please also check your system settings.</string>
 
     <!-- device messages for updates -->
-    <string name="update_1_46_android">New Onboarding 🐣 and More:\n\n⚡️ Sign up to secure fast chatmail servers (https://delta.chat/chatmail) or use classic e-mail servers\n\n👉 PUSH notifications for any chatmail server\n\n👤 Connect your friends securely via \"Attach contacts\"\n\n🥰 React with any emoji\n\n📌 Pin chats directly from search … and MORE at %1$s</string>
+    <string name="update_1_50_android">New Emoji Picker ❤️‍🔥 and More:\n\n🆕 New emoji picker with support for more emojis\n\n🎮 Webxdc apps can now trigger notifications and deep-link to sections directly on opening by tapping in-chat announcements\n\n👍 Notify reactions to your messages\n\n🔔 Change to more natural order of messages in grouped notifications\n\n🥸 Don\'t expose your real email address to Webxdc apps … and MORE at %1$s</string>
```
r10s, line 1184: we are not saying "webxdc" in a user-facing way, in general, the device message could be less technical, and more written for the normie, i.e. (feel free to pick only some things)

```suggestion
    <string name="update_1_50_android">What's new?\n\n❤️‍🔥 New emojis picker with more emoji\n\n🎮 Enhanced in-chat apps: Get notifications and open supporting apps in context\n\n📆 i.e. open an added calendar entry directly\n\n👍 Get notified about reactions to your messages ... \n\n... and bugfixes and MORE at %1$s</string>
```

i also stroke the last two point, i think they're minor and hard to explain the normal will probably have more questions when reading that.

adbenitez: "open supporting apps in context" even I am having a hard time to understand this :sweat_smile:

r10s: > and the explanation about calendar in a separated entry/topic I would put it together in the same item

+1

r10s: > "open supporting apps in context" even I am having a hard time to understand this 😅

agree, it was only the first that came into my mind. how can we say that easier?

PR #3463, opened 2024-12-03
add webxdc picker

adbenitez:
<br/>


src/main/java/org/thoughtcrime/securesms/WebxdcStoreActivity.java
```diff
@@ -0,0 +1,135 @@
...
+        intent.setData(uri);
+        setResult(Activity.RESULT_OK, intent);
+        finish();
+      } else {
+        ByteArrayInputStream data = new ByteArrayInputStream(httpResponse.getBlob());
+        res = new WebResourceResponse(mimeType, httpResponse.getEncoding(), data);
+      }
+    } catch (Exception e) {
+      e.printStackTrace();
+      ByteArrayInputStream data = new ByteArrayInputStream(("Error: " + e.getMessage()).getBytes());
```
Hocuri, line 113: Should show a nicer error message when there is no internet (or make things work without internet)

src/main/AndroidManifest.xml
```diff
@@ -376,6 +376,13 @@
               android:exported="true">
     </activity>
 
+    <activity android:name=".WebxdcStoreActivity"
+              android:label=""
+              android:theme="@style/TextSecure.LightTheme"
+              android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize|uiMode"
+              android:exported="true">
```
Hocuri, line 383: `exported` means that other apps can launch this activity; this isn't necessary here

src/main/AndroidManifest.xml
```diff
@@ -376,6 +376,13 @@
               android:exported="true">
     </activity>
 
+    <activity android:name=".WebxdcStoreActivity"
+              android:label=""
```
Hocuri, line 380: The label should probably be `R.string.webxdc_apps`

adbenitez: I just removed it, it is not present for FullMsgActivity and other some activities, I copied from WebxdcActivity where it is set to empty for some reason but probably also not necessary

src/main/java/org/thoughtcrime/securesms/WebxdcStoreActivity.java
adbenitez: working without internet is out of the scope of this PR, it would be possible if we used a webxdc app with internet access (which I recommend to do) then the webxdc is offline-first and will have something to load and show if the fetching of the remote metadata failed

Hocuri (approved):
Showing a nice error message when there is no internet is blocking, but everything else LGTM, so I'll approve already.

It's sad that it doesn't work without internet and that this introduces a soft dependency to a server, but you can't have everything in life, and in general, having a webxdc picker is great :tada:!

src/main/java/org/thoughtcrime/securesms/WebxdcStoreActivity.java
```diff
@@ -0,0 +1,135 @@
...
+        view.loadUrl(url);
+        return true;
+      }
+
+      @TargetApi(Build.VERSION_CODES.N)
+      @Override
+      public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request) {
+        view.loadUrl(request.getUrl().toString());
+        return true;
+      }
```
Hocuri, line 61: Would be nice to `openUrlInBrowser()` links that don't end on `.xdc`

adbenitez: I like the idea, but also the page shouldn't have all that big banners etc

src/main/java/org/thoughtcrime/securesms/WebxdcStoreActivity.java
r10s: yeah, that can be improved, however, for a first, experimental thing this should also be no blocker :) i have seen worse error messages :)

working without internet is out of scope of this PR, iirc, the idea is to let core handle that at some point

src/main/res/values/strings.xml
```diff
@@ -184,6 +184,8 @@
     <string name="webxdc_app">App</string>
     <!-- plural of "App"; used to present "Webxdc App" (https://webxdc.org) in a user friendly way. Please stay close to the original term and keep it short (it is used in menus with few screen space). -->
     <string name="webxdc_apps">Apps</string>
+    <string name="webxdc_store_url">Apps Picker URL</string>
+    <string name="webxdc_store_url_explain">If set, it will be used as the apps picker URL instead of the default one</string>
```
r10s, line 188: i think, "App Picker" is more correctly here. one would probably also not say "Images Picker" or "Files Selector" but "Image Picker" and "File Selector".

```suggestion
    <string name="webxdc_store_url">App Picker URL</string>
    <string name="webxdc_store_url_explain">If set, the URL will be used as the App Picker instead of the default one</string>
```

also tried to the explanation, the _"'App Picker URL'  will be used as the 'App Picker URL'"_ sounds too recursive to me

r10s (approved):
what an improvement, what an impact! 🛞

i do not think there are real blockers in this PR, sure things can be improved, but it is already discussed that this is only a first step and i.e. the idea is to route things through core, so maybe do not think too much about connectivity and thelike atm, if at the end core can provide things also when offline

the PR already has the flag "wait for merge window", that needs to be decided in detail

src/main/java/org/thoughtcrime/securesms/WebxdcStoreActivity.java
adbenitez: I just implemented it, now random URLs are not open in the webview but externally, only clicks in xdc files are intercepted, anyways the website should remove all that links but it is "good to have" for potential 3rd party URLs since the URL can be changed

src/main/java/org/thoughtcrime/securesms/WebxdcStoreActivity.java
adbenitez: about  the error, one of the problems is that the RPC returns that string instead of just the message, but yeah all that is out of scope of this PR

Hocuri:
```suggestion
      ByteArrayInputStream data = new ByteArrayInputStream(("Could not load apps. Are you online?\n\n" + e.getMessage()).getBytes());
```

src/main/java/org/thoughtcrime/securesms/WebxdcStoreActivity.java
r10s: well, staying inside the webview might also be useful for a static-html-store using several .html pages "alternative" - that could just be the file view from apache or nginx then. 

but yeah, just now it is better to open externally, these "alternatives" do not exist, but webxdc.org/apps with source-urls does

src/main/java/org/thoughtcrime/securesms/WebxdcStoreActivity.java
```diff
@@ -121,6 +121,15 @@ private WebResourceResponse interceptRequest(String url) {
     if (url.endsWith(".jpg") || url.endsWith(".jpeg") || url.endsWith(".png")) {
```
adbenitez, line 121: @link2xt isn't better to just use the mimeType returned by core in HttpResponse?

PR #3467, opened 2024-12-04
add "After 1 year" option to disappearing messages

### LINKED ISSUE #3437: Add "1 year" option for disappearing messages
There is a request to have an option longer than 5 weeks on the forum:
https://support.delta.chat/t/more-options-for-disappearing-messages/3344

This needs to be added to Android first because strings for translations are pulled by Desktop from Android.

adbenitez:
To do:
- [x] add "after 1 year" option to disappearing messages dialog
- [x] wait for  https://github.com/chatmail/core/pull/7084 to be merged
- [x] set new core stock strings
- [x] wait for a new core to be tagged
- [x] update changelog

close #3437


r10s (commented):
while it seems unquestionable that a longer period is needed, the 90 days seems somehow random, and maybe not sufficient, at least not alone, see newer requests at https://support.delta.chat/t/ux-ui-issue-disappearing-messages-seems-to-lack-the-common-choices-of-3-6-12-months/3392 (let's discuss things further  in the upstream issue)

link2xt:
Is this PR stuck?

Users keep asking on the forum https://support.delta.chat/t/more-options-for-disappearing-messages/3344/8

adbenitez:
> Is this PR stuck?

as "90 days" is quite arbitrary, I modify it to "1 year" which fits nicely and makes sense overall, but the core stock strings only expect weeks maximum, not even months, and shows "timer set to 52.1 weeks", so at least an "X months" string is needed

otherwise I could modify it to be something like 12 weeks or 52 weeks which might look a bit weird/arbitrary

cc @r10s

adbenitez:
also, should we remove the first option (1 minute)? to keep the same amount of options

r10s (approved):
lgtm when the new stock string is added and set

adbenitez:
> lgtm when the new stock strings are added and set

done, but we still need a new core release before this can be merged

PR #3468, opened 2024-12-04
Remove prepareMsg (and OUT_PREPARING message state)

Hocuri:
This PR removes prepareMsg, which put messages into the OUT_PREPARING state and was used while videos were recoded:
- It was buggy because when you forwarded a message while it was InPreparation, or when the app was killed while a message is InPreparation, the message would stay InPreparation forever.
- Android is the only UI using this InPreparation (according to @r10s, I didn't check this myself), so we can simplify some things in core, which will also make it easier to deduplicate blob files.

### DIFF
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -1572,12 +1572,6 @@ JNIEXPORT jboolean Java_com_b44t_messenger_DcMsg_isForwarded(JNIEnv *env, jobjec
 }
 
 
-JNIEXPORT jboolean Java_com_b44t_messenger_DcMsg_isIncreation(JNIEnv *env, jobject obj)
-{
-    return dc_msg_is_increation(get_dc_msg(env, obj))!=0;
-}
-
-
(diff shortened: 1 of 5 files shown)


adbenitez (approved):

r10s:
isIncreation() can be removed from dc_wrapper.c and from DcMsg.java as well - it was not used recently anyways, but now soon definetely pointness and removed from core

Hocuri:
> From the code it seems there is no way to cancel recoding in progress, right?

Yes, right. As long as we don't get feature requests for it, I think it's fine not to have it.

Hocuri:
This PR removes prepareMsg, which put messages into the OUT_PREPARING state and was used while videos were recoded. It also removes the now-unused method isIncreation().
- It was buggy because when you forwarded a message while it was InPreparation, or when the app was killed while a message is InPreparation, the message would stay InPreparation forever.
- Android is the only UI using this InPreparation (according to @r10s, I didn't check this myself), so we can simplify some things in core, which will also make it easier to deduplicate blob files.

PR #3473, opened 2024-12-05
fix starting of webxdc

r10s:
the issue is that unset `EXTRA_HREF` is passed around as an empty string, however `Bundle.getString(href, "index.html")` returns `index.html` only when `href` is `NULL`, which is mostly never the case.

this results in webxdc never started with index.html, and then not started at all


r10s:
there are more places where  webxdc is opened and "" might be passed, so it seemed easier, less intrusive to check and more reliable to  check for NULL as well as "" lately

PR #3477, opened 2024-12-06
Make the attachment background the same as the input bar background

Hocuri:
Two different users recently complained that they find drafting images confusing. One of them thought that they had already sent it, the other one didn't know how to send it (both of them figured it out after ~10s, but still, it's nicer to avoid such confusion). In both cases, the problem was that the background of the attached file is the same as the chat background, not the same as the input bar, even though the attached file belongs to the input bar and not to the chat.

We don't have to merge this now, but it's a super small tweak (just 1 line) and I think it can reduce confusion for new users. Also, it makes things look a little more similar to Deltachat-iOS (there, the attachment is on the left, not in the middle, which might look a bit nicer - should also be easy to change if we want to).

What do you think?

Before:

image_2024-12-06_14-41-16 | image_2024-12-06_14-41-13 | image_2024-12-06_14-41-14 | image_2024-12-06_14-41-12
-|-|-|-

After:
image_2024-12-06_14-39-49 | image_2024-12-06_14-39-46 | image_2024-12-06_14-39-48 | image_2024-12-06_14-39-47 
-|-|-|-

### DIFF
--- a/src/main/res/layout/conversation_activity_attachment_editor_stub.xml
+++ b/src/main/res/layout/conversation_activity_attachment_editor_stub.xml
@@ -4,6 +4,7 @@
              android:id="@+id/attachment_editor"
              android:layout_width="match_parent"
              android:layout_height="wrap_content"
+             android:background="?attr/input_panel_bg_color"
              android:gravity="center_horizontal"
              android:visibility="gone">
 


adbenitez:
for images it looks better but for files it is noticeable worse

r10s:
still, it is a step in the right direction. 

in another step, we could also swap quote/attachview, which it is more logical (reply below the quite, reply can be text, image sticker ...) and also what iOS and desktop are doing

to fix the file issue: maybe show the file left-aligned, the "x" button right-aligned, under the quote "x". this is what desktop/ios are doing. or, if easier, use the "outgoing green background" (which is more logical anyways). if would be fine with both

Hocuri:
I tried a bit to do both of these suggestions, but it's not easy to do either of them with the current code because it's a mess, the layout is changed from java code all the time, and it's not easy to change things without breaking others. Probably we need to re-write part of it in pure XML in order to swap quote/attachview, and at this point we can also move the "x" to the right, but until then I found a solution that looks OK without changing too much and without too much effort:

Screenshot_20241218-145658 | Screenshot_20241218-145709
-|-

src/main/res/drawable-ldrtl/message_bubble_background_sent_alone_with_border.xml
```diff
@@ -0,0 +1,17 @@
...
+        android:top="2px"
+        android:bottom="2px">
+        <shape android:shape="rectangle">
+            <stroke android:width="1dp" android:color="#55888888" />
+            <corners
+                android:topLeftRadius="@dimen/message_corner_radius"
+                android:topRightRadius="@dimen/message_corner_radius"
+                android:bottomRightRadius="@dimen/message_corner_radius" />
+            <!-- read issue: https://github.com/deltachat/deltachat-android/issues/480 before introducing an attr here. -->
+            <solid android:color="?android:attr/windowBackground" />
```
adbenitez, line 14: why not use `?attr/conversation_item_incoming_bubble_color` as the previous code?

adbenitez (approved):

gerryfrancis:
FYI, the comment of https://github.com/deltachat/deltachat-android/pull/3477/commits/beee472337ea7009cacca1313a6d614932fb831c is not correct, because we still support Lollipop, but the issue mentioned in the origin was in conjunction with pre-Lollipop generations.

So, nothing to do here, just a false comment. :)

Hocuri:
Oops, yes, the comment should be "we're not supporting pre-lollipop anymore"

PR #3478, opened 2024-12-07
center attachment icons

r10s:
this PR centers the icons in the attachment selector, the right-aligned ones looks esp. odd when no experiments are enabled.

before / after:

 &nbsp;  &nbsp;  &nbsp; 

the default now looks good,
and when all experiments are enabled, things are perfect as well. but even if only one experiment is enabled, things are better imo (but we anyway should optimize for default and not for experimental options):

 

i know there are disucssions about changing that at all, but this PR is an easy fix i could not resist :)

@adbenitez i like the new app-icon, btw 👌


adbenitez:
I was also thinking it will look weird when no experimental options are enabled, you were fast! thanks for taking care!

adbenitez (approved):

PR #3484, opened 2024-12-13
download .xdc files in background

### LINKED ISSUE #3479: Downloading large .xdc results in ANR
As reported in xdc dev group:
anr

My guess is that shouldOverrideUrlLoading should return quickly and not wait for the response here:
https://github.com/deltachat/deltachat-android/blob/a3a6919b08e354921f66b501e24803151ac726c6/src/main/java/org/thoughtcrime/securesms/WebxdcStoreActivity.java#L57
But I have not tested.

To reproduce on a fast phone some sort of sleep can be inserted inside getHttpResponse to simulate long download.

adbenitez:
close #3479


r10s (approved):
also here, code lgtm, did not test - and not sure how that looks to the user when a download really takes some 10 seconds. 

still, probably better than the ANR before

PR #3487, opened 2024-12-13
Prepare 1.50.3

adbenitez:


r10s (approved):
i would leave zhe device message as is, it does not hurt, and there is a very good chance user did not update to 1.50 yet, it is jot even on all platforms

otherwise lgtm

PR #3491, opened 2024-12-16
improve emoji pickers

### LINKED ISSUE #3490: Oversized Emojis
- Android version: 14
- Device: Nokia G42 5G
- Delta Chat version: 1.50.2
- Expected behavior: "normal" size of emoji
- Actual behavior: emojis became really huge (ladnscapemode
- Steps to reproduce the problem: seems to be persistent and only sometimes normal (smaller emojies). I don't know, when it disapears, but sometimes it looks ok. May be the Zoomlevel-Setup for fonts (one step larger, than default android settings) could be a problem.
- Screenshots:
Landscape-Mode-DeltaChat
Portrait-Mode-DeltaChat

adbenitez:
* adapt better to landscape mode
* fix bug: reload emoji picker after orientation changes

close #3490

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
@@ -42,6 +42,7 @@ import android.text.TextWatcher;
 import android.util.Log;
 import android.util.Pair;
 import android.view.KeyEvent;
+import android.view.LayoutInflater;
 import android.view.Menu;
 import android.view.MenuItem;
 import android.view.View;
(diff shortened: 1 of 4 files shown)


src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -321,10 +324,11 @@ public void onConfigurationChanged(Configuration newConfig) {
     composeText.setTransport(sendButton.getSelectedTransport());
     quickAttachmentDrawer.onConfigurationChanged();
 
-    if (emojiDrawerStub.resolved() && container.getCurrentInput() == emojiDrawerStub.get()) {
+    if (emojiPicker != null && container.getCurrentInput() == emojiPicker) {
       container.hideAttachedInput(true);
     }
 
+    reloadEmojiPicker();
```
adbenitez, line 331: potential optimization: keep track of current orientation in a variable and only call this if orientation `!=` previous saved orientation, to avoid reloading in other kind of configuration changes

Hocuri, line 331: Could just set emojiPicker to null instead of fully reloading it on orientation change, because it's hidden anyway

Issue #3488, opened 2024-12-14
FILL500 is sometimes slow if webxdc is closed and reopened multiple times in a row


link2xt:
It looks like WebRTC instances from previous run are not garbage collected and new FILL500 waits for them to get freed up or something like this.

Maybe if we don't reuse WebView and create a new one each time, the problem will be solved.
I.e. using this approach: https://developer.android.com/develop/ui/views/layout/webapps/webview#add-webview-oncreate
Currently Android code uses https://developer.android.com/develop/ui/views/layout/webapps/webview#add-webview-layout

adbenitez:
would be interesting to see if the WebxdcActivity is actually destroyed for you before you click the webxdc again

PR #3492, opened 2024-12-16
create WebView dynamically

adbenitez:
a try to see if this improves situation in #3488


adbenitez:
as expected, this didn't help, we might need to wait a bit before trying to load the webview if it was opened too few time ago

r10s:
thanks a lot for testing out!

PR #3495, opened 2024-12-17
don't throw IllegalStateException

### LINKED ISSUE #3494: CRASH in org.thoughtcrime.securesms.ConversationActivity.initializeResources
```
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:3755)
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:3929)
  at android.app.servertransaction.LaunchActivityItem.execute (LaunchActivityItem.java:103)
  at android.app.servertransaction.TransactionExecutor.executeCallbacks (TransactionExecutor.java:135)
  at android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:95)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2282)
  at android.os.Looper.loopOnce (Looper.java:210)
  at android.os.Looper.loop (Looper.java:299)
  at android.app.ActivityThread.main (ActivityThread.java:8250)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:556)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1045)
Caused by java.lang.IllegalStateException:
  at org.thoughtcrime.securesms.ConversationActivity.initializeResources (ConversationActivity.java:903)
  at org.thoughtcrime.securesms.ConversationActivity.onCreate (ConversationActivity.java:215)
  at org.thoughtcrime.securesms.PassphraseRequiredActionBarActivity.onCreate (PassphraseRequiredActionBarActivity.java:42)
  at android.app.Activity.performCreate (Activity.java:8238)
  at android.app.Activity.performCreate (Activity.java:8206)
  at android.app.Instrumentation.callActivityOnCreate (Instrumentation.java:1329)
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:3728)
```

adbenitez:
the problem is that `CHAT_ID_EXTRA` passed to `ConversationActivity` is `0` in the reported issue, I looked at several places and often it is checked that `chatId !=0` etc. so I couldn't really find a place in which it could be zero, so instead of a crash I changed it to be handle similar to other situations where the chatId is invalid or non-existing chat
 
close #3494

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
@@ -900,7 +900,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     }
     chatId = getIntent().getIntExtra(CHAT_ID_EXTRA, -1);
     if(chatId == DcChat.DC_CHAT_NO_CHAT)
-      throw new IllegalStateException("can't display a conversation for no chat.");
+      chatId = -1;
     dcChat           = dcContext.getChat(chatId);
     recipient        = new Recipient(this, dcChat);
     glideRequests    = GlideApp.with(this);


src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -900,7 +900,7 @@ private void initializeResources() {
     }
     chatId = getIntent().getIntExtra(CHAT_ID_EXTRA, -1);
     if(chatId == DcChat.DC_CHAT_NO_CHAT)
-      throw new IllegalStateException("can't display a conversation for no chat.");
+      chatId = -1;
```
adbenitez, line 903: mmh, actually this seems to be there to prevent from opening the trash chat that happens to be `chatId == 0`?

r10s: nope. the "trash" has id 6 == DC_CHAT_ID_TRASH.

and at comparable places, we're also using `getIntExtra(CHAT_ID_EXTRA, DC_CHAT_NO_CHAT)` 

still, not sure under which circumstances this "throw" can happen and what the gist of the check is

adbenitez: > and at comparable places, we're also using getIntExtra(CHAT_ID_EXTRA, DC_CHAT_NO_CHAT)

mmmh maybe it is a bug but we are using a lot `-1` instead of DC_CHAT_NO_CHAT, so in this line code section, if `CHAT_ID_EXTRA` was not present, it would have been set to `-1` and not throwing the exception then

r10s: yip, i was just mentioning, it is not an error to deal with -1 as well as 0.

not crashing is probably not that bad, but we could also do `finish(); return` - however, for that change, it may be good to have it reproducible (not that we're called when the activity is already in finishing or so ..)


why is the PR closed, btw?

PR #3503, opened 2024-12-17
do not lowercase string 'Bot'

r10s:
it is regarded as a bug
and esp. weird in translations.

it was regarded to be a bit 'nerdy' that time to write it lowercase, but it seems we have to get more serious now.

via https://support.delta.chat/t/two-typos-in-dc/3396 , but tbh, i was also stumbling upon that more than once. better use correct casing, don't make me think.


adbenitez (approved):
ftr, reasoning was also that Telegram uses lowercase for it, and it is the one more prominent mainstream app with good bot support and UIs around it

not that it makes much difference, Bot is also fine, the reported problem was by a German speaker, in German is common to have uppercase for nouns, so it is understandable why it was seen as a bug by that person

r10s:
yeah, however, i am using the english UI and am irritated as well, iirc, also internally there was a question here and there.

PR #3506, opened 2024-12-18
run SendRelayedMessageUtil.sendMultipleMsgs() in background

### LINKED ISSUE #3502: ANR in SendRelayedMessageUtil.sendMultipleMsgs
`dc_send_msg` blocks the main thread:

```
  
      #00  pc 0x000000000005823c  /apex/com.android.runtime/lib64/bionic/libc.so (syscall+28)
    
  
      #01  pc 0x0000000000cd0088  /data/app/~~0Q154b6xQGDTLUevPLXoUw==/chat.delta-o1eYDjDpzVMHK5_--aqNQg==/lib/arm64/libnative-utils.so (BuildId: 9a945705947528941f50411cc3e0b13821e656fb)
    
  
      #02  pc 0x0000000000ccfba8  /data/app/~~0Q154b6xQGDTLUevPLXoUw==/chat.delta-o1eYDjDpzVMHK5_--aqNQg==/lib/arm64/libnative-utils.so (BuildId: 9a945705947528941f50411cc3e0b13821e656fb)
    
  
      #03  pc 0x0000000000e1e8f0  /data/app/~~0Q154b6xQGDTLUevPLXoUw==/chat.delta-o1eYDjDpzVMHK5_--aqNQg==/lib/arm64/libnative-utils.so (BuildId: 9a945705947528941f50411cc3e0b13821e656fb)
    
  
      #04  pc 0x0000000000e1f1ec  /data/app/~~0Q154b6xQGDTLUevPLXoUw==/chat.delta-o1eYDjDpzVMHK5_--aqNQg==/lib/arm64/libnative-utils.so (BuildId: 9a945705947528941f50411cc3e0b13821e656fb)
    
  
      #05  pc 0x00000000008171cc  /data/app/~~0Q154b6xQGDTLUevPLXoUw==/chat.delta-o1eYDjDpzVMHK5_--aqNQg==/lib/arm64/libnative-utils.so (dc_send_msg+424) (BuildId: 9a945705947528941f50411cc3e0b13821e656fb)
    
  
      at com.b44t.messenger.DcContext.sendMsg (Native method)
    
  
      at org.thoughtcrime.securesms.util.SendRelayedMessageUtil.sendMultipleMsgs (SendRelayedMessageUtil.java:76)
    
  
      at org.thoughtcrime.securesms.ConversationActivity.lambda$onActivityResult$0$org-thoughtcrime-securesms-ConversationActivity (ConversationActivity.java:366)
    
  
      at org.thoughtcrime.securesms.ConversationActivity$$ExternalSyntheticLambda12.run (unavailable:4)
    
  
      at org.thoughtcrime.securesms.ConversationActivity.lambda$askSendingFiles$8 (ConversationActivity.java:689)
    
  
      at org.thoughtcrime.securesms.ConversationActivity$$ExternalSyntheticLambda0.onClick (unavailable:2)
    
  
      at androidx.appcompat.app.AlertController$ButtonHandler.handleMessage (AlertController.java:167)
    
  
      at android.os.Looper.loopOnce (Looper.java:243)
    
  
      at android.os.Looper.loop (Looper.java:338)
    
  
      at android.app.ActivityThread.main (ActivityThread.java:8470)
    
  
      at java.lang.reflect.Method.invoke (Native method)
    
  
      at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:600)
    
  
      at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1064)
    
  
```

adbenitez:
close #3502


src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -362,7 +362,7 @@ public void onActivityResult(final int reqCode, int resultCode, Intent data) {
             for (int i = 0; i < uriCount; i++) {
               uriList.add(multipleUris.getItemAt(i).getUri());
             }
-            askSendingFiles(uriList, () -> SendRelayedMessageUtil.sendMultipleMsgs(this, chatId, uriList, null));
+            askSendingFiles(uriList, () -> SendRelayedMessageUtil.sendMultipleMsgs(this, chatId, uriList));
```
r10s, line 365: i think, it would have been easier, to leave the existing function as is and dispatch to background from the caller:

```
            askSendingFiles(uriList, () -> {
                Util.runOnAnyBackgroundThread(() -> {
                    SendRelayedMessageUtil.sendMultipleMsgs(this, chatId, uriList, null));
                }
            }
```

the thing with  the newly added function is that it adds to the mess of not knowing if caller or implementation dispatch to background (as a rule of thumb, to separate concerns, i would go for caller if in doubt: the function remains simple and synchronous, it easier to understand, testable (no issue here) and flexible. the caller handles the complexity of concurrency)

also, we would have two function with nearly the same functionality but totally different names

r10s (approved):
as discussed above, i would _rethink_ who dispatches.

however, if there are good reasons to go the existing way, i am also fine with that, the suggestion is no blocker :)

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
adbenitez: that is what I did, they are two different functions now, one is synchronous and private, and the public one is async, it is only used in a single place at the moment, so I could do as suggested and call it in background from caller, that doesn't protect against making the same mistake again, but that is just precautions for hypothetical futures, so I will go with the simpler suggested solution for now

PR #3512, opened 2024-12-27
fix: Remove long stacktrace that flooded the logcat

Hocuri:
Everytime I attached an image, I got a super long stacktrace in the logcat. I found it to be [this problem](https://stackoverflow.com/questions/71746801/getting-failed-to-inflate-colorstatelist-leaving-it-to-the-framework-when-usi) and applied the fix described there.

### DIFF
--- a/src/main/res/layout/webxdc_view.xml
+++ b/src/main/res/layout/webxdc_view.xml
@@ -20,6 +20,7 @@
                 android:layout_width="128dp"
                 android:layout_height="128dp"
                 android:scaleType="centerCrop"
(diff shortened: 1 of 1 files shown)


adbenitez (approved):
sounds good, this is the webxdc_view tho, so it was not related to attaching an image?

r10s (approved):

Hocuri:
> sounds good, this is the webxdc_view tho, so it was not related to attaching an image?

IIRC, the webxdc view is included in the attachment view.

PR #3513, opened 2024-12-27
File deduplication, Android part

Hocuri:
Depends on https://github.com/deltachat/deltachat-core-rust/pull/6332.

Remaining bugs:
- ~~When sharing an image to DC (or when drafting and re-entering a chat), the dimensions aren't calculated correctly~~ Fixed thanks to adb
- ~~When sharing a video to DC (or when drafting and re-entering a chat) and then clicking on the draft to preview it, I only get a black screen~~ Works fine with https://github.com/deltachat/deltachat-core-rust/pull/6463

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/components/MediaView.java
+++ b/src/main/java/org/thoughtcrime/securesms/components/MediaView.java
@@ -3,10 +3,8 @@ package org.thoughtcrime.securesms.components;
 
 import android.content.Context;
 import android.net.Uri;
-
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
-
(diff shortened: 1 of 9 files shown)


Hocuri:
Probably I should bump the core submodule in this PR so that building the preview APK succeeds? (it's failing because Android now depends on the new `dc_msg_set_file_and_deduplicate()` API)

src/main/java/org/thoughtcrime/securesms/components/MediaView.java
adbenitez, line 1: changes to this file look unrelated to PR

adbenitez (approved):

PR #3516, opened 2025-01-04
avoid NullPointerException when using result of Rpc.getMsgReactions

adbenitez:
when there is no reactions yet for a message `Rpc.getMsgReactions` returns `null`, checking for this explicitly instead of putting noise in the logs with stack traces

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/reactions/AddReactionView.java
+++ b/src/main/java/org/thoughtcrime/securesms/reactions/AddReactionView.java
@@ -16,6 +16,7 @@ import com.b44t.messenger.DcContext;
 import com.b44t.messenger.DcMsg;
 import com.b44t.messenger.rpc.Reactions;
 import com.b44t.messenger.rpc.Rpc;
+import com.b44t.messenger.rpc.RpcException;
 
 import org.thoughtcrime.securesms.R;
 import org.thoughtcrime.securesms.connect.DcHelper;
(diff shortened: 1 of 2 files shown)


src/main/java/org/thoughtcrime/securesms/reactions/AddReactionView.java
```diff
@@ -124,12 +125,14 @@ private String getSelfReaction() {
...
-                result = selfReactions[0];
+            if (reactions != null) {
+                final Map<Integer, String[]> reactionsByContact = reactions.getReactionsByContact();
+                final String [] selfReactions = reactionsByContact.get(DcContact.DC_CONTACT_ID_SELF);
+                if (selfReactions != null && selfReactions.length > 0) {
+                    result = selfReactions[0];
+                }
             }
-        } catch(Exception e) {
+        } catch(RpcException e) {
```
r10s, line 135: why not continuing catching all exceptions?

there may still be unforseeable exceptions thrown. imu, and explicit exception-catch makes only sense when other exceptions are cough elsewhere - or if a crash is wanted. both seems not to be the case here.

adbenitez: The generic catch was due to the NPE that is explicit avoided, there is no other exception to be handled here and in any case explicit handling is always better than generic

adbenitez: in other parts of the code where `Rpc.getMsgReactions` is used, only RpcException is handled, java functions declare what exception they throw, trying to catch potentially future NPE in every single function call makes no sense to me, but if you want to go for it, create a new PR then wrapping every single line of code to be "future-proof", including other places where this same function is called, otherwise it is out of scope of this PR to change that usage just here when then it is not the same in all other parts of the code even when the same RPC function is called

r10s (approved):

src/main/java/org/thoughtcrime/securesms/reactions/AddReactionView.java
r10s: but - do we really want the app to crash in case a null-pointer or out-of-range is thrown? this can totally happen

still,  i am approving, it is more a general thing,  it is not worth letting this block the fix :)

adbenitez: we don't want the app to crash, that is why if Android Studio warns of a potential NPE we catch it, what doesn't make sense is to catch an exception in code that can't throw it, if in the future code change to return null, android studio will warn of unhandled NPE and THEN makes sense to handle it, otherwise code is hard to understand and new exceptions might get catch by "catch-all" when you actually want to handle them differently

adbenitez: > i am not sure, java best-practices agrees with that.

btw https://howtodoinjava.com/best-practices/java-exception-handling-best-practices/#3-3-do-not-catch-the-exception-class-rather-catch-specific-subclasses

it is the same in python, and common sense for any programming language, ofc there are situations where it might make sense, but that is the exception and not the rule, catching all exceptions generically is not recommended

r10s: well, we do not bubble exceptions up and handle them gracefully elsewhere. i just do not trust json-parser here, parsers are complex software 🤷‍♂️

PR #3520, opened 2025-01-06
[WIP] Core now adds a ~ by itself

Hocuri:
This PR is not meant to change user-visible behavior, we just shifted the responsibility to add the "~" to core. (without this PR, the name would be "~~Alice" sometimes)

WIP until https://github.com/deltachat/deltachat-core-rust/pull/6411 is merged.

### DIFF
--- a/src/main/java/com/b44t/messenger/DcMsg.java
+++ b/src/main/java/com/b44t/messenger/DcMsg.java
@@ -168,7 +168,7 @@ public class DcMsg {
     public String getSenderName(DcContact dcContact) {
         String overrideName = getOverrideSenderName();
         if (overrideName != null) {
-            return "~" + overrideName;
+            return overrideName;
         } else {
             return dcContact.getDisplayName();
         }


r10s:
> he clients could decide to use different ways to display this information

but they do not and did not all the years 🤷‍♂️ - cross platform consistency is also easier  by a prefix.  also, having a simple approach as a prefixed character, works better for notifications and other places we do not have so much control about the appearance (eg. reaction strings as "NAME reacted 🤷‍♂️ to MSG")

so, i think, that part is fine, if we agree to prefix virtually all names by `~`

adbenitez:
problems I can see coming: generation of the avatar letters if the name given by core always starts with ~

r10s:
> problems I can see coming: generation of the avatar letters if the name given by core always starts with ~

that's a point

Simon-Laux:
> problems I can see coming: generation of the avatar letters if the name given by core always starts with ~

we can have a new api returning the initials for avatar generation.

message parser has methods that can help us with taking exactly the first emoji (`parser::is_emoji::emoji` (nom parser eating first emoji) and `parser::is_emoji::get_first_emoji(text)`), we could make a function there that also works for other characters that consist of multiple utf8 chars - or copy the function or use some utf8 graphemes crate to do the job of correctly getting the first symbol/letter/emoji.

r10s:
closing stale PR, https://github.com/chatmail/core/pull/6411 is closed as well

Hocuri:
Ah yes, I forgot this, thanks for housekeeping!

PR #3521, opened 2025-01-06
fix: Use getFilename() instead of the actual filename on disk

Hocuri:
In preparation for https://github.com/deltachat/deltachat-core-rust/pull/6332, use the original name of the file (as returned by getFilename()) instead of the current filename on disk.

Most of the commits are cherry-picked from #3513.

What I did:
- Grep for usages of DcMsg.getFile(), check where it's passed, and make sure that all it's never used to check the file extension, to pass it to an external application, or to show it to the user.
- Look at places in the code where a filename is user-visible, like when viewing it in the UI or passing it to an external application (especially when sharing or opening it). Check where the filename comes from, and make sure that it comes from getFilename(), not from getName().
- Test that 
  1. sharing to and from DC
  2. opening
  3. saving ("Export attachment")
  4. drafting and re-entering the chat
  5. opening the drafted file once more
  6. sending
  
  still works for a) Images b) Videos c) vCards (attached contacts) d) attached files e) webxdc's. Also that editing and previewing drafted images works, and that previewing drafted videos works. (I hope I didn't forget to test any combination)

Known bugs:
- ~~The height isn't correctly for image and video drafts calculated because `getManuallyCalculatedSlideInfo()` isn't called, so that DcAttachment gets width and height 0~~
- ~~When tapping the preview of a draft image, a wrong image is shown~~

### DIFF
--- a/src/androidTest/java/com/b44t/messenger/uitests/offline/SharingTest.java
+++ b/src/androidTest/java/com/b44t/messenger/uitests/offline/SharingTest.java
@@ -93,7 +93,7 @@ public class SharingTest {
       }
     }
     Uri uri = Uri.parse("content://" + BuildConfig.APPLICATION_ID + ".attachments/" + Uri.encode(pngImage));
-    DcHelper.sharedFiles.put("/" + pngImage, 1);
+    DcHelper.sharedFiles.put(pngImage, 1);
 
     Intent i = new Intent(Intent.ACTION_SEND);
     i.setType("image/png");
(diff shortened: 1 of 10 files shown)


Hocuri:
I just noticed a bug that was introduced by this, marking as WIP again

Hocuri:
I fixed the bug, ready for review again

src/androidTest/java/com/b44t/messenger/uitests/offline/SharingTest.java
```diff
@@ -93,7 +93,7 @@ public void testShareFromScreenshot() {
       }
     }
     Uri uri = Uri.parse("content://" + BuildConfig.APPLICATION_ID + ".attachments/" + Uri.encode(pngImage));
-    DcHelper.sharedFiles.put("/" + pngImage, 1);
+    DcHelper.sharedFiles.put(pngImage, 1);
```
Hocuri, line 96: For explanation: Since `AttachmentsContentProvider.openFile()` (which checks `sharedFiles`) now uses `getPathSegments().get(0);` instead of `getPath();`, the leading `/` won't be included there. So, it's easier for  `AttachmentsContentProvider.openFile()` to have it without the leading `/`.

adbenitez (approved):

Issue #3517, opened 2025-01-05
Can't open attached PDF of message in extern application


MightyShadow702:
- Android version: 14
- Device: Ulefone 25T Pro
- Delta Chat version: 1.50.3
- Expected behavior: when the PDF opens in my PDF Viewer, it should get shown
- Actual behavior: the PDF Viewer opens but do show nothing, a other one crash
- Steps to reproduce the problem: open any send or received PDF
- Screenshots:
- Logs:

Gh0stz0x:
I can't reproduce the bug.
Have you tried MJ PDF?
If it works let me know, the bug needs to be fixed anyway.

MightyShadow702:
Hey, now I have tried it, with MJ PDF it works. but unfortunately with the Google Doc PDF Viewer or as an example with Collabora Office it does not work.

So a workaround for the moment is to use MJ PDF.

Hocuri:
@adbenitez and me tried to reproduce (& fix) this a bit, but got stuck. At first, it happened on his phone, too (Android 15, PDF viewer "Librera FD" from F-Droid), but then at some point it just didn't happen anymore (probably after he cleared the storage of "Librera FD" [which basically means he uninstalled and reinstalled it]). So, we're back to "can't reproduce".

adbenitez:
with "Libreara FD" it works, but I installed "Collabora Office" and in fact it fails to open the PDF from Delta Chat while from Telegram and WhatsApp it works, here is the log:

```
FATAL EXCEPTION: main
                                                                                                    Process: com.collabora.libreoffice, PID: 26415
                                                                                                    java.lang.RuntimeException: Unable to start activity ComponentInfo{com.collabora.libreoffice/org.libreoffice.androidlib.LOActivity}: java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.equals(java.lang.Object)' on a null object reference
                                                                                                    	at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:4129)
                                                                                                    	at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:4316)
                                                                                                    	at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:222)
                                                                                                    	at android.app.servertransaction.TransactionExecutor.executeNonLifecycleItem(TransactionExecutor.java:133)
[...]

PR #3528, opened 2025-01-10
improve bug report template

albjeremias:


Hocuri:
Thanks a lot for your PR!

I however think that the new template is too verbose, for many issues (e.g. https://github.com/deltachat/deltachat-android/issues/3517) it's not necessary to write a long issue like this.

r10s:
yes, i also had a look at the PR today, and i also had the feeling, it is a bit too verbose.

but thanks a lot, if we get the feeling that bug reports are too worse and we need more instructions, we'll remember this PR

closing it for now

PR #3529, opened 2025-01-10
Repair scripts/ndk-make.sh: Remove jsonrpc flag

Hocuri:
Adapt to https://github.com/deltachat/deltachat-core-rust/pull/6417


Hocuri:
Looks like we can only either build with the latest core release or with current core main until we either revert https://github.com/deltachat/deltachat-core-rust/pull/6417 (and make jsonrpc default, but don't remove it for now) or update the submodule.

PR #3558, opened 2025-01-23
pass account ID in chat shortcuts and properly switch to account

### LINKED ISSUE #3557: chat-shortcuts on the homescreen do not switch to correct profile
chat-shortcuts added to the homescreen as described eg at https://mastodon.social/@r10s/113874981888519009 do no switch to the correct profile, if you are using multiple profiles.

instead, one may end up in the weill-known ghost chat - or probably in a random chat if another profile has randomly the same chatId

for fixing, one may peek at what is done for the notifications, that got multi-profile support recently

Image

adbenitez:
close #3557

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ShareActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ShareActivity.java
@@ -56,6 +56,7 @@ public class ShareActivity extends PassphraseRequiredActionBarActivity implement
 {
   private static final String TAG = ShareActivity.class.getSimpleName();
 
+  public static final String EXTRA_ACC_ID = "acc_id";
   public static final String EXTRA_CHAT_ID = "chat_id";
   public static final String EXTRA_TITLE = "extra_title";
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/thoughtcrime/securesms/ShareActivity.java
```diff
@@ -206,13 +207,9 @@ public void onMediaResolved(Uri uri) {
...
 
   private void handleResolvedMedia(Intent intent) {
+    int       accId            = intent.getIntExtra(EXTRA_ACC_ID, -1);
     int       chatId           = intent.getIntExtra(EXTRA_CHAT_ID, -1);
 
-    String shortcutId = intent.getStringExtra(ShortcutManagerCompat.EXTRA_SHORTCUT_ID);
-    if (chatId == -1 && shortcutId != null) {
-      chatId = Integer.parseInt(shortcutId);
-    }
-
```
r10s, line 215: wondering where this came from, looks useless :)

r10s (approved):
nice! straight-forward, thanks a lot for fixing!

src/main/java/org/thoughtcrime/securesms/ShareActivity.java
adbenitez: Yes, couldn't find any similar usage in the Internet, maybe we were using the intent ID to pass the chat id at the beginning so this was kept as a migration step or it was from android <=4 times

adbenitez: And we don't even do it for other shortcuts we create

PR #3560, opened 2025-01-23
prepare 1.52.0

adbenitez:


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Delta Chat Android Changelog
 
+## v1.52.0
+2025-01
+
+* new group consistency algorithm
```
Hocuri, line 6: Not sure if this is too technical, most users won't understand what it means. Maybe it's possible to find sth everyone understands. In the general direction of "Avoid situations where some members of a group think that Alice is part of the group and others don't" but this is too long and also everyone will wonder "who is Alice??" - I can't think of a really good phrasing myself.

Hocuri (approved):

Hocuri:
@adbenitez probably we shouldn't use auto-merge on PRs that don't have a review yet :joy:

CHANGELOG.md
Hocuri: "Avoid situations where some members of a group think that someone is part of the group and others don't" or "Avoid situations where the members of a group have differing infos on who is part of the group" could work

adbenitez:
> @adbenitez probably we shouldn't use auto-merge on PRs that don't have a review yet 😂

oops!! yeah, in theory if one has something one wants to be changed before merged, the "request changes" or just comment should be used instead of approving, but it is hard to overlook, I honestly was not expecting any changes needed hence why I enabled auto-merge once someone just approved the PR

CHANGELOG.md
adbenitez: otoh, "algorithm" has become a more well known word in main-stream media, ex "twitter/facebook filter algorithm" etc. and "group consistency" is clear and anything having a "new" label should be good/better right??? :rofl:

adbenitez: some people might have never experienced these problems so mentioning it might work more as a negative point than positive "so this was actually that broken?"

PR #3562, opened 2025-01-27
remove unused com.soundcloud.android:android-crop dependency

adbenitez:
this dependency was necessary only for SDK<19 and not needed anymore that min SDK is 21


r10s (approved):
probably one of the last "android4" tags we add to and PR :)

Hocuri (approved):
Always a great feeling to get rid of stuff!

PR #3572, opened 2025-02-01
be more clear about the impact of blocked contacts

r10s:
this PR is about  to improve the wording of the dialog shown when a contact is blocked, aligning that more with the status quo

Screenshot 2025-02-01 at 15 07 35

ftr, as @adbenitez just reconfirmed, telegram also shows messages of blocked contacts in groups, i think, idea is to make discussions possible - otherwise there are lots of "i did not read the message" from various ppl. if the issue with a user is that serious, one should leave shared groups


adbenitez (approved):

r10s:
thinking it i over, the "existing" is still a bit misleading - others can create groups with blocked contacts

PR #3580, opened 2025-02-03
align avatar to message bubble instead of parent bottom

### LINKED ISSUE #3566: Avatar is shifted down when reaction is added
Adding reaction results in the avatar being shifted down:
Image

It should not move down and still line up with the bottom of the message bubble instead, like when the reaction is not there.
Delta Chat Desktop does not have this problem.

adbenitez:
close #3566

### DIFF
--- a/src/main/res/layout/conversation_item_received.xml
+++ b/src/main/res/layout/conversation_item_received.xml
@@ -43,7 +43,7 @@
             android:layout_height="wrap_content"
             android:layout_alignParentLeft="true"
             android:layout_alignParentStart="true"
-            android:layout_alignParentBottom="true">
+            android:layout_alignBottom="@id/body_bubble">
 
             <org.thoughtcrime.securesms.components.AvatarImageView
                 android:id="@+id/contact_photo"


adbenitez:
WARNING: don't approve if you don't want the PR to be merged

Hocuri (approved):
Looks good, didn't test

adbenitez:
yep I tested it, it works as expected, I tested with several situations like images, stickers, etc

Hocuri (approved):
OK, then it's fine to merge

Hocuri:
Honestly I'm not super fond of auto-merging because when I approve, I often remark _something_ the author should read before merging. I leave a small comment, I say that I didn't test it (which to me implies "please only merge if you tested it", I hope it also does for others). If clicking "approve" becomes a de-facto "merge", then I can't click "approve" until the author came back to me, which means an extra round trip (I comment->author comments->I approve&merge instead of just I approve->author comments&merges). And checking if auto-merge is enabled is an extra thing I have to keep in mind when approving. Additionally, sometimes it's nice to give others the time to look at a PR if they happen to have the time (not this one ofc, this one is super trivial), but if the PR is automatically merged after one review then this won't happen.

The only upside is saving 0.5 round trips when I actually don't have any comments, i.e. the PR is merged when it's approved, not when the PR author looks onto the PR next time.

adbenitez:
yes, I only enabled it because the change is trivial, and I also test things before pushing so I know I actually fixed the problem

PR #3581, opened 2025-02-04
fix location streaming

### LINKED ISSUE #3563: Location streaming does not work
When I enable location streaming, green location icon appears briefly in the status bar and then disappears. It seems this way no location is actually obtained unless I also run another app such as Organic Maps that properly requests location, because GPS does not have enough time to obtain even a single measurement.

When I start recording the track with Organic Maps with "Record track" feature, it constantly shows that app is accessing location in the status bar. I think Delta Chat should do the same.

I am testing with Android 14. I have location permission "Allowed all the time" for the app.

Our location code is at https://github.com/deltachat/deltachat-android/tree/main/src/main/java/org/thoughtcrime/securesms/geolocation
Android documentation for reference is at https://developer.android.com/develop/sensors-and-location/location/request-updates

adbenitez:
close #3563


adbenitez:
this was superseded by https://github.com/deltachat/deltachat-android/pull/4361

PR #3582, opened 2025-02-04
remove deprecated half-camera dependency

adbenitez:
the half-camera dependency is marked as deprecated and archived, unmaintained since 11 years ago, see: https://github.com/JakeWharton/NineOldAndroids

since a long while we switched to using system camera by default and option was buried in  advanced settings


r10s:
wasn't there a report that on some system the "camera activity" did not work but the "half camera" did?

iirc, we said, that was a bug in the OS that time, but that could strike back now. but, yeah, on OS where the "camera activity" is broken, probably much more things are questionable :)

adbenitez:
That was on CalyxOS, and it was fixed long ago, it is kind of a hacky system not your stock room, in any case it is not in the hand of Delta Chat to fix broken systems shipping a 11yo abandoned lib, which is not enabled by default so user has to discover it exist at all, even then user can still take pictures with camera and then select it in Delta Chat

There are closed issues related to this half-camera that we closed them precisely because we said we will remove it, in general it is not worth the efforts

r10s (approved):
makes sense, what a great feeling to throw away code!

did not text, however

Hocuri (approved):
Nice!!! Always feels great to get rid of stuff

PR #3585, opened 2025-02-05
fix: Opening files in external apps always works now

### LINKED ISSUE #3517: Can't open attached PDF of message in extern application

Hocuri:
The problem was that we returned `null` from `getType()`.

fix #3517


adbenitez (approved):

r10s (approved):
nice!

PR #3586, opened 2025-02-05
improve improved saved messages

adbenitez:


r10s (approved):
lgtm.

we should rename "star" to "bookmark" (or scarf :)

final review is easier in the original PR as this also reverts quite some changes or the original PR

PR #3587, opened 2025-02-07
Add "Past Members" string

link2xt:
It is needed to translate https://github.com/deltachat/deltachat-desktop/pull/4531


r10s (approved):

r10s:
i'll merging directly and push the string to transifex, so it can be pulled from desktio

PR #3588, opened 2025-02-08
fix: hide 'clone chat' if there is no chat

r10s:
before, 'clone chat' was hidden for contacts having a one-to-one-chats, but not for contacts _without_ a one-to-one-chat

opening a contact's profile for contacts without a chat, is not so unusual when eg. tapping the avatar of unknown members in a group

good news is that there was no crash or so, only an empty group was created :)


Hocuri (approved):
Nice!

adbenitez (approved):
Nice!

PR #3592, opened 2025-02-11
Prep 1.54.0

adbenitez:


build.gradle
```diff
@@ -27,8 +27,8 @@ android {
     useLibrary 'org.apache.http.legacy'
 
     defaultConfig {
-        versionCode 710
-        versionName "1.52.1"
+        versionCode 711
+        versionName "1.54.0"
```
Hocuri, line 31: Shouldn't it be 1.53.0 first? In general, odd numbers are prereleases and even numbers are releases, i.e. you make a 1.53.0, let test for a week or so, and then make 1.54.0 if things work fine. Or do we want to go directly to the stores with this?

Hocuri (approved):

build.gradle
adbenitez: yes, plan is directly going for the store, if things get bad we can make a 1.54.x as usual, the store takes some days to approve and we can control rolling, no need to do pre-release only,

I am going for 1.54.0 just to match with desktop numbering

r10s: yip, and one can rollout slowly. "real test-only releases" are done only where it was clear from the beginning on, that this will never hit the store. but yeah, there is always quite some confusion about versioning, there is not really a documentation or so about that :)

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # Delta Chat Android Changelog
 
+## v1.54.0
+2025-02
+
+* enhanced "Saved Messages" feature, now when forwarding a message to "Saved Messages" chat, it retains the sender information and a button to jump to the original message
```
r10s, line 6:
```suggestion
* enhanced "Saved Messages" feature, now when forwarding a message to "Saved Messages" chat, it retains the sender information and a button to jump to the original message
* Saved messages are marked by a bookmark sign
```

r10s (approved):

PR #3594, opened 2025-02-11
hide some option in Encryption category

### LINKED ISSUE #3591: Remove "Encryption" section from chatmail profiles
Android already has "Prefer End-to-End Encryption" option removed for chatmail, but the other two settings "Manage Keys" and "Send Autocrypt Setup Message" remain.

"Manage keys" is dangerous because export option exports the key into Downloads folder and then the key may be read by other applications, uploaded etc. And we really don't want users to import keys in their chatmail profiles because if they import the same key into multiple profiles, this will trigger AEAP if both profiles are added to the same chat at some point. We generally don't expect that multiple Delta Chat profile with the same key may exist.

"Send Autocrypt Setup Message" is slightly better, but it does not make sense for chatmail because users normally don't have other clients supporting Autocrypt on the other side. It is possible to setup Thunderbird by copying the password and then use Autocrypt Setup Message to move the key there as a developer option, but as a developer I can also setup desktop as a second device and extract the key on desktop, even with SQLite browser.

We could even remove the hacky "Manage keys" option from non-chatmail profiles because it does not have the way to see which key is now used and how many keys you have, but this will make users who like to manage keys and compare fingerprints unhappy, so better only change the chatmail for now.

adbenitez:
close #3591, similar to https://github.com/deltachat/deltachat-ios/pull/2636

- Remove "Prefer end-to-end encryption"
- Remove option to import key by file or autocrypt setup message (ASM), targets targets https://github.com/deltachat/interface/issues/80
- Remove option to export key file, which doesn't really make sense on mobile because it was exporting the key file without any passphrase to the world-readable Downloads folder
- Remove receiving ASM
- Deprecate unused strings

Sending ASM stays for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * add mute option "8 hours"
 * add menu option to easily save/unsave selected message
 * improve deletion confirmation for "Device Messages"
+* remove dangerous encryption options
 
 ## v1.54.4
 2025-03
--- a/src/main/java/com/b44t/messenger/DcMsg.java
+++ b/src/main/java/com/b44t/messenger/DcMsg.java
@@ -150,7 +150,6 @@ public class DcMsg {
     public native String  getWebxdcHref      ();
     public native boolean isForwarded        ();
     public native boolean isInfo             ();
-    public native boolean isSetupMessage     ();
     public native boolean hasHtml            ();
     public native String  getSetupCodeBegin  ();
     public native String  getVideochatUrl    ();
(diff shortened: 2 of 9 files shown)


r10s:
when this is merged, we need to file an issue/pr to adapt the faq

adbenitez:
updated to new requirements this needs a new core that ignores e2ee preference since now it is gone for all profiles

Hocuri (approved):

Hocuri (approved):

PR #3595, opened 2025-02-12
allow to add any chat to the home screen

adbenitez:
currently we already allow to add some chats to the home screen, but in a limited way, with the launchers that support some "quick actions" when long-pressing app icon and it only displays the 3 more resent chats + the "Saved Messages" chat, with this PR people can add to home screen any chat without the mentioned limitations:

 
image_2025-02-12_16-58-03

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/BaseConversationListFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/BaseConversationListFragment.java
@@ -11,6 +11,7 @@ import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
 import android.content.res.TypedArray;
+import android.graphics.Bitmap;
 import android.graphics.Color;
 import android.net.Uri;
 import android.os.AsyncTask;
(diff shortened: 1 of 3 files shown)


src/main/java/org/thoughtcrime/securesms/BaseConversationListFragment.java
```diff
@@ -316,9 +322,37 @@ private void handleSelectAllThreads() {
...
+        .setShortLabel(chat.getName())
+        .setIcon(IconCompat.createWithAdaptiveBitmap(avatar))
+        .setIntent(intent)
+        .build();
+    if (!ShortcutManagerCompat.requestPinShortcut(activity, shortcutInfoCompat, null)) {
+        Toast.makeText(activity, "ErrAddToHomescreen: requestPinShortcut() failed", Toast.LENGTH_LONG).show();
+    } else if (actionMode != null) {
+      actionMode.finish();
+      actionMode = null;
+    }
```
r10s, line 348: was first wondering why a PIN is requested :)

anyways, should't  `actionMode` also be finished in case of errors? not that it makes a big difference here, but this is what we're usually doing. not a blocker or so, just wondering :)

r10s (approved):
very nice!

code lgtm, also tested on api34 in emulator

src/main/java/org/thoughtcrime/securesms/BaseConversationListFragment.java
adbenitez: if it fails it is likely permission for shortcut was denied and user could go to app settings and enable it again, without needing to select the chat again, which is a relative small gain, selecting the chat again is not a big deal, 

but I think the biggest reason is: if it succeeds, it is finished, but if it fails, it is good that state remains as is to indicate the user that nothing changed and the action didn't succeeded, because otherwise a distracted user might think the shortcut was added to the home screen if they don't read and understand the toast in time and think it was a "added to home scree" toast etc

r10s: resonabe, thanks for explaining

r10s (approved):

PR #3596, opened 2025-02-16
cleanup ApplicationPreferencesActivity

adbenitez:


src/main/java/org/thoughtcrime/securesms/ApplicationPreferencesActivity.java
```diff
@@ -1,38 +1,14 @@
-/*
```
Hocuri, line 1: Why remove the copyright notice? I would have expected that we leave it in, and add "Delta Chat contributors" to the list?

adbenitez: mmmh I was considering it outdated/unnecessary I will revert it then, I really just wanted to remove the dead code, not start a discussion

Hocuri (approved):

PR #3600, opened 2025-02-19
use JSON-RPC for addAccount

adbenitez:


r10s (approved):
did not test, code lgtm

PR #3609, opened 2025-02-25
reword message deletion string

r10s:
that should get in when using a core with https://github.com/deltachat/deltachat-core-rust/pull/6573

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -369,14 +369,14 @@
     </plurals>
     <string name="ask_delete_named_chat">Delete chat \"%1$s\"?</string>
     <string name="ask_delete_message">Are you sure you want to delete this message?</string>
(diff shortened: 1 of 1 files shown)


src/main/res/values/strings.xml
```diff
@@ -375,8 +375,8 @@
         <item quantity="other">Delete %d messages here and on the server?</item>
     </plurals>
     <plurals name="ask_delete_messages_simple">
-        <item quantity="one">Delete %d message?</item>
-        <item quantity="other">Delete %d messages?</item>
+        <item quantity="one">Delete %d message on all devices?</item>
```
adbenitez, line 378: I think it is better if it says "on all YOUR devices" since otherwise it looks like this delete in all devices of all people

r10s:
merging that in already, to give translators time

PR #3616, opened 2025-02-25
cleanup AvatarImageView

adbenitez:


src/main/java/org/thoughtcrime/securesms/components/AvatarImageView.java
```diff
@@ -30,11 +29,6 @@ public AvatarImageView(Context context) {
   public AvatarImageView(Context context, AttributeSet attrs) {
     super(context, attrs);
     setScaleType(ScaleType.CENTER_CROP);
-
-    if (attrs != null) {
-      TypedArray typedArray = context.getTheme().obtainStyledAttributes(attrs, R.styleable.AvatarImageView, 0, 0);
-      typedArray.recycle();
-    }
```
adbenitez, line 37: pointless statement, loading some attributes but not actually using them, probably leftovers after the code actually using it got removed

src/main/java/org/thoughtcrime/securesms/components/AvatarImageView.java
```diff
@@ -64,8 +58,8 @@ public void clear(@NonNull GlideRequests glideRequests) {
     glideRequests.clear(this);
   }
 
-  private void setAvatarClickHandler(final Recipient recipient, boolean quickContactEnabled) {
-    if (!recipient.isMultiUserRecipient() && quickContactEnabled) {
+  private void setAvatarClickHandler(final Recipient recipient) {
+    if (!recipient.isMultiUserRecipient()) {
```
adbenitez, line 62: `quickContactEnabled` is always true

r10s (approved):

PR #3673, opened 2025-03-12
avoid NPE in processComposeControls

adbenitez:
#skip-changelog since this was never released


r10s (approved):
thanks for taking care!

PR #3674, opened 2025-03-12
add action menu icon option to toggle message bookmark

### LINKED ISSUE #3599: use dedicated save button
after some first feedback, in addition to "forward to saved" we want to have a dedicated "save" button.

this lets user save one tap and stay in context without screen switching, making the "save" function even nicer.

to make space for that, idea is to move the "info" button to the overflow menu (atm, "info" is a debug view anyways)

adbenitez:
close #3599

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationFragment.java
@@ -327,6 +327,7 @@ public class ConversationFragment extends MessageSelectorFragment
             menu.findItem(R.id.menu_context_edit).setVisible(false);
             menu.findItem(R.id.menu_context_reply_privately).setVisible(false);
             menu.findItem(R.id.menu_add_to_home_screen).setVisible(false);
+            menu.findItem(R.id.menu_toggle_save).setVisible(false);
         } else {
             DcMsg messageRecord = messageRecords.iterator().next();
             DcChat chat = getListAdapter().getChat();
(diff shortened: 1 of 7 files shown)


src/main/java/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -339,6 +340,12 @@ protected void setCorrectMenuVisibility(Menu menu) {
             boolean showReplyPrivately = chat.isMultiUser() && !messageRecord.isOutgoing() && canReply;
             menu.findItem(R.id.menu_context_reply_privately).setVisible(showReplyPrivately);
             menu.findItem(R.id.menu_add_to_home_screen).setVisible(messageRecord.getType() == DcMsg.DC_MSG_WEBXDC);
+
+            boolean saved = messageRecord.getSavedMsgId() != 0;
+            MenuItem toggleSave = menu.findItem(R.id.menu_toggle_save);
+            toggleSave.setVisible(messageRecord.canSave() && !chat.isSelfTalk());
+            toggleSave.setIcon(saved? R.drawable.baseline_bookmark_border_24 : R.drawable.baseline_bookmark_24);
```
r10s, line 347: let's do that the other way round, it feels wrong showing action but not state here.

also as the other  icons are less bold.

cmp. to the proxy settings, where we also have "filled" for enabled.

```suggestion
            toggleSave.setIcon(saved? R.drawable.baseline_bookmark_24 : R.drawable.baseline_bookmark_border_24);
```

PR #3676, opened 2025-03-12
feat: Move to addTransport / addTransportFromQr API

Hocuri:
cc @adbenitez 

Depends on https://github.com/chatmail/core/pull/6680, can be merged since we already updated core to v1.158.0

#skip-changelog


src/main/java/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -618,6 +620,26 @@ private void setupConfig() {
...
+                EnteredServerLoginParam.SocketSecurity.values()[imapSecurity.getSelectedItemPosition()],
+                getParam(R.id.imap_login_text, false),
+                getParam(R.id.password_text, false)
+        );
+        EnteredServerLoginParam smtpParam = new EnteredServerLoginParam(
+                getParam(R.id.imap_server_text, true),
+                Util.objectToInt(getParam(R.id.imap_port_text, true)),
+                EnteredServerLoginParam.SocketSecurity.values()[imapSecurity.getSelectedItemPosition()],
+                getParam(R.id.imap_login_text, false),
+                getParam(R.id.password_text, false)
```
Hocuri, line 636: TODO "imap" -> "smtp"

src/main/java/com/b44t/messenger/rpc/EnteredLoginParam.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+        throw new IllegalArgumentException("Invalid certificate position: " + position);
+    }
+
+    public enum SocketSecurity {
+        // Unspecified socket security, select automatically.
+        automatic,
+
+        // TLS connection.
+        ssl,
```
adbenitez, line 149: is this properly converted to string? sounds to me that the enum values would be converted to int while the socket security config expects a string, right? :thinking:

adbenitez (approved):

PR #3677, opened 2025-03-12
no 'delete on all devices' for device messages

r10s:
do not say "Delete on all your devices"
when deleting a message in the "Device Messages" chat.

while this is a minor,
it is one of the first things the user may see,
better be correct there to not give a false first impression of correctness :)

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/MessageSelectorFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/MessageSelectorFragment.java
@@ -75,7 +75,9 @@ public abstract class MessageSelectorFragment
       canDeleteForAll = false;
     }
 
-    String text = requireActivity().getResources().getQuantityString(R.plurals.ask_delete_messages, messageIds.length, messageIds.length);
+    String text = getActivity().getResources().getQuantityString(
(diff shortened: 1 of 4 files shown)


src/main/java/org/thoughtcrime/securesms/MessageSelectorFragment.java
```diff
@@ -61,9 +61,14 @@ protected void handleDeleteMessages(int chatId, final Set<DcMsg> messageRecords)
   }
 
   protected void handleDeleteMessages(int chatId, final int[] messageIds) {
-    DcChat dcChat = DcHelper.getContext(getActivity()).getChat(chatId);
+    boolean isDeviceTalk = false;
+    if (chatId != 0) {
+      isDeviceTalk = DcHelper.getContext(getActivity()).getChat(chatId).isDeviceTalk();
+    }
```
adbenitez, line 67: could have been
```suggestion
    boolean isDeviceTalk = chatId != 0 && DcHelper.getContext(getActivity()).getChat(chatId).isDeviceTalk();
```

adbenitez (approved):

PR #3687, opened 2025-03-18
avoid crash when clicking orphan info messages from webxdc

adbenitez:


r10s (approved):
lgtm, thanks for fixing!

PR #3689, opened 2025-03-18
don't use email address as contact unique identifier

adbenitez:
move away from using email address as contact identifier, try to use contact ID when possible

#skip-changelog: internal change, not user-visible change


r10s (approved):
lgtm so far, did not test.

we should merge that in, so things can be tested in nightlies etc.

PR #3694, opened 2025-03-20
use Rpc to set contact's name

### LINKED ISSUE #3692: Set display name by using `change_contact_name` JSON-RPC API
Here name is changed by creating a new contact with the same address and a new name:
https://github.com/deltachat/deltachat-android/blob/fbca401488a588b5a88f7160c0b8997b6676eb9c/src/main/java/org/thoughtcrime/securesms/ProfileActivity.java#L557-L558

Currently JSON-RPC API does the same thing, but I am going to change it so it works for PGP-contacts (https://github.com/chatmail/core/pull/6683). Using old method of attempting to create a new contact with the same email address will not change the name of PGP-contact, but change the name of the email contact or create a new one even.

adbenitez:
close #3692 

#skip-changelog: internal change from FFI to RPC

### DIFF
--- a/src/main/java/com/b44t/messenger/rpc/Rpc.java
+++ b/src/main/java/com/b44t/messenger/rpc/Rpc.java
@@ -129,6 +129,10 @@ public class Rpc {
         return getResult("get_account_file_size", accountId).getAsInt();
     }
 
+    public int createContact(int accountId, String email, String name) throws RpcException {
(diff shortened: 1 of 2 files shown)


adbenitez:
> But this is just using the same API over RPC, not `change_contact_name`.

ah there is a specific api nice! that was not said in the issue description, just "change to RPC" so I was thinking the existing api was going to change internally

Hocuri (approved):

PR #3698, opened 2025-03-25
feat: Remove AEAP message

Hocuri:
Based on #3676. AEAP is disabled in core until it is replaced by multi-transport; this PR removes the (now-wrong) info message.

PR #3707, opened 2025-04-01
tune down email wording

r10s:
this PR tunes down "e-mail"  in the primary UI,  main left overs are "classic login" and the welcome message that referenced email on purpose (we may want to change that, however)

moreover, where appropriate, we go already for "routing" instead of "server"

i also changed the translator hints and some wordings not in the primary UI, where not strictly needed - it makes a difference how we talk about that :)

largest part left, is  the "Show Classic E-Mail" option, that we should probably just hide in case of chatmail (it anyway defaults to "all") EDIT: did a PR: https://github.com/deltachat/deltachat-android/pull/3708

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -52,7 +52,7 @@
     <string name="mute">Mute</string>
     <string name="muted">Muted</string>
     <string name="ephemeral_messages">Disappearing Messages</string>
-    <string name="ephemeral_messages_hint">Applies to all members of this chat if they use Delta Chat; they can still copy, save, and forward messages or use other apps.</string>
+    <string name="ephemeral_messages_hint">Applies to all members of this chat; they can still copy, save, and forward messages.</string>
     <string name="save">Save</string>
     <string name="chat">Chat</string>
     <string name="media">Media</string>
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -312,7 +312,7 @@
     <string name="device_talk">Device Messages</string>
     <string name="device_talk_subtitle">Locally generated messages</string>
     <string name="device_talk_explain">Messages in this chat are generated on your device to inform about app updates and problems during usage.</string>
-    <string name="device_talk_welcome_message2">Get in contact!\n\n🙌 Tap \"QR code\" on the main screen of both devices. Choose \"Scan QR Code\" on one device, and point it at the other\n\n🌍 If not in the same room, scan via video call or share an invite link from \"Scan QR code\"\n\nThen: Enjoy your messages routed over the largest decentralized network that ever existed: E-mail. And in contrast to other popular apps, without central control or tracking or selling you, friends, colleagues or family out to large organizations.</string>
+    <string name="device_talk_welcome_message2">Get in contact!\n\n🙌 Tap \"QR code\" on the main screen of both devices. Choose \"Scan QR Code\" on one device, and point it at the other\n\n🌍 If not in the same room, scan via video call or share an invite link from \"Scan QR code\"\n\nThen: Enjoy your messages relayed over the largest decentralized network that ever existed: E-mail. And in contrast to other popular apps, without central control or tracking or selling you, friends, colleagues or family out to large organizations.</string>
     <string name="edit_contact">Edit Contact</string>
     <!-- Verb "to pin", making something sticky, not a noun or abbreviation for "pin number". -->
     <string name="pin_chat">Pin Chat</string>
@@ -939,7 +939,7 @@
     <!-- this may be shown instead of the chat's input field, with buttons "More Info" and "OK" -->
     <string name="chat_protection_broken">%1$s sent a message from another device.</string>
     <string name="chat_protection_enabled_tap_to_learn_more">Messages are guaranteed to be end-to-end encrypted from now on. Tap to learn more.</string>
-    <string name="chat_protection_enabled_explanation">It is now guaranteed that all messages in this chat are end-to-end encrypted.\n\nEnd-to-end encryption keeps messages private between you and your chat partners. Not even routers, servers or providers can read them.</string>
+    <string name="chat_protection_enabled_explanation">It is now guaranteed that all messages in this chat are end-to-end encrypted.\n\nEnd-to-end encryption keeps messages private between you and your chat partners. Not even servers, providers or relays can read them.</string>
     <string name="chat_protection_broken_tap_to_learn_more">%1$s sent a message from another device. Tap to learn more.</string>
     <string name="chat_protection_broken_explanation">End-to-end encryption cannot be guaranteed anymore, likely because %1$s reinstalled Delta Chat or sent a message from another device.\n\nYou may meet them in person and scan their QR code again to reestablish guaranteed end-to-end encryption.</string>
     <string name="invalid_unencrypted_tap_to_learn_more">⚠️ %1$s requires end-to-end encryption which is not setup for this chat yet. Tap to learn more.</string>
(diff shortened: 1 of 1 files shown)


src/main/res/values/strings.xml
```diff
@@ -52,7 +52,7 @@
     <string name="mute">Mute</string>
     <string name="muted">Muted</string>
     <string name="ephemeral_messages">Disappearing Messages</string>
-    <string name="ephemeral_messages_hint">Applies to all members of this chat if they use Delta Chat; they can still copy, save, and forward messages or use other e-mail clients.</string>
+    <string name="ephemeral_messages_hint">Applies to all members of this chat if they use Delta Chat; they can still copy, save, and forward messages or use other apps.</string>
```
Hocuri, line 55: I think that "if they use Delta Chat" and "or use other apps" may be unclear for users who are not used to being able to use different apps to participate in the same chat. Not sure what a better wording could be, but we could remove one of them:

```suggestion
    <string name="ephemeral_messages_hint">Applies to all members of this chat if they use Delta Chat; they can still copy, save, and forward messages.</string>
```

src/main/res/values/strings.xml
```diff
@@ -529,10 +529,10 @@
...
+    <!-- Placeholder %1$s will be replaced by the name of the contact changing their address. Placeholders %2$s and %3$s will be replaced by old/new addresses. -->
     <string name="aeap_addr_changed">%1$s changed their address from %2$s to %3$s</string>
     <!-- the explanation is shown (1) as a modal dialog with the buttons "Cancel" and "Continue" as well as (2) as a device message -->
-    <string name="aeap_explanation">You changed your email address from %1$s to %2$s.\n\nIf you now send a message to a verified group, contacts there will automatically replace the old with your new address.\n\nIt\'s highly advised to set up your old email provider to forward all emails to your new email address. Otherwise you might miss messages of contacts who did not get your new address yet.</string>
+    <string name="aeap_explanation">You changed your address from %1$s to %2$s.\n\nIf you now send a message to a verified group, contacts there will automatically replace the old with your new address.\n\nIt\'s highly advised to set up your old address to forward all messages to your new address. Otherwise you might miss messages of contacts who did not get your new address yet.</string>
```
Hocuri, line 535: I don't think it's necessary to change this string now (& require retranslations), since we will soon remove it: https://github.com/deltachat/deltachat-android/pull/3698

src/main/res/values/strings.xml
```diff
@@ -646,7 +644,7 @@
     <string name="proxy_share_link">Share Link</string>
 
     <string name="login_info_oauth2_title">Continue with simplified setup?</string>
-    <string name="login_info_oauth2_text">The entered e-mail address supports a simplified setup (OAuth 2.0).\n\nIn the next step, please allow Delta Chat to act as your Chat over E-mail app.\n\nDelta Chat does not collect user data, everything stays on your device.</string>
+    <string name="login_info_oauth2_text">The entered address supports a simplified setup (OAuth 2.0).\n\nIn the next step, please allow Delta Chat to act on your behalf.\n\nDelta Chat does not collect user data, everything stays on your device.</string>
```
Hocuri, line 647:
```suggestion
    <string name="login_info_oauth2_text">The entered address supports a simplified setup (OAuth 2.0).\n\nIn the next step, please allow Delta Chat to send and receive messages.\n\nDelta Chat does not collect user data, everything stays on your device.</string>
```

src/main/res/values/strings.xml
```diff
@@ -755,7 +752,7 @@
     <string name="pref_backup">Backup</string>
     <string name="pref_backup_explain">Back Up Chats to External Storage</string>
     <string name="pref_backup_export_explain">A backup helps you to set up a new installation on this or on another device.\n\nThe backup will contain all messages, contacts and chats and your end-to-end Autocrypt setup. Keep the backup file in a safe place or delete it as soon as possible.</string>
-    <!-- the placeholder will be replaced by the name of the profile's email address -->
+    <!-- the placeholder will be replaced by the name of the profile name -->
```
Hocuri, line 755: Duplicate "name"

```suggestion
    <!-- the placeholder will be replaced by the profile name -->
```

or 

```suggestion
    <!-- the placeholder will be replaced by the name of the profile -->
```

src/main/res/values/strings.xml
```diff
@@ -844,9 +840,9 @@
...
     <string name="autodel_device_ask">Do you want to delete %1$d messages now and all newly fetched messages \"%2$s\" in the future?\n\n• This includes all media\n\n• Messages will be deleted whether they were seen or not\n\n• \"Saved messages\" will be skipped from local deletion</string>
     <!-- %1$d will be replaced by the number of messages, you can assume plural/lots here. %2$s will be replaced by a timespan option. -->
-    <string name="autodel_server_ask">Do you want to delete %1$d messages now and all newly fetched messages \"%2$s\" in the future?\n\n⚠️ This includes e-mails, media and \"Saved messages\" in all server folders\n\n⚠️ Do not use this function if you want to keep data on the server\n\n⚠️ Do not use this function if you are using other e-mail clients besides Delta Chat</string>
+    <string name="autodel_server_ask">Do you want to delete %1$d messages now and all newly fetched messages \"%2$s\" in the future?\n\n⚠️ This includes messages, media and \"Saved messages\" in all server folders\n\n⚠️ Do not use this function if you want to keep data on the server or if you use other apps beside Delta Chat</string>
```
Hocuri, line 843: > Do not use this function if you want to keep data on the server or if you use other apps beside Delta Chat

Here I think the meaning becomes unclear by replacing "e-mail clients" with "apps". Like, of course the user is using other apps beside Delta Chat: Signal, a meditation app, their banking app...

Also, I think it is good to explicitly note that the deletion also includes "e-mails", not only (Delta Chat) messages.

This is a string that's only shown for non-chatmail profiles, where users are usually more aware that they are chatting over e-mail. So, I think we could just keep the previous string.

Hocuri (commented):
Some of the strings are generally only used if the user uses a classic e-mail server (like `login_error_cannot_login`), for these I don't think it would be necessary to change the wording, but I'm also not against it. For strings that are also used with chatmail servers (like `ephemeral_messages_hint`), +1 from me.

I have some small suggestions, only for `autodel_server_*` I'm concerned that the new string may be unclear.

src/main/res/values/strings.xml
r10s: yeah, i think, "copy, save, and forward" is good enough to make clear the option is not 100% and needs trust

also that is shorter, so chances are higher ppl read :)

src/main/res/values/strings.xml
r10s: retranslations won't be needed, translators will get a hint it should be re-reviewed (so they are aware of the change)

the thing i tuned also unrelated things more down is to _talk less_ also internally about email

(and it was easier when searching for email occurances while doing the PR :)

but yeah, i can re-add the original string, this is not primary UI anyways

adbenitez (approved):
Besides the comments of Hocuri it looks good to me, thanks for taking care! 🎖️

src/main/res/values/strings.xml
r10s: fine with staying with "e-mails" instead of "messages".

questionable is the second part:  "other email clients _besides_ Delta Chat implies that Delta Chat is just an email client. again, it is also how we internally talk about that, translator see it etc.

maybe say `Do not use this function if you are using classic e-mail clients as well`

or `Do not use this function if you are using classic e-mail clients on this profile`

r10s:
thanks for review, i'll wait with merging as it seems the "router" is a "relay" now, cmp https://github.com/chatmail/pages/pull/7

PR #3708, opened 2025-04-01
hide 'show classic email' for chatmail

r10s:
moreover for non-chatmail the option is moved to the other questionable (shared account) options as "watch sent"

first item at advanced is now "Show Log" - same as on desktop and iOS

when this is merged, we should file corresponding issues/pr for desktop/ios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-* 'Show all email' is the default for chatmail, no need for an option
+* hide superfluous "Show Classic E-mails" advanced setting for chatmail
 
 ## v1.56.1
 2025-03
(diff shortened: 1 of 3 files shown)


Hocuri (approved):
"View Log" really got promoted, from the last to the first entry!

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Delta Chat Android Changelog
 
+## Unreleased
+
+* 'Show all email' is the default for chatmail, no need for an option
```
adbenitez, line 5: Maybe frase it as: `hide superfluous "Show Classic E-mails" advanced setting for chatmail`

adbenitez (approved):

CHANGELOG.md
r10s: i changed that line, thanks for the suggestion

PR #3709, opened 2025-04-01
add stock string needed when securejoin takes longer than expected

r10s:
this PR is to make the string needed  for https://github.com/chatmail/core/pull/6722 translatable

there were some previous discussions about the wording at https://github.com/chatmail/core/pull/6722/files#r2022229455

the string shows up only when the securejoin process takes more than 15 seconds, here, see the second message in the screenshots:

once this is merged, the string should be pushed to transifex and used on desktop/ios as well

ftr: i did not change the existing string as a new ID is anyway needed and to avoid confusion and to allow smooth transition in other clients

#skip-changelog as the visible change comes from core update

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -1003,7 +1003,9 @@
     <!-- placeholder will be replaced by the name of the inviter. -->
     <string name="secure_join_replies">%1$s replied, waiting for being added to the group…</string>
     <string name="secure_join_wait">Establishing guaranteed end-to-end encryption, please wait…</string>
+    <!-- deprecated -->
     <string name="secure_join_wait_timeout">Could not yet establish guaranteed end-to-end encryption, but you may already send a message.</string>
+    <string name="secure_join_takes_longer">This takes longer than expected, maybe devices are offline…\n\nHowever, the process continues in background, you can do something else 🕺</string>
     <string name="contact_verified">%1$s introduced.</string>
     <string name="contact_not_verified">Cannot establish guaranteed end-to-end encryption with %1$s.</string>
     <!-- Shown in contact profile. The placeholder will be replaced by the name of the contact that introduced the contact. -->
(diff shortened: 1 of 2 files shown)


src/main/res/values/strings.xml
```diff
@@ -1003,7 +1003,9 @@
     <!-- placeholder will be replaced by the name of the inviter. -->
     <string name="secure_join_replies">%1$s replied, waiting for being added to the group…</string>
     <string name="secure_join_wait">Establishing guaranteed end-to-end encryption, please wait…</string>
+    <!-- deprecated -->
     <string name="secure_join_wait_timeout">Could not yet establish guaranteed end-to-end encryption, but you may already send a message.</string>
+    <string name="secure_join_takes_longer">This takes longer than expected, maybe devices are offline…\n\nHowever, the process continues in background, you can do something else 🕺</string>
```
iequidoo, line 1008: After the discussion in the core PR i suggest "maybe the contact or you are offline". This can guide the user to check their network i think. "Devices" may be not clear enough (whose devices are meant?)

adbenitez: I think "This is taking longer than expected" is more correct than "this takes longer than expected"

adbenitez (approved):

src/main/res/values/strings.xml
r10s: thanks for the suggestions, i did a commit!

for the "than expected": i stroke "than expected", this is not an error, we're offline first. it _is_ expected that it can take hours, _therefore_ we continues in background

r10s:
after some internal discussions, i stroke the emoji as well, that easily offend ppl ...

ftr, the string is comparable critical as this will happen often - and at  a time where ppl watch the screen with suspense, and probably _really_ read the message :)

thanks for all the feedback, here and previously in core! (when this is merged, i will also update the fallback string in core)

PR #3710, opened 2025-04-01
open contact in scope when tapping info messages

r10s:
just released core 1.159.0 won't show email addresses in info messages,  instead, the contact profile for the ID returned by [`dc_msg_get_info_contact_id()`](https://c.delta.chat/classdc__msg__t.html#a9c789529400bfa0ceccbed4a922c3665) should be opened when the message is tapped

this is how that looks like:

https://github.com/user-attachments/assets/f79301b1-1570-4d46-8d3e-5491e7323db7

counterpart of https://github.com/deltachat/deltachat-ios/pull/2652 , once this is merged, we should add jsonrpc and file an issue for desktop


adbenitez:
When it is "you added Too" it is clear but when "Foo added Bar" this makes it hard to know who was the one adding/removing members

adbenitez (approved):

r10s:
this PR does not look at reasoning or so, but only opens the profile given by core.

if there are several contacts mentioned in a message, as "Alice added by Bob", the person added/removed is of more interest  and returned by core (the person who did the change is usually known longer/better, and is also quite atop of the member list, one tap away), see [`dc_msg_get_info_contact_id()`](https://c.delta.chat/classdc__msg__t.html#a9c789529400bfa0ceccbed4a922c3665) and core discussions for more details.

it feels quite intuitive in practise and is a huge improvement - before, you had to search like crazy in larger groups, comparing email addresses in mind :)

r10s:
in theory, you can now also add two links in the "Alice added by Bob" message, the information are all there.

but in practise, i think, it is fine as is. we can always iterate :)

PR #3729, opened 2025-04-10
remove address from contact list and group member list

### LINKED ISSUE #3701: remove ADDR from contact lists
the contact lists for "new chat" and the member list still show the email address, which we want to tune down for known reasons (multi transport, non-telling addresses etc)

instead, the user can open the profile easily, which has more appropriate information and which is already possible today

on iOS we removed them at these places since half a year without issues.

we left them, however, in contact has no green checkmark, this might be a good compromise for classic email usage for now

adbenitez:
close #3701

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+* display "Last Seen" information instead of email address in contact list for contacts with green-checkmark, like in other popular messengers
 * hide superfluous "Show Classic E-mails" advanced setting for chatmail
 * remove mostly non-telling transport addresses when referring to a contact;
   the contact's profile gives a much better overview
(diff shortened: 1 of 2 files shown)


r10s:
let's just remove the address for now, and not replace it by the timestamp, though i know you really like and want that feature :)

- esp. in the contact list for creating a new contact, there is not much of a usecase that needs that information - the list is already sorted by time anyways, and if the contact was seen a minute or an hour ago does not make much of a difference if you want to create a chat with the contact

- in the member list of a group, if may be more useful, however, if was never missing the past

- the time clutter the UI

- consistency with other Delta Chat clients
 
- consistency with the title, where we do not show the time as well (these elements are pretty much the same otherwise)

- consistency with other messengers as mentioned in the changlog is not so clear - telegram shows the time, but none of whatsapp/signal/simplex

- that "bots" do not show time, sure, a minor, but also not nice

 although some of us do not have a big love for exact timings , i am not totally against introducing it at some point, then also together with the title bar - however, i would like to introduce that then together on all platforms and not let it sneak in and have inconsistent apps and no agreement about that part

r10s:
> should I always hide the address?

no. it is fine as is. hiding the address unconditionally (also in title bar then)  is another iteration

PR #3730, opened 2025-04-10
remove address from primary settings UI

r10s:
showing the address here prominently results in bad UX as ppl try to pass the address around,
which will not result in e2ee and will often not work.

see https://github.com/deltachat/deltachat-ios/pull/2664 for more resoning and discussion, eg. why remove it unconditionally

cmp https://github.com/deltachat/deltachat-desktop/pull/4970

### DIFF
--- a/src/main/res/layout/profile_preference_view.xml
+++ b/src/main/res/layout/profile_preference_view.xml
@@ -13,14 +13,31 @@
                android:contentDescription="@string/pref_profile_info_headline"
                android:transitionName="avatar"/>
 
-    <androidx.appcompat.widget.AppCompatTextView
-            android:id="@+id/profile_name"
-            android:layout_width="match_parent"
-            android:layout_height="wrap_content"
-            android:layout_marginEnd="16dp"
-            android:gravity="start"
-            android:textAlignment="viewStart"
-            android:textSize="20sp"
-            tools:text="Voltarine DeClyre"/>
+    <LinearLayout android:orientation="vertical"
+                  android:layout_width="match_parent"
+                  android:layout_height="wrap_content"
+                  android:layout_marginRight="16dp"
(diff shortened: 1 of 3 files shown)


adbenitez:
maybe it is interesting to show the bio there (a bit of it in a single line) I think WhatsApp and other do the same IIRC

r10s:
> maybe it is interesting to show the bio there (a bit of it in a single line) I think WhatsApp and other do the same IIRC

yip, that's a nice idea!

also fitting nicely with the idea to make the bio more prominent in the profiles

adbenitez (approved):

adbenitez:
> i am no longer on this PR, we can merge as is - or you can push here. or we close this PR and do a new one (as using bio would probably reuse the removed controls)

I will push to this one

adbenitez:
for the record, if address was super long, as in the case of yggmail addresses, it was not truncated but wrapped around and settings moved down etc. now this also solved that problem

src/main/res/layout/profile_preference_view.xml
```diff
@@ -29,12 +29,14 @@
...
 
-        <TextView android:id="@+id/number"
+        <TextView android:id="@+id/profile_status"
                   android:layout_width="match_parent"
                   android:layout_height="wrap_content"
                   android:gravity="start"
                   android:textAlignment="viewStart"
-                  tools:text="+14151231234"/>
+                  android:singleLine="true"
+                  android:ellipsize="marquee"
```
r10s, line 38: wondering, why we're not using `android:ellipsize="end"` here.

however, grepping for "marquee", we're using that at lots of places - but i have never seen sth scrolling (which is good i'd say) - so, as it is done as elsewhere, probably fine :)

r10s:
very nice! thank's for taking over the PR!

PR #3732, opened 2025-04-13
prepare 1.58.0

adbenitez:


r10s:
nice! will do an iOS 1.58 soon as well, https://github.com/deltachat/deltachat-ios/pull/2667

PR #3734, opened 2025-04-14
use Contact data from RPC

adbenitez:
- [x] migrate to Rpc in contact list
- [x] migrate to Rpc in member list
- [ ] migrate to Rpc in chat list

### DIFF
--- a/src/main/java/com/b44t/messenger/rpc/Contact.java
+++ b/src/main/java/com/b44t/messenger/rpc/Contact.java
@@ -1,0 +1,83 @@
+package com.b44t.messenger.rpc;
+
+public class Contact {
+    public final String address;
+    public final String color;
+    public final String authName;
+    public final String status;
+    public final String displayName;
+    public final int id;
+    public final String name;
+    public final String profileImage;
+    public final String nameAndAddr;
+    public final boolean isBlocked;
+    public final boolean e2eeAvail;
+
+    /// True if the contact can be added to verified groups.
+    ///
+    /// If this is true
+    /// UI should display green checkmark after the contact name
+    /// in contact list items,
+    /// in chat member list items
+    /// and in profiles if no chat with the contact exist.
+    public final boolean isVerified;
+
+    /// True if the contact profile title should have a green checkmark.
+    ///
+    /// This indicates whether 1:1 chat has a green checkmark
+    /// or will have a green checkmark if created.
+    public final boolean isProfileVerified;
+
+    /// The ID of the contact that verified this contact.
+    ///
+    /// If this is present,
+    /// display a green checkmark and "Introduced by ..."
+    /// string followed by the verifier contact name and address
+    /// in the contact profile.
+    public final int verifierId;
+
+    /// the contact's last seen timestamp
+    public final float lastSeen;
+    public final boolean wasSeenRecently;
+
+    /// If the contact is a bot.
+    public final boolean isBot;
+
+    public Contact(
+                   String address,
+                   String color,
+                   String authName,
+                   String status,
+                   String displayName,
+                   int id,
+                   String name,
+                   String profileImage,
+                   String nameAndAddr,
+                   boolean isBlocked,
+                   boolean e2eeAvail,
+                   boolean isVerified,
+                   boolean isProfileVerified,
+                   int verifierId,
+                   float lastSeen,
+                   boolean wasSeenRecently,
+                   boolean isBot
+                   ) {
+        this.address = address;
+        this.color = color;
+        this.authName = authName;
+        this.status = status;
+        this.displayName = displayName;
+        this.id = id;
+        this.name = name;
+        this.profileImage = profileImage;
+        this.nameAndAddr = nameAndAddr;
+        this.isBlocked = isBlocked;
+        this.e2eeAvail = e2eeAvail;
+        this.isVerified = isVerified;
+        this.isProfileVerified = isProfileVerified;
+        this.verifierId = verifierId;
+        this.lastSeen = lastSeen;
+        this.wasSeenRecently = wasSeenRecently;
+        this.isBot = isBot;
+    }
+}
--- a/src/main/java/com/b44t/messenger/rpc/Rpc.java
+++ b/src/main/java/com/b44t/messenger/rpc/Rpc.java
@@ -158,6 +158,10 @@ public class Rpc {
         getResult("add_transport", accountId, param);
     }
 
+    public Contact getContact(int accountId, int contactId) throws RpcException {
+        return gson.fromJson(getResult("get_contact", accountId, contactId), Contact.class);
+    }
+
     private static class Request {
         private final String jsonrpc = "2.0";
         public final String method;
--- a/src/main/java/org/thoughtcrime/securesms/ContactSelectionListFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/ContactSelectionListFragment.java
@@ -336,7 +336,7 @@ public class ContactSelectionListFragment extends    Fragment
         }
         return;
       }
-      int    contactId = contact.getSpecialId();
+      int    contactId = contact.getContactId();
       if (!isMulti() || !selectedContacts.contains(contactId)) {
         if (contactId == DcContact.DC_CONTACT_ID_NEW_CLASSIC_CONTACT) {
           Intent intent = new Intent(getContext(), NewContactActivity.class);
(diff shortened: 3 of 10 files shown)


r10s:
atm, playing around with the PR, what hit immediately my eyes was the "email sign" for a just not completely join, that seems  wrong, and will also happen quite often.

not sure what to show instead (probably nothing), and not sure, if this is related to core or to android, i just want to bring that up.

<br><i>here "Hh" is not an email contact, but a chatmail-to-chatmail pgp-contact that is offline for a moment or has bad network. as one cannot write anyways, it is probably better to show nothing

adbenitez:
> here "Hh" is not an email contact, but a chatmail-to-chatmail pgp-contact that is offline for a moment or has bad network. as one cannot write anyways, it is probably better to show nothing

that looks wrong but in practice it is an email contact at that point and in non-chatmail accounts you can start writing to them unencrypted I guess, but I guess the fingerprint is already know at that point so maybe core should already consider it a pgp contact? but in that case we shouldn't never allow to write to it until the process complete even for non-chatmail

r10s:
so the rpc is not used for chat rendering, right? asking, as there the overhead might become visible - at a first glance, jsonrpc's `get_contact()` has 6 database calls instead of 1 for cffi's `dc_get_contact()` - and one often really only needs the name there.

also for profile and contact lists, probably too much is loaded, eg. to set `isProfileVerified` the chat record is loaded, but that  bool is never used anywhere. in lists, some of them may be needed, but for rendering a name in the bubble, various loads would be superfluous.

i see that there is some incentive for the move, and i do not want to block here, but maybe we should do this by adding new low-level jsonrpc calls? so that we can really make a cut and use the new call anywhere without bad feelings. @Hocuri, @link2xt , @hpk42  and others did quite some effort to make the calls quick, it would be a pity if we start wasting them.

for the concrete `get_contact` - maybe leave the call as is as it is needed by desktop like that, and copy+paste the code to sth. as `cm_get_contact()` (prefix for `chatmail`), that mimics mostly the old `dc_*()`  calls? i

src/main/java/org/thoughtcrime/securesms/connect/DcHelper.java
```diff
@@ -422,9 +422,10 @@ public static ThreadRecord getThreadRecord(Context context, DcLot summary, DcCha
     long date = summary.getTimestamp();
     int unreadCount = getContext(context).getFreshMsgCount(chatId);
 
+    boolean isEmailThread = chatId > DcChat.DC_CHAT_ID_LAST_SPECIAL && !chat.isDeviceTalk() && !chat.isEncrypted();
```
r10s, line 425: to not show the mail icon for read-only threads, we could check for `can_send()` here. let isEncrypted() return true seems indeed wrong in this case.

or we can rename the function to sth. more UI-based, as showAsEmailChat() (as we did for showPadlock())

r10s (approved):
apart from the comments above, code and logic lgtm.

however, merging needs to wait for a core release

src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
```diff
@@ -165,24 +163,31 @@ public void bind(@NonNull ThreadRecord thread,
     DcContact contact = recipient.getDcContact();
     avatar.setSeenRecently(contact!=null? contact.wasSeenRecently() : false);
 
+    int iconRight = thread.isProtected()? R.drawable.ic_verified : thread.isEmailThread()? R.drawable.ic_outline_email_24 : 0;
     fromView.setCompoundDrawablesWithIntrinsicBounds(
         thread.isMuted()? R.drawable.ic_volume_off_grey600_18dp : 0,
         0,
-        thread.isProtected()? R.drawable.ic_verified : 0,
+        iconRight,
```
r10s, line 170: in general, the letter is a little big too big - maybe better go for 18dp as the volume icon? no blocker, however

adbenitez:
> so the rpc is not used for chat rendering, right?

yeah, I migrated to the RPC for `getContact()` but for the chatlist I was feeling it would be too much work to migrate everything right now plus potential performance problems so I asked link2xt to add a field to the `DcChat` object for now since that avoids a huge refactoring


@r10s, FTR, the `Rpc.getContact()` call is much more slower than the native `dcContext.getContact()` call, around 2-3ms and most of the time is spent in the rust/core side waiting for `dcJsonrpcInstance.getNextResponse()`, according to data collected by @Hocuri, we noticed because some bad/slow performace was visible in search results for contacts because inefficient code was fetching the same contact a lot of times just to display the "N Contacts" header in search results, this was not noticeable with the native bindings because it is much faster, anyways I fixed `SearchListAdapter.getHeaderId()` and now scrolling in the contacts section of search results is not slow

might be ok to use the RPC also in the chatlist performance-wise, because in the end only a few items are loaded at a time and even if it is "much slower than the C bindings" we are talking about a few milliseconds here, but the RPC is "much slower" than the C bindings as it currently is, but once again the main reason for not moving right now was to avoid huge refactoring, I think makes sense to first fully migrate DcContact to RPC completely in some future PR and get rid of DcContact class before migrating the chatlist

r10s:
> he Rpc.getContact() call is much more slower than the native dcContext.getContact() call, around 2-3ms and most of the time is spent in the rust/core side waiting for dcJsonrpcInstance.getNextResponse()

esp. to get a grip on that, it seems needed to have jsonrpc calls where the rust part does the same as the cffi.

apart from that, tbh, i was expecting that `getNextResponse()` needs some time and things are overall slower :) but if we have these low-level calls, would be interesting if calling _directly_ over `dc_jsonrpc_blocking_call()` will have comparable speed to cffi.

> but the RPC is "much slower" than the C bindings as it currently is, but once again the main reason for not moving right now was to avoid huge refactoring, I think makes sense to first fully migrate DcContact to RPC completely in some future PR and get rid of DcContact class before migrating the chatlist

yip makes sense to not do that everywhere directly, as mentioned above, things are also wasted in superfluous database calls, eg. _always and unconditionally_ the chat-object is loaded from the database, but not needed (only to set isProfileVerified which is not used at all on android)

i think, we definetly need more measurement and experience here, makes sense to move slow and only for comparable uncritical stuff (even 3ms slower, this sums up in a chat when some bubbles are shown, each bubble has several calls, so with 10 visible bubbles, 3+ calls each, this is already 100ms slower - and not yet about scrolling)

adbenitez:
> would be interesting if calling _directly_ over `dc_jsonrpc_blocking_call()` will have comparable speed to cffi.

I think hocuri tried that and there is not much difference, the main bottleneck is encoding/decoding JSON, see: https://github.com/deltachat/deltachat-android/pull/3739

src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
adbenitez: indeed it was a bit too big, set it to `18sp` as the verified icon and now it is smaller (it looks smaller than the verified icon because the icon has some empty space padding)

adbenitez:
> atm, playing around with the PR, what hit immediately my eyes was the "email sign" for a just not completely join, that seems wrong, and will also happen quite often.
> 
> not sure what to show instead. probably nothing, the email-symbol is probably not meant to show up for chatmail. also, not sure, if this is related to core or to android.
> 
> 
>
> here "Hh" is not an email contact, but a chatmail-to-chatmail pgp-contact that is offline for a moment or has bad network. as one cannot write anyways, it is probably better to show nothing


mmmh I can't reproduce this:

r10s:
i forced the "waiting chat" to check https://github.com/chatmail/core/pull/6810 -- and after that in upgrading the core to the branch needed for this PR - maybe it is not happening otherwise

adbenitez:
> i forced the "waiting chat" to check [chatmail/core#6810](https://github.com/chatmail/core/pull/6810) -- and after that in upgrading the core to the branch needed for this PR - maybe it is not happening otherwise

that sounds tricky, also how did it work at all for you with an existing account? to test this PR you need to setup account from scratch because migrations were missing :thinking: the normal way to force this state would be:

- create profile B
- copy invite link from profile B
- delete profile B
- in a profile A that you created on the pgp-branch core, scan the QR from B
- secure join will timeout because account B doesn't exist anymore and will not reply

r10s:
meanwhile, core shows the email-icon as the avatar, so all the additional icons are not needed and should be removed from the PR again.

the core-submodule is checked in accidentally and should be removed again

Hocuri:
The core-submodule is not checked in accidentally, but in order to get a working preview APK. But yes, it needs to be removed before merging.

adbenitez:
closed in favor of  #3793

PR #3739, opened 2025-04-17
 Experiment: Use blocking JsonRPC call 

Hocuri:
I wanted to try out how fast a blocking JsonRPC API would be. The performance advantage isn't big enough to be immediately visible, at least.

<details><summary>I also added some more logging to core</summary>
<p>

```diff
diff --git a/deltachat-jsonrpc/src/api.rs b/deltachat-jsonrpc/src/api.rs
index e60be8a23..5b83c65ae 100644
--- a/deltachat-jsonrpc/src/api.rs
+++ b/deltachat-jsonrpc/src/api.rs
@@ -1418,11 +1418,16 @@ async fn get_contact(&self, account_id: u32, contact_id: u32) -> Result<ContactO
         let ctx = self.get_context(account_id).await?;
         let contact_id = ContactId::new(contact_id);
 
-        ContactObject::try_from_dc_contact(
+        let start = std::time::Instant::now();
+        let get_by_id = deltachat::contact::Contact::get_by_id(&ctx, contact_id).await?;
+        info!(&ctx, "HOCURI get_by_id took {:?}", start.elapsed());
+        let res = ContactObject::try_from_dc_contact(&ctx, get_by_id).await;
+        info!(
             &ctx,
-            deltachat::contact::Contact::get_by_id(&ctx, contact_id).await?,
-        )
-        .await
+            "HOCURI get_contact in total took {:?}",
+            start.elapsed()
+        );
+        res
     }
 
     /// Add a single contact as a result of an explicit user action.
```

</p>
</details> 

In a random, somewhat representative sample,

- Calling `toJson()` in Java takes 236µs
- Loading the contact from the database takes 274.479µs
- Converting to Json with `ContactObject::try_from_dc_contact` in Rust then takes 379µs, so that the total `get_contact()` function in api.rs takes  653.125µs
- The total `blockingCall()` function takes 1050µs, the difference to the 653.125µs is probably encoding & decoding the request as Json
- Then there are two `fromJson()` calls in Java, which together take 615µs.

So, a _lot_ of time spent encoding & decoding Json.

<details><summary>Raw Output which only makes sense if you read the source</summary>
<p>

```
[415]toJson, now 236micros
[accId=3] deltachat-jsonrpc/src/api.rs:1423: HOCURI get_by_id took 274.479µs
[accId=3] deltachat-jsonrpc/src/api.rs:1425: HOCURI get_contact in total took 653.125µs
[415]blockingCall, now 1286micros
[415]fromJson#1, now 1632micros
[415]fromJson#2, now 1901micros
```

</p>
</details>

### DIFF
--- a/jni/dc_wrapper.c
+++ b/jni/dc_wrapper.c
@@ -2045,3 +2045,14 @@ JNIEXPORT jstring Java_com_b44t_messenger_DcJsonrpcInstance_getNextResponse(JNIE
     dc_str_unref(temp);
     return ret;
 }
+
+
+JNIEXPORT jstring Java_com_b44t_messenger_DcJsonrpcInstance_blockingCall(JNIEnv *env, jobject obj, jstring request)
+{
+CHAR_REF(request);
+char* temp = dc_jsonrpc_blocking_call(get_dc_jsonrpc_instance(env, obj), requestPtr);
+CHAR_UNREF(request);
+jstring ret = JSTRING_NEW(temp);
+dc_str_unref(temp);
+return ret;
+}
--- a/src/main/java/com/b44t/messenger/DcJsonrpcInstance.java
+++ b/src/main/java/com/b44t/messenger/DcJsonrpcInstance.java
@@ -14,6 +14,7 @@ public class DcJsonrpcInstance {
 
     public native void request(String request);
     public native String getNextResponse();
+    public native String blockingCall(String request);
 
     // working with raw c-data
     private long        jsonrpcInstanceCPtr;    // CAVE: the name is referenced in the JNI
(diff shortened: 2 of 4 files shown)


r10s:
> So, a lot of time spent encoding & decoding Json.

did not expect that.

still maybe jsonrpc is faster then expected. too get a fair comparison, we need to compare that to a bunch of cffi calls, so sth. as ...

```
c = dc_get_contact()
dc_contact_get_addr(c)
dc_contact_get_color(c)
dc_contact_get_name(c)
dc_contact_get_authname(c)
dc_contact_get_status(c)
dc_contact_get_displayname(c)
dc_contact_get_id(c)
dc_contact_get_name(c)
dc_contact_get_name_n_addr(c)
dc_contact_get_is_blocked(c)
dc_contact_get_last_seen(c)
dc_contact_was_seen_recently(c)
dc_contact_is_bot(c) // but _not_ get_profle_image(), is_profile_verified etc as that results in additional db calls that are not the baseline of normal loading speed
```

... would define the baseline (at least a handful properties are usually needed, and i expect, say, 8 cffi calls, even without additional database calls, to cost some time)

r10s:
> Converting to Json with ContactObject::try_from_dc_contact

this is not only converting, but includes ~5 additional database calls

PR #3744, opened 2025-04-22
fix some linter warnings

adbenitez:
#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationItem.java
@@ -269,11 +269,10 @@ public class ConversationItem extends BaseConversationItem
         R.attr.conversation_item_incoming_bubble_color,
         R.attr.conversation_item_outgoing_bubble_color,
     };
-    final TypedArray attrs      = context.obtainStyledAttributes(attributes);
-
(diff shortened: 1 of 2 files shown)


src/main/java/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -269,11 +269,10 @@ private void initializeAttributes() {
         R.attr.conversation_item_incoming_bubble_color,
         R.attr.conversation_item_outgoing_bubble_color,
     };
-    final TypedArray attrs      = context.obtainStyledAttributes(attributes);
-
-    incomingBubbleColor = attrs.getColor(0, Color.WHITE);
-    outgoingBubbleColor = attrs.getColor(1, Color.WHITE);
-    attrs.recycle();
```
r10s, line 276: where is `recycle()` gone?

that seems at least more than what the linter is telling me :)

adbenitez: The try block handles automatic closing of resources without needing to remember such manual calling to recycle with potential forgetting to do so, that is what the warning was about, calling recycle inside the try would cause a second attempt to recycle an already recycled resource results in an exception

r10s (approved):

PR #3750, opened 2025-04-25
make 'microphone' hint fitting for desktop as well 

r10s:
i think, it is even an improvement for android, as the hint is now also anchored by 'System Settings' which is more in focus by casual users than 'App Settings' (long tap icon). also, some android derivates may have different order,  which is captured better by a broader hint.

the hint is only shown when the access was initially denied.

i was first considering to have different hints, but that's a lot of work, now and to track in the future. i think, it is good enough that way

targets https://github.com/deltachat/deltachat-desktop/pull/4986#issuecomment-2824531972


adbenitez (approved):
not sure this is an improvement, find it a bit confusing ("this or that", then do "this or that") 

maybe it is possible to directly open the settings screen to enable microphone, we should consider to do this in the future

r10s:
yes, however, without saying "this or that", we woulld need to track quite some os and maintain the click paths

offering a "settings" button: iirc, i think we even do that

PR #3753, opened 2025-04-28
feat: Put "DeltaChatUI." before the tag of log messages

Hocuri:
Right now, I find it really hard to grep for Delta Chat in `adb logcat`.


Hocuri:
Can use `adb logcat --pid=`adb shell pidof -s com.b44t.messenger.beta` instead

PR #3756, opened 2025-05-03
fix crash in JSONRPC

r10s:
the fix could have been avoid by checking for null or not being overly specific with the exception.

as this is a very sensible area,
where any failure is dramatic,
we do both.

as said, this error pops up on monikas device,
unclear if related to an app update (she is on 1.58.2) or by some message.

~~did not test yet, still trying to get this patch on the phone~~ EDIT: tested, it fixes the issue at hand

### DIFF
--- a/src/main/java/com/b44t/messenger/rpc/Rpc.java
+++ b/src/main/java/com/b44t/messenger/rpc/Rpc.java
@@ -30,9 +30,13 @@ public class Rpc {
 
     private void processResponse() throws JsonSyntaxException {
         String jsonResponse = dcJsonrpcInstance.getNextResponse();
-        Response response = gson.fromJson(jsonResponse, Response.class);
 
-        if (response == null || response.id == 0) { // Got JSON-RPC notification/event, ignore
+        Response response = gson.fromJson(jsonResponse, Response.class);
+        if (response == null) {
+            Log.e(TAG, "Error parsing JSON: " + jsonResponse);
+            return;
+        } else if (response.id == 0) {
+            // Got JSON-RPC notification/event, ignore
             return;
         }
 
--- a/src/main/java/com/b44t/messenger/rpc/Rpc.java
+++ b/src/main/java/com/b44t/messenger/rpc/Rpc.java
@@ -32,7 +32,7 @@ public class Rpc {
         String jsonResponse = dcJsonrpcInstance.getNextResponse();
         Response response = gson.fromJson(jsonResponse, Response.class);
 
-        if (response.id == 0) { // Got JSON-RPC notification/event, ignore
+        if (response == null || response.id == 0) { // Got JSON-RPC notification/event, ignore
             return;
         }
 
@@ -63,7 +63,7 @@ public class Rpc {
             while (true) {
                 try {
                     processResponse();
-                } catch (JsonSyntaxException e) {
+                } catch (Exception e) {
                     e.printStackTrace();
                 }
             }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * fix: webxdc.selfName uses the name otherwise displayed
+* fix potential crash on startup
 
 ## v1.58.2
 2025-04


hpk42 (approved):
lgtm.

src/main/java/com/b44t/messenger/rpc/Rpc.java
```diff
@@ -32,7 +32,7 @@ private void processResponse() throws JsonSyntaxException {
         String jsonResponse = dcJsonrpcInstance.getNextResponse();
         Response response = gson.fromJson(jsonResponse, Response.class);
 
-        if (response.id == 0) { // Got JSON-RPC notification/event, ignore
+        if (response == null || response.id == 0) { // Got JSON-RPC notification/event, ignore
```
Hocuri, line 35: If we merge this, then we definitely need to log an error if `response == null`. Otherwise, DC may just hang on startup forever without any helpful message in the logs.

Hocuri:
While developing https://github.com/chatmail/core/pull/6818, I had a bug with the same symptom: DC crashed on startup with exactly this error message. The problem turned out to be:
- Android tries to get an account manager (via CFFI), which in turn tries to open the context. Opening the context fails because the migration fails (because of a bug, which I fixed since).
- The error message can't be printed to the logs, since there is no event loop yet; this is why there is no helpful error message in the logs.
- get_dc_accounts() returns null
- dc_jsonrpc_init() (i.e. getJsonrpcInstanceCPtr()) returns null
- get_dc_jsonrpc_instance() returns null
- dc_jsonrpc_next_response() (Rust) / getNextResponse() (Java) returns an empty string

If you can reproduce the problem and can install a re-compiled DC Android, I recommend replacing all calls to `eprintln!(()` in deltachat-cffi/lib.rs with `panic!()`. This way, the root error cause will show in the logs. You may be able to copy out DeltaChat's folder via adb.

About this PR: If the problem you had is similar to the problem I had, then this PR would not have made anything better. Instead of crashing on startup, DC would just hang forever on startup. I'm afraid there is not a lot the UI can do when `get_dc_accounts()` errors. 

What core could do is, make `get_dc_accounts()` return successfully if one of the profiles failed to open, so that the other profiles can be used. Not sure how much effort it would be to make this work nicely.

Hocuri:
I need to think a bit about general ways to improve this, maybe I'll come up with something.

r10s:
> If the problem you had is similar to the problem I had, then this PR would not have made anything better.

then i assume, you have another issue. the PR actually solves the issue, everything is working again as expected.

here was also nothing super-fancy on that account, so i assume, it can happen also in other cases

Hocuri:
Ok, that's great. As commented on the code, we need to log a warning / error when we ignore a null `response`, and should also log what `jsonResponse` for easier debugging (usually, it's probably an empty string, but maybe sometimes it's something else that fails to parse as JSON).

adbenitez:
> > If the problem you had is similar to the problem I had, then this PR would not have made anything better.
> 
> then i assume, you have another issue. the PR actually solves the issue at hand, things are working again as expected.
> 
> here was also nothing super-fancy on that account, so i assume, it can happen also in other cases

this is a bit bad, this didn't "solved the problem", in the end we don't even know now what the problem was, instead of just ignoring it and losing the opportunity to debug it and fix the actual problem it would be better to know why core returned null in this case, this shouldn't happen to start with if I got it right, and as @Hocuri said in other similar bugs where core would return null we then would just ignore the problem and potentially it can be a problem you can't just ignore :(

r10s:
_"better crash, the user will report and we will fix"_ is nice in theory, but unfortunately also comes with _"we do not care if the user's device is unusable for weeks"_ ;) 

i agree that in many flows it is better to crash, eg. if there is a concrete user action the user can not do as a workaround. but for such basic things, it just makes everything unusable.

we had that discussion often, i do not think, it makes much sense to warm that up again.

additional logging as @Hocuri suggested makes sense, though

adbenitez:
(Meant it in this particular case that you had access to the device)

Hocuri:
@r10s in any case, could you try to find out the root cause of this particular problem? You can try replacing all calls to `eprintln!(()` in deltachat-cffi/lib.rs with `panic!()`. This way, the root error cause should show in the logs.

src/main/java/com/b44t/messenger/rpc/Rpc.java
r10s: added a commit

r10s:
@adbenitez @Hocuri i added a commit to log the error case. if there are more things, please add to this PR, otherwise please merge; i am not really on android atm

PR #3774, opened 2025-05-23
point out that app drafts need to be sent

r10s:
we got some feedback,
that ppl were wondering if others can use an app
once drafted and opend.

the old title "Tap to open" might even underline that impression.

this PR changes the title to "Tap 'Send' to share" (we need to be super-short here, "Tap 'Send' to let others use the app" is already too much :)

the sending out of apps is the much more important step than to open, play a game and then wondering ...
ppl will figure out the latter themselves,
wondering if one can configure eg. a poll before sending - and if not, then it's that.


adbenitez (approved):
Mmmh tap what "send"??? It is a bit confusing but I agree with the point that it is might not be clear it needs to be sent

r10s:
> Mmmh tap what "send"???

yeah, i would be happy if we find a better wording, that is still short enough

hpk42:
i think it's fine to suggest the action, and not explain it.

PR #3776, opened 2025-05-23
show action button to see the list of apps sent in chat

adbenitez:
image

image


r10s (approved):
nice!

in a subsequent PR, we should alter the order of the tabs, so "Apps / Gallery / Files / Audio" - more important and more often needed things comes before by that.

(ftr, on iOS, we do not statically open the first one (so "Apps" then), but check if there is sth. in, and otherwise open the next tab. so  a chat with only images, opens at "Gallery", a chat with only pdf at "Files". i am not directly suggesting that, it is good also as now, let's first see how it works out on iOS in practise first)

PR #3780, opened 2025-05-26
add Web API's TTS support to webxdc

adbenitez:


r10s (approved):
great! source code looks easy enough then, did not test

PR #3783, opened 2025-06-04
Show channels in a proper "Channel" chat

Hocuri:
Depends on https://github.com/chatmail/core/pull/6901 but can already be reviewed, part of https://github.com/chatmail/core/issues/6884

- [x] Rename "broadcast list" to "channel" or "broadcast channel" both in UI and code (depending on whether it's important to be concise in the particular place). There are new stock strings in which `broadcast_list` is replaced with `channel` 
- [x] Adapt to the new chat types `InBroadcastChannel` and `OutBroadcastChannel` for incoming / outgoing channels, where the former is similar to a `Mailinglist` and the latter is similar to a `Broadcast` (which was removed)
- [x] Adapt to the new `create_broadcast_channel()` API
- [x] Wait for a core release with https://github.com/chatmail/core/pull/6901, and for https://github.com/deltachat/deltachat-android/pull/3793 to be merged


src/main/java/com/b44t/messenger/DcChat.java
```diff
@@ -54,15 +55,19 @@ public DcChat(int accountId, long chatCPtr) {
 
     public boolean isMultiUser() {
       int type = getType();
-      return type == DC_CHAT_TYPE_GROUP || type == DC_CHAT_TYPE_MAILINGLIST || type == DC_CHAT_TYPE_BROADCAST;
+      return type == DC_CHAT_TYPE_GROUP || type == DC_CHAT_TYPE_MAILINGLIST
```
adbenitez: at this point wondering if it isn't better to just check for `type != DC_CHAT_TYPE_SINGLE` instead of checking if it is any other type but single

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -441,12 +441,13 @@ public boolean onPrepareOptionsMenu(Menu menu) {
       menu.findItem(R.id.menu_show_map).setVisible(false);
     }
 
-    if (!dcChat.canSend() || dcChat.isBroadcast() || dcChat.isMailingList()) {
+    if (!dcChat.canSend() || dcChat.isMailingList()) {
```
adbenitez: `dcChat.isMailingList()` is `true` for incoming broadcasts (subscriber side) as of today? by looking at other parts where you introduced a IN_BROADCAST_CHANNEL chat type doesn't look like it to me, so here you would need to check if `isInBroadcastChannel()` and also hide the ephemeral timer option for them since they can't change the timer as a subscriber, right?

Hocuri: `canSend()` is `false` for incoming broadcasts, so, it's not necessary to check.

Hocuri:
Also here, I renamed `BroadcastChannel` to `Broadcast` [to match core](https://github.com/chatmail/core/pull/6901#issuecomment-3020056588)

r10s (approved):
code lgtm, did not test

adbenitez:
resolved merge conflict after profile rework that touched several areas in profile and moved code around to new files/classes

adbenitez:
@Hocuri so this is also ready for merging?

Hocuri:
I went through the UI testing list at https://github.com/chatmail/core/issues/6884, and there are some things wrong now:
- In the profile view, the subtitle says "Contact" rather than "Channel"
- When tapping on the avatar of an InBroadcast, it's possible to edit it (but ofc it doesn't work)
- It wasn't actually possible to leave a channel (anymore), though this is probably a core issue

But these issues are minor, I'll fix them in a follow-up PR.

PR #3785, opened 2025-06-10
improve the videocalls: open in internal webview

adbenitez:


adbenitez:
one problem is that jitsi meet page takes too long to close when hanging up, they keep you in the page for some seconds then forward you to another page, I am detecting that forwarding and then closing the webview

also if you close without hanging up it seems the ghost peers stay in the room for some while, maybe some more custom way to control and send the "disconnect me" right from DC when closing the activity would be needed but all that is custom for jitsi meet

src/main/java/org/thoughtcrime/securesms/videochat/VideochatUtil.java
```diff
@@ -23,7 +29,18 @@ public void invite(Activity activity, int chatId) {
...
+                if (unset) {
+                  dcContext.setConfig(DcHelper.CONFIG_WEBRTC_INSTANCE, DcHelper.DEFAULT_VIDEOCHAT_URL);
+                }
+
                 int msgId = dcContext.sendVideochatInvitation(dcChat.getId());
+
+                if (unset) {
+                  dcContext.setConfig(DcHelper.CONFIG_WEBRTC_INSTANCE, null);
+                }
+
```
adbenitez: this is an ugly hack, ideally the `sendVideochatInvitation()` API would accept an URL parameter, that would allow also passing different URLs depending on the UI wanting to start a video call or audio call with camera disabled by default

adbenitez:
the dialog that appears when inviting someone says that you need to have an app installed, we probably need a new string

r10s:
> the dialog that appears when inviting someone says that you need to have an app installed, we probably need a new string

for now, we can just _not_ use the string `videochat_invite_user_hint`

src/main/java/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
```diff
@@ -211,6 +205,22 @@ public void onCreate(Bundle paramBundle) {
       });
     }
 
+    Preference callsEnabled = this.findPreference("pref_calls_enabled");
+    if (callsEnabled != null) {
+      callsEnabled.setOnPreferenceChangeListener((preference, newValue) -> {
+        if ((Boolean)newValue) {
+          new AlertDialog.Builder(requireActivity())
+            .setTitle("Thanks for trying out \"Video Calls\"!")
+            .setMessage("• You can now call contacts\n\n"
```
WofWca: This seems to mean that you could call anyone except contacts previously.

WofWca (commented):
Regarding the "the other participant's video freezing" issue:
I think I have found the suspect: it might be the thing where Android pauses all media when you're on a call (or receive an "incoming call" notification?)

If you go to "saved messages" and start a call there, and then open your browser and try to play another video, you'll notice that it behaves in the same way. If you try to start playing the video, it will play for a moment, and then pause again. This is also the case for the incoming stream in our "calls" app.

However, I honestly have no idea why you can still play the video of your own camera inside the app, even when it's unmuted.

I have been debugging this with Eruda. Here is my little debug script, if you're interested:

<details><summary>Script</summary>

```javascript
// Hide the "Ringing..." div that covers the rest of the app.
document.querySelectorAll('div')[4].style.display =''

vids = document.querySelectorAll('video')

// Having this makes the new `localVideo` play.
if (1) {
	vids[0].srcObject = null
	vids[1].srcObject = null
	vids[0].remove()
	vids[1].remove()
}

// This makes both videos play.
const enableAudio = 1
const setSrcObjectOnEventIndex = 1

let localVideo
if (1) {
	localVideo = document.createElement('video');
	localVideo.autoplay = true;
	localVideo.muted = true;
	localVideo.controls = true;
	localVideo.style.width = '320px';
	localVideo.style.marginRight = '10px';

	document.body.appendChild(localVideo);
} else {
	localVideo = vids[1]	
}

let remoteVideo
if (1) {
	remoteVideo = document.createElement('video');
	remoteVideo.autoplay = true;
	remoteVideo.controls = true;
	remoteVideo.style.width = '320px';

	document.body.appendChild(remoteVideo);
} else {
	remoteVideo = vids[0]
	remoteVideo.controls = true
}

let localStream;
const pc1 = new RTCPeerConnection();
const pc2 = new RTCPeerConnection();

// When pc2 receives a track, display it in remoteVideo
let onTrackCount = 0
pc2.ontrack = (event) => {
	onTrackCount++
  if (onTrackCount !== setSrcObjectOnEventIndex) {
	  console.log('onTrackCount !== setSrcObjectOnEventIndex')
	  return
  }
  remoteVideo.srcObject = event.streams[0];
};

// Get local media and start the connection
navigator.mediaDevices.getUserMedia({ video: true, audio: enableAudio }).then(stream => {
  localStream = stream;
  localVideo.srcObject = localStream;

	if (0) {
		return
	}

  // Add local tracks to pc1
  localStream.getTracks().forEach(track => {
    pc1.addTrack(track, localStream);
  });

  // ICE candidates
  pc1.onicecandidate = e => {
    if (e.candidate) pc2.addIceCandidate(e.candidate);
  };
  pc2.onicecandidate = e => {
    if (e.candidate) pc1.addIceCandidate(e.candidate);
  };

  // Start offer/answer exchange
  pc1.createOffer().then(offer => {
    return pc1.setLocalDescription(offer);
  }).then(() => {
    return pc2.setRemoteDescription(pc1.localDescription);
  }).then(() => {
    return pc2.createAnswer();
  }).then(answer => {
    return pc2.setLocalDescription(answer);
  }).then(() => {
    return pc1.setRemoteDescription(pc2.localDescription);
  });
});
```

</details>

WofWca:
> the thing where Android (or is it Chromium) pauses all media when you're on a call

Or this might be this thing called "audio focus": https://developer.android.com/media/optimize/audio-focus

adbenitez:
in fact I tried disabling the use of the calls integration and now the video works, a bit weird because this used to work in the past, but I think for now I will not use the Android Telecom API at all, it is almost useless, only useful for calling from cars or Bluetooth devices

src/main/java/org/thoughtcrime/securesms/videochat/VideochatActivity.java
```diff
@@ -0,0 +1,126 @@
...
+    return true;
+  }
+
+  @Override
+  public void handleEvent(@NonNull DcEvent event) {
+    switch (event.getId()) {
+    case DcContext.DC_EVENT_OUTGOING_CALL_ACCEPTED:
+      if (event.getData1Int() == callId) {
+        String hash = "#answer=" + event.getData2Str();
+        webView.evaluateJavascript("window.location.hash = `"+hash+"`", null);
```
link2xt: This gets the answer from the network and injects it directly into javascript, may result in executing anything in the webview.

link2xt:
I looked into the payload sent in a place_call_info, and it is an SDP offer, but unnecessarily URL-encoded. Should be a plain SDP instead, there is no need to send all the `%` over the wire. Otherwise we are stuck with URL-encoded payload for compatibility.

Fixing this needs a fix https://github.com/chatmail/core/pull/7191 in the core first, because it turned out newlines were not tested and break the header.

src/main/java/org/thoughtcrime/securesms/videochat/VideochatActivity.java
link2xt: SDP sent over the wire should be not URL-encoded. Instead, base64 or whatever encoding should happen here before injecting into JS.

WofWca: FYI we do the same (base64) in desktop, for such untrusted content.

adbenitez:
> I found two issues with this:
> 
>     1. (maybe not an issue) you can receive incoming calls, despite the "Video Calls" experimental setting being turned off.
 
yep this is intentional

>     2. After installing the app for the first time, when you receive a call, it will not ask for "Microphone" permission, which results in an error in the calls-webapp console, something like "could not start audio stream". It only asks for microphone permission when you are the initiator of the call.

yes, I had it in my offline to-do list already, will take care of it soon

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -442,6 +442,8 @@ public boolean onPrepareOptionsMenu(Menu menu) {
       menu.findItem(R.id.menu_show_map).setVisible(false);
     }
 
+    menu.findItem(R.id.menu_start_call).setVisible(Prefs.isCallsEnabled(this) && dcChat.isEncrypted() && !isMultiUser());
```
WofWca: Should probably also check if it's "Saved messages" or a "Device chat", and `canSend`. See related desktop MR: https://github.com/deltachat/deltachat-desktop/pull/5484.

And the Core code:

https://github.com/chatmail/core/blob/0684810d38d0427a9635b40befc04f951b530106/src/chat.rs#L2918-L2938

https://github.com/chatmail/core/blob/738dc5ce197f589131479801db2fbd0fb0964599/src/calls.rs#L147

adbenitez: solved, thanks for pointing out!

src/main/res/values/strings.xml
```diff
@@ -368,6 +379,17 @@
...
+    <string name="start_call">Call</string>
+    <string name="calls">Calls</string>
+    <string name="answer_call">Answer</string>
+    <string name="end_call">Decline</string>
+    <string name="outgoing_call">Outgoing call</string>
+    <string name="incoming_call">Incoming call</string>
+    <string name="declined_call">Declined call</string>
+    <string name="canceled_call">Canceled call</string>
+    <string name="missed_call">Missed call</string>
+    <string name="call_date_and_duration">%1$s, %2$s</string>
```
r10s, line 392: this requires some translator not, we'll get lots of questions otherwise

```suggestion
    <!-- the first placeholder will be replaced by a date, the second placeholder by duration, example: "Thu, 08:12 pm, 2 minutes" -->
    <string name="call_date_and_duration">%1$s, %2$s</string>
```

src/main/java/org/thoughtcrime/securesms/util/DateUtils.java
```diff
@@ -119,6 +119,19 @@ public static String getFormatedDuration(long millis) {
...
+    if (seconds < 60) {
+      return c.getResources().getQuantityString(R.plurals.n_seconds_ext, seconds, seconds);
+    }
+
+    int mins = seconds / 60;
+    return c.getResources().getQuantityString(R.plurals.n_minutes_ext, mins, mins);
+  }
+
+  public static String getFormattedCallTime(final Context c, final long timestamp) {
+    return getFormattedDateTime(timestamp, DateFormat.is24HourFormat(c)? "HH:mm" : "hh:mm a");
```
r10s: this means, from three days old calls, we do not show the date but only the time.

i see the point, not showing sth. as "2 minutes, 1 minute", however, this should not remove displaying the full date for older messages.

(even though, we have headlines as "Today" or "Yesterday", to avoid confusion, we show the explicit date in each bubble - we had that differently in the past and changed it to be explicit on purpose. i think, we would do that even if the headlines are sticky, but currently they are not, which makes the date even more needed)

src/main/java/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
r10s: hm, not sure. but anyways, i created a more detailed suggestion below

src/main/java/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
```diff
@@ -211,6 +205,22 @@ public void onCreate(Bundle paramBundle) {
...
     }
 
+    Preference callsEnabled = this.findPreference("pref_calls_enabled");
+    if (callsEnabled != null) {
+      callsEnabled.setOnPreferenceChangeListener((preference, newValue) -> {
+        if ((Boolean)newValue) {
+          new AlertDialog.Builder(requireActivity())
+            .setTitle("Thanks for trying out \"Calls\"!")
+            .setMessage("• You can now call contacts\n\n"
+              + "• If you want to quit the experimental feature, you can disable it at \"Settings / Advanced\"")
```
r10s:
```suggestion
            .setTitle("Thanks for helping to debug \"Calls\"!")
            .setMessage("• You can now debug calls using the "phone" icon in one-to-one-chats\n\n"
              + "• The experiment is about making decentralised calls work and reliable at all, not about options or UI. We're happy about focused feedback at support.delta.chat\n\n")
```

gist is to lower expectations, and also, to not annoy others too much.

depending on how good or bad things are working, we may add better feedback channel. to move forward, for now, the plan is to make a post in the forum and to explain there how to get logs etc.

before getting broader, we should try out things more internally anyways

src/main/res/xml/preferences_advanced.xml
```diff
@@ -16,14 +16,15 @@
...
-            android:summary="@string/none"/>
-
         <Preference android:key="pref_webxdc_store_url"
             android:title="@string/webxdc_store_url"
             android:summary="@string/none"/>
 
+        <org.thoughtcrime.securesms.components.SwitchPreferenceCompat
+            android:defaultValue="false"
+            android:key="pref_calls_enabled"
+            android:title="@string/calls"/>
```
r10s: see above for reasoning and discussion

```suggestion
            android:title="Debug Calls"/>
```

src/main/res/values/strings.xml
```diff
@@ -368,6 +379,17 @@
     </plurals>
     <string name="file_saved_to">File saved to \"%1$s\".</string>
 
+    <string name="start_call">Call</string>
+    <string name="calls">Calls</string>
```
r10s, line 383: this string is not needed currently, option is called "Debug Calls" and this is only for the ~next release, so not worth translating. then we can see

```suggestion
```

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1458,7 +1467,7 @@ public void handleReplyMessage(DcMsg msg) {
     Recipient author = new Recipient(this, dcContext.getContact(msg.getFromId()));
 
     SlideDeck slideDeck = new SlideDeck();
-    if (msg.getType() != DcMsg.DC_MSG_TEXT) {
+    if (msg.hasFile()) {
```
r10s, line 1470: yeah, that seems better, i double checked that this is working :)

src/main/res/menu/conversation.xml
```diff
@@ -25,6 +25,11 @@
         android:visible="false"
         app:showAsAction="always"/>
 
+    <item android:id="@+id/menu_start_call"
+        android:title="@string/start_call"
+        android:icon="@drawable/baseline_call_24"
+        app:showAsAction="always" />
+
```
r10s, line 32: here is a little inconsistency with desktop, which has the order "apps - call" - as the "apps" section expands to multiple icons, and having "call" at unstable horizontal position looks bad and is bad wrt recognition.

i tried the "apps - call" order on android, however, that does look bad - and then also results in having "apps" at unstable horizontal position (groups eg. does not have "call" icon")

so, all in all, for now, it seems better to have this slight inconsistency between desktop and android/ios

src/main/java/org/thoughtcrime/securesms/components/CallItemView.java
```diff
@@ -0,0 +1,104 @@
...
+      }
+    } else {
+      int[] attrs = new int[]{
+        R.attr.conversation_item_incoming_text_secondary_color,
+      };
+      try (TypedArray ta = getContext().obtainStyledAttributes(attrs)) {
+        footer.setTextColor(ta.getColor(0, Color.BLACK));
+      }
+      icon.setColorFilter(getResources().getColor(R.color.delta_accent_darker));
+    }
```
r10s, line 89: i know, telegram is using the same colors for the phone icon, however, this looks pretty random (also on telegram)

while i can understand that the right bubble picks up "green" - the blue is even more random here.

one tries to make some sense of it, eg. is "missed" red - but there is not much, also as the colors are not shown in summary for technical reasons.

also, being a tappable icon, it also seems better to be not different without reasons 

so, better make them just "black" (or probably "white" in dark mode)

src/main/java/org/thoughtcrime/securesms/connect/DcHelper.java
```diff
@@ -398,11 +402,6 @@ public static String getBlobdirFile(DcContext dcContext, String path) {
 
   }
 
-  public static boolean isWebrtcConfigOk(DcContext dcContext) {
```
r10s, line 401: i wish, it would still be that simple 😅

src/main/java/org/thoughtcrime/securesms/util/AvatarUtil.java
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * @param filePath File path to image
+   * @return data URI like "data:image/jpeg;base64,..."
+   */
+  public static String asDataUri(String filePath) {
+    Bitmap bitmap = BitmapFactory.decodeFile(filePath);
+    if (bitmap == null) return null;
+
+    ByteArrayOutputStream baos = new ByteArrayOutputStream();
+    bitmap.compress(Bitmap.CompressFormat.JPEG, 90, baos);
```
r10s, line 24: recoding seems not needed here, but maybe simple than to mess around with `image/png` and whatnot :)

no blocker

src/main/java/org/thoughtcrime/securesms/util/AvatarUtil.java
```diff
@@ -0,0 +1,51 @@
...
+    return "data:image/jpeg;base64," + base64;
+  }
+
+  /**
+   * convert Drawable to data URI.
+   *
+   * @param drawable avatar image as Drawable
+   * @return data URI like "data:image/png;base64,..."
+   */
+  public static String asDataUri(Drawable drawable) {
```
r10s, line 36: very nice to take care of fallback avatars here!

this is missing on desktop, i filed an issue for that at https://github.com/deltachat/deltachat-desktop/issues/5529

src/main/res/values/strings.xml
adbenitez: actually the date is without `Thu, `

src/main/java/org/thoughtcrime/securesms/util/DateUtils.java
adbenitez: the thing is that showing full date gets too big and the duration will wrap around then (or overflow, didn't check)

src/main/java/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
adbenitez: this actually had a syntactic error with `"phone"` not using `\"`

src/main/java/org/thoughtcrime/securesms/components/CallItemView.java
adbenitez: I modified to your suggestion, this is how it looks:
image

src/main/res/values/strings.xml
r10s: well, on iOS, it is with :) - and also on android, we should show the date unless it is today

it is just to give some example

adbenitez: in iOS it is with if it is another day but today then it is only the time, so the time is a more neutral/flexible example but it is not that important, it is just an example so lets merge it like that, thanks

src/main/java/org/thoughtcrime/securesms/components/CallItemView.java
r10s: yip, this is better, "don't make me think" :) (i practise, there is not such a wall of phones, so the change is more subtile then the screenshot suggest :)

src/main/java/org/thoughtcrime/securesms/util/DateUtils.java
r10s: i think,  a typical date should still fit, sth like "Aug 24, 11:20 PM, 5 Minutes" - it is only for older calls, so the duration is anyway less important, even if it overflows

still, this is better than ppl seeing the bubble out of context - "wtf, someone called me at 13:00!!!" - and calling back, then realizing it was two days ago (we had that with other messages)

so, i would show the date mostly as usual, only for the same day, using the HH:MM instead of relative dates (to avoid "2 minutes, 1 minute")

src/main/java/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
r10s: ah sorry, i was just typing the suggestion on the web :)

src/main/java/org/thoughtcrime/securesms/util/DateUtils.java
adbenitez: yep, I agree showing only the time looks a bit weird because the other messages have different date format and we don't have floating/sticky date-markers headers, will tweak as suggested

r10s: great. as said, even with sticky/floating headers we would probably do the same

PR #3786, opened 2025-06-10
allow to react with same emoji as others

adbenitez:
this allows to easily react with same emoji as others from the reactions details fragment
image


r10s (approved):
yeah, let's give it  a try :)

PR #3788, opened 2025-06-13
add some DeltaTouch specific strings

r10s:


lk108:
Thanks! Maybe it should be "Include requests in counters and notification**s**"?

PR #3792, opened 2025-06-24
re-focus profile

r10s:
counterpart of https://github.com/deltachat/deltachat-ios/pull/2690 , see there for more information and reasoning


adbenitez:
thanks a lot! 

the last seen info looks pretty random positioned, would be better as subtitle of profile name (it seems you don't like this for some reason, so moving it at the bottom with the rest of similar text will be better in that case than current position, but I am ok with it if you decide to keep current position)

src/main/res/menu/profile_common.xml
```diff
@@ -8,7 +8,8 @@
 
     <item android:title="@string/global_menu_edit_desktop"
           android:id="@+id/edit_name"
-          app:showAsAction="never"/>
+          android:icon="@drawable/ic_create_white_24dp"
+          app:showAsAction="always"/>
```
adbenitez, line 12: was the edit pencil/action shown only for profiles and not for groups intentionally? or could I also make it available in a subsequent PR?

r10s: it is on purpose, that contact-name changes are easy accessible - while group name changes are not, and have another click path. 

these are two very different things.

for the contact-name, we want to encourage ppl to change them, they affect only the user making the change.  but this i not true for group-name changes, as they should be more stable, and ppl should really know what they're doing and affect lots of ppl.
for not set group avatar, however, tapping the group avatar makes changing it easy.

for iOS, that part is "wrong", btw, both views show "edit" where group should not. and for desktop it is even the other way round - contact-name-changes are complicated and group-name-changes are easy :)

we can think over that at some point, this is my gut feeling only :) but it is an improvement to before in any case, it is not harder to change group names than before - it is only easy to change contact-names

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Profiles focus on recognizing contacts
+* See the number of media directly in the profile, no need to tap around
+* Open "All Media" view with the tap that actually has content, if any
```
adbenitez: tap or tab?

r10s: tap the tab

Amzd:
I agree with adb about the last seen message. Might be fixed if it had less padding at the bottom.

I also think the two lines at the bottom (introduced by and email) look a bit lost, not sure exactly how to fix as idk anything about Android design language.

src/main/java/org/thoughtcrime/securesms/ProfileAdapter.java
```diff
@@ -0,0 +1,372 @@
...
+        itemData.add(new ItemData(ITEM_ADDRESS, dcContact.getAddr(), 0));
+      } else if (isMailingList) {
+        itemData.add(new ItemData(ITEM_ADDRESS, dcChat.getMailinglistAddr(), 0));
+      }
+    }
+
+    notifyDataSetChanged();
+  }
+
+  public int ALL_MEDIA_COUNT_MAX = 500;
```
adbenitez, line 350: why 500 and not 999 or 99? to me 99+ would already be good enough, not sure a higher precision number brings much utility here compared to potential performance cost of the calculation

adbenitez:
about the "send message" option, I think a chat bubble is more fitting there than the send-button icon, there are some option via android studio vector picker:
image

adbenitez:
> I agree with adb about the last seen message. Might be fixed if it had less padding at the bottom.
> 
> I also think the two lines at the bottom (introduced by and email) look a bit lost, not sure exactly how to fix as idk anything about Android design language.

maybe adding a divider to the bottom section so it is separated from chats in common would help, if it would be me I would remove the section completely:

* move last seen to top as subtitle
* move introduced-by to menu
* move addresses to encryption option or another "addresses" menu item (when multi-transport this will be a list not really a single address) since addresses are more advanced debugging information since the ID will be the encryption identity

then the screen would be much more clearer without weird info in random places detached from profile

adbenitez (approved):
overall looks good, "we can always iterate", thanks a lot! that was quite some work!

adbenitez:
btw, when muting a chat, there is no feedback of the muted state in the profile, is this intentional?

r10s:
btw, wondering why "add member" etc look grayish, looks disabled to me. will check that

src/main/res/layout/profile_avatar_item.xml
```diff
@@ -0,0 +1,57 @@
...
+
+        <org.thoughtcrime.securesms.components.AvatarView
+            android:id="@+id/avatar"
+            android:layout_width="152dp"
+            android:layout_height="152dp"
+            android:layout_marginTop="16dp"
+            android:layout_marginBottom="5dp"/>
+
+        <androidx.appcompat.widget.AppCompatTextView
+            android:id="@+id/name"
```
Hocuri, line 29: When the user taps on the contact's name, would be nice if the "Edit Name" dialog opens. Because I've had a user try to edit the name in this way, and it would be nice if this just worked.

src/main/res/layout/profile_text_item_button.xml
```diff
@@ -0,0 +1,34 @@
...
+        <TextView
+            android:id="@+id/value"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:layout_weight="1"
+            android:gravity="end|center_vertical"
+            style="@style/Signal.Text.Caption"
+            android:textColor="?attr/conversation_list_item_date_color"
+            android:visibility="gone" />
+
```
Hocuri, line 32: I'm wondering if we should show a image icon behind clickable list items ("Apps & Media", "Send Message", "Introduced by"), in order to communicate that they are clickable. For "Send Message" it's clear, for the others I'm not sure whether all users will understand, and neither am I sure whether this would help.

WhattsApp also does it here:

image

src/main/res/layout/profile_avatar_item.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Hocuri, line 1: Some small tweaks: Can we get
- a bit more spacing over&under the avatar, and a bit less spacing under "Last seen"?
- a bit more spacing left and right of the "Apps&Media" and "Send Message" icons?
- Slightly larger font for "Last seen", addresses, and "Introduced by"? I think they are hard to read esp. for people with impaired vision (and even myself).

src/main/java/org/thoughtcrime/securesms/ProfileAdapter.java
```diff
@@ -0,0 +1,372 @@
...
+        if (verifierId == DcContact.DC_CONTACT_ID_SELF) {
+          introducedBy = context.getString(R.string.verified_by_you);
+        } else {
+          introducedBy = context.getString(R.string.verified_by, dcContext.getContact(verifierId).getDisplayName());
+        }
+        itemData.add(new ItemData(ITEM_INTRODUCED_BY, introducedBy, 0));
+      }
+
+      if (dcContact != null) {
+        itemData.add(new ItemData(ITEM_ADDRESS, dcContact.getAddr(), 0));
```
Hocuri, line 341: BTW, for email-contacts, I'm not sure it's good to hide the address so much, but maybe you discussed about this already, so it's fine not to discuss again.

Hocuri (commented):
I like that the avatar is shown bigger, that it's easier to edit the name (one less tap), that "Send Message" got an icon, that there is more hierarchy between the buttons (e.g. "Send Message" is bigger than Lastseen), and that there is some white space which makes it look less cluttered.

About the gallery - I can see a few potential problems there, though I didn't user-test whether they actually lead to confusion:
- What happens when you click on "Apps & Media" is non-deterministic - sometimes, the "Apps" tab opens, sometimes the "Gallery" tab. This has potential for confusion, and also, I need to think more while using DC, because I can't always use the same tap pattern to arrive at the gallery.
- The happy path for opening the gallery is gone, as known from other messengers, esp. if there already is an app in the chat:
  - Ok, how do I open the gallery here?
  - Hm, there is no gallery icon.
  - _continues to look for something that may be what they need_
  - There are these four squares, and it says "Apps & Media", maybe this is what I need?
  - _click_
  - No, this is definitely not the gallery
  - _goes back without noticing that there was a "Gallery" tab

My favorite solution to these problems would be:
- If the user clicks on the "Apps" button in the chat titlebar, we _always_ open the "Apps" tab. 
  - At the top of the list of apps, we always show a button `⊕ Send App` (similar to how we show a `⊕ New Contact` button at the top of the contacts list).
  - Alternatively, and less effort to implement: If there are no apps, we show a help text explaining how to attach an app.
- Rename the "Apps & Media" button in the profile to "Media" or "Media & Apps", and use a gallery icon. If the user clicks on it, we _always_ open the "Gallery" tab.

I know that I'm late to the party, because things have already been implemented on iOS/Desktop. I still wanted to tell my impressions and ideas, but I recognize that it may not be helpful at this late stage, and would be OK with merging without (re-)discussing all of this.

adbenitez:
> About the gallery - I can see a few potential problems there, though I didn't user-test whether they actually lead to confusion:

I also don't like the "smart" opening of tabs that now makes it not deterministic which tab will open and it is a bit confusing and unexpected

src/main/res/layout/profile_avatar_item.xml
r10s: i also thought so, i had that first, see https://github.com/deltachat/deltachat-android/pull/3792/commits/030616ed5ffc884c30872e80bb06b8b4bdf9fc60, but i found it immediately confusing as we do not want to nudge for group-name changes in the same way

maybe okay to not edit group names that way or whatever, however, as the current state already is an improvement, i went for sth else instead of improving or thinking there too much

for your user - maybe they would have seen the pencil that was not there before. 
btw pencil could also be beside the name, but  that looks not so great, so i gave the current state a try first

src/main/java/org/thoughtcrime/securesms/ProfileAdapter.java
r10s: maybe we can show it in the subtitle for email contact, that would match the pattern of repeating the chat-title enlarged

if so, maybe hide it alltogether otherwise? (it is still at encrypt info)

r10s:
the "smart" thingie  - i was unsure there as well when i added it on ios as an experiment.
i came over that as most chats do not have apps but images. it is available there that way since some months. in practise, for me and user i watched, it is quite handy and no one complained so far. it makes the button more useful, and tbh, the gallery immediately looks different enough from app list, one does not even think about it

i would not die on that hill, but maybe worth a try on android as well

for different defaults and slightly different names if media is opened from chat or from profile - maybe, however, not sure if that would resolve confusion or add more. probably sth for an iteration

src/main/res/layout/profile_avatar_item.xml
Hocuri: > for your user - maybe they would have seen the pencil that was not there before.

True

src/main/java/org/thoughtcrime/securesms/ProfileAdapter.java
Hocuri: > maybe we can show it in the subtitle for email contact, that would match the pattern of repeating the chat-title enlarged

+1

> if so, maybe hide it alltogether otherwise? (it is still at encrypt info)

I think it would be kinda hard to figure where to look for this information, because it's not obvious to the user that you can get to the email address(es) by clicking on "Encryption"

Hocuri:
> for different defaults and slightly different names if media is opened from chat or from profile - maybe, however, not sure if that would resolve confusion or add more. probably sth for an iteration

My suggestion didn't include any slightly different names. My suggestion was to make the "Apps" button at the top of a chat (which only consists of an image, and has no name) always open the list of apps, and then make the list of apps more useful by either putting some help text there, or a button to attach an app.

And then change the button in the profile to be more similar to what users know from other messengers, and to just be a normal "open-the-gallery" button.

src/main/java/org/thoughtcrime/securesms/ProfileAdapter.java
r10s: it is random, sure. 

but having a bit more precision than "99" seems fine,  most chats would be 99+ otherwise. the calculation costs seem fine as well, also it is scheduled to background

src/main/res/layout/profile_text_item_button.xml
r10s: maybe. otoh, forward chevron is pretty iOS'ish and not that usual on android natively - if we want to go for it, we probably want to do it also elsewhere, otherwise, we have inconsistent UI

still, i doubt that chevron or not makes a big difference

src/main/res/layout/profile_avatar_item.xml
r10s: i tweaked spacing a bit

src/main/java/org/thoughtcrime/securesms/ProfileAdapter.java
r10s: sleeping over it,  having other display of non-chatmail contacts - well,  that results in non-deterministic UI as well, as the address is found at different places then, so not sure. if at all, it can go to another PR / discussion

as long as we have the display on the profile unconditionally, it seems fine. 
the address is now larger than before, btw

r10s:
thanks a lot for the feedback! in addition to the code-comments above:

- footers now repeat the normal font and spacings
- after sleeping over, i removed the "smart tab" in favour to deterministic UI. indeed we want to push for apps, and this is the most important tab. and having another tap for gallery seems fine. also less code and complexity
- otoh, having slightly different titles and meanings in Profile and Chat seems to add complexity as well, keep it simple for now
- for last seen, see discussion in opening topic 
- for moving "introduced by" to menu - i would love that, but that can go to another PR and being discussed there
- for showing mute: yes, maybe. otoh, it is shown at "chats in common", also maybe just remove the option from the profile menu, we also do not duplicate "disappearing messages". but that's for another PR as well
- a "+" button atop if the "Apps" tab is a neat idea!

Issue #2617, opened 2023-07-16
end-to-end encryption disabled even if necessary public keys are on device


monperrus:
- Android version: 13
- Delta Chat version:  1.38.2

- Expected behavior:
When public keys are on the device, end-to-end encryption is enabled for outgoing messages.

- Actual behavior:
Public keys are on the device (as per menu >> "Encryption",  "end-to-end encryption available",  [DC_STR_E2E_AVAILABLE](https://c.delta.chat/group__DC__STR.html#ga2025629ecb152101fca8347b97ce516d)), yet all outgoing messages are clear text. I don't find a switch to activate it (''end-to-end encryption preferred", [DC_STR_E2E_PREFERRED](https://c.delta.chat/group__DC__STR.html#gaaf68df9f876a78d007461d1dd269560b)).

- Steps to reproduce the problem:
Start a clear text chat between Alice (using an email client) and Bob (using DeltaChat), with a few messages, later Alice sends a message with public key in autocrypt header, then Bob sees "end-to-end encryption available", yet Bob continues to send clear text messages to Alice.

Hocuri:
If the last message you got from Bob was sent using a classical email client, then DC assumes that Bob can't read encrypted message and sends unencrypted. Since years, there is the plan to add a "Force-encrypted/Force-unencrypted" option when sending: https://support.delta.chat/t/force-once-un-encrypted-to-directly-reach-a-classic-mua-after-having-had-a-chat/227, but as so many things, we didn't have the time to implement it yet.

> * Start a clear text chat between Alice (using an email client) and Bob (using DeltaChat), with a few messages, later Alice sends a message with public key in autocrypt header, then Bob sees "end-to-end encryption available", yet Bob continues to send clear text messages to Alice.

Can you retry whether this actually happens even if the last message you got from Bob was sent with DeltaChat?

monperrus:
thanks for the fast answer @Hocuri .

do you mean that Deltachat checks some additional information in addition to the presence of the autocrypt header, such as the user-agent?

Hocuri:
For encryption, DC only checks the Autocrypt header, and inside the Autocrypt header the `prefer_encrypt` field.

monperrus:
OK, that's clear, thanks. It's consistent with the fact that Thunderbird does not set `prefer_encrypt` in the autocrypt header.

So a fix easier than adding a button would be to assume `prefer_encrypt=true` if `prefer_encrypt`is absent.

[...]

PR #3793, opened 2025-06-24
adapt to new key-contacts

adbenitez:
- [x] hide add-member, qr invite and ability to remove members from email threads/groups
- [x] show dialog if account migration fails
- [x] show email icon for unencrypted messages, don't show padlock for encrypted
- [x] disable ability to edit group for email threads/groups
- [x] disable chat cloning for email threads/groups
- [x] show email address in conversation subtitle for email contacts
- [x] hide "leave group" menu option for email threads/groups
- [x] don't allow to edit mailinglist name (it will get reset on next incoming message anyways)
- [x] properly set core submodule when a new core release with https://github.com/chatmail/core/pull/6796 is created
- [x] add changelog entry


adbenitez:
in #3507 we removed the email address from subtitle unconditionally because it was displayed for any non-verified contact which included vcard contacts, now that we have a more clearer distinction between email contact and contact with encryption this can be reintroduced and in fact helps to recognize contacts now that all have the same email icon, also because for email usage people expect to see the email address since it is the identity there, this also leverage the situation for email users to have some gains/features instead of worsening with the new pgp-contacts changes:

image

Hocuri (approved):
Let's go :tada: ~~(after updating core submodule to 2.0.0 of course)~~ Edit: Just saw that you already did, only the changelog is missing

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
 * Improve hint for app drafts
 * Add Text-To-Speech (TTS) support for in-chat apps
 * New icon for the QR icon
+* Improved separation between unencryted chats/contacts and encrypted ones, avoiding mixing of encrypted and unencrypted messages in the same chat
+* Removed padlocks, as encrypted is the default "normal" state, instead, unencryted email is marked as such
+* Classic email chats/threads now have an email letter icon making it easy to recognize
```
Hocuri: Some suggestions:
- Start a new sentence after `"normal" state`
- Typo: unencryted -> unencrypted
- Explain how unencrypted email is marked

```suggestion
* Removed padlocks, as encrypted is the default "normal" state. Instead, unencrypted email is marked with a small email / letter  (✉️) icon
* Classic email chats/threads get a big email / letter icon making it easy to recognize
```

Hocuri (approved):

monperrus:
Hi @adbenitez, all,

Thanks for the work in this PR.

I see in the release that padlocks have disappeared ("don't show padlock for encrypted"). I disagree with this change, it removes transparency and gives a false sense of security. WhatsApp, Signall all show meaningful padlocks.

Any chance to revert that particular change?

Hocuri:
> WhatsApp, Signall all show meaningful padlocks.



I can't find any padlocks in WhatsApp or Signal. Can you send a screenshot of this?

monperrus:
when you start a new chat 
whatsapp_encryption_1_new

since WhatsApp uses double-ratchet OTR encryption, it can never be lost.

on the contrary, DC stops encrypted if the recipient replies unencrypted (https://github.com/deltachat/deltachat-android/issues/2617), so one believes it's encrypted while it's not. Dangerous when encryption is mission critical.

Hocuri:
I see, the padlock is in the info message that is shown at the start of the chat. Probably makes sense for us to do the same; OTOH, it's not super important, because when you read the text, then it also says that messages are end-to-end encrypted.

> on the contrary, DC stops encrypted if the recipient replies unencrypted (#2617), so one believes it's encrypted while it's not. Dangerous when encryption is mission critical.

This is not the case anymore. If a chat is encrypted, it will be encrypted forever.

r10s:
> This is not the case anymore. If a chat is encrypted, it will be encrypted forever.

see https://delta.chat/en/2025-08-04-encryption-v2 and https://delta.chat/en/help#e2ee for details

monperrus:
Thanks for the pointers.

The property we're discussing is: "if I send an encrypted messsage to Bob, then I will send encrypted to Bob forever",

This is not the case, I've just tried this:

* Alice sends a message to Bob (on DC) encrypted
* Bob replies encrypted
* Alice sends a message to Bob (on DC) in clear
* Bob (on DC) replies to Alice in clear 

With the latest UI changes, we have two different chats with Alice, with no graphical icons to understand the difference between both.

r10s:
you are not the only one missing the locks :)

however, the chats with lower protection are clearly marker as "email" by the grey, boring, repeated avatars. it is so visibly different, that pplcompain it is too visible :)

but we really want usable security here, that ppl understand, these chats have a lower standard.

in user tests it turns out that locks, shieds, verification marks are often misunderstood and overseen.

therefore, we use a symbol more user understand, the email symbol. ppl may or may not intuitely think about e2ee, it is enough if they understand it is same as their classic mail.

finally note, that most users do no longer use delta chat for email, let alone on the same accout (as Bob from above, using eg. Delta Chat and Thinderbird), so they never see or get the grey avatars ot not e2ee.

this is also needed for our small team to keep focus, not diving into all cornercaes of pgp. if someone wants to do a fork with another focus, that'd be welcome, see forum

and for the advanced usere that do, it is easy enough to understand

monperrus:
> however, the chats with lower protection are clearly marker as "email" by the grey, boring, repeated avatars. i

ahah, I see! I was also surprised and confused by this borification.

Here is a proposal: we overlay a padlock on the colorful profile avatar / icon of the secure conversation.

PR #3794, opened 2025-06-25
[WIP] Make sending of statistics into a setting

Hocuri:
Counterpart of https://github.com/chatmail/core/pull/6851

### DIFF
--- a/src/main/res/xml/preferences_advanced.xml
+++ b/src/main/res/xml/preferences_advanced.xml
@@ -11,8 +11,11 @@
         android:summary="@string/enable_realtime_explain"
         android:title="@string/enable_realtime"/>
 
-    <Preference android:key="pref_self_reporting"
-        android:title="@string/send_stats_to_devs"/>
+    <org.thoughtcrime.securesms.components.SwitchPreferenceCompat
+        android:defaultValue="false"
+        android:key="pref_self_reporting"
+        android:title="@string/send_stats_to_devs"
+        android:summary="@string/send_stats_to_devs_explanation"/>
 
     <PreferenceCategory android:title="@string/pref_experimental_features">
 
--- a/src/main/java/org/thoughtcrime/securesms/connect/DcHelper.java
+++ b/src/main/java/org/thoughtcrime/securesms/connect/DcHelper.java
@@ -73,6 +73,7 @@ public class DcHelper {
     public static final String CONFIG_VERIFIED_ONE_ON_ONE_CHATS = "verified_one_on_one_chats";
     public static final String CONFIG_WEBXDC_REALTIME_ENABLED = "webxdc_realtime_enabled";
     public static final String CONFIG_PRIVATE_TAG = "private_tag";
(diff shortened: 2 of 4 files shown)


src/main/res/xml/preferences_advanced.xml
```diff
@@ -21,6 +21,11 @@
 
     </PreferenceCategory>
 
+    <org.thoughtcrime.securesms.components.SwitchPreferenceCompat
+        android:defaultValue="false"
+        android:key="pref_stats_sending"
+        android:title="@string/send_stats_to_devs" />
```
r10s: i think, a better position is inside the implicit "others" section  -  it is not related to "server", and if we add a section at that position, we would have another section, which is not needed

(more concretely, atop of "screen security" then)

the most important way to get ppl to send statistics is anyway the device message - the advanved settings are more to have a way to enable it again  - or maybe our delta friends, but they will find it nayways

adbenitez:
this doesn't compile

adbenitez (approved):

src/main/java/org/thoughtcrime/securesms/util/Prefs.java
```diff
@@ -65,6 +64,8 @@ public class Prefs {
   public  static final String WEBXDC_STORE_URL_PREF            = "pref_webxdc_store_url";
   public  static final String DEFAULT_WEBXDC_STORE_URL         = "https://webxdc.org/apps/";
 
+  public static final String PREF_STATS_SENDING                = "pref_stats_enabled";
```
adbenitez: this seems unused

src/main/java/org/thoughtcrime/securesms/util/Prefs.java
```diff
@@ -154,6 +155,14 @@ public static int getPrompteDozeMsgId(Context context) {
...
   }
 
+  public static void setStatsDeviceMsgId(Context context, int msg_id) {
+    setIntegerPreference(context, STATS_DEVICE_MSG_ID_PREF, msg_id);
+  }
+
+  public static int getStatsDeviceMsgId(Context context) {
+    return getIntegerPreference(context, STATS_DEVICE_MSG_ID_PREF, 0);
+  }
+
```
adbenitez, line 163: this looks a bit unnecessary, we already have device message chat APIs, the API is just a bit poor, there is `wasDeviceMsgEverAdded(label)` instead it a `getDeviceMessageId(label)` api could be added and that could replace/deprecate existing `wasDeviceMsgEverAdded` since `wasDeviceMsgEverAdded == getDeviceMessageId(label) != 0`

then we will not need this UI preferences

Hocuri: Core currently doesn't record the MsgId belonging to a device message label, it only has a sql table with a single column, i.e. a simple list of all device msgs. While we can change this, all of this makes this not straightforward, so, let's postpone it :)

FTR, the doze reminder device msg uses the same approach as the stats device msg.

PR #3797, opened 2025-07-04
another attempt at disabling webrtc, this time using `WebViewCompat.addDocumentStartJavaScript`

Simon-Laux:
problem is that it doesn't run on special urls like `about:blank`. I haven't found a solution for that. It works fine for all other frames, though.

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -1,5 +1,6 @@
 package org.thoughtcrime.securesms;
 
+import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
@@ -32,6 +33,8 @@ import androidx.core.app.TaskStackBuilder;
 import androidx.core.content.pm.ShortcutInfoCompat;
 import androidx.core.content.pm.ShortcutManagerCompat;
 import androidx.core.graphics.drawable.IconCompat;
+import androidx.webkit.WebViewCompat;
+import androidx.webkit.WebViewFeature;
 
 import com.b44t.messenger.DcChat;
 import com.b44t.messenger.DcContext;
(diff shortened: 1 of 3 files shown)


Simon-Laux:
closing, because there probably is no solution, I just made this pr for future reference.
I did quite some research before I gave up.

r10s:
how exactly is `about:blank` is related to loading our webxdc  - looking briefly, we're using `https://acc1234-msg4567.localhost`

is it as webxdc can open an `about:blank` iframe and change it later? can't we then use the new approach not at least for the initial frame, where we know it is not `about:blank` and keep FILL500 in place otherwise?

EDIT: answering myself, FILL500 fills global resources shared by all frames, so if another approach is not sufficient, we need to that that anyways before :/ 

/me obviously not that deep into that currently :)

r10s:
i assumed you tried that, but still i want to mention it: addDocumentStartJavaScript is not triggered as `about:blank` does not match `*` for some definition - but one might think, that it is triggered when another document is loaded, eg by `document.getElementById('myframe').src = 'foo.html'`

Simon-Laux:
> how exactly is about:blank is related to loading our webxdc 

if you run this pr and try it with the test app, you will find out, that it works in the frames, but other tests fail and expose the IP address. Those other tests use empty frames, which seem to load `about:blank`, which is somehow a special site.

Web extensions have an extra property for those pages `match_about_blank`, but I haven't found such an option in the android webview api.

There may also be one or two other special site cases that are similar to about:blank, but I haven't looked into those, https://github.com/w3c/webextensions/issues/575 may give more clues about the other cases.

I tried 3 approaches to fix the `about:blank` issue, that all were unsuccessful:
- `WebViewCompat.addDocumentStartJavaScript(this.webView, script, Collections.singleton("about:blank"));` - got an error that `about:blank` is not allowed there: `java.lang.IllegalArgumentException: allowedOriginRules about:blank is invalid`
- search for alternative api's to enable running the content script also on those special pages
- try to disable about blank frames via CSP, which also did not work

PR #3800, opened 2025-07-05
tweak new profile

### LINKED ISSUE #3799: long tap profile-address -> copy to clipboard
long tap on address should offer an option to copy it address to clipboard, this is a worsening introduced by  https://github.com/deltachat/deltachat-android/pull/3792 (before, you could long tap title)

r10s:
follow up for https://github.com/deltachat/deltachat-android/pull/3792 

closes #3799 

#skip-changelog as there was no release with the missing features

see commits for details

EDIT: copying was tuned down on purpose at https://github.com/deltachat/deltachat-android/pull/3175 , this PR just does the same as before, long tapping an address offers to copy it to the clipboard

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ProfileAdapter.java
+++ b/src/main/java/org/thoughtcrime/securesms/ProfileAdapter.java
@@ -8,6 +8,7 @@ import android.widget.TextView;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.fragment.app.Fragment;
 import androidx.recyclerview.widget.RecyclerView;
 
 import com.b44t.messenger.DcChat;
@@ -44,6 +45,7 @@ public class ProfileAdapter extends RecyclerView.Adapter
   public static final int ITEM_SHARED_CHATS = 60;
 
   private final @NonNull Context              context;
+  private final @NonNull Fragment             fragment;
   private final @NonNull DcContext            dcContext;
   private @Nullable DcChat                    dcChat;
   private @Nullable DcContact                 dcContact;
@@ -83,12 +85,13 @@ public class ProfileAdapter extends RecyclerView.Adapter
     }
   };
 
-  public ProfileAdapter(@NonNull  Context context,
+  public ProfileAdapter(@NonNull  Fragment fragment,
                         @NonNull  GlideRequests glideRequests,
                         @Nullable ItemClickListener clickListener)
   {
     super();
-    this.context        = context;
+    this.fragment       = fragment;
+    this.context        = fragment.requireContext();
     this.glideRequests  = glideRequests;
     this.clickListener  = clickListener;
     this.dcContext      = DcHelper.getContext(context);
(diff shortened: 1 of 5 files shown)


Hocuri:
With this PR, the click-animation when clicking on "Introduced by" is lost. Not sure if it's a problem, but it's a worsening, because it always feels weird when there is a button that doesn't give any feedback when pressed.

BTW, can we have it so that a short-click on the address also gives the option to copy?

r10s:
tbh, i have no idea where the animation comes from and how to control it. do you have a pointer?

for copy: i  think, long tap is fine and known. it'd be better to have a little menu (as in input field) but also there i have no good idea how to trigger (would need some research and time :)

IMG_9755

Hocuri:
No, no idea. Though I assume that if you remove the long-click logic, and just copy the address to clipboard when short-clicked (just as with all the other buttons in the list), the animation would stay; the downside would obviously be that it wouldn't work anymore to long-click to copy, which some users may expect.

r10s:
we had that other "small" menu in the title bar long-tap before, maybe one should peek there

src/main/java/org/thoughtcrime/securesms/ProfileAdapter.java
```diff
@@ -211,6 +211,9 @@ else if(holder.itemView instanceof ProfileTextItem) {
       if (data.viewType == ITEM_LAST_SEEN || data.viewType == ITEM_ADDRESS) {
         int padding = (int)((float)context.getResources().getDimensionPixelSize(R.dimen.contact_list_normal_padding) * 1.2);
         item.setPadding(item.getPaddingLeft(), item.getPaddingTop(), item.getPaddingRight(), padding);
+        if (data.viewType == ITEM_ADDRESS) {
+          item.setOnLongClickListener(view -> {clickListener.onAddressLongClicked(); return true;});
```
adbenitez: why long click and not just click? long click is harder to discover and simple click does nothing so it looks like this is just a label :thinking:

r10s: well, a long tap is pretty much known for the options cut-copy-paste, this is not really hard to discover, esp. as it is not the normal way we want to push ppl to. also, most of the time, ppl will not want to copy the address when single-tapping the item, so it will pop up accidentally.

also, for consistency with the bio, and other places a long tap is better (tapping a paragraph and offering "copy" is even more weird)

but yeah, main challenge is to get a grip on that, if single-tap is technically easier, i would not vote against it ...

r10s:
k, found out how the context menu works, we added it not so long ago to tune down copying addresses to not push ppl doing things that do not work at the end, https://github.com/deltachat/deltachat-android/pull/3175 - i do not see a need to revert that, nothing has changed since then

r10s:
> With this PR, the click-animation when clicking on "Introduced by" is lost.

this is fixed now.

we're doing just the same as introduced at https://github.com/deltachat/deltachat-android/pull/3175 now

adbenitez (approved):

PR #3803, opened 2025-07-07
modify QR icon

### LINKED ISSUE #3798: icons "qr code" and "apps" are too similar
this is esp. bad as they're at the same position in chatlist and in chat

Image<br>
Image

for the position, we could swap "search" and "qr code", otherwise, we can use "more pixels" and "less anchors" in the qr code icon

adbenitez:
image

close #3798


Hocuri:
This one has quite a lot of small details - is it maybe possible to use the one from Desktop?

In any case, looking forward to a nicer QR icon!

r10s:
nice!

r10s:
it is an improvement already.

the one from desktop is again too close to the app icon.

but i agree, we should have less details.

maybe got for only the upper left-corner being the "anchor" and all the rest being, sth as the following:

Screenshot 2025-07-07 at 19 05 10

or 

Screenshot 2025-07-07 at 19 05 10-ii

also, maybe swap "search" and "Qr code", so that it is no longer at the same position (also makes some sense as in chat-view, search is always in the overflow menu - and in chatlist it is close :)

Hocuri:
I like the second of the icons you sent. Not sure whether all users will recognize it, but I just asked one non-technical person, and she said that it's easily recognizable as a QR code.

Hocuri:
But maybe it would be possible to have even fewer details? The icon we're currently using has 9x9 pixels, while the icon in your screenshot has 11x11 pixels.

r10s:
i created a few-detailed, only one anchor, 9x9 pixel QR code icon at https://github.com/deltachat/interface/blob/main/icons/new-qr.svg - feel free to modify or play around as well (i tried smaller and larger, but 9x9 seems to be a nice spot :)

adbenitez:
in my opinion the icon proposed in this PR is fine, it is different enough from the apps icon without being totally uncommon icon as the only-one-square icon proposed here, also notice there is a "qr scanner icon" used in a button in the qr screen

I don't think the "let's make it different from the apps icon" needs to be overblown of proportion

adbenitez:
> in my opinion the icon proposed in this PR is fine, it is different enough from the apps icon without being totally uncommon icon as the only-one-square icon proposed here, also notice there is a "qr scanner icon" used in a button in the qr screen

I don't think the "let's make it different from the apps icon" needs to be overblown of proportion

Hocuri:
As I wrote in my first post, I think that the icon proposed in the PR has too many details.

But since this is one of the most important buttons in Delta Chat, I think that at this point it's worth it to make two different screenshots of Delta Chat's main screen with the two different icons, and then
- show them to ~10 different non-technical users (showing the screenshot on the phone, so that it feels real),
  - To half of them, we first show the icon proposed in the PR, to half of them, we first show the icon by @r10s, in order not to bias them
- ask whether they recognize what the icons are
- After we told them that it's two different QR code icons, and we're trying to decide which one to use, ask which one they like better.

adbenitez:
> As I wrote in my first post, I think that the icon proposed in the PR has too many details.

it is a QR, I think it is normal in this case and I also don't see any problem or blocker with the details, it is also the icon from the Material Design icons from Google, I think it is better to use something common, designed by some professional designers than creating something new created by some random dev

also showing it to 10 ppl around doesn't tell much on the big picture, and depends on who are the people around you and your location, nor should be taken as a fact, you would need much more people from different parts of the world to have some real valuable statistics

ex. even with the old one that is less(maybe too less) detailed, I have seen people saying "Click the chessboard icon" in tutorials/explanations to others

adbenitez:
further tweaks can be done in later PR or after feedback/complains of users, this is already good enough to solve the issue

PR #3807, opened 2025-07-10
fix: Small channel fixes

Hocuri:
Follow-up for #3783, part of https://github.com/chatmail/core/issues/6884.

- In the profile view of an InBroadcast, the subtitle said "Contact". I just removed the subtitle, because it already says "Channel" at the top.
- When tapping on the avatar of an InBroadcast, an "Edit" button was shown

#skip-changelog because this was not released.

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ProfileActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ProfileActivity.java
@@ -334,7 +334,10 @@ public class ProfileActivity extends PassphraseRequiredActionBarActivity
       Intent intent = new Intent(this, MediaPreviewActivity.class);
       intent.setDataAndType(profileImageUri, type);
       intent.putExtra(MediaPreviewActivity.ACTIVITY_TITLE_EXTRA, title);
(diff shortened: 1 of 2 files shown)


src/main/java/org/thoughtcrime/securesms/ProfileActivity.java
```diff
@@ -334,7 +334,10 @@ public void onEnlargeAvatar() {
       Intent intent = new Intent(this, MediaPreviewActivity.class);
       intent.setDataAndType(profileImageUri, type);
       intent.putExtra(MediaPreviewActivity.ACTIVITY_TITLE_EXTRA, title);
-      intent.putExtra(MediaPreviewActivity.EDIT_AVATAR_CHAT_ID, chatIsMultiUser ? chatId : 0); // shows edit-button, might be 0 for a contact-profile
+      intent.putExtra( // show edit-button, might be 0 for a contact-profile
+              MediaPreviewActivity.EDIT_AVATAR_CHAT_ID,
+              (chatIsMultiUser && !chatIsInBroadcast) ? chatId : 0
```
adbenitez: this should also be disabled for mailing lists! it is gone from menu already but not for this enlarged avatars

r10s (approved):

PR #3810, opened 2025-07-11
allow to "move to top" accounts in the accounts list

adbenitez:
this uses the new `set_accounts_order` JSON-RPC API introduced in https://github.com/chatmail/core/pull/6993

- [x] expose `set_accounts_order` via `Rpc.setAccountsOrder()`
- [x] add "move to top" option to account's context menu and use `Rpc.setAccountsOrder()`
- [x] wait for new core release with `set_accounts_order` support
- [x] update changelog

### DIFF
--- a/src/main/java/com/b44t/messenger/rpc/Rpc.java
+++ b/src/main/java/com/b44t/messenger/rpc/Rpc.java
@@ -164,6 +164,10 @@ public class Rpc {
         return gson.fromJson(getResult("create_broadcast", accountId, chatName), Integer.class);
     }
 
+    public void setAccountsOrder(List<Integer> order) throws RpcException {
+        getResult("set_accounts_order", order);
+    }
+
     private static class Request {
         private final String jsonrpc = "2.0";
         public final String method;
--- a/src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
@@ -41,6 +41,8 @@ import org.thoughtcrime.securesms.recipients.Recipient;
 import org.thoughtcrime.securesms.util.Util;
 import org.thoughtcrime.securesms.util.ViewUtil;
 
+import java.util.Arrays;
+
 public class AccountSelectionListFragment extends DialogFragment implements DcEventCenter.DcEventDelegate
 {
   private static final String TAG = AccountSelectionListFragment.class.getSimpleName();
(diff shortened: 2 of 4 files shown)


src/main/java/com/b44t/messenger/rpc/Rpc.java
```diff
@@ -164,6 +164,10 @@ public int createBroadcast(int accountId, String chatName) throws RpcException {
         return gson.fromJson(getResult("create_broadcast", accountId, chatName), Integer.class);
     }
 
+    public void setAccountsOrder(List<Integer> order) throws RpcException {
```
link2xt, line 167: Does it have to be a list of `Integer` instead of `int`?

src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
```diff
@@ -134,9 +136,35 @@ private void onContextItemSelected(MenuItem item, int accountId) {
...
+      onMoveToTop(accountId);
     }
   }
 
+  private void onMoveToTop(int accountId) {
+    Activity activity = getActivity();
+    if (activity == null) return;
+
+    int[] accountIds = DcHelper.getAccounts(activity).getAll();
+    Integer[] ids = new Integer[accountIds.length];
```
link2xt, line 149: If we use `List<int>`, could do this in place without allocating a new array.

src/main/java/com/b44t/messenger/rpc/Rpc.java
adbenitez: yes

src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
adbenitez: it doesn't matter, this list will never be so big that performance would be a problem, even if user had 1000 profiles it would be OK to have two arrays

src/main/java/com/b44t/messenger/rpc/Rpc.java
link2xt: Why? Also other methods are inconsistent, some take `int accountId` as the first argument, some take `Integer accountId`.

src/main/java/org/thoughtcrime/securesms/accounts/AccountSelectionListFragment.java
adbenitez: but for the record, the reason I used Integer, is because `List<int>` doesn't exist int is a primitive type and to use in `<>` you need a class

adbenitez:
> Drag&drop would be nicer than another menu item that can only move to the top.

drag and drop surely would be fancy, but the "move to top" as already implemented in iOS is good enough for sorting accounts without needing an "edit list mode" (otherwise user could accidentally drag&drop account while just scrolling or clicking) or more UI work/changes (which include changing the event listener that refresh the list when some event is received causing a redraw of the list which would mess up things if user is in the middle of dragging and dropping so you need to temporally ignore the events etc)

PR #3815, opened 2025-07-15
Add new option to create unencrypted email thread

adbenitez:


src/main/res/menu/group_create.xml
```diff
@@ -2,7 +2,7 @@
 
 <menu xmlns:android="http://schemas.android.com/apk/res/android" xmlns:app="http://schemas.android.com/apk/res-auto">
 
-    <item android:title="@string/group_create_button"
+    <item android:title="@string/perm_continue"
```
adbenitez, line 5: this was completely wrong, always showing "create group" also for broadcast/channel creation, and when editing existing group/broadcast was also saying "create group" which is wrong since it would be about saving changes to group name/avatar etc. for now I just changed it to a neutral string "continue" that fits for all, but if provided we could have more fancy string adaption for every situation

src/main/java/org/thoughtcrime/securesms/contacts/ContactSelectionListAdapter.java
```diff
@@ -268,6 +268,8 @@ public void onBindViewHolder(@NonNull ViewHolder viewHolder, int i) {
       itemMultiSelect = false; // the item creates a new contact in the list that will be selected instead
     } else if (id == DcContact.DC_CONTACT_ID_NEW_GROUP) {
       name = context.getString(R.string.menu_new_group);
+    } else if (id == DcContact.DC_CONTACT_ID_NEW_UNENCRYPTED_GROUP) {
+      name = context.getString(R.string.new_email);
```
Hocuri, line 272: I think that "New E-Mail"can be misunderstood to mean one of the following:
- Add an email account to Delta Chat
- Create a new email account and add it to Delta Chat
- Send a new encrypted email to a Thunderbird user (it is expected that mainly a bit more technical users will use this feature, who know what an encrypted email is)

Therefore, I think that it should be called "Send unencrypted email".

src/main/java/org/thoughtcrime/securesms/contacts/ContactSelectionListAdapter.java
```diff
@@ -268,6 +268,8 @@ public void onBindViewHolder(@NonNull ViewHolder viewHolder, int i) {
       itemMultiSelect = false; // the item creates a new contact in the list that will be selected instead
```
Hocuri, line 268: image


Can we remove the "Add Contact Manually" option from the screen that appears after clicking on `+` on the main screen? I think that 4 'new something' options are a bit overwhelming (5 with 'New Channel')

"Add Contact Manually" is available after clicking on "New E-Mail", I think this is enough.

BTW, it would anyways be nice to highlight "New Contact" a bit, because this is what most users need, but this is out of scope for this PR.

Hocuri (approved):

src/main/java/org/thoughtcrime/securesms/contacts/ContactSelectionListAdapter.java
adbenitez: also "send unencrypted email" doesn't really reflect what the option actually does, I already recommended calling it "new email thread" another even clearer option would be "new unencrypted group" because what the option does is not to create an email but a group/thread, btw actually in such groups it is clearer what to do and we could display subjects and show a subject field together with the message field

r10s: there are at least as many opinions as ppl on the word "e-mail" as well as the "envelope" :) see discussions at  https://github.com/deltachat/deltachat-ios/pull/2770, https://mastodon.social/@devSJR@fosstodon.org/114845881860226371 or internal chats for pros and cons

let's just stay with the simple thing for now and not open another discussion here :)

src/main/java/org/thoughtcrime/securesms/contacts/ContactSelectionListAdapter.java
r10s: +1, for the same reasons i removed "Add Contact Manually" from the "New Chat" on iOS, https://github.com/deltachat/deltachat-ios/pull/2770

it should continue existing below "Add recipient", "New Contact" and "Scan QR Code" (it think, it is already at all three places)

src/main/java/org/thoughtcrime/securesms/contacts/ContactSelectionListAdapter.java
adbenitez: about the mastodon user, I think that is totally overthinking, people that will see the email envelope is people using classic email and receiving classic email from people and newsletters/services they know are not even using Delta Chat, also it is never mixed between encrypted message as that person seems to misunderstand, so there is no risk that someone would think that messages are more secure than the ones without the symbol

adbenitez: (and I totally agree with not opening discussion here, I asked for review not to question the decisions that weren't taken by me I just did the same as iOS PR)

adbenitez:
> submodule update needs to be removed

why??? checking latest tagged core is necessary for the new API

adbenitez:
> before merging, as @Hocuri pointed out, "Add Contact Manually" needs to be removed from "New Chat"

done, by doing so I also fixed #3816

r10s:
> > submodule update needs to be removed
>
> why??? checking latest tagged core is necessary for the new API

wasn't there the idea to never update core in PR? otherwise, or it it fits randomly, just ignore my comment and move forward :) i just want to make sure, we do not downgrade

PR #3817, opened 2025-07-16
simplify e2ee strings

r10s:
as discussed with @hpk42, we want to simplify the terms used in UI, as there is no switch-forth-and-back, the "from now on" is superfluous, and the "guaranteed" is related, at least noisy and raises questions.

also, this allows us to use the same string when establishing via vcard contacts (core PR needs to be done), which is a good thing to move forward just now.

the faq will explain nuances then


adbenitez (approved):

r10s:
> I does break my heart a bit, because we discussed a very long time to arrive at this wording

i can feel you ... nothing lasts forever ❤️

adbenitez:
it evolved tho, so there is no reason to be sad but happy that now there is no need to mention encryption is guaranteed because it is always guaranteed and simpy can't be dropped or changed so no need to ask the user to trust a key change and constantly switching back and forth from protected to unprotected in some situations etc.

PR #3820, opened 2025-07-17
add device message

### LINKED ISSUE #3819: add a device message for 2.0
if possible, the device message should make it at least to some translations.

as the "key contact" is by far the largest change, we can probably use the same device message on all platforms (we usually point to CHANGELOG for "further changes" anyways)

adbenitez:
close #3819 

feel free to tweak or suggest changes

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -109,7 +109,7 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
     // it is not needed to keep all past update messages, however, when deleted, also the strings should be deleted.
     try {
       DcContext dcContext = DcHelper.getContext(this);
-      final String deviceMsgLabel = "update_1_50_0_android";
+      final String deviceMsgLabel = "update_2_0_0_android";
       if (!dcContext.wasDeviceMsgEverAdded(deviceMsgLabel)) {
         DcMsg msg = null;
         if (!getIntent().getBooleanExtra(FROM_WELCOME, false)) {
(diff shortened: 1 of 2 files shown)


r10s:
i applied the "tweaks" of @hpk42 and @Hocuri , looks as follows now:

Screenshot 2025-07-18 at 15 07 54<br>
Screenshot 2025-07-18 at 15 09 22

for the "donate", maybe stay with the ❤️ which is also used elsewhere

Hocuri:
I think that it's a bit fuzzy to say "End-to-end encryption is reliable and forever", tough I would be fine with it, it's good enough.

What about "An end-to-end encrypted chat will be end-to-end encrypted forever"? Although I recognize that this sentence is rather long and contains a lot of "end-to-end encrypted".

image

or (with less newlines)

image

Hocuri:
> for the "donate", maybe stay with the ❤️ which is also used elsewhere

+1

r10s:
it we stay with "forever", maybe go for for the 💎 emoji instead of 💯

Raiden-GH:
> An end-to-end encrypted chat will be end-to-end encrypted forever.

I think this sentence could give users the impression that messages in DC weren't reliably encrypted before, even though encryption was advertised. Verified chats were always encrypted, and chatmail messages anyway. Perhaps it would be better to point out that encrypted chats are now better separated from unencrypted ones.
The word "forever" could lead to the misconception that DC is now quantum-safe.

r10s:
> Perhaps it would be better to point out that encrypted chats are now better separated from unencrypted ones.

maybe, but a catchy, positive phrase to cover the complex situation before compared to today easily leaves out or simplifies some aspects.  concrete suggestions welcome :)

> The word "forever" could lead to the misconception that DC is now quantum-safe.

people often make their own conclusions, far away from reality - in case of Email and PGP very often too negative as based on far outdated stuff. so if they think by the pure word "forever", DC is now quantum-safe now, then it's that :) in fact, things are shifting towards pqc, btw, see https://github.com/rpgp/rpgp/pull/518

r10s:
ftr, i read "End-to-end encryption is reliable and forever now" that things, that _are_ marked as "End-to-end encryption" stay like that forever, which is 100% true - which was not always the case before, even though the state was never bad.

r10s:
all in all, i think, things are good enough.

still, i changed the ⌘ icon. i assume, the idea to use ⌘ was that is is somehow close to the four squares of the app icon - but  i think this is not really recognised and ppl will see a key or a shortcut symbol, at least if they have used mac at any point in t heir life :) a simple square is as far away and does not raise wrong associations - any may be read even as a button (but don't have to)

Screenshot 2025-07-19 at 14 05 03

r10s (approved):

PR #3822, opened 2025-07-19
allow tapping new e2ee-info-messages

r10s:
this PR allows to tap the new e2ee-info-messages introduced by https://github.com/chatmail/core/pull/7008
and removes now superfluous "protection broken" path.

(we now have new DC_INFO_CHAT_E2EE added to e2ee chats but otherwise unprotected chats. plus the old DC_INFO_PROTECTION_ENABLED, visually, only the letter gets the green checkmark, otherwise they're the same)

counterpart of https://github.com/deltachat/deltachat-ios/pull/2778

#skip-changelog as there was no official release


src/main/java/com/b44t/messenger/DcChat.java
```diff
@@ -70,7 +69,7 @@ public boolean isOutBroadcast() {
     }
 
     public boolean isHalfBlocked() {
```
adbenitez: now this is half blocked could be removed completely and just replace calls to it with isContactRequest()

r10s: i added a commit - one weird name less :)

PR #3828, opened 2025-07-21
tune down green checkmark

adbenitez:
- [x] remove green checkmarks from chatlist
- [x] remove green checkmark from contact lists
- [x] remove green checkmark from chat-protected info-message (DC_INFO_PROTECTION_ENABLED)
- [x] remove green checkmark from profile title
- [x] add green checkmark left of "Introduced by" line in the profile if contact is verified
- [x] when tapping chat-protected (DC_INFO_PROTECTION_ENABLED) or chat-e2ee (DC_INFO_CHAT_E2EE) info-message, open help at #e2ee instead of e2eeguaranteed
- [x] wait for https://github.com/deltachat/deltachat-pages/pull/1145/ to be merged and update offline help so opening `#e2ee` FAQ actually works

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * After some time, add a device message asking to donate. Can't wait? Donate today at https://delta.chat/donate
 * Allow to sort profiles up in the profile switcher
 * Add new option to create unencrypted email thread
+* Tune down green checkmarks that are often confused with "contact verification"
 * Update to core 2.3.0
 
 ## v1.58.4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 * After some time, add a device message asking to donate. Can't wait? Donate today at https://delta.chat/donate
 * Allow to sort profiles up in the profile switcher
 * Add new option to create unencrypted email thread
-* Tune down green checkmarks that were often confused with Twitter's/WhatsApp's "verified" badges
+* Green checkmarks are removed where they mostly refer to guaranteed encryption, which is the default now. They are still used for profile's "Introduced by"
 * Update to core 2.3.0
 
 ## v1.58.4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 * After some time, add a device message asking to donate. Can't wait? Donate today at https://delta.chat/donate
 * Allow to sort profiles up in the profile switcher
 * Add new option to create unencrypted email thread
-* Tune down green checkmarks that are often confused with "contact verification"
+* Tune down green checkmarks that were often confused with Twitter's/WhatsApp's "verified" badges
 * Update to core 2.3.0
 
 ## v1.58.4
--- a/src/main/java/org/thoughtcrime/securesms/ConversationUpdateItem.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationUpdateItem.java
@@ -26,7 +26,6 @@ public class ConversationUpdateItem extends BaseConversationItem
 {
   private DeliveryStatusView  deliveryStatusView;
   private AppCompatImageView  appIcon;
-  private AppCompatImageView  verifiedIcon;
   private int                 textColor;
 
   public ConversationUpdateItem(Context context) {
@@ -46,7 +45,6 @@ public class ConversationUpdateItem extends BaseConversationItem
     bodyText           = findViewById(R.id.conversation_update_body);
     deliveryStatusView = new DeliveryStatusView(findViewById(R.id.delivery_indicator));
     appIcon            = findViewById(R.id.app_icon);
-    verifiedIcon       = findViewById(R.id.verified_icon);
 
 
     bodyText.setOnLongClickListener(passthroughClickListener);
@@ -111,13 +109,6 @@ public class ConversationUpdateItem extends BaseConversationItem
       }
     } else {
       appIcon.setVisibility(GONE);
-    }
-
-    if (infoType == DcMsg.DC_INFO_PROTECTION_ENABLED) {
-      verifiedIcon.setVisibility(VISIBLE);
-      verifiedIcon.setImageResource(R.drawable.ic_verified);
-    } else {
-      verifiedIcon.setVisibility(GONE);
     }
 
     bodyText.setText(messageRecord.getDisplayBody());
--- a/src/main/java/org/thoughtcrime/securesms/ProfileAdapter.java
+++ b/src/main/java/org/thoughtcrime/securesms/ProfileAdapter.java
@@ -341,6 +341,9 @@ public class ProfileAdapter extends RecyclerView.Adapter
           introducedBy = context.getString(R.string.verified_by, dcContext.getContact(verifierId).getDisplayName());
         }
         itemData.add(new ItemData(ITEM_INTRODUCED_BY, introducedBy, dcContact.isVerified()? R.drawable.ic_verified : 0));
+      } else if (dcContact.isVerified()) {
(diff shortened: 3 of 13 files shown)


src/main/java/org/thoughtcrime/securesms/ProfileAdapter.java
```diff
@@ -339,7 +340,7 @@ public void changeData(@Nullable int[] memberList, @Nullable DcContact dcContact
         } else {
           introducedBy = context.getString(R.string.verified_by, dcContext.getContact(verifierId).getDisplayName());
         }
-        itemData.add(new ItemData(ITEM_INTRODUCED_BY, introducedBy, 0));
+        itemData.add(new ItemData(ITEM_INTRODUCED_BY, introducedBy, dcContact.isVerified()? R.drawable.ic_verified : 0));
       }
```
r10s, line 347: as discussed, the condition for showing the line should be `isVerified || getVerifierId` - one can happen sense without the other.

we need a new string "Introduced" then to cover the case where isVerified==true and getVerifierId==0

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * After some time, add a device message asking to donate. Can't wait? Donate today at https://delta.chat/donate
 * Allow to sort profiles up in the profile switcher
 * Add new option to create unencrypted email thread
+* Tune down green checkmarks that are often confused with "contact verification"
```
Hocuri: I think that people who are familiar with Delta Chat will wonder "but that's right? A green checkmark meant that the public key was verified?" And while we design for the many, it's not good to leave the nerds confused:

```suggestion
* Tune down green checkmarks that were often confused with Twitter's/WhatsApp's "verified" badges
```

src/main/java/org/thoughtcrime/securesms/ConversationUpdateItem.java
```diff
@@ -113,13 +111,6 @@ private void setGenericInfoRecord(DcMsg messageRecord) {
       appIcon.setVisibility(GONE);
     }
 
-    if (infoType == DcMsg.DC_INFO_PROTECTION_ENABLED) {
-      verifiedIcon.setVisibility(VISIBLE);
-      verifiedIcon.setImageResource(R.drawable.ic_verified);
-    } else {
-      verifiedIcon.setVisibility(GONE);
-    }
-
```
Hocuri, line 122: Not sure about removing the green checkmarks from the PROTECTION_ENABLED info message, because here they can't be confused with Twitter's/WhatsApp's "verified" icons, but I would also be fine with removing them, because only few people understand what they mean.

src/main/java/org/thoughtcrime/securesms/ProfileAdapter.java
Hocuri: At least in today's code, _getVerifierId!=0 => isVerified_ (but not the other way round, because sometimes the verifier is unknown but the contact is still verified). 

So, of course we can prepare already for a possible future where core stores a "verifier"/"introducer" for unverified contacts, but today, the database scheme doesn't even have a way to represent an unverified contact that has an introducer, so, it wouldn't be necessary to change things here.

> we need a new string "Introduced" then to cover the case where isVerified==true and getVerifierId==0

:+1:

r10s: k, thanks for clarification! then this simplifies things - i was about to adapt it towards "getVerifierId!=0 || isVerified" at https://github.com/chatmail/core/pull/7027 ,  but that needs more tweaking then

src/main/java/org/thoughtcrime/securesms/ConversationUpdateItem.java
r10s: the removal is not so much about that the green checkmark is confused with Twitter's/WhatsApp's "verified" icons.

it is about that it is far less needed than before, the nuances are hard to explain and do not justify to clutter UI and users to deal with it and need to understand it.

therefore, it should be removed there as well - otherwise, we again have things to explain that are not easy to explain (why some groups have the checkmark and others not...) and few is won.

adbenitez: but the message looks ugly and hard to understand now, I think the green checkmark transmited "this chat is encrypted, it is ok/safe to chat here" while now user needs to read more carefully the opaque info-message

adbenitez: doesn't whatsapp also shows such a badge for the "this chat is e2e encrypted"??

adbenitez: but yeah it is confusing if we use the green checkmark for "contact is verified" in the profile then, probably should be a green shield with a checkmark but that is out of scope for 2.0, but this PR is a deterioration overall

src/main/java/org/thoughtcrime/securesms/ProfileAdapter.java
r10s: if https://github.com/chatmail/core/pull/7027 is correct now, imu, the condition here still needs to be adapted in a way that isVerified() alone creates the line

adbenitez: I added a new "Introduced" string that is now displayed when verifier is unknown

src/main/java/org/thoughtcrime/securesms/ConversationUpdateItem.java
r10s: whatsapp uses a nicer/different background color, no big icon or so, i was considering that already, but thought, for now, other stuff is more important :)

so, i agree, it is good enough for now and out of scope of this PR. maybe it is also just fine as it is now, it looks different, but not bad. we can consider adding an icon or a color or whatever later



_whatsapp still has the shopping handbag in the message_

tbh, i am not sure of the yellowish background color is only for the e2ee message or just "their" info-message-color, so also used for member-added etc.

adbenitez: probably is not used for member added but for that kind of special message, I see they use the padlock, maybe that would be good enough, but it is not a big deal, "we can always iterate"

r10s (approved):

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 * After some time, add a device message asking to donate. Can't wait? Donate today at https://delta.chat/donate
 * Allow to sort profiles up in the profile switcher
 * Add new option to create unencrypted email thread
+* Tune down green checkmarks that were often confused with Twitter's/WhatsApp's "verified" badges
```
r10s: the removal is not really about that the green checkmark is confused with Twitter's/WhatsApp's "verified" icons.

it is about that they're far less needed than before,  the whole e2ee aspect is now default, even mitm aspect has changed a lot, and the nuances are hard to explain and do not justify to clutter UI and users to deal with it and need to understand it.

on iOS, i added that to the padlock thingie as `Removed padlocks and most green checkmarks, as encrypted is the default "normal" state. Instead, unencrypted email is marked with a small letter icon` arguably fuzzy as well

```suggestion
* Green checkmarks are removed where they mostly refer to guaranteed encryption, which is the default now. They are still used for profile's "Introduced by"
```

it is so hard to understand, it is even hard to explain in the CHANGELOG :)

adbenitez: mistly == mostly?

adbenitez: feel free to auto-commit changes directly

r10s:
merging in to make the new string available for other UI

PR #3836, opened 2025-07-23
update device message: "letter" -> "envelope"

WofWca:
"letter" can be confused with "alphabet letter". Envelope is more clear, to me at least.

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -1224,7 +1224,7 @@
     <string name="perm_enable_bg_already_done">You already allowed Delta Chat to receive messages in the background.\n\nIf messages still do not arrive in background, please also check your system settings.</string>
 
     <!-- device messages for updates -->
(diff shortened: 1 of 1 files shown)


r10s:
thanks for the thoughts!

technically, the PR is fine and only a `./scripts/tx-update-changed-sources.sh` would be needed afterwards.

ftr, at https://delta.chat/en/help and https://staging.delta.chat/1148/en/2025-07-25-encryption-v2 we name the thing "e-mail icon" - and at all places, including device message, we additionally show the icon, i think, that is the most important part.

idk, for me, the context is clear that it is not a glyph, also "letter" refers more to a real message (a love letter, eg. "You got a letter from Alice! Yay!") - where "envelope" is more a formal thing. 

but if everyone likes "envelope" more, we can change, but i doubt it makes a difference and is really important

WofWca:
Alright, let's close then.

PR #3847, opened 2025-07-30
improve classic email threads/chats UX

adbenitez:
disable non-functional options

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * hide contact email addresses in search results
 * disable non-functional message editing and ephemeral messages timer settings in classic email thread chat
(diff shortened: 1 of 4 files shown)


r10s:
> disable non-functional options

a little bit more depth would be cool - i was checking for "edit" in profiles and wondering :)

so for other readers:

this PR disables:

1. "Edit Message" when long tapping a message
2. "Disappearing Messages" in the chat's three-dot-menu

the first one is same for desktop/iOS already.

for the latter, yes, probably makes sense to disable that as well. however, we should then file issues to disable "Disappearing Messages" for desktop/ios as well

there may be more things not really useful for non-chatmail, eg. "attach app", but that can be discussed separately

r10s (approved):

PR #3856, opened 2025-08-02
don't enlarge avatars for special chats

### LINKED ISSUE #3855: no enlarging of "saved messages" and "device chat" avatars
same as the just fixed "email avatars", it does not make sense to allow enlarging these default icons.

it is even bad, as the user may unexpectedly open it and get confused or annoyed

cmp https://github.com/deltachat/deltachat-desktop/pull/5373

adbenitez:
close #3855


r10s (approved):
nice!

PR #3867, opened 2025-08-06
make apk builds more reproducible

### LINKED ISSUE #3608: Remove some non-determinism
ArcaneChat is reproducible in F-Droid, while your app is not built reproducible we continuously test older versions on https://verification.f-droid.org/?C=M;O=D and would like more and more apps to become repro

take this latest verification: https://verification.f-droid.org/com.b44t.messenger_7091.apk.diffoscope.html

this needs at least the build-id being removed: https://verification.f-droid.org/com.b44t.messenger_7091.apk.diffoscope.html#lib-armeabi-v-a-libnative-utils.so

and the build date: https://verification.f-droid.org/com.b44t.messenger_7091.apk.diffoscope.html#lib-armeabi-v-a-libnative-utils.so---strings---all---bytes-----

since this touches a bit more of the build process I've not opened a PR, but you can use the Arcane as a guide: eg. https://github.com/ArcaneChat/android/commit/2ce614dc556dd1d56b41d2809e970613d1dfdafe and https://github.com/ArcaneChat/android/commit/2376777bba18dd34185f301002b9b87eba525a35

adbenitez:
close #3608


scripts/ndk-make.sh
```diff
@@ -66,11 +74,9 @@ if test -z "$NDK_HOST_TAG"; then
 fi
 
 if test -z "$CARGO_TARGET_DIR"; then
-    CARGO_TARGET_DIR=target
+    export CARGO_TARGET_DIR=/tmp/deltachat-build
```
link2xt, line 77: Can this be a symlink as well? I don't think I have space for 4 builds in /tmp and don't want to rebuild from scratch each time /tmp is cleared.

adbenitez: I don't remember well, but I think it is pretty likely that I used that because cargo/rust was still getting the original path instead of symlink but if you could check that using symlink works I am more than happy to use that also for the same reasons

r10s (approved):

PR #3877, opened 2025-08-09
avoid ANR in ConversationListActivity.refreshTitle()

### LINKED ISSUE #3876: ANR in ConversationListActivity.refreshTitle()
```

  
  
      #00  pc 0x000000000005823c  /apex/com.android.runtime/lib64/bionic/libc.so (syscall+28)
    
  
      #01  pc 0x0000000000c16798  /data/app/~~YNiv_eVgPjOGhU6thKVLHA==/chat.delta-olZDFEzpdi1qJRxSHAZdcA==/lib/arm64/libnative-utils.so (BuildId: 1e2551f65d702b296e3f8d95c9bcc6f5ab849931)
    
  
      #02  pc 0x0000000000c16314  /data/app/~~YNiv_eVgPjOGhU6thKVLHA==/chat.delta-olZDFEzpdi1qJRxSHAZdcA==/lib/arm64/libnative-utils.so (BuildId: 1e2551f65d702b296e3f8d95c9bcc6f5ab849931)
    
  
      #03  pc 0x0000000000d7de4c  /data/app/~~YNiv_eVgPjOGhU6thKVLHA==/chat.delta-olZDFEzpdi1qJRxSHAZdcA==/lib/arm64/libnative-utils.so (BuildId: 1e2551f65d702b296e3f8d95c9bcc6f5ab849931)
    
  
      #04  pc 0x0000000000d7e5b8  /data/app/~~YNiv_eVgPjOGhU6thKVLHA==/chat.delta-olZDFEzpdi1qJRxSHAZdcA==/lib/arm64/libnative-utils.so (BuildId: 1e2551f65d702b296e3f8d95c9bcc6f5ab849931)
    
  
      #05  pc 0x000000000076c738  /data/app/~~YNiv_eVgPjOGhU6thKVLHA==/chat.delta-olZDFEzpdi1qJRxSHAZdcA==/lib/arm64/libnative-utils.so (dc_get_connectivity+236) (BuildId: 1e2551f65d702b296e3f8d95c9bcc6f5ab849931)
    
  
      at com.b44t.messenger.DcContext.getConnectivity (Native method)
    
  
      at org.thoughtcrime.securesms.connect.DcHelper.getConnectivitySummary (DcHelper.java:452)
    
  
      at org.thoughtcrime.securesms.ConversationListActivity.refreshTitle (ConversationListActivity.java:302)
    
  
      at org.thoughtcrime.securesms.ConversationListFragment.handleEvent (ConversationListFragment.java:337)
    
  
      at org.thoughtcrime.securesms.connect.DcEventCenter.lambda$sendToObservers$0 (DcEventCenter.java:105)
    
  
      at org.thoughtcrime.securesms.connect.DcEventCenter$$ExternalSyntheticLambda0.run (D8$$SyntheticClass)
    
  
      at android.os.Handler.handleCallback (Handler.java:958)
    
  
      at android.os.Handler.dispatchMessage (Handler.java:99)
    
  
      at android.os.Looper.loopOnce (Looper.java:243)
    
  
      at android.os.Looper.loop (Looper.java:338)
    
  
      at android.app.ActivityThread.main (ActivityThread.java:8399)
    
  
      at java.lang.reflect.Method.invoke (Native method)
    
  
      at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:589)
    
  
      at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1055)
    
  
```

adbenitez:
close #3876


r10s:
the PR is probably fine, but get_connectivity() should really not block, same for some of other api as get_config(), cc @link2xt @iequidoo @Hocuri

adbenitez:
for reference, here are the 2 users/devices affected by this issue:

image

iequidoo:
> the PR is probably fine, but get_connectivity() should really not block, same for lots of other api, cc @link2xt @iequidoo @Hocuri

`get_connectivity()` takes `SchedulerState`'s read lock. This isn't good, it's hard to maintain that a write lock isn't taken for a long period. I think `Connectivity` value should be cached in `Context` and read from there atomically. CC @link2xt

Hocuri:
Concretely, `start_io()`, `stop_io()` and `pause_io()` take a write lock of the scheduler, so, `get_connectivity()` has to wait if one of these are running concurrently.

Sure, we could store the information in `Context` rather than the scheduler. We need to make sure that while the whole scheduler is `Stopped` or `Paused`, `NotConnected` is returned correctly, possibly by setting the connectivity in `pause()` etc.

Only caching it in the context (and sometimes returning an outdated value) is probably error-prone, because if the UI reads an outdated value, and afterwards no more ConnectivityChanged events are sent, then the outdated value will be shown to the user for a long time.

adbenitez:
since it seems it is better to solve the actual problem on core instead of just ignoring it by moving the task to a thread (which might anyway get stuck there for long times and not refresh the expected info timely), lets close this PR

PR #3886, opened 2025-08-15
[WIP] feat: Adapt to channel-invite QR codes

Hocuri:
Companion PR for https://github.com/chatmail/core/pull/7042/

### DIFF
--- a/src/main/java/com/b44t/messenger/DcContext.java
+++ b/src/main/java/com/b44t/messenger/DcContext.java
@@ -49,6 +49,7 @@ public class DcContext {
 
     public final static int DC_QR_ASK_VERIFYCONTACT = 200;
     public final static int DC_QR_ASK_VERIFYGROUP   = 202;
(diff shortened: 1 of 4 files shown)


src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1152,7 +1152,10 @@ else if (!body.isEmpty()){
 
           if (doSend) {
             if (dcContext.sendMsg(dcChat.getId(), msg) == 0) {
-              Util.runOnMain(()-> Toast.makeText(ConversationActivity.this, dcContext.getLastError(), Toast.LENGTH_LONG).show());
+              String lastError = dcContext.getLastError();
+              if (!"".equals(lastError)) {
```
Hocuri, line 1156: This check is so that we don't show an empty toast when sending a message into a legacy channel fails.

PR #3898, opened 2025-08-22
default to permanent notification if push is not possible

### LINKED ISSUE #3894: add permanent notification in case PUSH is not available
with upcoming calls, we need even more reliable notifications.

if PUSH is not available (always the case eg. for f-droid or when non-chatmail is used), we currently ask for "background permission", which, however, is [known to be often insufficient](https://dontkillmyapp.com).

additionally, we should enable the permantent notification. 

ppl knowing their systems good enough to know that this is not needed can disable the option at "Settings / Notifications" or hide the notification (the latter is kind of known and common nowadays)

cmp https://github.com/deltachat/deltachat-android/issues/3893

adbenitez:
close #3894

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/util/Prefs.java
+++ b/src/main/java/org/thoughtcrime/securesms/util/Prefs.java
@@ -30,6 +30,7 @@ public class Prefs {
 
   private static final String TAG = Prefs.class.getSimpleName();
 
+  public  static final String RELIABLE_SERVICE_PREF         = "pref_reliable_service";
   public  static final String DISABLE_PASSPHRASE_PREF          = "pref_disable_passphrase";
   public  static final String THEME_PREF                       = "pref_theme";
   public  static final String BACKGROUND_PREF                  = "pref_chat_background";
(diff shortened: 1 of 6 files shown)


src/main/java/org/thoughtcrime/securesms/util/Prefs.java
```diff
@@ -212,13 +214,36 @@ public static void setChatRingtone(Context context, int accountId, int chatId, U
...
+    DcAccounts dcAccounts = DcHelper.getAccounts(context);
+    for (int accountId : dcAccounts.getAll()) {
+      DcContext dcContext = dcAccounts.getAccount(accountId);
+      if (!dcContext.isChatmail()) {
+        return false;
+      }
+    }
+    return true;
+  }
+
```
r10s: we now have to `isAllChatmail`, maybe move it to `DcAccounts` and use it from there in all placed needed

src/main/java/org/thoughtcrime/securesms/util/Prefs.java
```diff
@@ -212,13 +214,36 @@ public static void setChatRingtone(Context context, int accountId, int chatId, U
...
+    try {
+      value = getBooleanPreference(context, key, false);
+    } catch(Exception e) {}
+    boolean value2;
     try {
-      return getBooleanPreference(context, "pref_reliable_service", false);
+      value2 = getBooleanPreference(context, key, !value);
+    } catch(Exception e) {
+      value2 = !value;
     }
```
r10s: took me a while to understand, what is ongoing here, esp. as the whole class is super-simple otherwise :)

this is the kind of code, no one understands when reading out of context in one year :)

maybe better start from scratch with a new variable, `pref_reliable_service2` that has defaults to true.

if the user has really disabled it for some reasons in the past, these reasons may be invalid by the upcoming calls which require more reliable notifications.

otherwise, to check if a key was ever set, `contains()` should do the job as well

adbenitez: contains() is exactly what I was looking for! this whole complex stuff is just a workaround because I was not aware there was an API to check if it already exists! :sweat_smile:

PR #3908, opened 2025-09-01
show warning if background notifications will be unreliable

### LINKED ISSUE #3893: show prominently, if notifications are known to maybe have issues
by upcoming calls, reliable notification become much more important.

therefore, we show prominently if notifications are known to be maybe throttled or even disabled, eg.

- notifications disabled by the user in the system settings
- PUSH notifications are not available in general, and permanent notification is not active
- PUSH not connected temporarily
- maybe user using old, encrypted database
- iOS may disable notifications also for battery/data saving mode, we should check if that or other things apply for android
- ...

the information should be shown as a single status line inside the  "Notification" activity. the status line should be prefixed by ⚠️

moreover, show the warning emoji ⚠️ instead of on/off in the settings overview

there is also the idea around to use Device Messages for that, however, we should not spam the chat with things that are more temporary. but eg. if core would detect that messages are delayed, that could be an additional option, adding a warning once or few times. but let's start with the settings

adbenitez:
close #3893

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/preferences/NotificationsPreferenceFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/preferences/NotificationsPreferenceFragment.java
@@ -28,6 +28,7 @@ import org.thoughtcrime.securesms.ApplicationPreferencesActivity;
 import org.thoughtcrime.securesms.R;
 import org.thoughtcrime.securesms.connect.DcHelper;
 import org.thoughtcrime.securesms.connect.KeepAliveService;
+import org.thoughtcrime.securesms.notifications.FcmReceiveService;
 import org.thoughtcrime.securesms.util.Prefs;
 
 public class NotificationsPreferenceFragment extends ListSummaryPreferenceFragment implements Preference.OnPreferenceChangeListener {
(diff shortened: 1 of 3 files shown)


src/main/res/values/strings.xml
```diff
@@ -1035,6 +1035,7 @@
     <string name="notify_name_only">Name only</string>
     <string name="notify_no_name_or_message">No name or message</string>
     <string name="notifications_disabled">Notifications disabled</string>
+    <string name="unreliable_bg_notifications">Push-notifications unavailable, please enable Force Background Connection for reliable notifications</string>
```
r10s:
```suggestion
    <string name="unreliable_bg_notifications">Enable \"Force Background Connection\" for reliable notifications</string>
```

src/main/java/org/thoughtcrime/securesms/preferences/NotificationsPreferenceFragment.java
```diff
@@ -239,11 +243,21 @@ private void initializeRingtoneSummary(Preference pref) {
...
   public static CharSequence getSummary(Context context) {
+    return getSummary(context, context.getString(R.string.on));
+  }
+
+  public static CharSequence getSummary(Context context, String defaultOnValue) {
     NotificationManagerCompat notificationManager = NotificationManagerCompat.from(context);
     if (Build.VERSION.SDK_INT < Build.VERSION_CODES.TIRAMISU || notificationManager.areNotificationsEnabled()) {
-      return context.getString(DcHelper.getContext(context).isMuted() ? R.string.off : R.string.on);
+      if (DcHelper.getContext(context).isMuted()) {
+        return "⚠️ " + context.getString(R.string.notifications_disabled);
```
r10s: i'd just show "Off" in this case. this is what the user wants, and the warning cannot be fixed

PR #3925, opened 2025-09-26
use auto-generated RPC bindings

adbenitez:
and get rid of duplicate json parsing lib dependency: gson

for now I only included classes and Rpc methods that are actually used, not all the available generated API

**NOTE:** this branch is the base of https://github.com/deltachat/deltachat-android/pull/3785 so we need to merge this branch first

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -471,6 +471,8 @@
     <string name="chat_no_messages">No messages.</string>
     <string name="chat_self_talk_subtitle">Messages to Self</string>
     <string name="archive_empty_hint">Archived chats will appear here.</string>
+    <!-- Action to add a message to "Saved Messages". The longer form (instead of "Save" only) is needed esp. on desktop to make clear this is not about saving a file to disk -->
+    <string name="save_message">Save Message</string>
     <string name="saved_messages">Saved Messages</string>
     <string name="saved_messages_explain">• Forward messages here for easy access\n\n• Take notes or voice memos\n\n• Attach media to save them</string>
     <!-- Should match "Saved" from "Saved messages" -->


link2xt:
Where is the generator?

https://github.com/chatmail/core/pull/4603 has an unfinished generator for Python.

adbenitez:
> Where is the generator?
> 
> [chatmail/core#4603](https://github.com/chatmail/core/pull/4603) has an unfinished generator for Python.

generator is at:
https://github.com/deltachat/dcrpcgen

It was actually inspired by your draft PR y plan to add python and golang generators as well

Hocuri:
Nice to get rid of yet another dependency! I think it's better if @r10s reviews, I have been a bit out of Android developments lately.

Would be interesting to measure the performance impact of this - maybe the new library is faster or slower than the old one?

adbenitez:
I added a section to the readme explaining how to generate the RPC bindings

r10s:
> Nice to get rid of yet another dependency!

+1

> I think it's better if @r10s reviews, I have been a bit out of Android developments lately.

well, and i am not into jsonrpc and python :) but yeah, i will try to make sense out of it

> Would be interesting to measure the performance impact of this 

iirc, the most often called jsonrpc is about getting reactions. might indeed be interesting to gather some data here, however, as most things are still using ffi and  as this PR does not replace ffi, it should not be a big deal

adbenitez:
> > Would be interesting to measure the performance impact of this
> 
> iirc, the most often called jsonrpc is about getting reactions. might indeed be interesting to gather some data here, however, as most things are still using ffi and as this PR does not replace ffi, it should not be a big deal


yeah, in theory just-so this now should be faster because jackson lib is said to be a bit faster than gson according to some benchmarks last time me and @Hocuri looked into it, but besides that this is the same as before and the only additional thing is APIs to start/cancel calls so there should be no deterioration

r10s:
thanks a lot for going ahead, @link2xt <3

PR #3929, opened 2025-10-03
add 'Save Message' string

r10s:
the string can be used for the action to put a message to 'Saved Messages'.

it will be useful esp. on desktop,
where plain 'Save' often has the mindset of 'Files', esp. for older ppl :)
the added noun tries to break that.

additionally, desktop can resort menu items,
but that alone would not solve the issue.

android/ios does not really have that issue as it is far less about "Files" there, still, they can use the new string at some point.

there are also thoughs about renaming that alltogether, eg. to bookmark, but that is a larger and more breaking change, also the messages are not just 'bookmarked' on purpose, to eg. survive auto deletion.

once this is merged, we can pick the new string on desktop (maybe give it a bit time to be translated)

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -471,6 +471,8 @@
     <string name="chat_no_messages">No messages.</string>
     <string name="chat_self_talk_subtitle">Messages to Self</string>
     <string name="archive_empty_hint">Archived chats will appear here.</string>
+    <!-- Action to add a message to "Saved Messages". The longer form (instead of "Save" only) is needed esp. on desktop to make clear this is not about saving a file to disk -->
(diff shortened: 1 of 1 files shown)


adbenitez:
(side topic: I think in desktop what would be more useful is to have icons in front of the menú options, similar to the iOS app, it is way too confusing and even know that thanks God it is "export attachment" again I get less confuse but click the other option now and then instead of exporting attachment, I am rarely saving messages to saved messages I think the forwarding would have been enough without extra menu option and think this adds extra complexity and is a mistake UX-wise)

adbenitez (approved):

r10s:
yeah, icons in front of "Save" and others would be a win as well - disk-symbol for export then :)

PR #3946, opened 2025-10-13
prepare 2.20.0

adbenitez:


Raiden-GH:
Isn't this already merged?

adbenitez:
> Isn't this already merged?

why to ask what you can see here?

r10s:
@Raiden-GH @adbenitez 2.20 indeed was sent around already some days ago, however, we sometimes create .apk from PR, for tests esp. this may happen even before merging.

apart from that, could as well have been also a git hickup or we merge in other ways (sometimes happening at core, where PR are closed, but things still merged) or whatnot. so, all in all, no unreasonable question :)

PR #3950, opened 2025-10-16
don't show badge-counter in app icon for permanent notification

### LINKED ISSUE #3944: New permanent notification get treated like a normal new message with indication from the system
- Android version: Android 16
- Device: Samsung Galaxy S25 Ultra
- Delta Chat version: 2.17.0 (play-prerelease)
- Expected behavior: permanent notification without permanent message indication (like it does the syncthing-fork with its permanent notification)
- Actual behavior: the permanent notification get treated like a new message so I have always a indication of new message (above the DeltaChat icon as example), which causes the thought there could be a new message
- Steps to reproduce the problem: having permanent notification (for classic mail accounts now by default which is technically great) active.

adbenitez:
close #3944

sadly, for existing installations the channel already exist and Android doesn't allow to modify existing channels, but new installations will not get the dot or badge counter in the app icon in the launcher or home screen for the permanent notification

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Delta Chat Android Changelog
 
+## Unreleased
+
+* don't show badge counter in app icon for the permanent background notification
+
 ## v2.22.0
 2025-10
 
@@ -17,7 +21,6 @@
 * show dialog if user has permanently denied camera permission and tries to take picture for group avatar
 * several small fixes and improvements
 * add experimental built-in calls
-* don't show badge counter in app icon for the permanent background notification
 * update to core 2.22.0
 
 ## v2.11.0


r10s:
can't we add a number to the channel name? iirc, we did that elsewhere as well

disadvantage is that pll who have hidden the nktification, need to do that as well - but that is probably not masses, they know what to do and are actionable

otherwise, it is not fixable for the most (for all existing installs)

adbenitez:
> can't we add a number to the channel name? iirc, we did that elsewhere as well
> 
> disadvantage is that pll who have hidden the nktification, need to do that as well - but that is probably not masses, they know what to do and are actionable
> 
> otherwise, it is not fixable for the most (for all existing installs)

just changing number won't do, old channel needs to be deleted or it will still hang around in system settings, the channel is only created for people without push notifications, I don't think it is "the most" and for all the new users it will be no problem, also this is a cherry on top of the ice cream, the notification icon is still there this is only to hide badge/dot in some launchers, users might still want to hide the notification anyways, which is more disturbing, which this will revert if we now change the channel

r10s:
> the channel is only created for people without push notifications, I don't think it is "the most" 

after some estimation it will be easily several 100k of users not using push notifications. and having a "badge counter" among them is not rare. also the feedback we got, point to that direction. many user do not know about what to do upon that, so it is not that easy to fix for them, also strongly depending on the OS, so we cannot even give an easy tip. 

i fear, we will continue getting reports about the "1"

> just changing number won't do, old channel needs to be deleted or it will still hang around in system settings

it is not hard to delete a notificationChannel. but even if, having smth hanging around deep in the system settings is better than having a "1" hanging around on the app icon

adbenitez:
this is now ready for merging, I changed the channel name and added a migration to delete old channel, I also tested that the dot/badge disappeared from existing installation after updating to this branch, and confirmed that if the old channel name doesn't exist at all (ex. in new installations) this doesn't even cause a warning, deletion of invalid/non-existing channel ID is just silently ignored, also does nothing in old android APIs that don't support notifications channels

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Delta Chat Android Changelog
 
+## Unreleased
+
+* don't show badge counter in app icon for the permanent background notification
```
r10s, line 5:
```suggestion
* fix: don't show badge counter in app icon for the permanent background notification
```

PR #3956, opened 2025-10-21
feat: Add cant_decrypt_outgoing_msgs stock string

Hocuri:
This string was translatable in Core, but apparently we forgot to add it to Transifex.

Or, do we even still want to make it translatable? Since few people even know their password, it might be fine to just always show this error message in English, and remove the stock string from core?

See also https://github.com/deltachat/deltachat-desktop/pull/5587.


r10s:
we meanwhile consider "classic email login" an advanced usecase. the share of these users using multi device _and_ doing that wrong might indeed be too small  to justify a translation - esp as the translation may be inaccurate or raise even more questions. we are in a complex area here that we avoided alltogether in chatmail

as we did not have the string translated all the years where "classic email loging" was recommended workflow, it seems fine to leave it english only

it is  an advanced error that we usually do not expect and we usually do not try to translate them (it may be even hard, eg errors may come from server we cannot control language wise)

PR #3964, opened 2025-10-28
update to new dcaccount scheme

### LINKED ISSUE #3963: Use dcaccount scheme with a domain name instead of URL
Once the core is updated with https://github.com/chatmail/core/pull/7360
this line
https://github.com/deltachat/deltachat-android/blob/bc372d5406bef7a5875db5b327e467f89a2f6b10/src/main/java/org/thoughtcrime/securesms/InstantOnboardingActivity.java#L127
should be changed to use `DCACCOUNT + providerHost` as the default QR code when the core is updated.

This will make it use no-HTTPS onboarding.

Desktop PR: https://github.com/deltachat/deltachat-desktop/pull/5618
iOS issue: https://github.com/deltachat/deltachat-ios/issues/2855

adbenitez:
close #3963 

- [x] wait for core with https://github.com/chatmail/core/pull/7360

#skip-changelog


link2xt:
Core 2.23.0 is tagged

PR #3967, opened 2025-10-28
remove deprecated stock string 72

adbenitez:
core is printing:

```
[accId=1] deltachat-ffi/src/lib.rs:313: Invalid stock message ID 72
```


adbenitez:
@r10s can we also remove the string altogether? didn't do it in case some other client is still using it but in theory they shouldn't

PR #3968, opened 2025-10-28
don't force circle shape for shared shortcuts

### LINKED ISSUE #3948: Chat icons are shown as squares when long-pressing on the app icon
This bug report was sent to delta@merlinux.eu, look for "Ronin" for the original message and follow-up questions.
 
- Android version: HyperOS 2.0 / Android 15
- Device: Xiaomi 14
- Delta Chat version: 2.11.0, but problem is present "since forever"
- Expected behavior: The chat icons are either round, or cut-off in a pretty way, instead of being filled up with black corners.
- Actual behavior: To display them as squares, the round chat icons get filled up with black corners.
- Steps to reproduce the problem: long press on the Delta Chat app on the home screen, until chat previews are shown
- Screenshots:

Image

adbenitez:
close #3948 

leave image unmodified as square, don't force circle shape so system can apply their icon shapes without some black areas caused by the cropped circle avatar, example in my system with "pebble" shaped icons:

image


r10s (approved):
lgtm, did not test - but screenshot looks nice <3

PR #3995, opened 2025-11-05
fix: Don't crash when QR code can't be loaded

Hocuri:
Before this PR, when a QR code couldn't be loaded because of https://github.com/chatmail/core/pull/7399, Delta Chat crashed.

Now, the button does nothing.

It's not perfect, but easy to do, and only about an edge case where the user tried out the experimental broadcast channels in the past, and now tries to add a new member to a broadcast channel created with an old version of DC.

The hollistic alternative would be to switch to RPC for getSecurejoinQrSvg, so that the UI can get a proper error message.

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 # Delta Chat Android Changelog
 
-## v2.24.0
-2025-11
+## Unreleased
 
 * Make it possible to invite members into a channel via a QR code,
   and make channels more secure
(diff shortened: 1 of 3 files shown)


src/main/java/org/thoughtcrime/securesms/qr/QrShowFragment.java
```diff
@@ -94,8 +94,12 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
         try {
             SVG svg = SVG.getFromString(fixSVG(dcContext.getSecurejoinQrSvg(chatId)));
             imageView.setSVG(svg);
-        } catch (SVGParseException e) {
+        } catch (Exception e) {
```
adbenitez, line 99: I think this is problematic, if parsing SVG fails for whatever limitation of the SVG library, the user can't share the invite link anymore, I think it is better if the SVG exception handling is kept as in the past and an additional catch-all branch is created afterwards to catch then any other unexpected error

PR #4005, opened 2025-11-07
tweak RegistrationActivity for chatmail profiles

### LINKED ISSUE #4004: wrong title for chatmail in "password and accout"
also email login hint should be hidden

adbenitez:
close #4004

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Delta Chat Android Changelog
 
+## Unreleased
+
+* Tweak advanced section and wording of some advanced options
+
 ## v2.25.0
 2025-11
 


r10s:
thinking it over, the `login_advanced_hint` hint makes sense for editing chatmail as well - the dialog is advanced, you shall not chant it to an account used otherwise and if you change it and it is classic, it might become unencrypted

but we should add the following hint to strings.xml above `login_advanced_hint`, so on changes we are aware of that:
```
    <!-- the login is shown for creating a classic account as well as for editing classic/chatmail accounts - both are advanced -->
```

adbenitez:
I was not sure if it is word mentioning that we changed a title of an advanced option or the icon toggle on a password field in an advanced option etc :thinking: 

(update: tried to explain in some easy language)

PR #4006, opened 2025-11-07
unify e-mail to email

r10s:
once merged, we must to `./scripts/tx-update-changed-sources.sh`

counterpart of https://github.com/deltachat/deltachat-pages/pull/1192

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -100,8 +100,8 @@
     <string name="cannot_save_file">Cannot save %1$s.</string>
     <string name="copied_to_clipboard">Copied to clipboard.</string>
     <string name="contacts_headline">Contacts</string>
-    <string name="email_address">E-Mail Address</string>
-    <string name="bad_email_address">Bad e-mail address.</string>
+    <string name="email_address">Email Address</string>
+    <string name="bad_email_address">Bad email address.</string>
     <string name="password">Password</string>
     <string name="existing_password">Existing Password</string>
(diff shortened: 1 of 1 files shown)


src/main/res/values/strings.xml
```diff
@@ -640,13 +640,13 @@
     <string name="welcome_chat_over_email">Secure Decentralized Chat</string>
     <string name="scan_invitation_code">Scan Invitation Code</string>
     <string name="login_title">Log In</string>
-    <string name="login_advanced_hint">This login is for advanced users:\n\n• Do not use an account you\'re using in another app.\n\n• Classic e-mail server allow chats without end-to-end encryption marked by a mail icon.</string>
+    <string name="login_advanced_hint">This login is for advanced users:\n\n• Do not use an account you\'re using in another app.\n\n• Classic email server allow chats without end-to-end encryption marked by a mail icon.</string>
```
link2xt, line 643: While we are changing it, shouldn't "mail icon" be called "envelope icon"?

link2xt: It is called "mail icon" in the FAQ though.

link2xt: But we can also put emoji here, like "✉️ icon".

r10s: i just tried - depending on the android version, the emoji is yellow or white, but does not really look like the actual thing.

and showing the real thing is some effort.

as it is advanced, it seems better to just show nothing here

PR #4007, opened 2025-11-07
remove deprecated/broken `send_autocrypt_setup_message` advanced preference

adbenitez:


r10s (approved):

r10s:
reasoning: while using classic email is totally fine,  asm is about "shared usage", which is getting unsupported, cmp https://delta.chat/en/help#classic-email

moreover, this removes one potential thing that a user can be tricked to do - which is worthful esp. as it is a very rarely used option

PR #4009, opened 2025-11-08
fix typo [fr]

sowdowdow:
Position sharing has a typo [french] for the reciever.

-> missing space


adbenitez:
cc @r10s

r10s:
thanks for the fixes, i added them to transifex, where we do our translations - changes on github's strings.xml are not persistent and would be overwritten, see https://delta.chat/en/contribute

(the changes here on github will be visible on next translation pull, usually before a release)

PR #4012, opened 2025-11-10
transport strings etc

r10s:
this PR changes the wording for classic email login towards "transport" and adds the strings needed for adding/removing transports eg. at https://github.com/deltachat/deltachat-desktop/pull/5680 - it's mostly the strings @nicodh already prototyped

cc @hpk42 @adbenitez @nicodh @link2xt - this refers to things recently discussed

  

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/RegistrationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/RegistrationActivity.java
@@ -127,7 +127,7 @@ public class RegistrationActivity extends BaseActionBarActivity implements DcEve
         ActionBar actionBar = getSupportActionBar();
         if (actionBar != null) {
             actionBar.setTitle(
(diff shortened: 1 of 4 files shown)


adbenitez (approved):
it is pretty long for  a title etc. more when translated to other languages, pretty sure it will not show fully on several phones,

wondering if the "classic" part could be strike, most people don't know or need to know that chatmail relays technically are also "just email" and what people know as "email" is what we call "classic email" 

then the hint below could be more explicit `use classic email as transport, this login is for advanced users: ...`

r10s:
well, we use the "classic email" term a lot, everywhere.

but yeah, it is comparable long, to mitigate it, i suggested a shorter form to the translators.

but +1 for picking up the term "classic email" also for the hint as `use classic email as transport, this login is for advanced users: ...` - then even if the title is truncated (that happens sooner than the button), full context is there

PR #4013, opened 2025-11-10
don't linkify numbers

adbenitez:
Android's default number linkifier regex is pretty bad, it is better to disable it, more now that link color is changed:

image

this is more consistent with desktop, in iOS long numbers are linkified, should probably just not be linkified there

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Change color of links in text messages
+* Do not linkify numbers
 * Tweak advanced section and wording of some advanced options
 
 ## v2.25.0


r10s:
>  in iOS long numbers are linkified, should probably just not be linkified there

not sure - if the detection is good enough, it is a handy to be able to tap in the number.

also not sure about not linkifying numbers on android, it is not hard to assume that ppl are using that and that it is handy

so, not sure about the PR as such - yeah, some numbers are more outstanding now, but maybe that is the better tradeoff than not being able to tap - or even copy - a phone number someone sent you

i would ~~tend to~~ close the PR

r10s:
also, phone  linkifier is different across android versions - eg. this is the rendering on android14:

Screenshot 2025-11-10 at 16 01 34

adbenitez:
in my experience I only had bad experiences with the number linkifying, it has been always annoying/inappropriate, I am not sure how often people are using Delta Chat to exchange phone numbers, that is something more common in ex. WhatsApp where that is the way to exchange contacts

but feel free to close this PR

PR #4017, opened 2025-11-11
only show three-dots if self is not in chat

adbenitez:
followup for #4014 

#skip-changelog


src/main/java/org/thoughtcrime/securesms/ConversationTitleView.java
```diff
@@ -73,7 +73,7 @@ public void setTitle(@NonNull GlideRequests glideRequests, @NonNull DcChat dcCha
     } else if (dcChat.isOutBroadcast()) {
       subtitleStr = context.getResources().getQuantityString(R.plurals.n_recipients, chatContacts.length, chatContacts.length);
     } else if( dcChat.isMultiUser() ) {
-      if (dcChat.canSend()) {
+      if (dcChat.isSelfInChat(dcContext)) {
```
r10s: while thinking about that we waste a database call (members are already read some lines above), i am wondering if we cannot improve the situation with left groups by doing the check only if there is only one member.

```suggestion
      if (chatContact.length > 1 || dcChat.isSelfInChat(dcContext)) {
```

that would show member count for left groups in the vast majority of cases - and for most cases, no database call is wasted.

same for the other condition.

could also become a dedicated function as `maybeInaccurateMembercount()` or so.

just some thoughts, i would also be fine with the existing approach.

adbenitez: the member count for chats in secure-join process will be zero, so we just need to wait for core

r10s: > the member count for chats in secure-join process will be zero, so we just need to wait for core

nope, it is "one" already now. when eg. offline, the person you scanned the QR code from will be added as a member

adbenitez: notice that I said **it WILL be zero** I know right now it is 1, but this will change in the next core release, cc @link2xt

r10s: i  see, but the suggested suggestion, `chatContact.length > 1 || dcChat.isSelfInChat(dcContext)` should still be fine, or?

or do you mean that we do not need to check for isSelfInChat() at all and only check for `chatContact.length > 0` to display "N members"?

adbenitez: yep, since the member count for incomplete secure-join groups will be zero, then we have a good enough way to determine/discriminate specifically when it is a secure-join process (except the edge special case of a group with zero members and then it is a good compromise to have ... there)

link2xt: If you have left the group and rejoining, the member count will not be zero.

adbenitez: oh darn it! ok then lets go with r10s's suggestion, at some point would be nice to have the s"ecure-join in progress in this chat" exposed so clients can display things in a reasonable way without being half-baked

adbenitez:
@r10s I changed the code, optimized it to use available member-list, and changed the logic to use `memberCount>1` as you suggested

r10s:
there should be follow-ups for android/iOS, i will take care for that.

PR #4019, opened 2025-11-12
unify multi-device into a single setting

### LINKED ISSUE #3829: Remove the option to delete messages "at once" for chatmail
We already delete messages at once for non-multidevice (bcc_self=0) case.
Setting the option to "at once" manually breaks multi-devices setups:
https://support.delta.chat/t/can-only-see-messages-on-one-device/3902

Seems there is no reason to have the option. We still need to offer a way for users to switch this back to "automatic" if they currently have the option set to "at once".

adbenitez:
close #3829 

this is an effort into simplifying/unifying the settings related to multi-device usage into a single setting/toggle that is easier to understand to the user than several indirect and more complex settings (delete from server + send copy to self)

image

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -317,7 +317,8 @@
     <string name="menu_scroll_to_bottom">Scroll to the Bottom</string>
     <string name="menu_scroll_to_top">Scroll to the Top</string>
     <string name="menu_help">Help</string>
-    <string name="what_is_webxdc">What is Webxdc?</string>
+    <!-- use the same term for "Apps" as elsewhere -->
+    <string name="what_is_webxdc">About Apps</string>
     <string name="privacy_policy">Privacy Policy</string>
     <string name="menu_select_all">Select All</string>
     <string name="select_chat">Select Chat</string>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Improve handling of video recoding
 * Tweak advanced section and wording of some advanced options
 * Expose new "Multi-device mode" option instead of "Delete from server" for chatmail profiles
+* Opened in-chat apps got a 'About Apps' menu item
 
 ## v2.25.0
 2025-11
--- a/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -296,6 +296,8 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
     if (itemId == R.id.menu_add_to_home_screen) {
       addToHomeScreen(this, dcAppMsg.getId());
       return true;
+    } else if (itemId == R.id.webxdc_help) {
+      DcHelper.openHelp(this, "#webxdc");
     } else if (itemId == R.id.source_code) {
       IntentUtils.showInBrowser(this, sourceCodeUrl);
       return true;
--- a/src/main/res/menu/webxdc.xml
+++ b/src/main/res/menu/webxdc.xml
@@ -2,10 +2,12 @@
 
 <menu xmlns:android="http://schemas.android.com/apk/res/android">
 
-    <item android:title="@string/add_to_home_screen"
-        android:id="@+id/menu_add_to_home_screen" />
     <item android:id="@+id/show_in_chat"
         android:title="@string/show_in_chat" />
+    <item android:title="@string/add_to_home_screen"
+        android:id="@+id/menu_add_to_home_screen" />
+    <item android:title="@string/what_is_webxdc"
+        android:id="@+id/webxdc_help" />
     <item android:title="@string/source_code"
         android:id="@+id/source_code" />
 


src/main/java/org/thoughtcrime/securesms/preferences/ChatsPreferenceFragment.java
```diff
@@ -81,11 +81,7 @@ public void onCreate(Bundle paramBundle) {
       autoDelServer.setOnPreferenceChangeListener(new AutodelChangeListener("delete_server_after"));
     }
     if (dcContext.isChatmail()) {
-      autoDelServer.setEntries(new CharSequence[]{getString(R.string.automatic), getString(R.string.autodel_at_once)});
```
link2xt, line 84: "automatic" string can probably be deprecated once not used on other platforms? Not sure how deprecation works in this case when you need to track if it is still used by other platforms. It can be mentioned at least in the translator comments that the string is not going to be used anywhere so they don't try to find what is the context for this string.

src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -110,6 +110,11 @@ public void onCreate() {
           e.printStackTrace();
         }
       }
+
+      // 12.11.25: this is needed until core starts ignoring "delete_server_after" for chatmail
```
link2xt: nitpick: I prefer ISO dates like 2025-11-12, otherwise there are a lot of ways to write a date with dots.

This code part will go anyway when isChatmail is deprecated.

src/main/res/values/strings.xml
```diff
@@ -803,8 +803,8 @@
     <string name="pref_imap_folder_warn_disable_defaults">If you change this option, make sure, your server and your other clients are configured accordingly.\n\nOtherwise things may not work at all.</string>
     <!-- deprecated -->
     <string name="pref_watch_sent_folder">Watch Sent Folder</string>
-    <string name="pref_send_copy_to_self">Send Copy to Self</string>
-    <string name="pref_send_copy_to_self_explain">Required when using this account on multiple devices.</string>
+    <string name="pref_send_copy_to_self">Multi-device Mode</string>
```
link2xt, line 817: This string is difficult to translate, "Multi-" prefix does not have a clear meaning. "Mode" sounds like if it is something that you can turn on and off such as "Airplane mode" or "Do not disturb mode" to get additional capabilities or disable some features.

I'd call it "Multiple devices" instead.

src/main/java/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
```diff
@@ -76,9 +76,9 @@ public void onCreate(Bundle paramBundle) {
       });
     }
 
-    bccSelfCheckbox = (CheckBoxPreference) this.findPreference("pref_bcc_self");
-    if (bccSelfCheckbox != null) {
-      bccSelfCheckbox.setOnPreferenceChangeListener((preference, newValue) -> {
+    multiDeviceCheckbox = (CheckBoxPreference) this.findPreference("pref_bcc_self");
+    if (multiDeviceCheckbox != null) {
+      multiDeviceCheckbox.setOnPreferenceChangeListener((preference, newValue) -> {
```
link2xt, line 81: Since this is now exposed to default chatmail profiles that most user use, it could use a warning message when turning the option off explaining that you should only turn it off if you actually don't use this profile on multiple devices. With "delete at once" there is a warning because it says it will now delete messages from the server. It's not a big deal that turning off this option will result in message deletion on chatmail server, but the text could explain that leaving this option off will result in not seeing your own messages on other devices, messages being delivered only on one device etc. Otherwise users will just switch it off and forget that they did.

src/main/res/values/strings.xml
```diff
@@ -803,8 +803,8 @@
...
+    <string name="pref_send_copy_to_self_explain">Required when using this profile on multiple devices.</string>
```
link2xt: This now just repeats the option name. Looking at the screenshot, this does not look good, the text below does not explain anything.

Could be something like "Send outgoing message copies and synchronization messages to your other devices."

"Required when using this profile on multiple devices." could go into a scary warning that is displayed when the option is turned off to save space.

link2xt (commented):
Technically looks good, i.e. it should work with chatmail and non-chatmail profiles.

But the translation text is better discussed with @r10s and the warning while can be added later is probably easier to add it already because then the text in the explanation does not need to tell that this is required for multiple devices.

src/main/res/values/strings.xml
adbenitez: the translations don't have to be literal, about just "multiple devices" I am not sure, let's wait for r10s if he has some suggestion or prefers "Multiple devices", anyways the setting has a subtitle explaining when it needs to be enabled

src/main/res/values/strings.xml
adbenitez: > "Send outgoing message copies and synchronization messages to your other devices."

this is too advanced for most users I would say :thinking: they don't need to know what the core is doing internally to make multi-device work

link2xt: It does not need to tell everything that the core does (sending sync messages, sending messages to self, enable deletion "at once" for chatmail etc.) but say at least something to explain what it does. Maybe "Send background messages for device synchronization. Required when using this profile on multiple devices." or so. Otherwise it says "Multi-device mode: needed for multi-device" basically.

The problem I see is that the setting is off by default, you can turn it on, nothing visibly changes and you have no idea what it does. This is annoying for the users who want to understand why the setting is needed and if it is working at all when looking through the settings. If the setting description at least says that it enables setting of some background/invisible messages, then user will get some idea of what it will do if they just enable it (waste their traffic and server quota) and not something else like "developer mode" that will add new features and options here and there. See also the forum topic: https://support.delta.chat/t/more-information-in-settings/4273

src/main/res/values/strings.xml
r10s: we are speaking of "multi-device" [already in the help](https://delta.chat/en/help), so that seems fine. 

but i would still add a translator hint, explaining what it is and that translations do not need to be literal

src/main/res/values/strings.xml
r10s: maybe use the subtitle to explain that it is turned on automatically as needed - that user do not think, they have to care about that. and also are not wondering "i **swear**, i have not enabled that!!!!!"

and also the other way round, say when it makes sense to turn it off again.

maybe say smth as:

```
Multi-Device Mode

Automatically enabled on adding a second device or on backup.
Can be disabled if using only one device and the relay runs full.
```

could be shorter, but this is the shortest i am able to do :)

adbenitez: I sympathize with tweaking the summary, I agree currently is not saying much, but talking about "the relay" will get outdated quite soon,

trying to come up with something better I get this from the top of my head:

"This is needed for the app to work properly when you use your profile in several devices. It will be auto-enabled when you add a second device or create a backup"

but not sure, feel free to directly commit string suggestions

link2xt: > Automatically enabled on adding a second device or on backup.
> Can be disabled if using only one device and the relay runs full.

This still does not explain what the setting does. Just saying that it is needed/required for the app to work properly is not a replacement for this.

I'd also not talk about the relay or quota, with multi-transport you will get multiple relays running full. We don't need to encourage user to disable it, it's fine to keep it enabled even if you have a single device.

Regarding automatic enabling see discussion in https://github.com/chatmail/core/pull/7440, it makes sense to explain it here. I'd not talk about backup, maybe we want to not do this later.

Another proposal:

Name: "Multi-Device Mode"
Description: "Send background messages for device synchronization. Automatically enabled on adding a second device."
Warning when disabling: "Multi-device Mode must be enabled when using the same profile on multiple devices. Only disable this setting if you have removed this profile from all your other devices. Disabling multi-device mode while using the profile on multiple devices will result in missed messages and other problems."

src/main/res/values/strings.xml
link2xt, line 817: This needs a comment saying that the string should not be translated literally, and can be translated e.g. as "Mulitple devices" and should be translated similarly to the FAQ. We have a section https://delta.chat/en/help#can-i-use-delta-chat-on-multiple-devices-at-the-same-time but FAQ translations are currently outdated.

adbenitez:
I added a confirmation dialog, just re-used the existing warning from other advanced settings feel free to suggest something or push a string directly

(EDIT: just saw the latest recommendation in the comments, I pushed with the suggested behavior and strings)

src/main/java/org/thoughtcrime/securesms/preferences/ChatsPreferenceFragment.java
adbenitez: automatic string is still used in advanced configuration for IMAP/SMTP security

src/main/res/values/strings.xml
r10s: ftr, i added the comment in a subsequent PR at https://github.com/deltachat/deltachat-android/pull/4031

PR #4021, opened 2025-11-13
tune down display of phone-number-links

### LINKED ISSUE #4020: A lot of numbers are turned into phone links
Since https://github.com/deltachat/deltachat-android/pull/4011 is merged it is much more visible that phone number highlighting selects way more numbers than needed:

Image

Clicking on the numbers opens the call app. On desktop there is no such problem because it does not try to highlight phone numbers.

r10s:
phone numbers are recognized too eagerly on some android, hilight them using standard color.

ftr, on my tests in the emulator, i could not reproduce the issue @link2xt and @adbenitez  pointed out at all - **stock android5 as well as stock android13** seem render the string fine (only one hilite, with 5+ characters, yes also questionable, but far away from the tens of hilights in the issue #4020)

maybe **some android derivate** (CalyxOS?) tries to "improve" make phone number highlighting here?
if this turns out to be true, the issue is even smaller than initially though, and more a bug of the android derivate

closes #4020 

#skip-changelog as the original change was not released yet


link2xt:
My screenshot is from Android 16.

adbenitez:
I think the actual problem should be fixed in CalyxOS, can someone confirm this also happens in signal? if it is something that happens system-wide there is no point on doing hacks in delta chat that deteriorate things for other people, CalyxOS and the like are responsive fixing things and providing new releases/updates for users, in the past we had a problem with the system camera and they fixed it

hpk42:
On Tue, Nov 18, 2025 at 11:51 -0800, adb wrote:
> adbenitez left a comment (deltachat/deltachat-android#4021)
> 
> I think the actual problem should be fixed in CalyxOS, can someone confirm this also happens in signal? if it is something that happens system-wide there is no point on doing hacks in delta chat that deteriorate things for other people, CalyxOS and the like are responsive fixing things and providing new releases/updates for users, in the past we had a problem with the system camera and they fixed it

FTR CalyxOS is in limbo.  They told all users to transition to a different system, while they are revamping everything.

link2xt:
GrapheneOS has the same issue and we still don't understand where it comes from because neither CalyxOS nor GrapheneOS have a custom repo for the code involved. Could be libphonenumber pulls in data about phone numbers that is not available on builds without blobs from Google or Google services.

link2xt:
#4020 is already closed, so this workaround is not needed.

PR #4022, opened 2025-11-14
 feat: Add the possibility to withdraw channel invite links and QR codes 

Hocuri:
Add the ability to withdraw broadcast invite codes; companion PR for [#7439](https://github.com/chatmail/core/pull/7439)

#skip-changelog because withdrawing will only work when updating core


Hocuri:
Actually, this doesn't use any new core API, so, can be merged already

PR #4023, opened 2025-11-14
feat: Rephrase "Establishing end-to-end encryption" -> "Establishing connection"

Hocuri:
Companion PR for https://github.com/chatmail/core/pull/7444

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -1030,7 +1030,7 @@
     <string name="secure_join_started">%1$s invited you to join this group.\n\nWaiting for the device of %2$s to reply…</string>
     <!-- placeholder will be replaced by the name of the inviter. -->
     <string name="secure_join_replies">%1$s replied, waiting for being added to the group…</string>
-    <string name="secure_join_wait">Establishing end-to-end encryption, please wait…</string>
+    <string name="secure_join_wait">Establishing connection, please wait…</string>
     <string name="contact_verified">%1$s introduced.</string>
     <!-- Shown in contact profile. The placeholder will be replaced by the name of the contact that introduced the contact. -->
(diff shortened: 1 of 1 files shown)


src/main/res/values/strings.xml
```diff
@@ -1030,7 +1030,7 @@
     <string name="secure_join_started">%1$s invited you to join this group.\n\nWaiting for the device of %2$s to reply…</string>
     <!-- placeholder will be replaced by the name of the inviter. -->
     <string name="secure_join_replies">%1$s replied, waiting for being added to the group…</string>
-    <string name="secure_join_wait">Establishing end-to-end encryption, please wait…</string>
+    <string name="secure_join_wait">Establishing connection, please wait…</string>
```
adbenitez, line 1033: don't we have the inviter name already as for groups? why not to do it as for groups? (but I agree this is already an improvement)

r10s: we have the name already, but it is already shown in the title, so not sure if redundancy is needed or wanted here. in contrast to group-joins it is pretty clear to whom the connection is established.

but yeah, let's leave that up for a possible future iteration

r10s:
merging that in to get the string translated

src/main/res/values/strings.xml
adbenitez: @r10s I am talking about **Channels** here not 1:1 chats, now that you mention it, then the problem is that channels are just using the same string as 1:1 chats instead of having a dedicated string like groups do

r10s: ah, yes, for channels, it makes sense to point to the inviter device as well.

and i agree, that probably justifies a new string - we duplicated so many strings for channels, so that super-user-facing info should be explicit about channels as well

PR #4025, opened 2025-11-16
use app-term instead of webxdc-term;  add "About Apps" to three dot menu

r10s:
~~it is used when an in-app is on the screen (on desktop), however, instead of helping, immediately raises another question, as the user just did not know about webxdc at all at that point (we are not using the term in the apps at all).~~
~~better just use the string 'Help' and open
offline help at #webxdc - the entry is also just improved :)~~

EDIT: this PR now changes the string and adds the corresponding entry to android:

Screenshot 2025-11-16 at 12 43 41

will do a issue/PR for desktop to open the offline help also there

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 * Do not show "1 member" when the process of joining the group is not finished
 * Improve handling of video recoding
 * Tweak advanced section and wording of some advanced options
+* Opened in-chat apps got a 'About Apps' menu item
+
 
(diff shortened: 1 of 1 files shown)


adbenitez (approved):

r10s:
@adbenitez i changed the PR, keeping smth as "About Apps" - only help seems to be ambiguous here - help about tetris in the image above?

adbenitez (approved):

PR #4031, opened 2025-11-17
streamline Multi-Device Mode wording

r10s:
came over that while filing issues for iOS/desktop ...

we are saying Real-Time as well, so i think that should be upper cased as well, in general we upper case most key words

more tricky is the other change, i think "send background messages" may be too confusing, also hard to translate. user may think, real messages are exchanged, maybe smth visible, maybe even some action is needed "Send".

not totally sure if "Exchange data" is better, but want to bring up the idea. i will not die on that hill :)

successor of https://github.com/deltachat/deltachat-android/pull/4019

cc @adbenitez @link2xt

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -818,7 +818,7 @@
     <string name="pref_watch_sent_folder">Watch Sent Folder</string>
     <!-- No need to be literal here, you can also use "Use Multiple Devices", "Support Multiple Devices" or other fitting terms. However, it should fit to the wording or your language at https://delta.chat/help  -->
     <string name="pref_send_copy_to_self">Multi-Device Mode</string>
-    <string name="pref_send_copy_to_self_explain">Exchange data for device synchronization. Automatically enabled on adding a second device</string>
+    <string name="pref_send_copy_to_self_explain">Synchronize your messages with your other devices. Automatically enabled on adding a second device</string>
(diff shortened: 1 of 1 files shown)


link2xt:
> i think "send background messages" may be too confusing, also hard to translate

"Background" has a translation in the android UI, it in common to download updates from google play in background: https://support.google.com/googleplay/answer/1663315

Exchange data sounds like it may use some other way than the one used to send messages, e.g. not relay but p2p, google cloud, bluetooth and so on.

r10s:
> "Background" has a translation in the android UI, it in common to download updates from google play in background: https://support.google.com/googleplay/answer/1663315

thanks for the link!

that speaks of "background data" 

what about: "Use background data for device synchronization"? that'd also be nice in german, "Hintergrunddaten für Gerätesynchronisation verwenden" - and it avoids "send messages" wording

"data" still it makes clear, or even clearer, that this is smth eating your data plan or filling server

adbenitez:
> what about: "Use background data for device synchronization"?

even to myself this kind of already tells me nothing, "what is background data??" etc. at this point I think it is better not to explain the means for what core do the synchronization, (background, data, messages, maybe p2p device connection in the future etc) and just tell what it actually does "if enabled this device will synchronize messages (or state?) with other devices" or something like that

r10s:
i would also like smth like that better.

maybe smth. as 

```
Multi-Device Mode
Synchronize your messages with your other devices. Automatically enabled on adding a second device
```

if that is background message or data or whatever does not matter.

wrt ["both tell the same"](https://github.com/deltachat/deltachat-android/pull/4019#discussion_r2521739017): kind of true, in theory, the term "Multi-Device Mode" could also be removed then, and the option could read "Synchronize your messages with your other devices". but for referring to it, already in the warning, it may still be handy - and shorter

link2xt (approved):

PR #4035, opened 2025-11-18
customize phone number regex in Linkifier

### LINKED ISSUE #4020: A lot of numbers are turned into phone links

adbenitez:
close #4020 

the regex is the same from java source code (`Patterns.PHONE`) with some small tweaks to avoid some weird links in the middle of text or short too numbers, if the number is short also the value of linkifying is low, as they are easily to memorize and type directly

alternative to #4021

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/connect/DcHelper.java
+++ b/src/main/java/org/thoughtcrime/securesms/connect/DcHelper.java
@@ -159,6 +159,11 @@ public class DcHelper {
     dcContext.setStockTranslation(118, context.getString(R.string.secure_join_replies));
     dcContext.setStockTranslation(119, context.getString(R.string.qrshow_join_contact_hint));
 
+    // HACK: svg does not handle entities correctly and shows `&quot;` as the text `quot;`.
+    // until that is fixed, we fix the most obvious errors (core uses encode_minimal, so this does not affect so many characters)
+    // cmp. https://github.com/deltachat/deltachat-android/issues/2187
+    dcContext.setStockTranslation(120, context.getString(R.string.qrshow_join_group_hint).replace("\"", ""));
+    dcContext.setStockTranslation(121, context.getString(R.string.connectivity_not_connected));
     dcContext.setStockTranslation(124, context.getString(R.string.group_name_changed_by_you));
     dcContext.setStockTranslation(125, context.getString(R.string.group_name_changed_by_other));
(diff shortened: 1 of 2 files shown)


adbenitez:
just as documentation, I am not sure we should merge this, it seems most stock phones provide their own (proprietary?) implementation that linkifies correctly, while if you just use the `Patterns.PHONE` regex in the AOSP the results are much worse, it seems that is what CalyxOS uses and that is why there is much more crazy nonsensical linkification, 

while I took the official regex and improved it, there is still edge cases like https://some-555555-thing.com if you click in the number it will open the phone calling instead of the link

src/main/java/org/thoughtcrime/securesms/util/Linkifier.java
```diff
@@ -33,8 +41,12 @@ public static SpannableString linkify(SpannableString messageBody) {
       replaceURLSpan(messageBody); // replace URLSpan so that it is not removed on the next addLinks() call
     }
 
+    if (Linkify.addLinks(messageBody, PHONE_PATTERN, "tel:", (s, start, end) -> s.subSequence(start, end).toString().replaceAll("[\\- \\.]", "").length() >= 5, null)) {
```
link2xt:
```suggestion
    if (Linkify.addLinks(messageBody, PHONE_PATTERN, "tel:", Linkify.sPhoneNumberMatchFilter, Linkify.sPhoneNumberTransformFilter)) {
```

If this works, we can as well use these standard match and transform filters.

adbenitez: i used it, but i dont see any difference with my previous results, still http://aaa-555555-bbb.com gets the number as link, 0.9999 is also linkified while this doesn't happen in a stock xiaomi

adbenitez:
@link2xt @r10s great news! came up with a new idea! now the custom linkifier is only used if it is detected that the system linkifier is messed up (by testing that it converts to phone number the string "a100b") so we can keep the good linkifiers in the systems that already have them and apply our enhanced custom linkifier only when needed as a best effort

with this now this PR has much better outcome than https://github.com/deltachat/deltachat-android/pull/4021

link2xt (approved):
Tested, this is how it looks like now for me:
Screenshot with a fix

Issue #3970, opened 2025-10-29
use list_transports


r10s:
- [x] as a preparation, get rid of all oauth2 things; they are unused and clutter code #3977 
- [x] remove legacy ["companion app"](https://github.com/deltachat/deltachat-android/pull/2088) code #3981 
- [x] to get prepared for mutli-transport, use `list_transport` instead of `CONFIG_` in RegistrationActivity (#4039)
- [x] rename `RegistrationActivity` to `EditTransportActivity` (#4046)
- [x] add a `ListTransportActivity` similar to the proxy view that allows removing transport and selecting default transport (#4080)
- [x] adding transport, for now, should be done by scanning an DCACCOUNT:-QR code from the normal scanner (instead of adding a new profile) #4080
- [x] changing the address should not be possible in the settings of the transport. If you want to change the address, then create a new transport and remove old one.

cc @link2xt does that make sense for the current state of multi-transport? otherwise, feel free to edit this issue directly :)

PR #4037, opened 2025-11-18
add full rpc bindings

adbenitez:
this adds the `listTransports()` RPC API required by #3970 

#skip-changelog


Hocuri (approved):
Makes sense, and is nicer than needing to cherry-pick individual functions all the time

r10s:
background was also the addition of ProGuard, which now removes dead code

PR #4039, opened 2025-11-19
use listTransports

adbenitez:
part of #3970

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/RegistrationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/RegistrationActivity.java
@@ -122,9 +122,7 @@ public class RegistrationActivity extends BaseActionBarActivity implements DcEve
         try {
             List<EnteredLoginParam> relays = rpc.listTransports(accId);
             if (!relays.isEmpty()) config = relays.get(0);
-        } catch (RpcException ignored) {
-            Log.e(TAG, "listTransport failed: ", e);
-        }
+        } catch (RpcException ignored) {}
 
         ActionBar actionBar = getSupportActionBar();
         if (actionBar != null) {


src/main/java/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -122,19 +118,23 @@ public void onCreate(Bundle bundle) {
             startActivity(new Intent(this, ProxySettingsActivity.class));
         });
 
-        boolean isConfigured = DcHelper.isConfigured(getApplicationContext());
+        EnteredLoginParam config = null;
+        try {
+            List<EnteredLoginParam> relays = rpc.listTransports(accId);
+            if (!relays.isEmpty()) config = relays.get(0);
+        } catch (RpcException ignored) {}
```
r10s: why not do a `Log.e(TAG, "listTransport failed: ", e);` as in similar cases?

src/main/java/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -122,19 +118,23 @@ public void onCreate(Bundle bundle) {
             startActivity(new Intent(this, ProxySettingsActivity.class));
         });
 
-        boolean isConfigured = DcHelper.isConfigured(getApplicationContext());
+        EnteredLoginParam config = null;
+        try {
+            List<EnteredLoginParam> relays = rpc.listTransports(accId);
+            if (!relays.isEmpty()) config = relays.get(0);
```
r10s, line 124: for understanding, in the future, this will not always be `0` but the actually edited transport

src/main/java/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -430,6 +418,20 @@ private EnteredCertificateChecks certificateChecksFromInt(int position) {
...
+    private int certificateChecksToInt(EnteredCertificateChecks check) {
+        if (check == null) return 0;
+
+        switch (check) {
+            case automatic:
+                return 0;
+            case strict:
+                return 1;
+            case acceptInvalidCertificates:
+                return 2;
```
r10s, line 428: ftr, `acceptInvalidCertificates2` (comment `// 3 is a deprecated alias for 2` above) is handled by the jsonrpc bindings, i double checked that :)

src/main/java/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -430,6 +418,20 @@ private EnteredCertificateChecks certificateChecksFromInt(int position) {
...
+        }
+        throw new IllegalArgumentException("Unknown EnteredCertificateChecks type: " + check);
```
r10s: hm, where is this error caught? this can happen when jsonrpc adds a new value and the check is not exhaustive therefore. 

or can we somehow force java to use exhaustive enums? then a throw would not be required

src/main/java/org/thoughtcrime/securesms/RegistrationActivity.java
```diff
@@ -444,6 +446,22 @@ public static Socket socketSecurityFromInt(int position) {
...
+        case automatic:
+            return 0;
+        case ssl:
+            return 1;
+        case starttls:
+            return 2;
+        case plain:
+            return 3;
+        }
+        throw new IllegalArgumentException("Invalid socketSecurity type: " + security);
```
r10s: same question here

src/main/java/org/thoughtcrime/securesms/RegistrationActivity.java
adbenitez: this is bound to immediately change with my next PRs, was considering adding a // TODO there but in the end deemed it not even worth doing

src/main/java/org/thoughtcrime/securesms/RegistrationActivity.java
adbenitez: mmmh, actually it is impossible to have a value that we don't support, rpc parsing will just fail if it is a new value that it can't convert to the Enum, (or return null otherwise), so that condition is never reached, unless we add support for the new value (update auto-generated bindings) but forget to handle the new type, but even then it is impossible to set the unsupported value as well so it should never happen that that line is reached

src/main/java/org/thoughtcrime/securesms/RegistrationActivity.java
adbenitez: mainly because this code will be moved out to another activity that is the one that will be calling listTransports and passing a single transport to this activity, so adding TAG here would become unused quickly

PR #4043, opened 2025-11-20
remove all png files before help creation

r10s:
this PR removes all `*.png` files from `assets/help` dir; needed/new files are copied from the help generation script directly afterwards.

(this will remove the old comic and the old padlock)

(cleaning whole dir comes to mind as well, but there is *.css in, the dir is filled by both, android and pages, so this line fixes the issue at hand)

### DIFF
--- a/scripts/create-local-help.sh
+++ b/scripts/create-local-help.sh
@@ -1,1 +1,2 @@
+rm src/main/assets/help/*.png
 ../deltachat-pages/tools/create-local-help.py ../deltachat-pages/result src/main/assets/help


Hocuri:
Specifically, there were a lot of warnings printed when running `scripts/create-local-help.sh`

r10s:
> btw, recently there was a problem generating help, some old translations still included the old hand-drawn image

this is fixed now (help was pushed to translators only recently)

> Specifically, there were a lot of warnings printed when running scripts/create-local-help.sh

they come from checking if the links are working - github is recently unhappy about the ~20 requests in short time. no idea what to do about that, maybe only a temporary hickup

PR #4044, opened 2025-11-20
feat: Add missing core stock strings

Hocuri:
I added all core stock strings from `stock_str.rs` >= 178 that were missing previously; most importantly, all the channel and statistics strings.

Some of these are not in a core release yet, but this is not a problem; core only prints a warning if we set an unknown string.

While there are some other strings in `stock_str.rs` that are missing here, it's not clear that these should be translated; probably we can just remove these from core.

I also moved strings 120 and 121 to the correct place.

#skip-changelog


adbenitez (approved):
thanks a lot!

PR #4045, opened 2025-11-20
allow to open all links (asking before with dialog)

### LINKED ISSUE #4054: in webxdc, open dialog on clicking `https:` etc.
currently, when clicking an `https:`-link in an webxdc, nothing happens.

however, this is too strict and eliminates a lot of legitimate usecases, as having an "about" page, a "donate" link - user-entered links in an editor or spreadsheet.

there are lots of ideas around about whitelisting links somehow or even checking the code, after lots of internal discussion, however, it seems fine to show an dialog as 

```
Do you want to open this link?

https://this-is-some-link/?skdhfkh+foo+bar

[ Cancel ]     [ Copy ]  [ Open ]
```

when targeting this issue, 

- [x] add the dialog on opening https: links - show the raw ASCII domains, and not some unicode representation [^1] 
- [x] consider which other protocols we want to support (https://github.com/deltachat/deltachat-android/pull/4066)
- [x] make sure, the links cannot be opened silently without the dialog being shown
- [x] adapt the testing tool at https://github.com/webxdc/webxdc-test to trigger all protocols we want to support

[^1]: showing plain ascii is needed to avoid [homograph attack](https://en.wikipedia.org/wiki/IDN_homograph_attack) - eg. the "а" in `wikipediа.org` is _not_ an ASCII-`a` this kind of stuff is used to trick users. by showing raw ASCII punycode, this is discoverable, `wikipediа.org` would be shown as `xn--wikipedi-86g.org`

adbenitez:
**UPDATE:** actually this PR now allows opening all links and showing a dialog before the app proceeds to open them
close #4054

image

## Old description
add geo: support and support for i.delta.chat links

geo-urls are URLs with coordinates that are understood by map apps to display a location in maps, ex `geo:123,124`
recently got a user that was creating a "find vegan restaurants" webxdc requesting this feature because then the restaurant links doesn't work/open in the user's map app

the i.delta.chat links are similar to the old `OPENPGP4FPR:` we already allow, we just didn't get to update the clients to support the new format, this fixes that

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/LocalHelpActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/LocalHelpActivity.java
@@ -15,9 +15,6 @@ public class LocalHelpActivity extends WebViewActivity
   public static final String SECTION_EXTRA = "section_extra";
 
   @Override
-  protected boolean shouldAskToOpenLink() { return false; }
-
-  @Override
   protected boolean allowInLockedMode() { return true; }
 
   @Override
--- a/src/main/java/org/thoughtcrime/securesms/WebViewActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/WebViewActivity.java
@@ -37,7 +37,7 @@ public class WebViewActivity extends PassphraseRequiredActionBarActivity
 
   protected WebView webView;
 
-  protected boolean shouldAskToOpenLink() { return true; }
+  protected boolean shouldAskToOpenLink() { return false; }
 
   protected void toggleFakeProxy(boolean enable) {
     if (WebViewFeature.isFeatureSupported(WebViewFeature.PROXY_OVERRIDE)) {
--- a/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -339,9 +339,12 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
     }
   }
 
+  @Override
+  protected boolean shouldAskToOpenLink() { return true; }
+
   // This is usually only called when internetAccess == true or for mailto/openpgp4fpr scheme,
   // because when internetAccess == false, the page is loaded inside an iframe,
-  // and WebViewClient.shouldOverrideUrlLoading is not called for HTTP(S) links inside the iframe
+  // and WebViewClient.shouldOverrideUrlLoading is not called for HTTP(S) links inside the iframe unless target=_blank is used
   @Override
   protected boolean openOnlineUrl(String url) {
     Log.i(TAG, "openOnlineUrl: " + url);


r10s:
imu, this is not only about classic `<a href>` links the user has tapped, but also javascript could fire these urls without user action with arbitrary data (geo is not only lat/lng), invoking other android apps that register to the protocol.

not sure if that is a problem, it was arguably with openpgp4fpr always like that.

apart from that, the comment above openOnlineUrl() is outdated - and discussion at https://github.com/deltachat/deltachat-desktop/issues/5750 needs to be resolved before merging

in general, last time we discussed at bb33, there was the very clear idea to show an dialog for all links the user has tapped, and then allow opening them. that could include geo: as well - it was unclear in that discussion, if that is doable on android, but maybe check that direction first?

adbenitez:
showing a dialog is nice, but geo urls are even safer than the other urls we already allow so this PR doesn't make things worse for that super hypothetical case when someone first trick you into installing a malicious map app (or handling the other type of links)

could be merged as is, it is already an improvement

the dialog stuff can come in a separated PR on-point

adbenitez:
> and discussion at deltachat/deltachat-desktop#5750 needs to be resolved before merging

I don't think that is a blocker, this can be merged and worse case as result of that discussion, another pr needs to be done anyways because even if this one doesn't get merged, we already have in main code that accepts invite links in the old format that is the exact same situation as allowing i.delta.chat links

Simon-Laux:
I doubt that https://en.wikipedia.org/wiki/Geo_URI_scheme can be reliably used to do damage, for that the attacker would need to know what app the victim user opens those links in and know some exploit in the parsing logic of `geo:` URIs in that maps app.
It's not free form links that can link to a collection/recording address and directly embed user data into the link.
But if we are paranoid we could also add an API to core to sanitize the geo links before opening them, or we ask every-time in which supported map app the user wants to open the link, with an option to choose a default.

r10s:
yeah, the PR is probably fine

thanks for analysis, @Simon-Laux and @adbenitez

r10s:
EDIT: reopened, i was too quick on closing, i agree, that it is mostly unrelated to https://github.com/deltachat/deltachat-android/pull/4054 - sorry for the additional noise

r10s:
ah, during testing i realize that openOnlineUrl is also used for other things (eg. hardcoded delta chat homepage link in help), we do not want to have the new AlertDialog there, but really only for webxdc

we probably want to use IntentUtils.showInBrowser() for these links.

adbenitez:
> ah, during testing i realize that openOnlineUrl is also used for other things (eg. hardcoded delta chat homepage link in help), we do not want to have the new AlertDialog there, but really only for webxdc
> 
> we probably want to use IntentUtils.showInBrowser() for these links.

I also realized this but intentionally kept it, there were people complaining already of links opening suddenly without warning in the (otherwise offline) help, I think it doesn't hurt to show the dialog before proceeding to opening the external links in this case, this also includes the incoming html messages etc.

but I am also fine with further adapting and adding an exception just for the offline help if you think it is better that way

r10s:
let's do not add the warning in front of internal and existing links that get opened. 

the PR should only change behaviour or webxdc

we usually avoid these kind of warnings wherever possible, and here it is possible. also, it is already kind of decided to go for an arrow as on desktop/iOS instead -

adbenitez:
I modified it so the confirmation dialog only appears in webxdc and is opened directly otherwise, I think it did make sense to show the link for html emails, but that is a discussion for another day I guess, and a bit outside the main target of this PR

PR #4050, opened 2025-11-21
garbage-collect storage of deleted webxdc apps

### LINKED ISSUE #4049: Clear unused local storage for Webxdcs
Android has APIs to list all origins, how much data they use and clearing them: https://developer.android.com/reference/android/webkit/WebStorage.html

Delta Chat should from time to time garbage-collect origins that do not have any corresponding webxdc instance, by going through all the origins and then converting them to webxdc instance ID. Origins are constructed deterministically for webxdc message IDs, so should be possible to convert them back. If some origin does not have a corresponding webxdc, it can be cleared.

adbenitez:
close #4049 

- [x] wait for a new core with https://github.com/chatmail/core/pull/7487 to be released and use the new api

to completely fix the problem, it requires a new core with https://github.com/chatmail/core/pull/7531 but can be merged already

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/webxdc/WebxdcGarbageCollectionWorker.java
+++ b/src/main/java/org/thoughtcrime/securesms/webxdc/WebxdcGarbageCollectionWorker.java
@@ -12,7 +12,6 @@ import chat.delta.rpc.RpcException;
 import com.google.common.util.concurrent.ListenableFuture;
 
 import java.util.Collections;
-import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import org.thoughtcrime.securesms.connect.DcHelper;
@@ -43,20 +42,11 @@ public class WebxdcGarbageCollectionWorker extends ListenableWorker {
           return;
         }
 
-        List<Integer> accounts;
-        try {
-          accounts = rpc.getAllAccountIds();
-        } catch (RpcException e) {
-          Log.e(TAG, "error calling rpc.getAllAccountIds()", e);
-          completer.set(Result.failure());
-          return;
-        }
-
         for (Object key : origins.keySet()) {
           String url = (String)key;
           Matcher m = WEBXDC_URL_PATTERN.matcher(url);
-          int accId = m.matches()? Integer.parseInt(m.group(1)) : 0;
-          if (accId != 0 && accounts.contains(accId)) {
+          if (m.matches()) {
+            int accId = Integer.parseInt(m.group(1));
             int msgId = Integer.parseInt(m.group(2));
             try {
               if (rpc.getExistingMsgIds(accId, Collections.singletonList(msgId)).isEmpty()) {


src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -218,6 +221,44 @@ public void onReceive(Context context, Intent intent) {
...
+              }
+            } else { // old webxdc URL schemes, etc
+              webStorage.deleteOrigin(url);
+              Log.i(TAG, String.format("Deleted unknown origin: %s", url));
+            }
+          }
+
+          Log.i(TAG, "Done running Webxdc storage garbage collection.");
+        });
+        Util.sleep(60*60*1000); // 1h
```
link2xt: `while` loop with sleep is not a good option for scheduling tasks. It will run at the app start and slow down first start after update or reboot. If you run multiple tasks this way they will tend to synchronize over time.

Android has dedicated APIs for this: https://developer.android.com/reference/androidx/work/PeriodicWorkRequest
It can also have constraints like requiring that battery is not low etc.

adbenitez: I knew about worker, but wanted to keep it simple, the worker seems synchronous while the WebStorage api is async receiving callback, worker doWork() would exit while the callback is still not executed, there probably is some hack but not sure it is worthy, other tasks could go in worker but this one is not worth doing while the app is killed, btw the worker stuff requires a dependency, currently we have it only for a single worker that is not always used it, it could be potentially removed at some point so I didn't want to add more dependency on that lib, also because google is always changing apis and adding constrains, maybe then they decide to limit the worker similar to what happens with services already, if they are not doing it already, the the worker gets ratelimited and whatnot

adbenitez: to avoid potential initial congestion, it could probably sleep for 10 seconds

adbenitez: UPDATE: actually right in the Worker/Tasks docs they point to just use java threads (or kotlin corutines) in the "asynchronous work" section: https://developer.android.com/develop/background-work/background-tasks#asynchronous-work)

link2xt: "Asynchronous work" is about doing something while in foreground without blocking the UI, reloading chatlist etc.

As for the worker, https://developer.android.com/develop/background-work/background-tasks/persistent/threading/worker says that ListenableWorker should be used in this case. It has startWork() method that returns a future that can be set from the callback.

adbenitez: that requires a new dependency ([councurrent-futures](https://developer.android.com/jetpack/androidx/releases/concurrent#declaring_dependencies)) just for this simple task, I think it is not worthy

adbenitez: and once again, we don't need to wake up the app to do this work/task, which is what workers do, and the way we use it in other places is to wake up and do a background fetch, for this chore it is actually better if is run only when the app is in foreground

link2xt: Is it a large dependency? It's not some third-party dependency, just google's backport.

The whole source code of it is here, it is just a few small java classes: https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-main/concurrent/concurrent-futures/

We can also get rid of Signal code implementing the same thing here then:
https://github.com/deltachat/deltachat-android/tree/9a6abed1e4f4f23b910f2bf168c311d569b0416a/src/main/java/chat/delta/util

link2xt: > we don't need to wake up the app to do this work/task, which is what workers do

What is the disadvantage of waking the app, likely when the phone is not used, and doing this housekeeping without slowing down the UI?

> for this chore it is actually better if is run only when the app is in foreground

I don't know if the timer of sleep advances in background, likely not because linux timers for older kernels that android supports don't advance. So the task will actually not even run every hour unless you keep the app in foreground.

If the timer does advance, then it means the timer will expire when you open the app and one hour has passed, in which case the task will slow down opening the app, exactly when you try to refresh the chatlist it will start accessing the database and loading webxdc messages.

link2xt: https://stackoverflow.com/questions/17561380/android-thread-sleep-sometimes-waits-far-too-long says the timer does not advance when the screen is off, so this is likely to sleep way longer than 1 hour. Thread.sleep called from Util.sleep is not meant for scheduling.

adbenitez: > We can also get rid of Signal code implementing the same thing here then:

we can't that is needed by the rpc bindings which must not depend on android deps



> What is the disadvantage of waking the app, likely when the phone is not used, and doing this housekeeping without slowing down the UI?

what is the problem with current solution? in any case the worker will be executed also when app is in foreground, this solves the problem at hand and it doesn't make anything slower, getting the few urls of apps etc is super cheap, if you want to overoptimize something we could have a more lightweight api for checking if a message exists, but even with `getMessage` everything is super fast, desktop is anyways using it to load a ton of messages constantly

I want just to fix the bug, not to over-optimize and use complex solutions and add dependencies just for a simple cheap check run 24 times a day

src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -218,6 +221,45 @@ public void onReceive(Context context, Intent intent) {
...
+          for (Object key : origins.keySet()) {
+            String url = (String)key;
+            Matcher m = WEBXDC_URL_PATTERN.matcher(url);
+            if (m.matches()) {
+              int accId = Integer.parseInt(m.group(1));
+              int msgId = Integer.parseInt(m.group(2));
+              try {
+                rpc.getMessage(accId, msgId);
+                Log.i(TAG, String.format("Existing webxdc origin: %s", url));
+              } catch (RpcException ignore) {
```
link2xt: Here is an API that does not rely on ignoring errors to check if the message exists: https://github.com/chatmail/core/pull/7487

src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
link2xt: > what is the problem with current solution?

The code is not doing what it appears to do, `Util.sleep(60*60*1000); // 1h` will not sleep for one hour.

> if you want to overoptimize something 

I don't want to overoptimize anything. Using sleep for scheduling is not "not optimal", it is not correct. It pretty much never works as expected on any OS.

If you don't want to fix it, I think it is better to remove the loop and say it only runs when the app is started.

adbenitez: I can adapt it to avoid the sleep problem, sleeping for 10sec and checking last-checked timestamp

link2xt: > I can adapt it to avoid the sleep problem, sleeping for 10sec and checking last-checked timestamp

The app then will be suspended most of the time and discover that timeout expired ~5 seconds after being unsuspended when push notification arrives. `WorkManager` uses `JobScheduler` system service that it interacts via IPC internally, you cannot emulate this with sleep.

I did a PR using ListenableWorker: https://github.com/deltachat/deltachat-android/pull/4051

adbenitez: @link2xt oops, I merged your PR but the package/folder where you put the worker has nothing to do with the task it does, this is not about connectivity/network, probably should have gone into a new webview or webxdc package

src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
adbenitez: @link2xt to use the new api we need a new core release

link2xt: Core 2.28.0 is tagged, so can be updated on main branch already. As long as users cannot change the address and add new transport this way, 2.28.0 is releasable.

adbenitez: @link2xt btw, in the changelog is not mentioned: did core also reset delete_from_server_after for chatmail? because currently we have a manual migration for this in android and I would like to now if it can be removed:

```
      // 2025.11.12: this is needed until core starts ignoring "delete_server_after" for chatmail
      if (ac.isChatmail()) {
        ac.setConfig("delete_server_after", null); // reset
      }
```

link2xt: No, core does not do anything about it, see https://github.com/deltachat/deltachat-desktop/issues/5752#issuecomment-3562969965
This "migration" should be kept for as long as `delete_server_after` is hidden. Ideally `isChatmail` should be removed and all profiles should have the same settings, then this will stop compiling and we will notice.

src/main/java/org/thoughtcrime/securesms/webxdc/WebxdcGarbageCollectionWorker.java
```diff
@@ -0,0 +1,76 @@
...
+          Matcher m = WEBXDC_URL_PATTERN.matcher(url);
+          if (m.matches()) {
+            int accId = Integer.parseInt(m.group(1));
+            int msgId = Integer.parseInt(m.group(2));
+            boolean delete;
+            try {
+              delete = rpc.getExistingMsgIds(accId, Collections.singletonList(msgId)).isEmpty();
+            } catch (RpcException ignore) {
+              delete = true; // account doesn't exist anymore, clean storage
+            }
```
adbenitez: @link2xt I am correct that the api will still cause an exception if accId is from an account that no longer exists instead of returning empty list in that case? it is a bit bad that this cleans storage in any rpc error assuming it is due to deleted account

src/main/java/org/thoughtcrime/securesms/webxdc/WebxdcGarbageCollectionWorker.java
```diff
@@ -0,0 +1,76 @@
...
+          String url = (String)key;
+          Matcher m = WEBXDC_URL_PATTERN.matcher(url);
+          if (m.matches()) {
+            int accId = Integer.parseInt(m.group(1));
+            int msgId = Integer.parseInt(m.group(2));
+            boolean delete;
+            try {
+              delete = rpc.getExistingMsgIds(accId, Collections.singletonList(msgId)).isEmpty();
+            } catch (RpcException ignore) {
+              delete = true; // account doesn't exist anymore, clean storage
```
link2xt: Ignoring `getExistingMsgIds` exceptions does not look good, the whole idea of `getExistingMsgIds` was that errors don't need to be ignored as it does not throw an error when the message does not exist.

Since account may disappear, I think this should be handled by getting a set of valid account IDs in the beginning of the worker and then filter out origins that belong to non-existing account IDs before calling `getExistingMsgIds`. Then you don't need to catch any errors, if there is any exception (e.g. account disappered in the middle of the task or actual I/O error) the worker will just crash and will be retried.

src/main/java/org/thoughtcrime/securesms/webxdc/WebxdcGarbageCollectionWorker.java
link2xt: `get_existing_msg_ids` does not handle missing account.
I commented here: https://github.com/deltachat/deltachat-android/pull/4050#discussion_r2569855871
Generally ignoring exceptions and using them for control flow is not good, better not ignore any errors and in the rare case that account disappears in the middle of the garbage collection it will be restarted.

We can also change the core so `get_existing_msg_ids` just returns empty list when account does not exist, that will need another core upgrade.

link2xt: I opened core PR: https://github.com/chatmail/core/pull/7531
Probably easier to wait for it, then just call `getExistingMsgIds` and not ignore errors.

src/main/java/org/thoughtcrime/securesms/webxdc/WebxdcGarbageCollectionWorker.java
adbenitez: that is what I was asking [above](https://github.com/deltachat/deltachat-android/pull/4050#discussion_r2554385128), but implemented the manual check for account IDs in the java code as suggested

src/main/java/org/thoughtcrime/securesms/webxdc/WebxdcGarbageCollectionWorker.java
```diff
@@ -0,0 +1,88 @@
...
+
+    return CallbackToFutureAdapter.getFuture(completer -> {
+      final String LABEL = "Webxdc Garbage Collector";
+
+      List<Integer> accounts;
+      try {
+        accounts = rpc.getAllAccountIds();
+      } catch (RpcException e) {
+        Log.e(TAG, "error calling rpc.getAllAccountIds()", e);
+        completer.set(Result.failure());
```
link2xt: Don't know if this manual catching of exceptions and setting the result is needed. Thought only the worker will crash if there is no `catch`, but don't know.

adbenitez: I don't know if android would rate-limit workers that crash unexpectedly without using the api properly setting the result, I think this doesn't hurt

PR #4053, opened 2025-11-22
improve building and releasing docs

adbenitez:


scripts/update-rpc-bindings.sh
```diff
@@ -0,0 +1,11 @@
...
+# you need to have dcrpcgen
+
+# generate schema.json
+ROOT_DIR=$PWD
+cd ./jni/deltachat-core-rust/deltachat-rpc-server
+cargo run -- --openrpc > "$ROOT_DIR/schema.json"
+cd "$ROOT_DIR"
+
+# generate code
+dcrpcgen java --schema schema.json -o ./src/main/java/
```
link2xt: Somehow this file has no trailing newline, you might want to fix your editor config.

link2xt (approved):

scripts/update-rpc-bindings.sh
adbenitez: I had it disabled because it was annoying when editing files then causing diffs besides my actual changes (talking about .xml files from this android project IIRC)

PR #4061, opened 2025-11-24
only open links directly in offline help

adbenitez:
follow-up to #4045 

only allow to open links directly without asking user in trusted offline help, otherwise it might be unexpected for some potentially malicious crafted HTML email

#skip-changelog


r10s:
as discussed, closing for now, this would require sync and agreement with other ui before

PR #4064, opened 2025-11-25
Make  edge-to-edge great again

adbenitez:
### The problem

current edge-to-edge implementation is bad:
- whole window background is set to main-color to draw color behind system status bar
- bottom navigation bar has main-color due to the previous point, this looks too off and doesn't follow edge-to-edge guidelines
- due to window background being main color it can happen that when a screen is loading (during activity launch) the whole screen has main-color instead of white
- opening keyboard reveals main-color background for a second that looks like a weird  flickering
- no proper edge-to-edge in lists with scroll, the items are covered by bottom navigation bar which is not how edge-to-edge is supposed to work
- lost drop-shadow in toolbars

This is a follow up of #3814 and reverts some of the hacks introduced there

### Fixed Activities
- [x] NewContactActivity
- [x] ChatBackgroundActivity
- [x] ProxySettingsActivity
- [x] BackupTransferActivity
- [x] RegistrationQrActivity
- [x] ApplicationPreferencesActivity
- [x] BlockedContactsActivity
- [x] ConversationActivity
- [x] ConversationListArchiveActivity
- [x] CreateProfileActivity
- [x] GroupCreateActivity
- [x] ConversationListActivity
- [x] InstantOnboardingActivity
- [x] LogViewActivity
- [x] EditTransportActivity
- [x] WebViewActivity
- [x] WebxdcStoreActivity
- [x] WelcomeActivity
- [x] QrActivity
- [x] AllMediaActivity
- [x] ContactSelectionActivity
- [x] ProfileActivity
- [x] QrShowActivity
- [x] ShareActivity

### Additional tasks

- [x] enable edge-to-edge only in SDK 30+ (android 11), it is not working properly in android<11
 
### Not targeted in this patch

- Image editor: (ScribbleActivity, StickerSelectActivity) this is already special/different-looking always black screen, not worth to touch 
- MediaPreviewActivity: same as the image editor
- There is still the (apparently android) bug that when entering selection mode in activities that use a custom Toolbar component, the system bar turns black, at some point we could try to remove the use of Toolbar and use the system toolbar + setting custom view, as it is widely used in most parts of the app, but that is out of scope of this (already big) patch
- In the settings (ApplicationPreferencesActivity) the layout is generated in an intricate way, after some attempts I didn't manage to make the navigation bar not to cover the settings list while scrolling, but at least it is not colored with main-color anymore  
 
### Before vs After

image
image

image
image

image
image

### Before vs After (Android<11, SDK<30)

<details><summary><b>Show screenshots...</b></summary>

### Android 5

image
image

### Android 7

image
image

### Android 10

image
image

</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 * Send .webm videos as file, they are not supported by all platforms
 * Tweak advanced section and wording of some advanced options
 * Fix: avoid crash in push notifications handling
+* Fix: clean up web storage of deleted in-chat apps
 * Expose new "Multi-device mode" option instead of "Delete from server" for chatmail profiles
 * Opened in-chat apps got a 'About Apps' menu item
 * Avoid gray avatar on profile creation
--- a/build.gradle
+++ b/build.gradle
@@ -152,6 +152,7 @@ android {
 }
 
 dependencies {
+    implementation 'androidx.concurrent:concurrent-futures:1.3.0'
     implementation 'androidx.sharetarget:sharetarget:1.2.0'
     implementation 'androidx.webkit:webkit:1.14.0'
     implementation 'androidx.multidex:multidex:2.0.1'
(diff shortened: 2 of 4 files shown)


r10s (approved):
very nice!

thanks a lot for spending several nights to get that done ❤️

the code lgtm, as far as i can tell - and i also checked in practise on android5 and android16 (both in emulator)

no blocker, but one thing that hit my eye is that read-only chats are not really edge to edge but continue having the not-so-nice lower bar:

Image

this is esp. visible at the _very first impression_ of the device chat message. but also the new channels would look better if the wallpaper would continue to the edge.

adbenitez:
ah! I also noticed that but forgot about it when I got busy solving the situation with android 5 etc. created an issue to follow that detail so this can be merged and doesn't block other PRs

PR #4068, opened 2025-11-28
update advanced settings

r10s:
this PR updates the advanced settings to the new priorities:

- move 'server' up
- move other options down, after experimental
- introduce 'legacy' section

counterpart of https://github.com/deltachat/deltachat-ios/pull/2922

once this is settled and merged, corresponding issue should be filed for desktop

 &nbsp; 

<details>
<summary>before</summary>

 &nbsp;  

</details>

#skip-changelog as covered by "Tweak advanced section and wording of some advanced options"

### DIFF
--- a/src/main/res/xml/preferences_advanced.xml
+++ b/src/main/res/xml/preferences_advanced.xml
@@ -83,20 +83,17 @@
         <org.thoughtcrime.securesms.components.SwitchPreferenceCompat
             android:defaultValue="true"
             android:key="pref_mvbox_move"
-            android:title="@string/pref_auto_folder_moves"
-            android:summary="@string/pref_auto_folder_moves_explain"/>
+            android:title="@string/pref_auto_folder_moves"/>
 
         <org.thoughtcrime.securesms.components.SwitchPreferenceCompat
             android:defaultValue="false"
             android:key="pref_only_fetch_mvbox"
-            android:title="@string/pref_only_fetch_mvbox_title"
-            android:summary="@string/pref_only_fetch_mvbox_explain" />
+            android:title="@string/pref_only_fetch_mvbox_title" />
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/thoughtcrime/securesms/preferences/AdvancedPreferenceFragment.java
adbenitez: if the whole category is hidden, there is no need to change the individual items anymore

src/main/res/xml/preferences_advanced.xml
```diff
@@ -55,33 +65,21 @@
...
     </PreferenceCategory>
 
-    <PreferenceCategory android:title="@string/pref_server">
-
-        <Preference android:key="password_account_settings_button"
-            android:title="@string/edit_transport"/>
-
-        <Preference android:key="proxy_settings_button"
-            android:title="@string/proxy_settings"/>
+    <PreferenceCategory android:key="pref_legacy" android:title="Legacy">
```
adbenitez: I would call it `pref_category_legacy` or `pref_cat_legacy` so it is clear in the java code this is a category and not an individual preference like the rest, makes it easier to spot problems like the one I commented above

adbenitez (approved):
nice improvement!

src/main/res/xml/preferences_advanced.xml
r10s: i went for `pref_category_*` that we're using also elsewhere

r10s:
@adbenitez i added your suggestions - and moreover removed the explanations from the now legacy options - they are referring to shared usecases, that we want to get rid of. i updated the screenshots to reflect that

ftr, subsequent idea is to show legacy options not for newly added profiles; for the old ones add a device message. this will be done in core, however

PR #4071, opened 2025-11-28
update comment about recommended way of enabling edge-to-edge

adbenitez:
now that we target SDK 36, it is possible to use the recommended way of enabling edge-to-edge

#skip-changelog


adbenitez:
@r10s actually the use of `WindowCompat.enableEdgeToEdge(getWindow());` made things worse, displaying white-on-white (because nav bar has no background at all then) for the navigation bar when user has the 3-buttons navigation mode instead of gestures, the gesture bar adapts to background but the 3-buttons navigation bar doesn't and one has to manually care about setting the proper color and it just looks odd/hard to read without a background, the current approach (`EdgeToEdge.enable(this);`) actually takes care of set a translucent background for the navigation bar in 3-buttons mode so it is always readable

PR #4075, opened 2025-11-29
adapt wording to 'delete contact' question

r10s:
we no longer check the address book; this PR removes that part from the question when deleting a contact

successor of https://github.com/deltachat/deltachat-android/pull/4074


r10s:
merging directly in to push to transifex

PR #4080, opened 2025-12-01
add screen to list/add/remove transports/relays

### LINKED ISSUE #3970: use list_transports

adbenitez:
close #3970

- [x] add new activity to `AndroidManifest.xml`
- [x] create RelayListActivity with basic listing of relays/transports
- [x] allow to set main transport
- [x] allow editing transports
- [x] allow removing transports
- [x] allow to add transports
- [x] ~~final considerations about when to allow to add transports~~ (core should return error)
- [x] update changelog

image
image

### DIFF
--- a/src/main/res/layout/relay_list_item.xml
+++ b/src/main/res/layout/relay_list_item.xml
@@ -21,7 +21,7 @@
         android:layout_height="24dp"
         android:layout_marginEnd="16dp"
         android:src="@drawable/ic_check_white_24dp"
-        android:contentDescription="@string/main_transport"
+        android:contentDescription="@string/def"
         android:visibility="gone"
         app:tint="?attr/colorAccent"
         tools:visibility="visible" />
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -681,7 +681,6 @@
     <string name="proxy_enabled_hint">You are using a proxy. If you\'re having trouble connecting, try a different proxy.</string>
     <!-- "Transports", these are the relay servers used, can also be translated as "Transport Servers", if only "Transport" sounds too weird in the destination language -->
     <string name="transports">Transports</string>
-    <string name="main_transport">Main transport</string>
     <string name="add_transport">Add Transport</string>
     <string name="remove_transport">Remove Transport</string>
     <string name="edit_transport">Edit Transport</string>


r10s:
a high-level comment: why switch from "transport" wording to "relay"? 

this adds confusion - already now, wrt title, but even more in the long run -, in the UI we call it "transport", as well as the rpc calls are worded like "transport". and also desktop (and android before this PR) calls it "transport" in the code. same for translation keys. and we talks about "multi transport" all the time ("chatmail relay" can be regarded as a form of transport then, yes this is confusing as well, but no need to add even more :)


let's  stay with the "transport" wording in java code as well

adbenitez:
there were classes already called transport, this was confusing, so I used relay for the servers, I think it is an internal technical name and it is fine to use relay for the servers in the android code

tomorrow we realize calling it transport in the UI was a bad idea and then I dont think we would rename all internal classes anyways

r10s:
> there were classes already called transport

interestingly, this is at least partly old signal code where they refer to "sms" as transport.

but well, also "RelayUtil" and the like exist and is not related to transport, so that is a weak arg

> tomorrow we realize calling it transport in the UI was a bad idea and then I dont think we would rename all internal classes anyways

no, we would not rename api. similar to "profile" called "account" in api. but having a 3rd name adds even more confusion [^1]

i am bringing that up, as someone switching a lot between all ~6 languages (rust/cffi/typescript/jsonrpc/java/swift), and having somehow consistent namings on key things is king. if one is doing things mainly in one codebase, this is surely not so much of an issue.

but i will not die on that hill ...

[^1]: unless we call it "relay" in the UI then - bets are open :)

adbenitez:
> but well, also "RelayUtil" and the like exist and is not related to transport, so that is a weak arg

yeah only 1 file that could be renamed to ShareUtil actually, but anyways I am just explaining why I decided to change the name while working on it, I even started this PR going for transport naming but then got into all that transport* files and realized it was not a good idea to name all like that, more if the classes would be outside in the same dir like the rest, that is also why I created a dedicated folder as extra separation

if you really want I can go all over the place renaming everything again it will be a bit annoying but I can do it (not on the weekend tho)

r10s:
no, i do not insist. this is no blocker. it is indeed a nice PR and as far as i dived to the code while working on iOS counterpart all lgtm ❤️

however, i wanted to explain (for the next time then :) why it also matters having similar wordings across rust/cffi/typescript/jsonrpc/java/swift and that this is not without reasons as well.

r10s:
some minors, that i came over while playing around:

- i first did not fine the "add transport" - the floating button button is not used often by is, eg. for proxy it is different. just saying, maybe it is fine and we should just use floating more. let's leave it as is and iterate over that at some point.

- also no blocker: the current layout (left) appears fuzzy, as hidden options shift other elements. a suggestion on the right:<br>


- again no blocker: it is very easy to accidentally change the default transport, maybe okay, but it seems less a thing than eg. proxy. i am not sure about that point, what we really want here. at least it is consistent with desktop :)

unrelated to this PR, we might also want to somehow remove confusion with the email addresses shown here, now more prominently as before. ppl will try and expect things that are not working with them

but then, we're in advanced regions here. and we can iterate. it is cool that things start working

adbenitez:
> unrelated to this PR, we might also want to somehow remove confusion with the email addresses shown here, now more prominently as before. ppl will try and expect things that are not working with them

I was thinking already about that, I think it wiill be better to split the email addresses and use the domain as title and the name/address part as subtitle, this has the advantages:

* the list then list more prominently transports/relays than addresses
* it is easier to see the domain/relay/host address because if addresses are long then the domain will not be visible at all
* it makes it different than en email address so people dont try to share them as email

r10s:
> split the email addresses and use the domain as title and the name/address part as subtitle

like that idea, and reasoning make sense.

on ios, it looks like the following then (edit/delete are as usual in the menu or reachable via swipe there):

adbenitez:
that looks really nice!

adbenitez:
@r10s I see you are using "default" to mark the "main" relay, I introduced "main" string in this pr, should I use "default" instead?

r10s:
yes, maybe "default" is good enough for now, also desktop is using "default" so also better wrt consistency

both anyway do not explain much, next iteration we need probably an faq entry or so. but that is after that release. we're at advanced :)

PR #4094, opened 2025-12-07
remove realtime preference toggle

adbenitez:
it happens often that people disable it in the past because they got scared by the description and forgot about it then complain that apps are not working as expected and need to be told to enable it

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
+++ b/src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
@@ -168,7 +168,8 @@ public class ApplicationContext extends MultiDexApplication {
               }
             }
 
-            // 2025.12.07: revert to enabled since the setting was removed
+            // 2025-12-16: The setting was removed.
+            // Revert it to the default if it was changed in the past.
             ac.setConfigInt("webxdc_realtime_enabled", 1);
 
             // 2025-11-12: this is needed until core starts ignoring "delete_server_after" for chatmail


Raiden-GH:
Sorry to interrupt. I think there will probably be many complaints that this functionality can't be disabled. After all, it also poses a certain security risk. If you disable it, you know that an app can only sync via email, thus keeping the IP address secret.

adbenitez:
whatever you change for sure someone will complain, but I don't think there will be many complains, most people don't even know that option exists nor understand it

also, it is already the same for calls for example, or when clicking a link someone sent to you in the chat

you are also sharing your IP with the chatmail servers, etc. sharing it with your family and friends is the least of the concerns, this is nothing otherworldly, normally people don't need to worry about this, and if they do, then the proper solution is to use a VPN or Tor (and leak the IP then to that providers, you see to use the internet someone has to know your IP) not to have several settings in Delta Chat

src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -168,6 +168,9 @@ public void onCreate() {
               }
             }
 
+            // 2025.12.07: revert to enabled since the setting was removed
```
Hocuri:
```suggestion
            // 2025-12-16: The setting was removed.
            // Revert it to the default if it was changed in the past.
```

in order to follow ISO 8601 (which we're mostly using when writing dates in the code).

Also, this makes the comment more easily understandable.

src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -168,6 +168,9 @@ public void onCreate() {
...
+            ac.setConfigInt("webxdc_realtime_enabled", 1);
+
             // 2025.11.12: this is needed until core starts ignoring "delete_server_after" for chatmail
```
Hocuri:
```suggestion
            // 2025-11-12: this is needed until core starts ignoring "delete_server_after" for chatmail
```

link2xt:
If we merge this, need issues on iOS, Ubuntu Touch and Desktop repos.

r10s:
ftr, the FAQ are updated as well meanwhile, summing up reasoning: https://delta.chat/en/help#who-sees-my-ip-address

link2xt:
Brought up on the forum, linking here for reference: https://support.delta.chat/t/weaker-security-controls-for-webxdc-apps/4654

PR #4095, opened 2025-12-08
change "transport" to "relay" wording

r10s:
the wording change is the outcome of internal discussions.

moreover, the PR removes the "email" reference from the very exposed "welcome" message


r10s:
once merged, also faq needs adaption

PR #4103, opened 2025-12-09
avoid NPE in WebxdcGarbageCollectionWorker

### LINKED ISSUE #4102: crash in WebxdcGarbageCollectionWorker
the relevant line:
```
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.List chat.delta.rpc.Rpc.getExistingMsgIds(java.lang.Integer, java.util.List)' on a null object reference
at d6.b.onReceiveValue(SourceFile:85)
```

so it seems that due to the way Worker service work, in some situations it is not possible to access `ApplicationContext` and get its rpc field, hence DcHelper.getRpc() returns null and calling rpc.getExistingMsgIds() crashes with NPE (see @link2xt why I was skeptical of using a worker service for this)
 

2.33.0 Crash Report, Android 10:
```
org.chromium.base.JniAndroid$UncaughtExceptionException: Native stack trace:
#00 pc 0x032fc3d5 /data/app/com.google.android.trichromelibrary_744417130-bwbUR0wSKBsZjb3trt9NoA==/base.apk (offset 0x8a8000)
#01 pc 0x02b92097 /data/app/com.google.android.trichromelibrary_744417130-bwbUR0wSKBsZjb3trt9NoA==/base.apk (offset 0x8a8000)
#02 pc 0x02b8d21f /data/app/com.google.android.trichromelibrary_744417130-bwbUR0wSKBsZjb3trt9NoA==/base.apk (offset 0x8a8000)
#03 pc 0x00f9a063 /data/app/com.google.android.trichromelibrary_744417130-bwbUR0wSKBsZjb3trt9NoA==/base.apk (offset 0x8a8000)
#04 pc 0x00f9a337 /data/app/com.google.android.trichromelibrary_744417130-bwbUR0wSKBsZjb3trt9NoA==/base.apk (offset 0x8a8000)
#05 pc 0x02bb4f29 /data/app/com.google.android.trichromelibrary_744417130-bwbUR0wSKBsZjb3trt9NoA==/base.apk (offset 0x8a8000)
#06 pc 0x02bb3177 /data/app/com.google.android.trichromelibrary_744417130-bwbUR0wSKBsZjb3trt9NoA==/base.apk (offset 0x8a8000)
#07 pc 0x035abc67 /data/app/com.google.android.trichromelibrary_744417130-bwbUR0wSKBsZjb3trt9NoA==/base.apk (offset 0x8a8000)
#08 pc 0x035abc35 /data/app/com.google.android.trichromelibrary_744417130-bwbUR0wSKBsZjb3trt9NoA==/base.apk (offset 0x8a8000)
#09 pc 0x020e8259 /data/app/com.google.android.trichromelibrary_744417130-bwbUR0wSKBsZjb3trt9NoA==/base.apk (offset 0x8a8000)
#10 pc 0x000068df /system/lib/libutils.so
#11 pc 0x000065d5 /system/lib/libutils.so
#12 pc 0x000545d3 /system/lib/libandroid_runtime.so

at org.chromium.base.JniAndroid.handleException(chromium-TrichromeWebViewGoogle.aab-stable-744417130:21)
at android.os.MessageQueue.next(MessageQueue.java:336)
at android.os.Looper.loop(Looper.java:197)
at android.app.ActivityThread.main(ActivityThread.java:7948)
at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1075)
Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.List chat.delta.rpc.Rpc.getExistingMsgIds(java.lang.Integer, java.util.List)' on a null object reference
at d6.b.onReceiveValue(SourceFile:85)
at WV.d13.onResult(chromium-TrichromeWebViewGoogle.aab-stable-744417130:39)
at org.chromium.android_webview.AwQuotaManagerBridge.onGetOriginsCallback(chromium-TrichromeWebViewGoogle.aab-stable-744417130:12)
... 7 more
```

adbenitez:
close #4102

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/webxdc/WebxdcGarbageCollectionWorker.java
+++ b/src/main/java/org/thoughtcrime/securesms/webxdc/WebxdcGarbageCollectionWorker.java
@@ -29,12 +29,6 @@ public class WebxdcGarbageCollectionWorker extends ListenableWorker {
   public @NonNull ListenableFuture<Result> startWork() {
     Log.i(TAG, "Running Webxdc storage garbage collection...");
 
-    Rpc rpc = DcHelper.getRpc(context);
-    if (rpc == null) {
-        Log.e(TAG, "Failed to get access to RPC, Webxdc storage garbage collection aborted.");
-        return;
-    }
-
     final Pattern WEBXDC_URL_PATTERN =
       Pattern.compile("^https?://acc(\\d+)-msg(\\d+)\\.localhost/?");
 
@@ -46,6 +40,13 @@ public class WebxdcGarbageCollectionWorker extends ListenableWorker {
           Log.i(TAG, "Done, no WebView origins found.");
           completer.set(Result.success());
           return;
+        }
+
+        Rpc rpc = DcHelper.getRpc(context);
+        if (rpc == null) {
+            Log.e(TAG, "Failed to get access to RPC, Webxdc storage garbage collection aborted.");
+            completer.set(Result.failure());
+            return;
         }
 
         for (Object key : origins.keySet()) {


src/main/java/org/thoughtcrime/securesms/webxdc/WebxdcGarbageCollectionWorker.java
```diff
@@ -29,6 +29,11 @@ public WebxdcGarbageCollectionWorker(Context context, WorkerParameters params) {
   public @NonNull ListenableFuture<Result> startWork() {
     Log.i(TAG, "Running Webxdc storage garbage collection...");
 
+    if (rpc == null) {
```
link2xt: Is the worker created each time from scratch, or is it created once and `startWork` is called multiple times later?

If the worker is reused, then this will not work properly because if it returns here once, it will keep returning later. Then we should store the `context` and call `DcHelper.getRpc(context)` each time instead of storing `rpc` as I wrote in #4102 

If the worker is not reused, then this is simpler and fine, but this needs to be checked.

adbenitez: ok, modified to get rpc later

src/main/java/org/thoughtcrime/securesms/webxdc/WebxdcGarbageCollectionWorker.java
```diff
@@ -18,17 +18,23 @@
...
   }
 
   @Override
   public @NonNull ListenableFuture<Result> startWork() {
     Log.i(TAG, "Running Webxdc storage garbage collection...");
 
+    Rpc rpc = DcHelper.getRpc(context);
+    if (rpc == null) {
+        Log.e(TAG, "Failed to get access to RPC, Webxdc storage garbage collection aborted.");
+        return;
```
link2xt: CI fails to build this. startWork returns future, this should be a future that succeeds immediately. Could be moved directly before WebStorage.getInstance() probably.

adbenitez: actually moved it inside webStorage.getOrigins(), that is even better because then it even has more time for Rpc to get initialized

PR #4107, opened 2025-12-10
improve UX of adding new relays via QR or clicking links

adbenitez:
#skip-changelog


r10s (approved):

link2xt:
Not nice that this is a skip-changelog, as far as I understand this solves a problem with QR code scanner present in 2.33.0 but fixed in 2.33.1. Also 2.33.0 changelog is completely missing so it is impossible to tell what changed between 2.33.0 and 2.33.1.

Copy-pasting here from the chat:
> now it is possible to add new relay to existing account by clicking the QR link on the same device without needing to scan qr from another device
>
> so to add a new profile user has to really click on "add profile" and then if a link is opened while in this stage where there is no profile selected (actually there is a dummy unconfigured profile selected in the case of android) then it is account creation and not adding relay to existing profile

adbenitez:
> Not nice that this is a skip-changelog, as far as I understand this solves a problem with QR code scanner present in 2.33.0 but fixed in 2.33.1. Also 2.33.0 changelog is completely missing so it is impossible to tell what changed between 2.33.0 and 2.33.1.

there was no 2.33.0 release, it was never tagged and never released, it was only shared for testing in the forum, that is why there is no change added to the changelog since it was fixed before it was released to any store

PR #4119, opened 2025-12-12
avoid NPE in InstantOnboardingActivity.onCreate

### LINKED ISSUE #4118: NullPointerException  at org.thoughtcrime.securesms.InstantOnboardingActivity.onCreate
```
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:4111)
  at android.app.ActivityThread.handleLaunchActivity (ActivityThread.java:4277)
  at android.app.servertransaction.LaunchActivityItem.execute (LaunchActivityItem.java:103)
  at android.app.servertransaction.TransactionExecutor.executeCallbacks (TransactionExecutor.java:135)
  at android.app.servertransaction.TransactionExecutor.execute (TransactionExecutor.java:95)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2443)
  at android.os.Looper.loopOnce (Looper.java:226)
  at android.os.Looper.loop (Looper.java:313)
  at android.app.ActivityThread.main (ActivityThread.java:8751)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:571)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1135)
Caused by java.lang.NullPointerException: Attempt to invoke virtual method 'java.lang.String android.net.Uri.toString()' on a null object reference
  at org.thoughtcrime.securesms.InstantOnboardingActivity.onCreate (SourceFile:67)
  at android.app.Activity.performCreate (Activity.java:8290)
  at android.app.Activity.performCreate (Activity.java:8270)
  at android.app.Instrumentation.callActivityOnCreate (Instrumentation.java:1329)
  at android.app.ActivityThread.performLaunchActivity (ActivityThread.java:4085)
```

adbenitez:
close #4118

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/InstantOnboardingActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/InstantOnboardingActivity.java
@@ -119,6 +119,7 @@ public class InstantOnboardingActivity extends BaseActionBarActivity implements 
         finish();
         return;
       }
+      // if URI is unexpectedly null, then fallback to new profile creation
       AccountManager.getInstance().beginAccountCreation(this);
     }
 


src/main/java/org/thoughtcrime/securesms/InstantOnboardingActivity.java
```diff
@@ -111,12 +111,15 @@ public void onCreate(Bundle bundle) {
...
-      return;
+      Uri uri = getIntent().getData();
+      if (uri != null) {
+        Intent intent = new Intent(this, RelayListActivity.class);
+        intent.putExtra(RelayListActivity.EXTRA_QR_DATA, uri.toString());
+        startActivity(intent);
+        finish();
+        return;
+      }
+      AccountManager.getInstance().beginAccountCreation(this);
```
r10s, line 123: when exactly is this line reached?

i do not doubt, it is correct, but i like to understand better. maybe that line also deserves a comment

adbenitez: to be honest, I do not know how could it happen that the activity got opened via intent but the intent data doesn't provide the URI/URL, but it happened as per the crash report, if there is no URI (dcaccount: / dclogin:) then we have to fallback to old behavior of "just open the activity to create a new profile, as a best effort, will add a comment

PR #4128, opened 2025-12-15
chats preferences is called just 'Chats'

r10s:
it was renamed first on iOS from 'Chats and Media' to 'Chats' to not confuse with 'All Apps and Media' -
however, that point is also valid for android,
but even if not, it is good for inner consistency and easier documentation.

moreover, the settings are in reality also about contacts and more, so just a broad chats is fine. and it is shorter :)

but also for outer consistency,
this seems better - also in quite some other apps, it is just called 'chats'

Screenshot 2025-12-15 at 15 24 45

#skip-changelog as a minor change


adbenitez (approved):
notice that "Chats and Media" is mentioned in `devicemsg_storage_exceeding` that also needs update, also probably the local help needs update, it is mentioned in https://delta.chat/en/help#backup

r10s:
i changed the click path in `devicemsg_storage_exceeding` and will do for faq, thanks a lot for the pointer.

for `devicemsg_storage_exceeding`: the message is no longer accurate anyways, the storage options just do not exist for chatmail, so for the vast majority of cases. not sure, what is an actionable item there. but that is another PR

PR #4134, opened 2025-12-18
Harden security around relays and profile management

### LINKED ISSUE #4124: protect with lock profile/relay deletion
deleting profile/relays is a dangerous action and should be protected by phone lock pin

adbenitez:
- [x] require authorization for relays management
- [x] don't expose activities unnecessarily
- [x] protect profile deletion

close #4124


r10s (approved):
lgtm

PR #4151, opened 2026-01-07
avoid crash in FetchForegroundService.onCreate()

### LINKED ISSUE #4133: ForegroundServiceStartNotAllowedException in FetchForegroundService.onCreate
Image

```
  at android.app.ActivityThread.handleCreateService (ActivityThread.java:5354)
  at android.app.ActivityThread.-$$Nest$mhandleCreateService (Unknown Source)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:2682)
  at android.os.Handler.dispatchMessage (Handler.java:107)
  at android.os.Looper.loopOnce (Looper.java:282)
  at android.os.Looper.loop (Looper.java:387)
  at android.app.ActivityThread.main (ActivityThread.java:9500)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:600)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:1005)
Caused by android.app.ForegroundServiceStartNotAllowedException: Time limit already exhausted for foreground service type dataSync
  at android.app.ForegroundServiceStartNotAllowedException$1.createFromParcel (ForegroundServiceStartNotAllowedException.java:54)
  at android.app.ForegroundServiceStartNotAllowedException$1.createFromParcel (ForegroundServiceStartNotAllowedException.java:50)
  at android.os.Parcel.readParcelableInternal (Parcel.java:5105)
  at android.os.Parcel.readParcelable (Parcel.java:5087)
  at android.os.Parcel.createExceptionOrNull (Parcel.java:3267)
  at android.os.Parcel.createException (Parcel.java:3256)
  at android.os.Parcel.readException (Parcel.java:3239)
  at android.os.Parcel.readException (Parcel.java:3181)
  at android.app.IActivityManager$Stub$Proxy.setServiceForeground (IActivityManager.java:7193)
  at android.app.Service.startForeground (Service.java:776)
  at org.thoughtcrime.securesms.service.FetchForegroundService.onCreate (SourceFile:43)
  at android.app.ActivityThread.handleCreateService (ActivityThread.java:5329)
```

adbenitez:
in https://github.com/deltachat/deltachat-android/pull/3312 the `FetchForegroundService` was introduced as a way for the app to be considered in foreground to avoid some systems not giving background data connection (or perhaps users restricting background network/data access to the app) but besides that push notifications were working already fine for most people, hence if the system now decides it doesn't want to allow to start the foreground notification, it seems good enough to just log the error and then fallback to the old behavior of doing nothing, since at that point the delta chat core is already started etc. 

close #4133

### DIFF
--- a/src/gplay/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
+++ b/src/gplay/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
@@ -91,7 +91,19 @@ public class FcmReceiveService extends FirebaseMessagingService {
   @Override
   public void onMessageReceived(@NonNull RemoteMessage remoteMessage) {
     Log.i(TAG, "FCM push notification received");
-    FetchForegroundService.start(this);
+    // Note: The system can downgrade the high priority messages to normal priority
+    // if the app is not using the high priority messages for surfacing time sensitive
+    // content to the user. If the message's priority is downgraded, your app cannot
+    // start a foreground service and attempting to start one results in a
+    // ForegroundServiceStartNotAllowedException.
+    // So, it's recommended to check the result of RemoteMessage.getPriority() and
+    // confirm it's PRIORITY_HIGH() before attempting to start a foreground service.
+    // source: https://developer.android.com/develop/background-work/services/fgs/restrictions-bg-start
+    if (remoteMessage.getPriority() == RemoteMessage.PRIORITY_HIGH) {
+      FetchForegroundService.start(this);
+    } else {
+      FetchForegroundService.fetchSynchronously();
+    }
   }
 
   @Override
--- a/src/main/java/org/thoughtcrime/securesms/service/FetchForegroundService.java
+++ b/src/main/java/org/thoughtcrime/securesms/service/FetchForegroundService.java
@@ -80,12 +80,11 @@ public final class FetchForegroundService extends Service {
         } // else we stop FetchForegroundService on DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE
       });
     } catch (Exception e) {
-      Log.w(TAG, "Error calling startForeground(): " + e + ", fetching in background.");
-      fetchSynchronously();
+      Log.e(TAG, "Error calling startForeground()", e);
     }
   }
 
-  private static void fetchSynchronously() {
+  public static void fetchSynchronously() {
     // According to the documentation https://firebase.google.com/docs/cloud-messaging/android/receive,
     // we need to handle the message within 20s, and the time window may be even shorter than 20s,
     // so, use 10s to be safe.


src/main/java/org/thoughtcrime/securesms/service/FetchForegroundService.java
```diff
@@ -88,14 +70,40 @@ public void onCreate() {
...
+
+      Util.runOnAnyBackgroundThread(() -> {
+        Log.i(TAG, "Starting fetch");
+        if (!ApplicationContext.getDcAccounts().backgroundFetch(300)) { // as startForeground() was called, there is time
+          FetchForegroundService.stop(this);
+        } // else we stop FetchForegroundService on DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE
+      });
+    } catch (Exception e) {
+      Log.w(TAG, "Error calling startForeground(): " + e + ", fetching in background.");
+      fetchSynchronously();
```
adbenitez: this is a last resort, I am not sure it will not be killed immediately and not able to run, ideally instead of throwing exception inside onCreate, the system should have gave us error in the call to `ContextCompat.startForegroundService` in `FetchForegroundService.start()` there we already then fallback to the synchronous fetching (that happens in the FCM (push notification call back))

Hocuri:
As discussed in 1:1, it seems unlikely to me that this helps. Instead of a crash, we'll probably get an ANR because Android doesn't want you to do work in `onCreate()`.

**EDIT by @adbenitez:** I updated the code to avoid getting stuck in `onCreate()`

src/gplay/java/org/thoughtcrime/securesms/notifications/FcmReceiveService.java
```diff
@@ -91,7 +91,19 @@ public static String getToken() {
...
     Log.i(TAG, "FCM push notification received");
-    FetchForegroundService.start(this);
+    // Note: The system can downgrade the high priority messages to normal priority
+    // if the app is not using the high priority messages for surfacing time sensitive
+    // content to the user. If the message's priority is downgraded, your app cannot
+    // start a foreground service and attempting to start one results in a
+    // ForegroundServiceStartNotAllowedException.
+    // So, it's recommended to check the result of RemoteMessage.getPriority() and
+    // confirm it's PRIORITY_HIGH() before attempting to start a foreground service.
+    // source: https://developer.android.com/develop/background-work/services/fgs/restrictions-bg-start
```
adbenitez, line 101: potentially this could be the cause of the crash and maybe it solves it

src/main/java/org/thoughtcrime/securesms/service/FetchForegroundService.java
```diff
@@ -88,14 +70,39 @@ public void onCreate() {
...
+      startForeground(NotificationCenter.ID_FETCH, notification);
+
+      Util.runOnAnyBackgroundThread(() -> {
+        Log.i(TAG, "Starting fetch");
+        if (!ApplicationContext.getDcAccounts().backgroundFetch(300)) { // as startForeground() was called, there is time
+          FetchForegroundService.stop(this);
+        } // else we stop FetchForegroundService on DC_EVENT_ACCOUNTS_BACKGROUND_FETCH_DONE
+      });
+    } catch (Exception e) {
+      Log.e(TAG, "Error calling startForeground()", e);
```
Hocuri, line 83: The "ideal" solution here would be to signal back to `start()` that it needs to call `fetchSynchronously()`. But since these things are hard to test, let's not do this unless there are concrete problems with the current solution.

PR #4156, opened 2026-01-08
truncate file names in the middle

r10s:
that way, the end is always visible,
which contains important information more often than the middle.

this is common knowledge since the 90's or so
however, as quite some previously stable ux patterns, lost on the way. (cmp. Scott Jensen, https://www.youtube.com/watch?v=1fZTOjd_bOQ )

sidenote: this issue might have been part of some ticket confusion, cc @hpk42

counterpart of https://github.com/deltachat/deltachat-ios/pull/2978

# before

# after

it is not only about the extension (which was visible before as well), but also about suffixes as names, numbers.

the uppercase "type" can be removed probably, esp. if that is the extension only. but that is another PR/issue

### DIFF
--- a/src/main/res/layout/document_view.xml
+++ b/src/main/res/layout/document_view.xml
@@ -37,9 +37,8 @@
                           android:layout_height="wrap_content"
                           style="@style/Signal.Text.Body"
                           android:singleLine="true"
-                          android:maxLines="1"
                           android:clickable="false"
-                          android:ellipsize="end"
+                          android:ellipsize="middle"
                           android:textColor="@color/document_icon_text"
                           tools:text="The-Anarchist-Tension-by-Alfredo-Bonanno.pdf"/>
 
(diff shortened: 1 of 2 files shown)


adbenitez:
now some space is wasted by the extension that is already visible in the subtitle

src/main/res/layout/document_view.xml
```diff
@@ -37,9 +37,8 @@
                           android:layout_height="wrap_content"
                           style="@style/Signal.Text.Body"
                           android:singleLine="true"
-                          android:maxLines="1"
```
r10s, line 40: the linter says, `maxLines="1"` should be removed resp. replaced by `singleLine`, so i did ...

Image

r10s:
> now some space is wasted by the extension that is already visible in the subtitle

yip, see comment above. it is not only about the extension, but also about suffixes in the name. we can probably just remove the extension from the subtitle.

PR #4164, opened 2026-01-14
Remove email address from profile

### LINKED ISSUE #4161: remove email address from profile
the usual reasons, the email address should be removed from the profile.

it stays accessible in the "encryption info"

it can be removed unconditionally (for non-chatmail, it is shown in the subtitle, that is good enough)

wchen342:
See #4161

closes #4161


adbenitez (approved):
lgtm, thanks!

PR #4165, opened 2026-01-14
Fix dialogs not showing when pasted QR codes are not valid

### LINKED ISSUE #4160: adding relay: no error dialog on pasting invalid link
- Android version: 11
- Device: OnePlus A3003
- Delta Chat version: 2.35-foss
- Expected behavior: visible error dialog like when scanning invalid QR codes elsewhere in the app
- Actual behavior: no user feedback, just goes back to relay overview and looks like it is not working
- Steps to reproduce the problem: try to add a relay copying some invalid link (neither `dcaccount` nor `dclogin`)

I got there because I copied the link to a QR code image of the chatmail relay and not the actual link of the image and at first I was a bit annoyed because I thought there would be a bug in the app, but fortunately it is "just" a missing error dialog.

wchen342:
The dialogs are built correctly, but are not actually shown.

closes #4160

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Allow to see inbox quota for all relays in connectivity screen
 * Truncate file names in the middle, not at the end; important information are more often at the end
 * Remove email address from user profile
+* Fix: Show dialog if pasted QR codes are invalid
 * Update to core 2.36.0
 
 ## v2.35.0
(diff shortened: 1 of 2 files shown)


src/main/java/org/thoughtcrime/securesms/NewConversationActivity.java
```diff
@@ -138,10 +138,10 @@ protected void onActivityResult(int requestCode, int resultCode, Intent data) {
     super.onActivityResult(requestCode, resultCode, data);
     if (resultCode != RESULT_OK) return;
 
+    QrCodeHandler qrCodeHandler = new QrCodeHandler(this);
     switch (requestCode) {
       case IntentIntegrator.REQUEST_CODE:
         IntentResult scanResult = IntentIntegrator.parseActivityResult(resultCode, data);
-        QrCodeHandler qrCodeHandler = new QrCodeHandler(this);
         qrCodeHandler.handleOnlySecureJoinQr(scanResult.getContents(), SecurejoinSource.Scan, SecurejoinUiPath.NewContact);
```
adbenitez: this change seems unnecessary or leftovers of experimenting/testing

wchen342: I was thinking the initialization can be moved out of switch. I will keep it then.

PR #4169, opened 2026-01-15
Select correct account when returning from conversations launched through shortcut

### LINKED ISSUE #3913: wrong chatlist after opening chat's homescreen shortcut
when opening a chat from a shortcut in the home screen, and while the current selected account is not the one the chat belongs to, when returning to the chatlist, it will not be refreshed, it will still show the chats of the previously selected account before opening the chat via shortcut

wchen342:
Previously, the current selected account is not changed, if the home screen shortcut launches into a conversation under a different account than the current one.

Now, if we see a conversation that is launched from a shortcut, we always reset the current selected account based on which account the conversation is under.

closes #3913

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Mark external links with " ↗" to make them clear
 * Make QR code larger on "Add Second Device" screen
 * Fix: Show dialog if pasted QR codes are invalid
+* Fix: Select correct account when returning from conversation launched through home screen shortcut from different account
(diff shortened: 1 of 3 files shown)


adbenitez:
I created a more simple alternative solution at #4170 that doesn't require adding new extras to the shortcuts, existing shortcuts, ex. chat shortcuts already added to the home screen, will not have that new extra so this wouldn't fix the problem for existing shortcuts

PR #4171, opened 2026-01-16
Show hint during Contact search if results are empty

### LINKED ISSUE #3896: no hint in empty contact search results when adding member to chat
- Android version: 
- Device:
- Delta Chat version: 2.10
- Expected behavior: there should be an indication that this is not possible
- Actual behavior: nothing happens, there is no indication that adding the address was not successful
- Steps to reproduce the problem: try to add an email address to an encrypted group
- Screenshots:
- Logs:

wchen342:
Added a "No results found for..." text when results are empty, see below:

image

closes #3896

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ContactSelectionListFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/ContactSelectionListFragment.java
@@ -22,6 +22,7 @@ import static org.thoughtcrime.securesms.util.ShareUtil.isRelayingMessageContent
 import android.app.Activity;
 import android.content.Intent;
 import android.os.Bundle;
+import android.text.TextUtils;
 import android.util.SparseIntArray;
 import android.view.LayoutInflater;
 import android.view.Menu;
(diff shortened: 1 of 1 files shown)


src/main/res/values/strings.xml
```diff
@@ -255,6 +255,7 @@
     <string name="channels">Channels</string>
     <!-- consider keeping the term "channel" as in WhatsApp or Telegram -->
     <string name="new_channel">New Channel</string>
+    <string name="filter_no_result">No results found for %1$s</string>
```
adbenitez: actually we already have a string for this, check `search_no_result_for_x`

wchen342: Got it, I have updated the PR.

adbenitez:
in general I had a bad feeling about the approach of using the adapter with the new item, because it seems like even if not searching but the list of elements is just empty this would also show such "result not found" item, this doesn't happen normally because usually there is always at least one element in the list by default, but even then it is causing some other issues:

* the text is not centered
* when using a classic email account, the avatar of the item is not hidden, a QR icon is shown
* when clicking the "result not found" item you get offered to chat with invalid contact

see screenshots below, I recommend instead to do something similar to what we are doing in the chatlist, check `ConversationListFragment.loadChatlist()` there a dedicated `emptySearch` view is show when the a search term exist and there is not results:
image


----

### this are the screenshots of the errors described above:

image

  
image

wchen342:
Ah ok, I saw the other search but thought it will be overkill to add a whole view for this. But you are right, if it causes other problems then we shall not use the item.

wchen342:
I changed to use the empty view already in `ContactSelectionListFragment`. Now it looks like this:

image

PR #4173, opened 2026-01-16
Do not launch WelcomeActivity as singleTask

### LINKED ISSUE #3686: RegistrationActivity opens after adding new profile, pressing "Back", and re-opening DC
- Android version: 13
- Device: FP3
- Delta Chat version: Current main
- Steps to reproduce:
  - Add a new profile
  - Press the "Back" button
  - Open DC again from "recent apps"
- Expected behavior:
  - The ConversationListActivity opens
- Actual behavior:
  - The activity for adding a new profile opens again

wchen342:
My understanding of what happened:

 - `AccountManager` launches `WelcomeActivity` with `singleTask`, which creates a new task
 - This makes `WelcomeActivity` the root of current task
 - When creation is finished, `finishAffinity()` is called, which removes the activity and everything below it, but this doesn't change the task root
 - When exited and resumed from Recent, Android tried to restore task root, thus launching `WelcomeActivity` instead
 
 Since we probably just want to prevent multiple `WelcomeActivity` occur at once, `singleTop` is enough.
 
 closes #3686

#skip-changelog

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -221,7 +221,7 @@
               android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize"/>
 
     <activity android:name=".WelcomeActivity"
-              android:launchMode="singleTask"
+              android:launchMode="singleTop"
               android:theme="@style/TextSecure.LightNoActionBar"
               android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize"/>
 


wchen342:
> some people claim sigleTop doesn't prevent ending with two instances of the activity in the stack

It's technically true, but I think in this case we cannot have both because how task works. I didn't find a way to set the task root manually. Either we will end up with the wrong resume state because of `singleTask` or we need to make sure `WelcomeActivity` won't end up in a weird state ourselves.

PR #4179, opened 2026-01-19
avoid empty profiles after using "add as second device" from welcome screen

### LINKED ISSUE #4175: Phantom profile appears after transferring a profile to DC-Android
When transferring a profile to DC-Android via QR code, an additional phantom profile appears in the profile switcher alongside the transferred one. Selecting it takes you to the new profile screen. Afterward, it disappears. 

It can also be deleted directly and is 4 KiB in size. I was unable to reproduce this problem in DC-iOS 2.33.0 and DC-Desktop 2.35.0, so I suspect it's a bug in DC-Android.

Android 15
DC-Android 2.35.0

Image

adbenitez:
since #4134 the code to "add second device" from QR got merged into a single place both for the QR scanner in the chatlist and the scanner from WelcomeActivity (see https://github.com/deltachat/deltachat-android/pull/4134/changes#diff-a5aa788e8d9641e7f161e419bb4920805e19c049206f757c8a66fb29c332ac15R342)

the problem is that the code used was handling the case of the chatlist scanner where there is already an existing configured account and then first it switch to a new created account, when calling from WelcomeActivity we already have a new not configured account so the code was creating yet another new account and selecting that, leaving behind the already created empty account

note: switching accounts and adding second device had a shared logic in `switchAccountAndStartActivity(Activity activity, int destAccountId, @Nullable String backupQr)` this was hard to read and reason about so I removed and added more on-point code in the 2 places where it was used

close #4175

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Show hint for empty contact search results
 * Fix: Show dialog if pasted QR codes are invalid
 * Fix: Refresh chat list when returning from conversation if selected profile changed
+* Fix: Avoid empty profiles after using "add as second device" from welcome screen
 * Update to core 2.36.0
 
 ## v2.35.0


wchen342:
~I somehow find a new bug:~

So if you add as a second device when the app is first run (i.e. there were no accounts before), and then finish adding the account, you will see the account name is not shown correctly:

image

However, if you add another account, then switch back to the first account, you will see the name becomes correct:

image

But if you delete the second new account, the name of the first account will be wrong again:

image

I think something is probably not refreshed correctly?

EDIT: intentional behavior

adbenitez:
@wchen342 that is intentional, when there is a single profile we show the app name, when there are several then we show the profile name

this was the behavior prefered by @r10s, when discussing this topic I was voting for always show your profile name, as you confusion shows, it can be perceived as a bug and showing your profile name would make it more obvious you can tap it to open the app switcher etc., fun fact, in ArcaneChat client (a dc fork I have) I always show the display name

however, changing this behavior is out of the scope from this pull request

PR #4180, opened 2026-01-20
allow to switch profile while sharing/forwarding

### LINKED ISSUE #4055: in "Share to Delta", allow selecting target profile
currently, when sharing information to Delta Chat, only changes from the last selected profile are selectable.

this issue is about adding a profile switcher to the "Share To" dialog.

it probably makes sense to use already known same layout and ux elements, so smth like:

Image

tapping the avatar allows to switch ones account (when using the same account switcher, creating/deleting/editing a new account can be disabled, these things do not really make sense at this point)

when there are no multiple accounts, the account switcher can be hidden.

ftr, the same layout can be used for forwarding messages, however, that requires some work in core first

https://github.com/deltachat/deltachat-ios/issues/2912

adbenitez:
- [x] account switcher: make switching accounts smoother
- [x] allow to switch account while sharing
- [x] allow to switch account while forwarding (using https://github.com/chatmail/core/pull/7491)
- [x] wait for https://github.com/chatmail/core/issues/7724 to be resolved and a new core available
- [x] update changelog

close #4055 

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 * Fix: Update menu when using "select all" in contact selection
 * Fix: Avoid empty profiles after using "add as second device" from welcome screen
 * Fix multi-device seen messages synchronization when using multiple relays
-* Update to core 2.38.0
+* Update to core 2.39.0
 
 ## v2.35.0
 2026-01


adbenitez:
mainly waiting for a new chatmail core with a bug fix for forwarding files, but otherwise the code can be reviewed already

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1535,6 +1558,9 @@ public void onAttachmentChanged() {
 
   @Override
   public void handleEvent(@NonNull DcEvent event) {
+    DcContext dcContext = DcHelper.getContext(context);
+    if (event.getAccountId() != dcContext.getAccountId()) return;
```
wchen342, line 1569: I have a few questions about this part. If I understand correctly, the rest of the events will not be handled if `accountId` doesn't match? In that case:
1. Are they just left to be handled by other places in the app?
2. For events like `DC_EVENT_CHAT_EPHEMERAL_TIMER_MODIFIED`, etc. how can they be triggered when on this screen (for testing)?

adbenitez: that was just a measure to avoid bad behavior, but I think the proper solution would actually be to register for listening to that events on all accounts, because currently if the account changes then the events are not processed anymore

it seems this was always a bit wrong, or it can't never happen that the old event listener is there if the accounts changed, in theory it could happen if the activity is open and then you open a chat shortcut (from other account) from the home screen and onNewIntent is called instead of destroying old activity and creating a new one, this never seemed to be a problem or was too much of a corner case that we never noticed it

anyways I think I will go with the approach of listening on all accounts and then with the current check to ensure we process only events of the selected account

I knew I wanted to change something but forgot about it, so good you bring it up! we need to carefully review and test PR because I don't feel confident that there are not bug introduced

wchen342: I agree it is better to be handled by some listener. Will you do that in this PR too?

I also want to mention something I noticed after working on the code (as a whole) for a while, that a lot of things are tangled together, like UI is doing business logic (i.e. `View` doing stuff that is supposed to be handled by `Model`) and vice versa. I think in the future when we are adding new things, we need to at least start trying to make things separated if possible. It will also make code easier to understand and more testable, reducing potential bugs I believe.

adbenitez: ok, tested a bit, in fact there was already bug in this event-registering in ConversationActivity, steps to reproduce:

1. first  get a DC with two profiles A and B
2. in A select a chat and use the option in the menu to create a shortcut in home screen
3. go to profile B then open a chat, minimize the app with that chat open, and go to the home screen
4. click the shortcut to open the chat in the other account
5. now the ConversationActivty gets onNewIntent(), onCreate() is not called so the event listener is registered for the other chat, including the `isMultiuser()` check that adds some event listener conditionally and might not apply for the currently opened chat

will push a commit with a fix

adbenitez:
> we may want to have avatar instead of three-dot at some point

the menu is also used for "export attachment" when coming from a webxdc, that is why I used menu instead of avatar button

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
wchen342: Is `isMultiuser` for detecting group chat? If so, I think we need to also check if the current user is in the group?

wchen342:
@r10s I feel like we need to add a warning before sharing message to a different profile. When I first tried it I was actually expecting the message to be sent by the current profile, which surprised me to see it is actually sent by a different profile (it makes sense programmatically, but not everyone is expecting it).

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
adbenitez: what for in the context of the event listener?

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -247,17 +245,23 @@ public void handleOnBackPressed() {
...
+    eventCenter.removeObservers(this);
+
+    eventCenter.addMultiAccountObserver(DcContext.DC_EVENT_CHAT_MODIFIED, this);
+    eventCenter.addMultiAccountObserver(DcContext.DC_EVENT_CHAT_EPHEMERAL_TIMER_MODIFIED, this);
+    eventCenter.addMultiAccountObserver(DcContext.DC_EVENT_CONTACTS_CHANGED, this);
 
     if (!isMultiUser()) {
-      eventCenter.addObserver(DcContext.DC_EVENT_INCOMING_MSG, this);
-      eventCenter.addObserver(DcContext.DC_EVENT_MSG_READ, this);
+      eventCenter.addMultiAccountObserver(DcContext.DC_EVENT_INCOMING_MSG, this);
```
wchen342, line 262: Sorry, wrong place, I meant the `isMultiUser` here.

adbenitez:
> @r10s I feel like we need to add a warning before sharing message to a different profile. When I first tried it I was actually expecting the message to be sent by the current profile, which surprised me to see it is actually sent by a different profile (it makes sense programmatically, but not everyone is expecting it).

the message is sent by the profile you select, I think if the original profile would be the one sending to a chat/contact from a different profile that would be even more unexpected

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
adbenitez: yep that is the place I was thinking you were talking about, but I don't think being a member or not is related to adding that events, one can get removed and added back etc.

but if there is something wrong please point out

wchen342:
> the message is sent by the profile you select, I think if the original profile would be the one sending to a chat/contact from a different profile that would be even more unexpected

Yes, but I think the problem here is not which profile to send, and there is no clear indication which profile is sending it, thus the confusion (different people will have different interpretations).

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
wchen342: Oh so the listeners are only for non-group messages? Then it makes sense.

PR #4181, opened 2026-01-20
avoid account rollback in InstantOnboardingActivity

adbenitez:
rolling back account creation is not needed because now `AccountManager.getInstance().beginAccountCreation(this);` is never called

this code was problematic because when user clicked `switch profile > add profile` and then go to nine.testrun.org and click the QR, then InstantOnboardingActivity is opened, but `fromWelcome` is false since the activity gets open by the `dcaccount:` deeplink hence the `handleOnBackPressed()` would be called, and account rolled back

nothing terrible happens because then user ends in the chatlist of the previous selected account, but the code is legacy code from when account creation would start from already configured account, hence the need for rollback since there would be no WelcomeActivity to return to, but now if account is already configured we just offer to add relay, so there is no need to have all this code that is only triggered by chance in some corner case

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/InstantOnboardingActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/InstantOnboardingActivity.java
@@ -23,7 +23,6 @@ import android.widget.ImageView;
 import android.widget.TextView;
 import android.widget.Toast;
 
-import androidx.activity.OnBackPressedCallback;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.AlertDialog;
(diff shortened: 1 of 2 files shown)


src/main/java/org/thoughtcrime/securesms/InstantOnboardingActivity.java
```diff
@@ -106,35 +103,21 @@ public void onCreate(Bundle bundle) {
...
       Uri uri = getIntent().getData();
       if (uri != null) {
         Intent intent = new Intent(this, RelayListActivity.class);
         intent.putExtra(RelayListActivity.EXTRA_QR_DATA, uri.toString());
         startActivity(intent);
-        finish();
-        return;
       }
-      // if URI is unexpectedly null, then fallback to new profile creation
-      AccountManager.getInstance().beginAccountCreation(this);
```
adbenitez, line 123: this was questionable, user click a link and they expect that link to be used, no just to fallback to the default onboarding server, so better just ignore the deeplink if it is invalid somehow, this should never happen tho

wchen342: I tested an invalid link, it shows the invalid QR dialog:

image

The text can be changed, otherwise nothing bad happens.

PR #4182, opened 2026-01-20
Migrate deprecated `onBackPressed` calls

### LINKED ISSUE #4140: Migration to predictive back required
for now/temporarily there is a way to opt-out of "predictive back" setting `android:enableOnBackInvokedCallback="false"` but eventually this will be removed and predictive back enforced, when it is enforced:

> For apps targeting Android 16 (API level 36) or higher and running on an Android 16 or higher device, the predictive back system animations (back-to-home, cross-task, and cross-activity) are enabled by default. Additionally, onBackPressed is not called and [KeyEvent.KEYCODE_BACK](https://developer.android.com/reference/android/view/KeyEvent#KEYCODE_BACK) is not dispatched anymore.

so we need to migrate all calls to `onBackPressed`

wchen342:
Here is a list of Activities changed and I have tested navigation on them:

- [x] `WelcomeActivity`
  - [x] When there were no existing account
  - [x] When there are existing accounts
- [x] `LocalHelpActivity`
  - [x] WebView back navigation
  - [x] Go back to last Activity
- [x] ~`InstantOnboardingActivity`~
  - [x] ~Open from `dcaccount` link, app has no existing account~
  - [x] ~Open from `dclogin` link, app has existing accounts~
- [x] `CreateProfileActivity`
  - [x] Up button works correctly
  - [x] Go back to last Activity
- [x] `ConversationListArchiveActivity`
  - [x] Go back to `ConversationListActivity`
  - [x] Go back when forwarding message
- [x] `ConversationListActivity`
  - [x] Collapse Search Bar
  - [x] Go Back to Home Screen
  - [x] Go Back whe forwarding message
  - [x] Export Attachment from WebXDC
  - [x] `Up` button works correctly
- [x] `ConversationActivity`
  - [x] Go Back to `ConversationList`
  - [x] `emojiPicker` closes correctly
- [x] `ContactSelectionActivity`/`ContactMultiSelectionActivity`
  - [x] `Up` button works correctly
- [x] `BackupTransferActivity`
  - [x] (Sender) Warning dialog before go back
  - [x] (Receiver) Activity finishes normally

Some other notes:
- `isInputOpen` branch in `CreateProfileActivity` seems unreachable, since we cannot open emoji picker from there
- `StickerSelectActivity` doesn't have a three dot menu, so `onOptionsItemSelected` cannot be reached

@adbenitez @r10s If you can help test if there is anything I missed, it will be great. Thanks!

closes #4140

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/WelcomeActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/WelcomeActivity.java
@@ -15,6 +15,7 @@ import android.view.View;
 import android.widget.Button;
 import android.widget.TextView;
 
+import androidx.activity.OnBackPressedCallback;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.ActionBar;
@@ -90,6 +91,19 @@ public class WelcomeActivity extends BaseActionBarActivity implements DcEventCen
         registerForEvents();
         initializeActionBar();
 
+        getOnBackPressedDispatcher().addCallback(this, new OnBackPressedCallback(true) {
+            @Override
+            public void handleOnBackPressed() {
+                AccountManager accountManager = AccountManager.getInstance();
+                if (accountManager.canRollbackAccountCreation(WelcomeActivity.this)) {
+                    accountManager.rollbackAccountCreation(WelcomeActivity.this);
+                } else {
+                    setEnabled(false);
+                    getOnBackPressedDispatcher().onBackPressed();
+                }
+            }
+        });
+
         DcHelper.maybeShowMigrationError(this);
     }
 
@@ -135,7 +149,7 @@ public class WelcomeActivity extends BaseActionBarActivity implements DcEventCen
 
         switch (item.getItemId()) {
         case android.R.id.home:
-            onBackPressed();
+            getOnBackPressedDispatcher().onBackPressed();
             return true;
         }
 
@@ -382,16 +396,6 @@ public class WelcomeActivity extends BaseActionBarActivity implements DcEventCen
             }
         } catch (Exception e) {
             e.printStackTrace();
-        }
-    }
-
-    @Override
-    public void onBackPressed() {
-        AccountManager accountManager = AccountManager.getInstance();
-        if (accountManager.canRollbackAccountCreation(this)) {
-            accountManager.rollbackAccountCreation(this);
-        } else {
-            super.onBackPressed();
         }
     }
 }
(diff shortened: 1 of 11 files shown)


src/main/java/org/thoughtcrime/securesms/WelcomeActivity.java
adbenitez, line 1: there is still a call to `onBackPressed()` in WelcomeActivity, the intention was that it would then call our custom code and rallback account creation etc

wchen342: Ah good catch, I missed that one.

I also updated with the changes from #4181

src/main/java/org/thoughtcrime/securesms/qr/BackupTransferActivity.java
```diff
@@ -200,7 +207,10 @@ public void doFinish() {
             startActivity(new Intent(this, ApplicationPreferencesActivity.class));
             overridePendingTransition(0, 0);
         }
-        finish();
+        if (backPressedCallback != null) {
+            backPressedCallback.setEnabled(false);
+        }
+        getOnBackPressedDispatcher().onBackPressed();
```
adbenitez: why was necessary to change the behavior here?

adbenitez:
> * `isInputOpen` branch in `CreateProfileActivity` seems unreachable, since we cannot open emoji picker from there

`isInputOpen` calls internally `KeyboardAwareLinearLayout.isKeyboardOpen()` it seems to be not only about emoji picker but also the soft-keyboard, anyways I tried commenting that code and in fact it still seems to work without it, when you press back the keyboard is hidden first, and it doesn't go back immediately, so not sure if we need to take care about the system keyboard ourselves, seems we don't, but I didn't test in android 5, maybe in older androids we need to handle that manually

src/main/java/org/thoughtcrime/securesms/qr/BackupTransferActivity.java
wchen342: It actually has the same effect because the default `onBackPressed` just calls `finish`, but let me change it back for clearance.

wchen342:
> when you press back the keyboard is hidden first

My understanding is on modern Android the soft keyboard is handled by system so we cannot intercept back press on them anyway, so that left only the emoji picker being relevant.

src/main/java/org/thoughtcrime/securesms/qr/BackupTransferActivity.java
wchen342: Updated.

adbenitez:
> > when you press back the keyboard is hidden first
> 
> My understanding is on modern Android the soft keyboard is handled by system so we cannot intercept back press on them anyway, so that left only the emoji picker being relevant.

I just tested in android 5 and it seems to work without that code, anyways, let's keep the PR focused on its target and not do code clean-up and other side-effects, just in case

PR #4185, opened 2026-01-23
fix: Change maximum height and width of avatar-images to the maximum allowed by Chatmail Core

72374:
Avatar-images with a resolution larger than 512x512 will be resized by Chatmail Core, while media-quality is set to [balanced](https://github.com/chatmail/core/blob/1b8c732611b6b12d74cc59acac7d4506911645db/src/constants.rs#L200C1-L200C50), which is currently the highest quality-setting.

---

I did not test this, but this change could improve the quality of avatar-images.
If i understand correctly, images will be resized twice, once by the image-editor and once by Chatmail Core, which can make images blurry; especially if the original image has a resolution lower than 640x640 pixels and will be upscaled.

A better solution would probably be to use the resolution of the image after editing, and let Chatmail Core do the resizing, so that no unnecessary resizing is done.
I think removing the overrides for the resolution ([there](https://github.com/deltachat/deltachat-android/blob/b9c5e741e83868588ffd5088495d9c234db18ca2/src/main/java/org/thoughtcrime/securesms/CreateProfileActivity.java#L150), [there](https://github.com/deltachat/deltachat-android/blob/b9c5e741e83868588ffd5088495d9c234db18ca2/src/main/java/org/thoughtcrime/securesms/GroupCreateActivity.java#L376), and [there](https://github.com/deltachat/deltachat-android/blob/b9c5e741e83868588ffd5088495d9c234db18ca2/src/main/java/org/thoughtcrime/securesms/InstantOnboardingActivity.java#L281)) could be enough to do that, but i currently cannot test that.

Image-comparison
---
Opening the following images in new browser-tabs and switching between them with Ctrl + Tab works well for seeing the differences.

<details>

1 - Original (512x512 - 4:2:0 colour-subsampling):
1 - Original (512x512 - 4:2:0 colour-subsampling)

2 - Without this change, after setting the original image as an avatar with the Android-app of Delta Chat, and then downloading the avatar-image with the desktop-app of Delta Chat:
2 - Delta Chat (avatar)

</details>

fix, or at least improve, #4188

#skip-changelog because we can add the changelog entry when we update core, which also improved avatar quality

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Improve switch speed when changing profiles
+* Allow to switch profile when sharing or forwarding
 * Don't notify notification-to-all from in-chat apps if the chat is muted
 * Allow to see inbox quota for all relays in connectivity screen
 * Truncate file names in the middle, not at the end; important information are more often at the end
@@ -15,7 +17,7 @@
 * Fix: Update menu when using "select all" in contact selection
 * Fix: Avoid empty profiles after using "add as second device" from welcome screen
 * Fix multi-device seen messages synchronization when using multiple relays
-* Update to core 2.38.0
+* Update to core 2.39.0
 
 ## v2.35.0
 2026-01
--- a/src/main/java/org/thoughtcrime/securesms/AllMediaDocumentsFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/AllMediaDocumentsFragment.java
@@ -54,7 +54,6 @@ public class AllMediaDocumentsFragment
   public void onCreate(Bundle bundle) {
     super.onCreate(bundle);
 
-    dcContext = DcHelper.getContext(getContext());
     chatId = getArguments().getInt(CHAT_ID_EXTRA, -1);
     viewtype1 = getArguments().getInt(VIEWTYPE1, 0);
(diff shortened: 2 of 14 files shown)


Hocuri:
I tried out this PR. It certainly improves image quality, but the avatar is still worse quality than original.

Probably, in order to get even better quality, we would need to save the original image file, and then pass it to core, rather than taking this bitmap.

> I think removing the overrides for the resolution ([there](https://github.com/deltachat/deltachat-android/blob/b9c5e741e83868588ffd5088495d9c234db18ca2/src/main/java/org/thoughtcrime/securesms/CreateProfileActivity.java#L150), [there](https://github.com/deltachat/deltachat-android/blob/b9c5e741e83868588ffd5088495d9c234db18ca2/src/main/java/org/thoughtcrime/securesms/GroupCreateActivity.java#L376), and [there](https://github.com/deltachat/deltachat-android/blob/b9c5e741e83868588ffd5088495d9c234db18ca2/src/main/java/org/thoughtcrime/securesms/InstantOnboardingActivity.java#L281)) could be enough to do that, but i currently cannot test that.

I tried commenting out these lines in addition to this PR here. Then, the avatar looks as good as with only the PR here, and a 1024x1024px image is passed to core. Which is weird, because the original image is definitely 512x512 quality.

r10s:
ftr, cmp https://github.com/deltachat/deltachat-desktop/issues/4914

PR #4198, opened 2026-01-29
Make device messages and subscribed channels edge to edge

### LINKED ISSUE #4069: read-only chats are not really edge to edge
read-only chats are not really edge to edge but continue having the not-so-nice lower bar:

Image

this is esp. visible at the _very first impression_ of the device chat message. but also the new channels would look better if the wallpaper would continue to the edge.

(copied from https://github.com/deltachat/deltachat-android/pull/4064#pullrequestreview-3518973823)

wchen342:
Removed bottom edge on
- Device Messages
- Incoming Channels

See screenshot:
image

closes #4069

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
@@ -1009,10 +1009,8 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     if (dcChat.canSend()) {
       composePanel.setVisibility(View.VISIBLE);
       attachmentManager.setHidden(false);
-      if (isInitialization) {
-        ViewUtil.forceApplyWindowInsets(findViewById(R.id.root_layout), true, false, true, true);
-        fragment.handleRemoveBottomInsets();
-      }
+      ViewUtil.forceApplyWindowInsets(findViewById(R.id.root_layout), true, false, true, true);
+      fragment.handleRemoveBottomInsets();
     } else {
       composePanel.setVisibility(View.GONE);
       attachmentManager.setHidden(true);


adbenitez:
the message list is not respecting the bottom navigation bar insets, bottom padding is missing, with gesture navigation it just looks a bit bad with the bar covering a bit of the message but it is not so bad, but if you go to settings and switch to the 3-buttons style navigation then a bigger part of the message will be covered by the bar

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -995,6 +995,10 @@ private void initializeResources() {
     if(chatId == DcChat.DC_CHAT_NO_CHAT)
       throw new IllegalStateException("can't display a conversation for no chat.");
     dcChat           = DcHelper.getContext(context).getChat(chatId);
+    if (dcChat.isDeviceTalk() ||
+      isMultiUser() && dcChat.isInBroadcast()) {
```
adbenitez: `isMultiuser()` seems unnecessary  here, `isInBroadcast` should be sufficient 🤔

wchen342: Right, I think I saw the check below and just borrowed it😂

wchen342:
I have updated the way paddings are set. It shall now work correctly with 3-button navigation too:

image

src/main/java/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -100,15 +100,17 @@ public class ConversationFragment extends MessageSelectorFragment
...
     private TextView                    noMessageTextView;
     private Timer                       reloadTimer;
 
     public boolean isPaused;
     private Debouncer markseenDebouncer;
     private Rpc rpc;
+    private boolean pendingHideBottomDivider;
 
-    @Override
+  @Override
```
adbenitez: looks like an accidental change

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -995,6 +995,11 @@ private void initializeResources() {
     if(chatId == DcChat.DC_CHAT_NO_CHAT)
       throw new IllegalStateException("can't display a conversation for no chat.");
     dcChat           = DcHelper.getContext(context).getChat(chatId);
+    if (dcChat.isDeviceTalk() || dcChat.isInBroadcast()) {
+      ViewUtil.applyWindowInsets(findViewById(R.id.root_layout), true, false, true, false);
+      ViewUtil.applyWindowInsets(findViewById(R.id.fragment_content), false, true, false, true);
+      fragment.hideBottomDivider();
```
adbenitez: sadly this is not edge to edge, this just adds a padding and then the list doesn't show behind the bar just as without this change, only difference is that the navigation bar doesn't have the white background

also it is not really only about "device messages" chat and incoming channels, it can happen that there is no input bar when user is no longer part of a group, it is better to tie this logic to the place where the input bar is decided to be shown/hidden

this is the expected behavior when scrolling in edge-to-edge:

Image

current behavior of this PR:

Image

wchen342: I am a bit confused, I thought the message shouldn't go behind the system bar? Because the previous commit will make them go behind the system bar.

Or do you only want the messages to go behind the bar when the OS is not in 3-button mode? I feel like that's not the best idea, the behavior shall be consistent between the two modes.

adbenitez: it should go behind the bar **while scrolling** the problem with the initial state of the PR is that when you scroll to the bottom of the list there was no enough bottom padding/margin and the last item gets covered, compare with the chatlist or "All Media" activity, there you can see the proper behavior while last item is not covered if you scroll to bottom

wchen342: I have made the scroll go under the bar while keep the last item above it. Can you check if it looks good now?

wchen342: Also changed the condition to test `canSend`.

src/main/java/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -158,6 +161,12 @@ public View onCreateView(@NonNull LayoutInflater inflater, ViewGroup container,
         // with hardware layers, drawing may result in errors as "OpenGLRenderer: Path too large to be rendered into a texture"
         list.setLayerType(View.LAYER_TYPE_SOFTWARE, null);
 
+        if (pendingAdjustBottomLayout) {
+          bottomDivider.setVisibility(View.GONE);
+          ViewUtil.applyWindowInsets(list, false, true, false, true);
+          pendingAdjustBottomLayout = false;
```
adbenitez: if this boolean is there only to avoid re-applying insets, this is not a problem, original size is stored and calling this several times will not increase insets every time

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
adbenitez: super nice! it is working as expected now, but there is a small detail: if you leave a group, then the input bar disappear but it doesn't become edge-to-edge, if you go back of the chat and enter again then it appears properly

the setting of the edge-to-edge should happen in `setComposePanelVisibility()`

src/main/java/org/thoughtcrime/securesms/ConversationFragment.java
wchen342: It is for delayed apply of the changes, because the function can be called before the fragment has finished creating the view, thus need a delayed execution.

adbenitez (approved):

PR #4201, opened 2026-01-30
fix mailto handling and remove unused code

### LINKED ISSUE #4195: mailto link stopped working
several users that use bots like the "www bot" are complaining that the links in the html messages of the bot stopped working, these links are mailto links that forward to the chat of the bot with a draft, now when clicking it opens other email apps or complains that there are no apps to handle that links

adbenitez:
at https://github.com/deltachat/deltachat-android/pull/4134/changes#diff-6305bbf50570f0946de077f04e8ea56c53d8f6e1b53ec403c8c2ddb10d531db1L215-L220
the mailto handling got removed from `NewConversationActivity` because it was already present in `ShareActivity` but in `ShareActivity` only `SENDTO` was being handled, `VIEW` was missing

also some dead code handling mailto was left in  `NewConversationActivity` 

close #4195

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ShareActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ShareActivity.java
@@ -104,24 +104,23 @@ public class ShareActivity extends PassphraseRequiredActionBarActivity implement
     resolvedExtras = new ArrayList<>();
 
     List<Uri> streamExtras = new ArrayList<>();
+    if (MailtoUtil.isMailto(getIntent().getData())) {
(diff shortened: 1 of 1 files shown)


wchen342:
I tested and it seems the intents are not directly handled, but instead just shows a "Share with..." screen? Is this intentional?

adbenitez:
> I tested and it seems the intents are not directly handled, but instead just shows a "Share with..." screen? Is this intentional?

in fact this is unexpected! thanks for finding out!

adbenitez:
> I tested and it seems the intents are not directly handled, but instead just shows a "Share with..." screen? Is this intentional?

fixed!

r10s (approved):
code lgtm, did not test further

PR #4205, opened 2026-02-03
remove deselected members

### LINKED ISSUE #4202: Cannot remove members via "add members" screen
With Delta Chat 2.35, open a group profile, click "Add members", in the screen uncheckmark some members. Then click on the checkmark in the top right corner to confirm. Nothing happens as a result, member is not removed.

Either it should be impossible to uncheckmark existing members, or uncheckmarking them via "add members" screen should remove them.

adbenitez:
this also makes it possible to search by name what members to remove, which was not possible before, making it really hard to find someone in the list in +50 members group ex. when someone has full inbox or expired/deleted account

close #4202

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * Fix: Refresh chat list when returning from conversation if selected profile changed
 * Fix: Update menu when using "select all" in contact selection
 * Fix: Avoid empty profiles after using "add as second device" from welcome screen
+* Fix: Remove from group deselected members in the contact selection list
 * Fix multi-device seen messages synchronization when using multiple relays
 * Avoid crash when the app is minimized with profile switcher or reactions dialogs open
 * Update to core 2.39.0


wchen342:
I think I found a bug: if you long press in the group add member screen, and click the bin icon to delete them, then you either click the tick or go back:
- The group members are still the same
- But if you go back into the add member list now, only "Me" is still there, and the rest are gone, even though the group is not empty

EDIT: it is probably not caused by the PR though, but I wonder is this a bug?

EDIT 2: I can't get it back even after a force stop..

adbenitez:
> I think I found a bug: if you long press in the group add member screen, and click the bin icon to delete them, then you either click the tick or go back:
> 
>     * The group members are still the same
> 
>     * But if you go back into the add member list now, only "Me" is still there, and the rest are gone, even though the group is not empty

mmh that is indeed confusing, what is happening here is that that screen allows to remove the contacts from your contact list (it is the exact same list/view than in the contact list when you click the + button in the chatlist) so what it is doing is removing the contact from contact list, not from the group, but in this context where the list is used to add/remove members the trash icon would be expected to be used to remove the member from group not to remove from contact list, lets just hide the trash icon there, but that is in fact a separate existing issue, will create a dedicate issue

PR #4210, opened 2026-02-04
show menu to select audio/video call modes

adbenitez:
- [x] show menu with option to select to start in video/audio call mode instead of directly start video call
- [x] when audio-only mode is selected start the webview app with video muted
- [x] when user answers a call start with camera disabled if it is audio call
- [x] use the new core APIs to send the video/audio mode of the call to the remote peer
- [x] display the camera icon if the in-chat message is of video-call mode instead of always showing the audio-call icon
- [x] update calls web app
- [x] update changelog
- [x] adapt icons in menu to light/dark modes

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/calls/CallActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/calls/CallActivity.java
@@ -43,14 +43,14 @@ public class CallActivity extends WebViewActivity implements DcEventCenter.DcEve
   public static final String EXTRA_CHAT_ID = "chat_id";
   public static final String EXTRA_CALL_ID = "call_id";
   public static final String EXTRA_HASH = "hash";
-  public static final String EXTRA_AUDIO_ONLY = "audio_only";
+  public static final String EXTRA_HAS_VIDEO = "has_video";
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/thoughtcrime/securesms/calls/CallActivity.java
```diff
@@ -94,7 +98,7 @@ public void onPermissionRequest(PermissionRequest request) {
       .withPermanentDenialDialog(getString(R.string.perm_explain_access_to_camera_denied))
       .onAllGranted(() -> {
         String url = "file:///android_asset/calls/index.html";
-        webView.loadUrl(url + hash);
+        webView.loadUrl(url + query + hash);
```
wchen342, line 101: Just a small note, this is fine in this case because the text is fixed, but generally it is better not concatenate url directly in string.

wchen342 (approved):
Tested the UI except for "when user answers a call start with camera disabled if it is audio call" because I can't get the call to go through for some reason (I tried both real device and emulator). The UI itself works fine.

Also in audio only mode, the call still uses video permission, which can be improved in the future.

PR #4213, opened 2026-02-05
more detailed call strings

r10s:
the new stock strings will be used when core includes https://github.com/chatmail/core/pull/7830 , however, can also already merged before to push the new strings to translators

#skip-changelog as this is a change in an experimental feature, also minor


r10s:
reviewed by @adbenitez sitting beside me, in new york

PR #4217, opened 2026-02-07
Allow scanning invite codes from onboarding

### LINKED ISSUE #4041: allow scanning invite codes from onboarding
- when scanning a QR-invite-code (1to1, group, channel) from the onboarding scanner, show an alert with the text `instant_onboarding_confirm_contact` resp. `instant_onboarding_confirm_group` and the buttons "OK" and "Cancel"

- on confirmation, go the the normal name-enter-screen, show the same text above "Set a name ..." there

- after account creation, create the corresponding chat, desktop even opens it directly

desktop already implemented that and the strings are right there, on android we cut that path that time for $reasons, to move forward. but it should be added eventually. 

current:

Image

desired:

counterpart of https://github.com/deltachat/deltachat-ios/issues/2906

<details>

<summary>background</summary>

for the user, it is unclear, that there are securejoin-invites and createaccount-invites.

from the beginning on, we wanted to allow to scan both codes from "Create New Profile / Use Other Server / Scan Invitation Code" - even though currently "Use Other Server" is in between (might be that the onboarding  "Scan Invitation Code" also gets moved one level up btw, but that is out of scope of this issue)

moreover, this part is also needed to make i.delta.chat instructions error resistant.
</details>

wchen342:
See  #4041 for details.

The prompts uses the default Android style since we don't use, e.g. button with background, on Android.

Here is how it looks like:

image

closes  #4041

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
@@ -1059,16 +1059,20 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     if (dcChat.canSend()) {
       composePanel.setVisibility(View.VISIBLE);
       attachmentManager.setHidden(false);
-      ViewUtil.forceApplyWindowInsets(findViewById(R.id.root_layout), true, false, true, true);
-      fragment.handleRemoveBottomInsets();
+      // FIXME: disabled for now to avoid problems with chat scrolling and keyboard covering input bar
+      // ViewUtil.forceApplyWindowInsets(findViewById(R.id.root_layout), true, false, true, true);
+      // fragment.handleRemoveBottomInsets();
     } else {
       composePanel.setVisibility(View.GONE);
       attachmentManager.setHidden(true);
       hideSoftKeyboard();
+      // FIXME: disabled for now to avoid problems with chat scrolling and keyboard covering input bar
+      /*
       if (isInitialization) {
         ViewUtil.forceApplyWindowInsets(findViewById(R.id.root_layout), true, false, true, false);
         fragment.handleAddBottomInsets();
       }
+      */
     }
   }
 


src/main/java/chat/delta/rpc/types/SecurejoinSource.java
```diff
@@ -1,6 +1,7 @@
 /* Autogenerated file, do not edit manually */
 package chat.delta.rpc.types;
 
+// FIXME: the sources are not used correctly
```
adbenitez: this file should not be modified manually

src/main/java/chat/delta/rpc/types/SecurejoinUiPath.java
```diff
@@ -10,4 +10,8 @@ public enum SecurejoinUiPath {
 
   /** The user first clicked on the `+` button in the main screen, and then on "New Contact" */
   NewContact,
-}
\ No newline at end of file
+
+  /** The user scanned a join QR code when creating a new profile from Instant Onborading Screen */
+  // FIXME: need to be added in core
+  // InstantOnboarding,
```
adbenitez: same here, rpc package souldn't be edited manually, I guess you could use `Unknown` as it is already the case IIRC for other links clicked outside the app and handled via deep-link, if something needs to be added must be added in chatmail core

maybe @Hocuri  can comment because to be honest I don't really know well the purpose of this

src/main/java/chat/delta/rpc/types/SecurejoinSource.java
wchen342: Ok I have reverted the changes.

src/main/java/chat/delta/rpc/types/SecurejoinUiPath.java
wchen342: I suppose the RPC is generated from core? Then these shall be modified in core I assume. I have reverted the changes for now.

Hocuri: Passing `Unknown` is fine. This statistic will anyways never be counted, because the user can only enable stats-sending after they created a profile. At the time the user is in InstantOnboarding, there is on way stats-sending is enabled already.

adbenitez (approved):
lgtm

PR #4226, opened 2026-02-13
Fix video player crash

wchen342:
Change PlayerView to StyledPlayerView to avoid conflict between ExoPlayer2 and Media3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Add context menu in long-pressing relays items instead of showing buttons
+* Enhanced video player UI
 
 ## v2.42.0
 2026-02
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Add context menu in long-pressing relays items instead of showing buttons
+* Fix: Crash in video player
```
adbenitez: since 2.42.0 will not be released we don't need to report this to users, maybe instead say "enhanced video player" since that is the user-visible change

r10s:
ftr, a corresponding report in the forum at https://support.delta.chat/t/android-app-crashes-v2-42-0/4765/2

thanks for fixing the bug already❤️

PR #4230, opened 2026-02-17
add basic sticker picker

adbenitez:
- [x] add basic sticker picker
- [x] add changelog entry

for now just the sticker picker similar to how the picker works in Delta Chat Desktop, you add stickers to the collection by saving received (or sent via keyboard) stickers, clicking on a sticker message in the chat offers to save to collection, to delete a saved sticker, long press the sticker and a dialog will appear (in DC desktop actually there is no way to delete inside delta chat)
 
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Allow to set chat description
 * Unified date display in call bubbles
 * Explain at "Settings / Chats / Outgoing Media Quality" how to send original quality
+* Add a basic sticker picker
 * Fix: keep original sent timestamp for resent messages
 * Fix: make clicking on broadcast member-added messages work always
 * Fix: remove notification when a message is deleted by sender


r10s:
THIS IS GREAT STUFF!

- i first had a hard way to find "add to collection", but i was thinking too complex, trying long tap or so... was then wondering if a single tap is not too annoying, but at least whatsapp on iOS also shows smth similar when tapping a sticker. probably okay, but would still be interesting how whatsapp/signal/telegram handle that on android

- the new tab bar eats quite some vertical screen space, esp. when _not_ using stickers - wondering, if that cannot be integrated smarter somehow. before / after: <br>Screenshot 2026-02-18 at 16 41 25 Screenshot 2026-02-18 at 16 40 03
iOS solves that by adding the sticker button to the top bar, to the very left, on opening, still "recent" is selected.
again, we should check what whatsapp/signal/telegram  is doing here

- stickers are cross-profile, which is great! backup-wise, it seems not-so-important, if ppl really care and curate an collection, they probably do that outside of any messenger. also, the really important stickers _are_ backed up, as sent in tons of chats.  and, if needed, we can iterate :)

- for the storage layout: as probably different collections are coming, one should keep that in mind. maybe having added stickers in `stickers` and collections in `stickers/subdir` is good enough, tho

adbenitez:
I did tested Signal, WhatsApp and Telegram (on android) while doing this and used them as guide

>     * i first had a hard way to find "add to collection", but i was thinking too complex, trying long tap or so... was then wondering if a single tap is not too annoying, but at least whatsapp on iOS also shows smth similar when tapping a sticker. probably okay, but would still be interesting how whatsapp/signal/telegram handle that on android

that is how it works in WhatsApp/Telegram/Signal, btw something interesting I noticed while testing on Signal: Monocles Chat seems to support importing signal's sticker packs!!! this is what happens when you click a sticker in signal:

image


>     * the new tab bar eats quite some vertical screen space, esp. when _not_ using stickers - wondering, if that cannot be integrated smarter somehow. 

I was thinking about hiding it if there is no stickers but then it is harder for people to discover they can save stickers, a next step would also be to put a button there to import sticker from gallery/image/filesystem so not sure it is worth the effort hiding, the hint that appears when no sticker is added could be more useful explaining how to add stickers there tho 

>       iOS solves that by adding the sticker button to the top bar, to the very left, on opening, still "recent" is selected.
>       again, we should check what whatsapp/signal/telegram  is doing here

I basically copied what the others were doing, they have a dedicated tab bar with emoji/sticker/GIF tabs (Telegram recently changed the whole UI and now such tab bars are floating tho but before they also had a normal tabs bar):

Signal:
image

WhatsApp looks pretty similar to what we have, but I just notice they use this component you can pull up to expand! we should just consider using that as well in the future, that way user can see more emojis/stickers:

image
 

>     * stickers are cross-profile, which is great! backup-wise, it seems not-so-important, if ppl really care and curate an collection, they probably do that outside of any messenger. also, the really important stickers _are_ backed up, as sent in tons of chats.  and, if needed, we can iterate :)

similar with chat wallpaper, if we have some core API at some point stickers then could live in the blobs dir and be backed up, but that would make them per-profile, while that has some advantages, in my experience of using Delta Chat desktop, it has been a major annoyance, I ended up having a global stickers folder and using symbolic links on each profile so it is a common folder in the end and I can use the stickers in every profile

>     * for the storage layout: as probably different collections are coming, one should keep that in mind. maybe having added stickers in `stickers` and collections in `stickers/subdir` is good enough, tho

I was considering using a "Saved" folder for custom saved stickers as desktop but when thinking "what is someone then have a pack named "Saved"? having them in the root folder while other packs are folders is not a problem, we check already if they are files before adding them to the "Saved" list and it is also possible to migrate them to a dedicated folder in the future if wanted

r10s:
thanks for sharing your thoughts, all sounds reasonable!

src/main/res/values/strings.xml
```diff
@@ -178,6 +178,9 @@
     <string name="add_to_sticker_collection">Add to Sticker Collection</string>
     <string name="add_stickers_instructions">To add stickers, tap "Open Sticker Folder", create a subfolder for your sticker pack, and drag image and sticker files there</string>
     <string name="open_sticker_folder">Open Sticker Folder</string>
+    <string name="ask_add_sticker_to_collection">Add this sticker to your collection?</string>
+    <string name="ask_delete_sticker">Delete this sticker?</string>
+    <string name="sticker_picker_empty_hint">Saved stickers will appear here.</string>
```
r10s: "save" is be a bit unclear, as we actually have "saved message" - i was indeed trying that! before realising that it just should be tapped :)

so, the following seems clearer:

```suggestion
    <string name="sticker_picker_empty_hint">Tap a sticker to add it here.</string>
```

we could also say "Tap a sticker in the message view to add it here.", but that raises the question what the "message view" is ... it is only a hint, ppl will realise that tapping the sticker on the traffic light is not working

src/main/java/org/thoughtcrime/securesms/components/emoji/StickerPickerView.java
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  private List<File> getSavedStickers() {
+    List<File> stickerFiles = new ArrayList<>();
+    
+    if (!stickerDir.exists()) {
+      return stickerFiles;
+    }
+
+    File[] files = stickerDir.listFiles();
```
wchen342, line 66: It is probably better to move file listing and searching to background to avoid ANR?

src/main/java/org/thoughtcrime/securesms/components/emoji/StickerPickerView.java
wchen342, line 1: Since this is a new view, I think it is better to at least separate UI from business logic. Proper MVVM will be even better.

Of course, you can decide if you want to change it now since it will take some time 🙂 Just a suggestion since doing this on new view shall be simpler.

src/main/java/org/thoughtcrime/securesms/components/emoji/StickerPickerView.java
```diff
@@ -0,0 +1,187 @@
...
+      View view = layoutInflater.inflate(R.layout.sticker_picker_item, parent, false);
+      return new StickerViewHolder(view);
+    }
+
+    @Override
+    public void onBindViewHolder(@NonNull StickerViewHolder holder, int position) {
+      File stickerFile = stickerFiles.get(position);
+      holder.stickerFile = stickerFile;
+
+      glideRequests.load(stickerFile)
```
wchen342, line 120: ~Can this be done outside of the adapter?~

Sorry, wrong comment, I think we can use `GlideApp.with(holder.itemView.getContext())` instead of passing `glideRequests` so the requests are bind to each view's lifecycle. Then there shall be no need to clear it in `onViewRecycled` and the loading will stop themselves when the item is out of view.

src/main/java/org/thoughtcrime/securesms/components/emoji/MediaKeyboard.java
```diff
@@ -51,13 +62,90 @@ public void show(int height, boolean immediate) {
 
   public void show() {
     if (emojiPicker == null) {
-      emojiPicker = findViewById(R.id.emoji_picker);
-      emojiPicker.setOnEmojiPickedListener(this);
+      setupViews();
```
wchen342: Is there a reason this is not done in constructor or `onFinishInflate`/`onAttachedToWindow`?

src/main/java/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -534,6 +536,51 @@ private void handleForwardMessage(final Set<DcMsg> messageRecords) {
...
+        // Copy sticker to stickers directory
+        String extension = stickerFile.getName();
+        int dotPos = extension.lastIndexOf('.');
+        if (dotPos >= 0) {
+          extension = extension.substring(dotPos + 1);
+        }
+        String fileName = System.currentTimeMillis() + "." + extension;
+        File destFile = new File(stickersDir, fileName);
+
+        try (java.io.FileInputStream in = new java.io.FileInputStream(stickerFile);
```
wchen342, line 565: This part needs to be moved to background thread.

wchen342:
Tested and works! There is a tiny issue I am not sure if it is intended: when adding the same sticker for multiple times, I will get multiple copies.

As I mentioned in private chat, you can ignore the architectural comment if it's too complicated. But I think the threading need to be fixed at least.

adbenitez:
> Tested and works! There is a tiny issue I am not sure if it is intended: when adding the same sticker for multiple times, I will get multiple copies.

yep that is intentional, right now we don't have uniqueness, maybe if the sticker's filename would be the file hash then we could avoid duplicates but we can leave the blame on the user, if you add the same sticker twice you get what you asked for

src/main/java/org/thoughtcrime/securesms/components/emoji/StickerPickerView.java
adbenitez: I just imitated similar existing code, if you have a better recommendation that doesn't make the code over-complicated you could provide a PR with a patch for this PR

wchen342: As I mentioned it is just a suggestion. I just feel it is counter productive to add things then refactor.

Anyway, we can ignore this for now, but can you have a look at the rest of the comments?

src/main/java/org/thoughtcrime/securesms/components/emoji/StickerPickerView.java
adbenitez: mmmh, I am not deep into Glide, could you explain how that would help? I did a quick test:

in `init()` we do:
`glideRequests = GlideApp.with(context);`

so I saved that `context` and then in `onBindViewHolder` did the following check: 
`holder.itemView.getContext().equals(context)`

and the condition is true, so in theory `GlideApp.with(holder.itemView.getContext())` is the same as that `GlideApp.with(context)`

what is the difference that would make it bind to each view's lifecycle? I guess it wouldn't hurt to do what you say I just want to understand what is going on

wchen342: That is interesting, I was assuming that by using the view, Glide will cancel the loading requests when the view is scrolled out of the screen and being detached. But if the contexts are the same then the problem does not really exist.

src/main/java/org/thoughtcrime/securesms/components/emoji/MediaKeyboard.java
adbenitez: I moved it to `onFinishInflate()`

src/main/java/org/thoughtcrime/securesms/components/emoji/StickerPickerView.java
adbenitez: moving to background is not so easy without doing a bigger refactoring, I tried to just use `Util.runOnBackground()` and then I get this error when saving a sticker:

`android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.`

wchen342: How are you doing it? I just tried it myself and it works:

```
private final ExecutorService executor = Executors.newSingleThreadExecutor();

public void loadStickers() {
    executor.execute(() -> {
      final List<File> stickerFiles = getSavedStickers();

      post(() -> setAdapter(new StickerAdapter(getContext(), glideRequests, stickerFiles, listener)));
    });
  }
```

`AsyncTask` is deprecated if I remember correctly, so the `runOnBackground` method need to be updated if we want to keep using it.

adbenitez: the problem is that we also have that "empty list hint" that we show when there is no sticker and you then can't just run this in background and immediately call the code that decides to show/hide the hint, then I needed to also move that to the thread to wait for the list to be loaded

adbenitez:
merging as it is "good enough" but this anyways needs other iterations (like allowing to list different packs and import a pack) and probably needs to be refactored using `androidx.lifecycle.AndroidViewModel` which should make it easier to move things to background

Issue #3928, opened 2025-10-02
videos not always recoded


r10s:
when sharing or attaching multiple videos to Delta Chat, videos are not recoded.

for two files, this easily results in 200 mb instead of 2 mb being sent out. difference may then be that sending videos seems not to work at all on mobile data.

there is a warning, however, ppl do not read - and if, it is not nice to work around that. came over that during holidays several times :)

the warning was added that time to move forward and have sending multiple files _at all_. however, we should iterate over that at some point and always recode videos therefore, _unless_ they're sent as files.

adbenitez:
we used to have the "in preparation state" were the message would be already created and have the "sending..." spinning wheel and user could leave the chat, but that got removed and now it is a blocking dialog, for one video this can already be a long time, if there are several it would look really suspicious/stuck to the user that will give up / cancel after some time

r10s:
> for one video this can already be a long time

a blocking dialog is fine as long as "cancel" is working. this is what iOS is doing as well.
and one huge video can already take longer than two short clips

(in fact, it was not possible at all for monet to send  the 30 second baltic see swimming video when combined with another 5 second clip)

yeah, it is probably a little bit effort, there was a reason to not do recoding that time.

still, i consider this as a bug, it is inconsistent and unexpected, and causes confusion in real world, i have seen this recently again, and also before

we might need the permanent notification, iirc, it js pretty generic meanwhile, eg also used for backups

it is not a high-prio bug, however, i do not want it to be forgotten, therefore this issue

> Sometimes recoding is not desirable

sure. you can always attach as file, this is the recommended way also for images.

see https://delta.chat/en/help#does-delta-chat-support-images-videos-and-other-attachments

this issue is about making things consistent

adbenitez:
yes I also would like if all videos and probably all images (this one should be easier!) get recorded/compressed and not conditionally if it is 1 or many, what I am pointing out is that in the current state this has become worse UX and doing it for several videos would be not as pain-free as it was in the past when we had that "in preparation" state and the user could keep going around without waiting

r10s:
"in-preparation" had lots of other issues and will not come back anytime soon. impact was low for lots of effort. 

so a blocking dialog (as now) is the best we can do. 

but it is fine, it is not that often long recoding is needed

AeliusSaionji:
Related to this behavior, I often see webm videos being sent as-is, and invariably the iOS users on the other end will be unable to play said video.

[...]

PR #4235, opened 2026-02-17
clearify, how to recode multiple videos

r10s:
two translators, independently, had no idea what "attach them separately" mean, tho using Delta Chat.

the wording from the first commit was regarded as clearer, but i modified that as well and avoided "recode" in a second commit.

best would be if we would just recode multiple selected videos, and would not need this string, but iirc, this is a bit more complex - if not @adbenitez @wchen342 please create an issue :)

#skip-changelog

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -712,7 +712,7 @@
     <!-- first placeholder is replaced by the number of files (always 2 or more); second placeholder is replaced by a chat name -->
     <string name="ask_send_files_to_chat">Send %1$d files to \"%2$s\"?</string>
     <string name="ask_send_files_to_selected_chats">Send %1$d file(s) to %2$d chats?</string>
-    <string name="videos_sent_without_recoding">(Videos are sent as original, big files. To make them smaller and save data, send them as single messages, one at a time)</string>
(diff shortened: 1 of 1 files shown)


src/main/res/values/strings.xml
```diff
@@ -712,7 +712,7 @@
     <!-- first placeholder is replaced by the number of files (always 2 or more); second placeholder is replaced by a chat name -->
     <string name="ask_send_files_to_chat">Send %1$d files to \"%2$s\"?</string>
     <string name="ask_send_files_to_selected_chats">Send %1$d file(s) to %2$d chats?</string>
-    <string name="videos_sent_without_recoding">(Videos are sent as original, big files. To send videos as smaller files, attach them separately)</string>
+    <string name="videos_sent_without_recoding">(Videos are sent as original, big files. To make them smaller and save data, send them as single messages, one at a time)</string>
```
wchen342: What about
`Sending multiple videos will send them as-is, which may be large. If you'd like smaller file sizes, you can send each video in its own message.`?

adbenitez:
> best would be if we would just recode multiple selected videos, and would not need this string, but iirc, this is a bit more complex - if not @adbenitez @wchen342 please create an issue :)

we already have an issue, created by you! :smile: https://github.com/deltachat/deltachat-android/issues/3928

Issue #4231, opened 2026-02-17
I can no longer add people to groups


mschunte2:
I am using Delta 2.42.0 on Android. For some reason, I can no longer add people to groups:
1. When I choose "add member", I can select a person to add by adding a checkmark to the person.
2. When I then click the OK checkmark, the window closes.
3. I get back to the "group info page" - BUT the member is not added...

This is the same for all profiles on my Android phone. I can add members to the same group on my desktop.

adbenitez:
hi, more information is required, could you try to add members, after it fails go to advanced settings and share the log you will find there here

mschunte2:
I created a new ID and tried again. Still no change. I am reluctant to post the log here. If you ping me on delta, I can share the log with you 1:1: https://i.delta.chat/#8D71EAC9273E3E6B9DB35CE1FB19DA6147DE27E6&i=WwdELP6f1nNWnyZ6SG16pb8i&s=ryP0ast_4ykESVEuq4Vc99H1&a=rurvmwf4p%40nine.testrun.org&n=Testing

adbenitez:
talking with @mschunte2 a bit more it seems it works during group creation (`GroupCreateActivity`) but not later (`ProfileFragment`) the only difference I see is that `onActivityResult()` is deprecated for `Fragment` class that would explain why handling the result of the `ContactMultiSelectionActivity` works in the activity class but not in the fragment, but then this is not a new problem, because it was like that in the past, I can not reproduce the issue in an even more recent version of android tho, so not sure 

I tested in an emulator with SDK 35 as the device of the user (sdk 35, android 15, Sony XQ-DC54)

wchen342:
Maybe add some logging and also make a quick workaround to pass things around manually to test if the theory is correct?

adbenitez:
user confirmed this also happens with 2.36 in that device so it is not a new issue

adbenitez:
even with the latest update the problem keeps happening, there is nothing special with what we are doing and it seems the problem is of that specific special company device, it is also affecting other apps, so let's close this since there is no apparent solution on our side and testing is also hard because that kind of device only allows installing from google play so every test requires a whole release cycle

PR #4238, opened 2026-02-18
migrate away from deprecated `Fragment.onActivityResult()`

adbenitez:
this might solve #4231 or at least help debugging
 
- [x] avoid deprecated `ProfileFragment.onActivityResult()`
- [x] add more logging to handling of contact selection activity results in `ProfileFragment` 
- [x] avoid deprecated `ConversationFragment.onActivityResult()`
- [x] avoid deprecated `AdvancedPreferenceFragment.onActivityResult()`
- [x] add support for new `ActivityResultLauncher` API to `ScreenLockUtil`
- [x] avoid deprecated `NotificationsPreferenceFragment.onActivityResult()`
- [x] avoid deprecated `ContactSelectionListFragment.onActivityResult()`

## What to test when trying out this branch?

* adding members to group via "add members"
* changing ringtone in Notifications settings
* any place where you are asked for screen lock/pin:
   - after scanning a relay account QR from chatlist and clicking "add relay"
   - creating a backup in Settings > Chats > export backup
   - clicking "add second device" in Settings
   - delete profile

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/relay/RelayListActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/relay/RelayListActivity.java
@@ -51,6 +51,7 @@ public class RelayListActivity extends BaseActionBarActivity
   /** QR provided via Intent extras needs to be saved to pass it to QrCodeHandler when authorization finishes */
   private String qrData = null;
   private ActivityResultLauncher<Intent> screenLockLauncher;
+  private ActivityResultLauncher<Intent> qrScannerLauncher;
 
   /** Relay selected for context menu via onRelayLongClick() */
   private EnteredLoginParam contextMenuRelay = null;
@@ -60,6 +61,15 @@ public class RelayListActivity extends BaseActionBarActivity
     super.onCreate(savedInstanceState);
     setContentView(R.layout.activity_relay_list);
 
+    qrScannerLauncher = registerForActivityResult(
+      new ActivityResultContracts.StartActivityForResult(),
+      result -> {
+        if (result.getResultCode() == RESULT_OK) {
+          IntentResult scanResult = IntentIntegrator.parseActivityResult(result.getResultCode(), result.getData());
+          new QrCodeHandler(this).handleOnlyAddRelayQr(scanResult.getContents(), null);
+        }
+      }
+    );
     screenLockLauncher = registerForActivityResult(
       new ActivityResultContracts.StartActivityForResult(),
       result -> {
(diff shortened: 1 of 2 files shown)


src/main/java/org/thoughtcrime/securesms/relay/RelayListActivity.java
```diff
@@ -223,22 +242,9 @@ public boolean onOptionsItemSelected(@NonNull MenuItem item) {
...
     super.onActivityResult(requestCode, resultCode, data);
-    if (resultCode != RESULT_OK) {
-      // if user canceled unlocking, then finish
-      if (requestCode == ScreenLockUtil.REQUEST_CODE_CONFIRM_CREDENTIALS) finish();
-      return;
-    }
-
-    QrCodeHandler qrCodeHandler = new QrCodeHandler(this);
-    if (requestCode == IntentIntegrator.REQUEST_CODE) {
+    if (resultCode == RESULT_OK && requestCode == IntentIntegrator.REQUEST_CODE) {
```
wchen342: Why this is separated from the other branch?

adbenitez: this comes from handling the result of 3rd party QR code scanner library (via its `IntentIntegrator` class) so for now it is left as is

the newer ActivityResultLauncher API is used now in `ScreenLockUtil` which makes it more clear and less indirect that the result needs to be handled by the calling class

wchen342: But it can also be moved to `registerForActivityResult` right? The `IntentIntegrator` actually do not use the REQUEST_CODE if you call it with `parseActivityResult(int resultCode, Intent intent)`, so there is no need to keep it here I think?

wchen342: Actually I think you are right, I missed that the `IntentIntegrator` launch its own activity with the request code.

adbenitez: yep, but actually looking closely `IntentIntegrator.initiateScan()` just does this: 

```
public final void initiateScan() {
        startActivityForResult(createScanIntent(), requestCode);
    }
```

I could call `createScanIntent()` directly and then handle with the new API

wchen342: Yes, that is a good idea. That will work around the requestCode problem completely.

PR #4240, opened 2026-02-21
don't set channel description on channel creation if it is empty

adbenitez:
when user creates a channel, only call rpc.setChatDescription if there is actually some description to set, otherwise there will be a "you changed the description" info-message when the user didn't set anything

in groups this was not noticed because they are not promoted on creation, but channels are created promoted then the call to set the description causes a system message, this looks a bit like a core issue because description didn't change (well from unset to empty string) and core should probably ignore it as it does when it is called with the same string/description twice 

#skip-changelog: not released feature


r10s (approved):
it is fine to have a second check here.

and, i agree, this is also a core issue. if not done yet, please file an issue there as well

PR #4241, opened 2026-02-23
Add full VoIP call support with all common functionalities

### LINKED ISSUE #3914: Better call notifications
this issue is to collect some information about the proper way of implementing call notifications being introduced by https://github.com/deltachat/deltachat-android/pull/3785

* guide: Create a call style notification for call apps https://developer.android.com/develop/ui/views/notifications/call-style#java
* the notification requires to be launched by a foreground service of type `android:foregroundServiceType="phoneCall"` which in turn needs `android.permission.FOREGROUND_SERVICE_PHONE_CALL`
* this approach of binding the notification to a service doesn't seem to support several incoming calls at the same time
* proper way to place incoming/outgoing calls is via Telecom API: https://developer.android.com/reference/android/telecom/ConnectionService
* when creating the `CallStyle` notification, if you set actions manually via `addAction` the buttons get duplicated it is not clear yet to me if relaying on the api is sufficient to warranty that in older phones the buttons will appear without manually adding them
* calling `setFullScreenIntent` is required on the notification, this is what will allow showing an incoming call in full screen when the phone is locked, ~~for this the calls-webapp needs to be expanded to include a new screen similar to the incoming call screens of other apps with accept and decline buttons~~ (the calls-webapp already has this screen)

### LINKED ISSUE #4244: I can't hear my partner when he calls me from his iPhone.
When I make a call from Android to iPhone, the call goes through normally on any network, and you can hear each other well. 

But when I get a call from an iPhone to Android, I can't hear the person talking to me from the iPhone, but the sound from Android to iPhone goes well.

Calls were checked on different networks (WiFi/LTE/VPN)

iOS 17.5.1 
Android 12.5.4
iOS Delta Chat version are the latest at the moment
Android Delta Chat ver 2.43.0

wchen342:
close #4244
close #3914 

Added functionalities:

 - Revamped in-call screen
   - Camera control
   - Mic control
   - Switch front/back camera
   - Full screen (except for the status bar, which is by design on Android)
 - Native WebRTC support
 - New notifications (CallStyle on Android 12+, high priority notification on Android 8-12):
   - Incoming:
     - Full screen notification over lock screen
     - User can answer call *without* unlocking the phone
     - High priority notification when phone is unlocked
   - During active calls (both incoming/outgoing):
     - Active call notification
     - Chip/pill over status bar showing call duration
 - PiP (Picture-in-Picture) mode support for call screen
 - Proper signaling through RPC
 - Android system integration through CallsManager/Telecom

Because this is a testing branch, I raised the min SDK level to 23:
  - The `core-telecom` library requires min sdk of 23
  - We can add a building flavor to switch source sets and libraries for 21-23 , which will not include the library

For now, please test if there are major breaking functionalities. Minor problem can be fixed later.

## To do:

- [x] when ending a call via "end call" red button, it shows "unknown" in the name for a second before the screen is closed
- [x] it is not possible anymore to disable camera or mic before answering a call in the "full screen call offer" screen
- [x] avatar of the contact not displayed anymore when making an outgoing call, the screens looks a bit empty/black
- [x]  missing speaker button: when starting in video call mode, you can disable video but if you put the phone in the ear it will not switch and will keep using loud speaker, and when starting in audio call it is not possible to switch to loud speaker when incoming/outgoing video is enabled and the phone is not in the ear
- [x] it is not possible to answer call clicking message bubble (show full-screen call offer when clicking in the incoming call message
- [x] incoming video call notification is shown as audio call, no camera icon shown in the answer button etc, probably it is missing to set the property to tell android that the call has video
- [x] when starting a call if you press back the call is not minimized, (~~then if you click the notification a new call is started~~ this part is solved)
- [x] avatar has a gray border
- [x] when going to picture-in-picture and back with self-video disabled then you can see your self-thumbnail like a transparent frame with a drop shadow (in the case of call that started as audio, video-call get an old thumbnail from self-video even if it is disabled, so my guess is that it is getting visible without checking if self-video is enabled)
- [x] the audio call doesn't start in "earpiece" mode

### DIFF
--- a/src/gplay/AndroidManifest.xml
+++ b/src/gplay/AndroidManifest.xml
@@ -3,10 +3,6 @@
     xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:tools="http://schemas.android.com/tools">
 
-    <!-- force compiling emojipicker on sdk<21 and firebase on sdk<19; runtime checks are required then -->
-    <uses-sdk
-        tools:overrideLibrary="androidx.emoji2.emojipicker, com.google.firebase.messaging, com.google.android.gms.cloudmessaging" />
-
     <meta-data
         android:name="firebase_analytics_collection_deactivated"
         android:value="true" />
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -47,9 +47,8 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MICROPHONE" />
     <uses-permission android:name="android.permission.MANAGE_OWN_CALLS"/>
 
-    <!-- force compiling emojipicker on sdk<21; runtime checks are required then -->
-    <uses-sdk tools:overrideLibrary="androidx.emoji2.emojipicker,
-    androidx.core.telecom,im.conversations.webrtc"/>
+    <!-- force compiling libs on older sdk than supported; runtime checks are required -->
+    <uses-sdk tools:overrideLibrary="androidx.core.telecom,im.conversations.webrtc"/>
 
     <application android:name=".ApplicationContext"
                  android:icon="@mipmap/ic_launcher"


adbenitez:
got this crash:
```
2026-02-24 14:30:27.598  8824-8824  CallUtil                com.b44t.messenger.beta              E  Failed to load caller icon
                                                                                                    java.lang.IllegalArgumentException: You must call this method on a background thread
                                                                                                    	at com.bumptech.glide.util.Util.assertBackgroundThread(Util.java:183)
                                                                                                    	at com.bumptech.glide.request.RequestFutureTarget.doGet(RequestFutureTarget.java:192)
                                                                                                    	at com.bumptech.glide.request.RequestFutureTarget.get(RequestFutureTarget.java:130)
                                                                                                    	at org.thoughtcrime.securesms.calls.CallUtil.getIconFromChat(CallUtil.java:69)
                                                                                                    	at org.thoughtcrime.securesms.calls.CallCoordinator.initiateOutgoingCall(CallCoordinator.java:874)
                                                                                                    	at org.thoughtcrime.securesms.calls.CallUtil.startCall(CallUtil.java:48)
                                                                                                    	at org.thoughtcrime.securesms.calls.CallUtil.startVideoCall(CallUtil.java:35)
                                                                                                    	at org.thoughtcrime.securesms.ConversationActivity.onOptionsItemSelected(ConversationActivity.java:625)
                                                                                                    	at android.app.Activity.onMenuItemSelected(Activity.java:4821)
                                                                                                    	at androidx.activity.ComponentActivity.onMenuItemSelected(ComponentActivity.java:528)
                                                                                                    	at androidx.fragment.app.FragmentActivity.onMenuItemSelected(FragmentActivity.java:265)
                                                                                                    	at androidx.appcompat.app.AppCompatActivity.onMenuItemSelected(AppCompatActivity.java:256)
                                                                                                    	at androidx.appcompat.view.WindowCallbackWrapper.onMenuItemSelected(WindowCallbackWrapper.java:109)
                                                                                                    	at androidx.appcompat.app.AppCompatDelegateImpl.onMenuItemSelected(AppCompatDelegateImpl.java:1235)
                                                                                                    	at androidx.appcompat.view.menu.MenuBuilder.dispatchMenuItemSelected(MenuBuilder.java:833)
                                                                                                    	at androidx.appcompat.view.menu.SubMenuBuilder.dispatchMenuItemSelected(SubMenuBuilder.java:90)
                                                                                                    	at androidx.appcompat.view.menu.MenuItemImpl.invoke(MenuItemImpl.java:157)
                                                                                                    	at androidx.appcompat.view.menu.MenuBuilder.performItemAction(MenuBuilder.java:984)
                                                                                                    	at androidx.appcompat.view.menu.MenuPopup.onItemClick(MenuPopup.java:128)
                                                                                                    	at android.widget.AdapterView.performItemClick(AdapterView.java:330)
                                                                                                    	at android.widget.AbsListView.performItemClick(AbsListView.java:1292)
                                                                                                    	at android.widget.AbsListView$PerformClick.run(AbsListView.java:3314)
                                                                                                    	at android.widget.AbsListView.onTouchUp(AbsListView.java:4305)
                                                                                                    	at android.widget.AbsListView.onTouchEvent(AbsListView.java:4046)
                                                                                                    	at androidx.appcompat.widget.DropDownListView.onTouchEvent(DropDownListView.java:230)
                                                                                                    	at androidx.appcompat.widget.MenuPopupWindow$MenuDropDownListView.onTouchEvent(MenuPopupWindow.java:138)
                                                                                                    	at android.view.View.performOnTouchCallback(View.java:16820)
                                                                                                    	at android.view.View.dispatchTouchEvent(View.java:16773)
                                                                                                    	at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3137)
                                                                                                    	at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2811)
                                                                                                    	at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3143)
                                                                                                    	at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2826)
                                                                                                    	at android.view.ViewGroup.dispatchTransformedTouchEvent(ViewGroup.java:3143)
                                                                                                    	at android.view.ViewGroup.dispatchTouchEvent(ViewGroup.java:2826)
                                                                                                    	at android.widget.PopupWindow$PopupDecorView.dispatchTouchEvent(PopupWindow.java:2568)
                                                                                                    	at android.view.View.dispatchPointerEvent(View.java:17110)
                                                                                                    	at android.view.ViewRootImpl$ViewPostImeInputStage.processPointerEvent(ViewRootImpl.java:8349)
                                                                                                    	at android.view.ViewRootImpl$ViewPostImeInputStage.onProcess(ViewRootImpl.java:8103)
                                                                                                    	at android.view.ViewRootImpl$InputStage.deliver(ViewRootImpl.java:7480)
                                                                                                    	at android.view.ViewRootImpl$InputStage.onDeliverToNext(ViewRootImpl.java:7537)
                                                                                                    	at android.view.ViewRootImpl$InputStage.forward(ViewRootImpl.java:7503)
                                                                                                    	at android.view.ViewRootImpl$AsyncInputStage.forward(ViewRootImpl.java:7674)
                                                                                                    	at android.view.ViewRootImpl$InputStage.apply(ViewRootImpl.java:7511)
                                                                                                    	at android.view.ViewRootImpl$AsyncInputStage.apply(ViewRootImpl.java:7731)
                                                                                                    	at android.view.ViewRootImpl$InputStage.deliver(ViewRootImpl.java:7484)
                                                                                                    	at android.view.ViewRootImpl$InputStage.onDeliverToNext(ViewRootImpl.java:7537)
                                                                                                    	at android.view.ViewRootImpl$InputStage.forward(ViewRootImpl.java:7503)
                                                                                                    	at android.view.ViewRootImpl$InputStage.apply(ViewRootImpl.java:7511)
                                                                                                    	at android.view.ViewRootImpl$InputStage.deliver(ViewRootImpl.java:7484)
2026-02-24 14:30:27.602  8824-8824  CallUtil                com.b44t.messenger.beta              E  	at android.view.ViewRootImpl.deliverInputEvent(ViewRootImpl.java:10697)
                                                                                                    	at android.view.ViewRootImpl.doProcessInputEvents(ViewRootImpl.java:10648)
                                                                                                    	at android.view.ViewRootImpl.enqueueInputEvent(ViewRootImpl.java:10616)
                                                                                                    	at android.view.ViewRootImpl.processRawInputEvent(ViewRootImpl.java:11039)
                                                                                                    	at android.view.ViewRootImpl$WindowInputEventReceiver.onInputEvent(ViewRootImpl.java:10825)
                                                                                                    	at android.view.InputEventReceiver.dispatchInputEvent(InputEventReceiver.java:277)
                                                                                                    	at android.os.MessageQueue.nextLegacy(MessageQueue.java:985)
                                                                                                    	at android.os.MessageQueue.next(MessageQueue.java:1094)
                                                                                                    	at android.os.Looper.loopOnce(Looper.java:197)
                                                                                                    	at android.os.Looper.loop(Looper.java:367)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:9331)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:566)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:945)
```

adbenitez:
another crash in the logs:
```
2026-02-24 14:32:33.993  8824-8824  AndroidRuntime          com.b44t.messenger.beta              E  FATAL EXCEPTION: main
                                                                                                    Process: com.b44t.messenger.beta, PID: 8824
                                                                                                    android.app.RemoteServiceException$CannotPostForegroundServiceNotificationException: Bad notification for startForeground
                                                                                                    	at android.app.ActivityThread.throwRemoteServiceException(ActivityThread.java:2510)
                                                                                                    	at android.app.ActivityThread.-$$Nest$mthrowRemoteServiceException(Unknown Source:0)
                                                                                                    	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2888)
                                                                                                    	at android.os.Handler.dispatchMessage(Handler.java:132)
                                                                                                    	at android.os.Looper.dispatchMessage(Looper.java:333)
                                                                                                    	at android.os.Looper.loopOnce(Looper.java:263)
                                                                                                    	at android.os.Looper.loop(Looper.java:367)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:9331)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:566)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:945)
```

adbenitez:
some other details:

- [ ] when I start a call there is a noisy notification for myself on the outgoing call notification, would be better if the notification for outgoing call (the one to hang up) is added without any sound/silent 
- [ ] the call activity remains in recent apps screen after finishing a call instead of merging back with main app activity
- [ ] the hang-up button for outgoing call notification doesn't seem to work
- [ ] when starting in video call mode, you can disable video but if you put the phone in the ear it will not switch and will keep using loud speaker, and when starting in audio call it is not possible to switch to loud speaker when incoming/outgoing video is enabled and the phone is not in the ear
- [ ] it says "connecting..." when already ringing the other device, it should say "ringing..." when the core api is called to deliver the outgoing call

wchen342:
> some other details:
> 
> * [x]  when I start a call there is a noisy notification for myself on the outgoing call notification, would be better if the notification for outgoing call (the one to hang up) is added without any sound/silent
> * [x]  the call activity remains in recent apps screen after finishing a call instead of merging back with main app activity
> * [x]  the hang-up button for outgoing call notification doesn't seem to work
> * [ ]  when starting in video call mode, you can disable video but if you put the phone in the ear it will not switch and will keep using loud speaker, and when starting in audio call it is not possible to switch to loud speaker when incoming/outgoing video is enabled and the phone is not in the ear
> * [x]  it says "connecting..." when already ringing the other device, it should say "ringing..." when the core api is called to deliver the outgoing call

I fixed them except for the load speaking one, which needs a new feature.

The app needs to be uninstalled for the sound problem to be fixed because of the notification channel setting.

wchen342:
Also fixed the first crash, the second one is likely caused by the first one, so hopefully it is fixed too.

wchen342:
I have added:

 - A button to select audio only or video call, before answering an incoming call
 - Preview of local video during making an outgoing call, before the call is connected
 - A speaker button for switching between: speaker, earpiece, wired headphone (if available), bluetooth headphone (if available), or casting devices (if available). I did not add proximity sensor because it is not very compatible with this list of devices.

adbenitez:
nice! but now the audio call doesn't start in "earpiece" mode as in the past, that is the expected default for voice call while the speaker for videocall, 

I added other problems to the issue checklist

wchen342:
> it is not possible to answer call clicking message bubble

This shall be fixed too, if you click on notification it will show the full screen activity. Or did you mean something else?

wchen342:
- Fixed the problem when PiP is not entered during call initialization.
- Fixed the problem when call can be initialized twice from the notification.

adbenitez:
> > it is not possible to answer call clicking message bubble
> 
> This shall be fixed too, if you click on notification it will show the full screen activity. Or did you mean something else?

I meant clicking the message bubble in chat, not notification, ex. you open chat and there is an "incoming call" message and you click it

wchen342:
> I meant clicking the message bubble in chat, not notification, ex. you open chat and there is an "incoming call" message and you click it

Ah ok that one, I still think it shall not get into a call this way, but I will see if I can make it work in some way.

wchen342:
- Removed border from avatar on incoming call screen
 - Add video incoming call notification style. Note that the style is *not* consistent and depends on the OEM. For example, in my emulator, the style changes the "Answer" button to "Video", instead of adding a camera icon.

wchen342:
- Made ongoing audio call starts with earpiece
 - Fixed a UI layout problem (hopefully) related to local video window.
 - Fixed a bug with PiP where home button doesn't work.
 - Made it possible to answer/decline an incoming call by clicking on the corresponding message

wchen342:
- Added permission checks for camera, microphone, notification and full screen intent
- Fixed the brief glitch of caller/callee name when disconnecting

nicodh:
> **To test the changes in this pull request, install this apk:** [📦 app-preview.apk](https://github.com/deltachat/deltachat-android/suites/59389040266/artifacts/5769640719)

This still crashes on my phone (like the previous apks of this PR)

adbenitez:
> * Made ongoing audio call starts with earpiece

it works only when you start the outgoing audio call, but on receiver side it starts with speaker mode

> * Made it possible to answer/decline an incoming call by clicking on the corresponding message

but now it is not possible to "call back" by clicking on an (already ended or missed) incoming call message bubble, only clicking outgoing call message bubble triggers the starting of a call now


Other pending problems:

- [x] we used to have an outgoing tone for the "ringing..." state from the caller point of view
- [x] the outgoing call screen is pretty empty/black would be nice to have the avatar in the middle and the status ("ringing...", etc) below
- [x] when going to PiP then you get a black screen in audio-only call, would be nice if the avatar of the other contact would be shown there if possible instead of just a black square

r10s:
this is impressive stuff! 🕶️

at a first glance,  things work as expected, i also managed to build the PR in the emulator as well as on a real device.

for the UI, not sure if it is a good place to discuss this here, in a huge PR, UI-wise i would be fine with merging and iterate in subsequent PR. iirc, there are some issues wrt building and size, that needs to be figured out before diving into details of the UI. same if the code and build.gradle is fine as well.

but as i have them in mind, some comments, again, they are blocking nothing:

-  the bar with the buttons is not matching the overall UI in android. ii'd remove the whole background and just have some round buttons, this also fits better to overall UI of desktop and iOS.

- when video is disabled, it seems better, instead of hiding the video, showing a placeholder and maybe a text, saying "you disabled video" or so. this is true for other systems as well.

but we need to look a bit more closely what other messengers are doing here.

again, these are high-level comments and must not block anything here

wchen342:
> but now it is not possible to "call back" by clicking on an (already ended or missed) incoming call message bubble, only clicking outgoing call message bubble triggers the starting of a call now
> 
> Other pending problems:
> 
> * [x]  we used to have an outgoing tone for the "ringing..." state from the caller point of view
> * [x]  the outgoing call screen is pretty empty/black would be nice to have the avatar in the middle and the status ("ringing...", etc) below
> * [x]  when going to PiP then you get a black screen in audio-only call, would be nice if the avatar of the other contact would be shown there if possible instead of just a black square

All shall be fixed now.

> there are some issues wrt building and size, that needs to be figured out before diving into details of the UI. same if the code and build.gradle is fine as well.

I have made some changes to how things are built:

 - Each ABI is now built into their own apk, i.e. you will get `*-arm64-v8a-*.apk`, `*-x64-64-*.apk`, etc. instead of one monolith apk. Each one is ~40 MB uncompressed, and ~20-25 MB download size (=compressed by Google Play).
 - It now build two flavors: one Legacy for Android 5 to 7.1, and one for Android 8+. This is needed because AndroidX Telecom library only support API 23+ and most functionalities only works API 26+ (=Android O). The Legacy version uses the old code unchanged, and the Modern version uses this PR.
 - CI is updated to build both Legacy and Modern flavor, and will build `arm64`.
 
There shall no longer be problems with apk sizes (22 MB download on Google Play shall be no problem).

adbenitez:
this needs to be rebased

adbenitez:
it is a bit bad this is now a single commit, ex. I wanted to see the split-apk changes separately since it is not clear if we really want to go with that path, having so many separated apks complicates things

wchen342:
> it is a bit bad this is now a single commit, ex. I wanted to see the split-apk changes separately since it is not clear if we really want to go with that path, having so many separated apks complicates things

It is mainly `build.gradle` changes. I can make it build monolith apk again, or just build aab.

The legacy/modern split is mandatory, unless we want to raise min sdk to 26.

wchen342:
> this needs to be rebased

Done, also I split the last commit from the rest.

wchen342:
I removed the ABI-based splits for APKs so it now goes back to building one monolith apk.

For API levels, I am strongly against overriding library's min-sdk and relying solely on runtime API level checks. That requires dynamic switching, API level checks over multiple components, and need to keep all of them in sync over multiple places. Anything goes wrong and we will potentially get a runtime crash. This is not trivial unlike a simple addition where one check is enough.

Also, similar problem will likely arise more and more often in the future, so at some point it has to be done anyway, if we want to keep supporting old devices. It cannot be pushed down the line indefinitely.

For Google Play they allow multiple versions for the same app based on API levels. The only problem will be F-Droid where we need to submit a new id for the new version of app.

adbenitez:
> For API levels, I am strongly against overriding library's min-sdk and relying solely on runtime API level checks. That requires dynamic switching, API level checks over multiple components, and need to keep all of them in sync over multiple places. Anything goes wrong and we will potentially get a runtime crash. This is not trivial unlike a simple addition where one check is enough.

it is also fine to just not offer calls in android 5-7 if they don't support the telecom API, but having a totally different apk and asking then android 5-7 people to move from current apk to that one with different app-id and submitting it to f-droid is too much effort better to just not support calls there then

wchen342:
> it is also fine to just not offer calls in android 5-7 if they don't support the telecom API, but having a totally different apk and asking then android 5-7 people to move from current apk to that one with different app-id and submitting it to f-droid is too much effort better to just not support calls there then

That actually still requires runtime checks and override, but at least we are not having two versions of code so that is an acceptable compromise.

wchen342:
I have added the override and runtime checks, so the call will only run on Android 8+ (API 26+). The call button is hidden now if the system is below 8.

I will mark this as ready now. Since the main functionality is working, the rest can be fixed in future issues or commits.

adbenitez:
@WofWca are you sure you are testing the latest version? that problems you mention I also had them but they are fixed in the meantime :thinking: 

(point 6 is a different thing not related to this PR and it is a problem I have already noticed, can happen in several situations)

WofWca:
Yes, turns out that was using an old version.

wchen342:
@adbenitez Also remind me to switch strings to use resource files before merge, I almost forgot that.

adbenitez:
it is looking really nice!!!

sadly now there is a cut shadow probably because the size of the container of the buttons is too small:

image

adbenitez:
also there is a problem with the video thumbnail, when incoming video is disabled, it looks nice with round corners, but if the other side also have video enabled, then our outgoing video preview thumbnail looks like this:

image

wchen342:
Hmm, maybe let us fix visual problems in a separate follow-up? I feel like we shall not keep this open and adding new fixes, if there is no major breaks.

adbenitez:
> Hmm, maybe let us fix visual problems in a separate follow-up? I feel like we shall not keep this open and adding new fixes, if there is no major breaks.

makes sense, I moved them to an issue

src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java
```diff
@@ -0,0 +1,1523 @@
...
+        }
+      }
+    );
+  }
+
+  // Helper (CallService)
+
+  public synchronized String fetchIceServers() throws RpcException {
+    if (activeAccId == null) {
+      throw new RpcException("No active account");
```
adbenitez: note: this is a bit confusing, RPC exceptions are meant to be raised by the RPC, I got already confused by this when I got this error and was thinking it was coming from core

src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java
```diff
@@ -0,0 +1,1523 @@
...
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+      Person person = new Person.Builder()
+        .setName(displayName)
+        .setIcon(icon)
+        .setImportant(true)
+        .build();
+
+      builder = new Notification.Builder(this.appContext, CHANNEL_ID_ONGOING)
+        .setSmallIcon(R.drawable.icon_notification)
+        .setContentTitle("Ongoing call")
```
adbenitez, line 1401: this shouldn't be hard-coded strings but use `R.string`

src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java
```diff
@@ -0,0 +1,1523 @@
...
+        .setContentText(statusText)
+        .setContentIntent(contentIntent)
+        .setStyle(Notification.CallStyle.forOngoingCall(person, hangupPendingIntent))
+        .addPerson(person)
+        .setOngoing(true)
+        .setCategory(Notification.CATEGORY_CALL);
+    } else {
+      builder = new Notification.Builder(this.appContext, CHANNEL_ID_ONGOING)
+        .setSmallIcon(R.drawable.icon_notification)
+        .setContentTitle("Ongoing call")
```
adbenitez, line 1411: same here: this shouldn't be hard-coded strings but use R.string

src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java
```diff
@@ -0,0 +1,1523 @@
...
+        .setCategory(Notification.CATEGORY_CALL)
+        .addAction(new Notification.Action.Builder(
+          Icon.createWithResource(this.appContext, R.drawable.baseline_call_end_24),
+          "Decline", declinePendingIntent).build())
+        .addAction(new Notification.Action.Builder(
+          Icon.createWithResource(this.appContext, R.drawable.baseline_call_24),
+          "Answer", answerPendingIntent).build());
+    }
+
+    notificationManager.notify(NOTIFICATION_ID_CALL, builder.build());
```
adbenitez, line 1370: all this hard-coded strings should be localized

src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java
```diff
@@ -0,0 +1,1523 @@
...
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+      if (!hasNotificationPermission()) {
+        Log.w(TAG, "POST_NOTIFICATIONS permission not granted, cannot update notification");
+        return;
+      }
+    }
+
+    String name = displayName.getValue();
+    if (name == null) {
+      name = "Unknown";
```
adbenitez, line 1441: also here: needs to be localized

src/main/java/org/thoughtcrime/securesms/calls/CallActivity.java
```diff
@@ -1,207 +1,871 @@
...
+        statusText.setText("Ringing...");
+        incomingCallPrompt.setVisibility(View.GONE);
+        bottomLayoutContainer.setVisibility(View.VISIBLE);
+        callerIconContainer.setVisibility(View.GONE);
+        answerModeSelector.setVisibility(View.GONE);
+        remoteAvatarView.setVisibility(View.VISIBLE);
+        break;
+
+      case CONNECTING:
+        statusText.setText("Connecting...");
```
adbenitez, line 524: all this statuses need to be localized instead of hard-coded

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -842,13 +745,13 @@ public Bitmap getAvatar(DcChat dcChat) {
...
-            notificationManager.cancel(tag, callId);
-        } catch (Exception e) { Log.w(TAG, e); }
-    }
+//    public void removeCallNotification(int accountId, int callId) {
+//        try {
+//            NotificationManagerCompat notificationManager = NotificationManagerCompat.from(context);
+//            String tag = "call-" + accountId;
+//            notificationManager.cancel(tag, callId);
+//        } catch (Exception e) { Log.w(TAG, e); }
+//    }
```
adbenitez: I guess this could be just removed

src/main/res/values/colors.xml
```diff
@@ -73,4 +73,16 @@
...
+    <color name="signal_dark_background">#1B1C1F</color>
+    <color name="signal_dark_surface">#1B1C1F</color>
+    <color name="signal_dark_surface_1">#23242A</color>
+    <color name="signal_dark_surface_2">#272A31</color>
+    <color name="signal_dark_surface_3">#2C2F37</color>
+    <color name="signal_dark_on_surface">#E2E1E5</color>
+    <color name="signal_dark_on_surface_variant">#BEBFC5</color>
+    <color name="signal_primary">#B6C5FA</color>
+    <color name="signal_red">#BA1B1B</color>
+    <color name="signal_green">#4CAF50</color>
```
adbenitez: it is better to prefix this with `call_` or `callscreen_` instead of `signal_` that doesn't say much where it is actually used

src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java
```diff
@@ -825,9 +825,6 @@ public void resumeWith(@NonNull Object result) {
   // Helper (CallService)
 
   public synchronized String fetchIceServers() throws RpcException {
-    if (activeAccId == null) {
-      throw new RpcException("No active account");
```
wchen342, line 829: It needs to return, I'll fix it later.

PR #4250, opened 2026-03-04
Remove notification when message is deleted

### LINKED ISSUE #4242: Notification stays after its message is deleted
- Android version: ~16
- Device:
- Delta Chat version: 2.35.0
- Expected behavior: Notification gets removed
- Actual behavior: It stays
  1. Have Delta Chat in the background
  2. Get a message notification
  3. Wait for the sender to delete the message.
- Screenshots:
- Logs:

On desktop we have an MR for this: https://github.com/deltachat/deltachat-desktop/pull/6059.

A part of this issue could be also the fact that editing a message doesn't update the notification.
But I anticipate hearing that "editing is only for fixing typos", which I don't think is a strong argument because we can't police people to use message editing only for fixing typos.

wchen342:
closes #4242


WofWca (approved):
Looks trivial. I didn't verify the event number and didn't test this.

src/main/java/org/thoughtcrime/securesms/connect/DcEventCenter.java
```diff
@@ -210,6 +210,7 @@ public long handleEvent(@NonNull DcEvent event) {
         break;
 
       case DcContext.DC_EVENT_MSGS_NOTICED:
+      case DcContext.DC_EVENT_MSG_DELETED:
         DcHelper.getNotificationCenter(context).removeNotifications(accountId, event.getData1Int());
```
adbenitez: `removeNotifications` removes ALL notifications of the given chatId, while the `DC_EVENT_MSG_DELETED` is triggered when a single message was deleted, then you are removing all notifications of that chat, instead just the notification of the specific message should be removed

adbenitez (changes requested):

src/main/java/org/thoughtcrime/securesms/connect/DcEventCenter.java
wchen342: The problem is there is one notification for the whole chat, so when a message is deleted, what do we do? Either remove the whole notification (which is equivalent to removing the notification by `chatId`), or do we rebuild the notification with a previous message?

wchen342: I made the notifications rebuild when a message is deleted, which is probably nicer.

adbenitez:
btw, at some point we probably should use the proper android API for chat apps notifications, that probably would have made this easier and allow people to expand notification of a chat and reply to specific message, right now this merging of notification makes it impossible to react to or mark as read a specific message etc.

https://developer.android.com/social-and-messaging/guides/communication/notifications-conversations

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -683,69 +725,85 @@ private void maybeAddNotification(int accountId, DcChat dcChat, int msgId, Strin
...
 
-            // add notification, we use one notification per chat,
-            // esp. older android are not that great at grouping
-            try {
-              notificationManager.notify(String.valueOf(accountId), ID_MSG_OFFSET + chatId, builder.build());
-            } catch (Exception e) {
-              Log.e(TAG, "cannot add notification", e);
-            }
+            // Show notification
+            notificationManager.notify(String.valueOf(accountId), ID_MSG_OFFSET + chatId, builder.build());
```
adbenitez: if this fails now the next attempt for sdk >=24 will be skipped, if there was a reason to remove the try-catch please add a comment

wchen342: None of the methods throw exceptions, so why is there a try...catch in the first place? Can you explain?

adbenitez: well, it is bad that this code didn't have any comment saying why the exception handling was there, in this particular case it was just copy-pasted from other parts of this class that also have that handling

but if it was there maybe it was for a reason, if you search the web you will find stuff about this throwing exception: https://stackoverflow.com/questions/72491592/notificationmanager-notify-throwing-securityexception-only-on-android-11-and-1

even if `notify()` is not declared to throw any exception, in practice it actually does under certain circumstances in certain OEM implementations

in general when doing a PR is better to only touch the exact lines that need to be touched to solve the problem without doing any other unrelated change, those better go in a dedicated separated PR

wchen342: Do we know if that actually occurred in the past?

Anyway, I added it back.

adbenitez: I remember seeing such security exception error but don't remember well or the context, but since the function doesn't document any exception throwing, it is likely it was added because of the problem actually happened

PR #4251, opened 2026-03-05
Feat: Unifiedpush

p1gp1g:
This PR adds UnifiedPush support.

UnifiedPush support is added only for flavors that does not support the Play Services right now. We can easily make it available for that flavor, if the play services aren't installed too - just let me know if you want.

If the user doesn't have any UnifiedPush service, nothing changes for them.
If the user has one UnifiedPush service (this is the case for most users), or many but they have a default distributor, the application uses it silently.
If the user has many UnifiedPush services, and hasn't selected their default yet, then a dialog is shown to inform about the incoming OS picker - to avoid context-less request. The OS picker shows that it is related to UnifiedPush but everybody may not *see* it. A user with many UnifiedPush services is definitely looking for this feature

Regarding the settings:
- I haven't added any setting.
- *The Notifications setting's description* was either a warning, or empty (or No/Yes if verbosity is enabled). It is slightly modified for user understanding: If the app uses FCM the description is *"with Google push"*, and if the app uses UnifiedPush, the description is *"with UnifiedPush"*. Not showing this information is often cause of misunderstanding from the users.
- When the user selects *Force Background Connection*, UnifiedPush is explicitly disabled

I hope and think this implementation doesn't add any friction to the user


build.gradle
```diff
@@ -52,6 +55,7 @@ android {
         buildConfigField("String", "TEST_ADDR", buildConfigProperty("TEST_ADDR"))
         buildConfigField("String", "TEST_MAIL_PW", buildConfigProperty("TEST_MAIL_PW"))
         buildConfigField("String", "NDK_ARCH", getNdkArch())
+        buildConfigField("String", "VAPID_KEY", "\"BG3RlodkBbX8YOu_3ZYCLuwJEkNKYo4QhS2iNp29Yon0Aw5m9uIgaAK7a0b5GfrC9cSLAR3JrcY5sQLZ0Xn9gCw\"")
```
p1gp1g: Change it with your notifier VAPID key (cf. https://github.com/chatmail/notifiers/pull/53)

src/main/java/org/thoughtcrime/securesms/connect/FetchWorker.java
```diff
@@ -10,6 +10,10 @@
 import org.thoughtcrime.securesms.util.Util;
 
 public class FetchWorker extends Worker {
+    /**
+     * Tag used for the periodic worker if push isn't available
+     */
+    public static final String periodicWorkTag = "FetchWorker";
```
p1gp1g: We move it to a constant there, because the registration and cancel of the periodic worker need the same tag, but doesn't occur in the same place

src/main/java/org/thoughtcrime/securesms/util/Prefs.java
```diff
@@ -161,7 +164,7 @@ public static int getStatsDeviceMsgId(Context context) {
     return getIntegerPreference(context, STATS_DEVICE_MSG_ID_PREF, 0);
   }
 
-  public static boolean isPushEnabled(Context context) {
+  public static boolean isFcmPushEnabled(Context context) {
```
p1gp1g, line 202: I've renamed because FCM isn't the only way to get push now, and this is about FCM

p1gp1g:
I've rebased the PR to fix the conflicts

link2xt:
I have tested a bit, got counter in "Sunup" next to Delta Chat at "1" and then "2", so it seems to work, pushed real VAPID key as a commit on top.

p1gp1g:
So I can try with a public chatmail server now ?

link2xt:
> So I can try with a public chatmail server now ?

Yes, https://github.com/chatmail/notifiers/pull/53 is deployed.

build.gradle
link2xt: This is set to the real key with a single commit on top.

src/main/java/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -90,6 +92,8 @@ public void onCreate(Bundle icicle) {
     eventCenter.addObserver(DcContext.DC_EVENT_REACTIONS_CHANGED, this);
     eventCenter.addObserver(DcContext.DC_EVENT_CONNECTIVITY_CHANGED, this);
     eventCenter.addObserver(DcContext.DC_EVENT_SELFAVATAR_CHANGED, this);
+
+    updateReminders();
```
p1gp1g, line 91: When `updateReminders` was in onResume, it was called multiple times, but 2 shared prefs were avoiding concurrent requests to the user: 

We had either:
- `Prefs.ASKED_FOR_NOTIFICATION_PERMISSION=true` => `DozeReminder.mayAskDirectly` is called
    1. Among other, check if `Prefs.DOZE_ASKED_DIRECTLY` is set
    2. if it isn't set, `Prefs.DOZE_ASKED_DIRECTLY` is set to `true` => so we don't ask twice for the exemption
- `Prefs.ASKED_FOR_NOTIFICATION_PERMISSION=false`:
    1. Set to `true` (so this branch effectively run only once)
    2. Ask the user for POST_NOTIFICATIONS permission
    3. When the permission is granted or denied, **onResume is called** => another instance of `updateReminders` is called, with `Prefs.ASKED_FOR_NOTIFICATION_PERMISSION=true`, cf above. => `mayAskDirectly` is called
    4. `DozeReminder.mayAskDirectly` is then called a 2nd time from `onAllGranted`, but ignored as it just has been called, and if doze exemption should be asked, `Prefs.DOZE_ASKED_DIRECTLY` is already `true`

To avoid the concurrent runs, it is better to run onCreate.

src/main/res/values/strings.xml
```diff
@@ -1178,4 +1178,8 @@
 
     <!-- device messages for updates -->
     <string name="update_2_0">What\'s new?\n\n💯 End-to-end encryption is reliable and forever now. Padlocks 🔒 are gone!\n\n✉️ Classic email without end-to-end encryption is marked with a letter symbol\n\n😻 New enhanced profile screen for all your contacts\n\n🔲 New button for quick access to apps used in a chat\n\n❤️ Please donate to help us remain independent and continue to bring improvements: %1$s</string>
+    <string name="dialog_introduce_unifiedpush_selection">You are about to select your UnifiedPush service.</string>
+    <string name="notification_reconfigure_notifications">The application has been unregistered for notifications. Open the app to reconfigure.</string>
+    <string name="pref_notification_desc_using_google_push">Using Google push</string>
```
r10s: if one would asks ppl on the street what "google push" is, most would probably guess a new fitness app for the next workout, or maybe a way to beam videos onto a wall :)

delta aims to be designed for average ppl, not just experts. if in doubt, it is fine if things are more complex for experts to understand.

that said, let's stick with a simple "on" when fcm is used. and only specify "using unified push" in case of unified push. experts need to know what  "on" means in details then, yes.

p1gp1g: OK, I've removed it

Note to myself: It would have been better using "Using Play Services" btw

src/main/java/org/thoughtcrime/securesms/service/UnifiedPushService.java
```diff
@@ -0,0 +1,162 @@
...
+  private static String TAG = "UnifiedPushService";
+
+  @Override
+  public void onNewEndpoint(@NonNull PushEndpoint pushEndpoint, @NonNull String _s) {
+    // We could have one endpoint/publicKey+auth per account, just like we could have one FCM key
+    // per account, but we follow FCM architecture. At least chatmail server can't know, with the push
+    // token, that 2 accounts are from the same device. With this unique push token, encrypted, the
+    // notifier server could know that a user is on 2 different servers.
+    Log.d(TAG, "New endpoint received");
+    String token = serializeForNotifiers(pushEndpoint);
```
link2xt, line 40: How stable is this token?

We print FCM token to the log in `src/main/java/org/thoughtcrime/securesms/LogViewFragment.java`:
https://github.com/deltachat/deltachat-android/blob/5fdf319cdaaca31137c919946470e45e699d5914/src/main/java/org/thoughtcrime/securesms/LogViewFragment.java#L266
But nothing about UnifiedPush is currently logged there, so I cannot see it from the log.

If the token may be different on each application start, it could be a problem that we will store a lot of tokens for the user on a chatmail relay, and it will have to notify all these tokens each time a message arrives.

This can be solved by either invalidating the token quickly, so it gets removed from the relay due to error, or by making sure that the token Delta Chat gives to chatmail/core does not change too often. Core makes sure to reuse old OpenPGP-encryption of the token if it does not change.

On Apple we have a problem that developers who reinstall the application get the same token each time, but encrypted differently with OpenPGP and then the same token is stored multiple times but encrypted differently on the relay. Relay then notifies the same token multiple times. The problem worked around by deduplicating in https://github.com/chatmail/notifiers/, but if decrypted token stays the same and does not become invalid, we will have to notify the same user many times for real.

p1gp1g: The push endpoint\* *should* be stable across re-registrations (From the specs: *The distributor SHOULD NOT create a new endpoint if a valid registration exist for the token and nothing has been updated.*). It isn't a *MUST NOT* because it doesn't break things: therefore, the implementation must work even if we have a distributor that does not follow this recommendation. But that isn't the expected behavior, and I don't know any distributor that doesn't follow that reco.

As you say, even with a stable endpoint, the encrypted token is always different on the relay, because the endpoint (or token) is re-encrypted, and I didn't see multiple push notification even with re-registration. After looking into core lib, I see that the library compare the new (clear text) token with the already-registered one, and update on the server only if it has changed: https://github.com/chatmail/core/blob/a95bf778687ce5770c753abbb30fb9bbf2d53c12/src/imap.rs#L1574 : this is why I didn't see multiple push notifications.

I don't see from the code if the *old_encrypted_device_token* is removed from chatmail if the token is updated. It would be great to do so. But even if it isn't removed, the old token is wiped after one push notification. If the old token isn't wiped:
- relay send old_token and new_token to notifiers:
- notifiers push to both endpoint
- the request to the old endpoint ends with 404/410/403/401 and the request to the new with 201
- the old endpoint is removed from the relay

=> To conclude, a same endpoint isn't stored twice, I don't know if an outdated token is removed from the client, but it will be wiped after the first notification if it isn't. So it works if the endpoint is rotated. And we don't expect lot of rotation.

\* The equivalent of the token for FCM or APNS with UnifiedPush is the push endpoint.

p1gp1g: After some tests, I confirm that the old_encrypted_device_token is not removed by the client. I'll see if I can do something on the core lib - this will be useful for other push systems too

link2xt: There is no API to remove tokens from the relay, they are removed because of errors or because they are outdated. `old_encrypted_device_token` and all the comparison code is to avoid encrypting the same token differently each time the app starts. As long as the app does not generate new tokens each time, it's fine.

With logging this can be tested easier by looking at plaintext token in the log, restarting the app, then checking if the token is not changed.

p1gp1g: I see. I've added logs for UnifiedPush: https://github.com/deltachat/deltachat-android/pull/4251/commits/bb4f2c77050fe1880d8815217ed6a9f412ebd3b2

adbenitez:
this PR needs to be rebased

p1gp1g:
The PR is rebased

adbenitez:
@p1gp1g hi, sorry but we just started using a code formatter so there are some merge conflicts again

instructions on how to run the code formatter are at:
https://github.com/deltachat/deltachat-android/blob/main/CONTRIBUTING.md#coding-conventions

p1gp1g:
No problem, I understand that it necessarily breaks some PR when it is deployed, and the conflicts are usually easy to resolve :+1:

p1gp1g:
Rebased and lint applied !

src/main/java/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
```diff
@@ -60,8 +61,9 @@ public static boolean isEligible(Context context) {
       Log.e(TAG, "Error calling getChatlist()", e);
     }
 
-    return !isPushAvailableAndSufficient(); // yip, asking for disabling battery optimisations makes
-    // sense
+    return !isPushAvailableAndSufficient(
+        context); // yip, asking for disabling battery optimisations
```
adbenitez: I think it is better to just put the comment above the line of code

p1gp1g: OK, I'm waiting for more comments to resolve this one in the same time :+1:

src/main/java/org/thoughtcrime/securesms/service/UnifiedPushService.java
```diff
@@ -0,0 +1,179 @@
...
+  public void onUnregistered(@NonNull String _s) {
+    Log.w(TAG, "Unregistered");
+    ResolvedDistributor res = UnifiedPush.resolveDefaultDistributor(this);
+    // We are using a single registration for all accounts, we don't need
+    // to wait a few seconds to get other instance unregistrations.
+    // We can process this unregistration now
+    if (res instanceof ResolvedDistributor.Found) {
+      String newDistrib = ((ResolvedDistributor.Found) res).getPackageName();
+      UnifiedPush.saveDistributor(this, newDistrib);
+      register(this);
```
wchen342, line 103: Can this cause a register/unregister loop if (1) the default is still on the system; (2) user logged out of the default distributor?

src/main/java/org/thoughtcrime/securesms/service/UnifiedPushService.java
```diff
@@ -0,0 +1,179 @@
...
+import org.unifiedpush.android.connector.FailedReason;
+import org.unifiedpush.android.connector.PushService;
+import org.unifiedpush.android.connector.UnifiedPush;
+import org.unifiedpush.android.connector.data.PublicKeySet;
+import org.unifiedpush.android.connector.data.PushEndpoint;
+import org.unifiedpush.android.connector.data.PushMessage;
+import org.unifiedpush.android.connector.data.ResolvedDistributor;
+
+public class UnifiedPushService extends PushService {
+  private static String TAG = "UnifiedPushService";
```
wchen342: Shall be `final`.

src/main/java/org/thoughtcrime/securesms/service/UnifiedPushService.java
p1gp1g: If the distributor sends UNREGISTER because the user logs out, it must disable UnifiedPush related Broadcast Receiver. And even if it doesn't, it would reply to `register` with REGISTRATION_FAILED, the function above. So we can't have a loop here

p1gp1g:
Once merged, I've a commit to enable UnifiedPush on the gplay flavor if the user doesn't have the Play Services installed: https://github.com/p1gp1g/deltachat-android/commit/7fac912ca4c76681de4f28f6180e68924b7fe42d

If you prefer, I can push it to this PR

src/main/java/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
p1gp1g: I forgot to do this patch with my last rebase

p1gp1g:
That's good, I see 5% (of the 40%) battery drain for the official build, <1% for the one with UnifiedPush :)

Note: Those numbers are informative, it isn't extremely reliable

adbenitez:
update on this: we were busy with a release deadline and now comes a bugfix and stabilization release so we can't merge any big feature for the next release either and are busy with that, sorry for the delays with your PR we have the hands full

p1gp1g:
I saw the recent big releases :+1: 

I will ping again in a few weeks to see if you are less busy

src/main/java/org/thoughtcrime/securesms/ConversationListFragment.java
adbenitez: I think it was in onResume for reasons, if it is only called on onCreate then this is never re-validated ex. if now a new classic email profile is added and you need then to ask for allowing to run in background because otherwise the classic email account doesn't support push notifications, even if you have google/unified push notifications

src/main/res/values/strings.xml
```diff
@@ -1176,6 +1176,10 @@
     <!-- disabling "Reliable service" will hide a the maybe annoying permanent-notification with the drawback that new-message-notifications get potentially unreliable -->
     <string name="perm_enable_bg_reminder_title">Tap here to receive messages while Delta Chat is in the background.</string>
     <string name="perm_enable_bg_already_done">You already allowed Delta Chat to receive messages in the background.\n\nIf messages still do not arrive in background, please also check your system settings.</string>
+    <string name="dialog_introduce_unifiedpush_selection">You are about to select your UnifiedPush service.</string>
+    <string name="notification_reconfigure_notifications">The application has been unregistered for notifications. Open the app to
```
adbenitez, line 1180: I was testing some unified push provider app, after I uninstalled it, fennec (firefox fork) was constantly bothering me about something like this from time to time every time I opened the app, hopefully this will not be the case here?

p1gp1g: Fennec and IronFox moved to a new co-maintained Firefox-module for the feature, and Fennec introduced a bug while trying to enable by default the feature. It won't happen there

src/main/java/org/thoughtcrime/securesms/ConversationListFragment.java
p1gp1g: cf my previous comment that explain that `updateReminders` is called many times but effectively run only once based on 2 shared prefs. It may have been in onResume before the behavior got updated to the current implementation, but it no longer make sense to have it there.

The current implementation (not this PR) used to have some bugs too. The code suggest that you want to run DozeReminder only if POST_NOTIFICATIONS permission is granted:
- there is a mechanism to ask for POST_NOTIFICATIONS permission only once. If the user deny POST_NOTIFICATIONS, DozeReminder isn't run (cf. `onAnyDenied`).
- but, if the permission is already asked (and denied), DozeReminder is run automatically (cf `else { DozeReminder.maybeAskDirectly(activity); }`). And `onResumed` is always called after the OS dialog to request POST_NOTIFICATIONS: it is always run, even if the user deny POST_NOTIFICATIONS
- So we have `onCreate` -> `onResumed` -> `updateReminders` (first instance) -> OS ask for POST_NOTIFICATIONS -> user deny -> `updateReminders` (first instance) doesn't call DozeReminder + `onResumed` -> `updateReminders` (second instance) -> `DozeReminder.maybeAskDirectly(activity)` (even without POST_NOTIFICATIONS)

p1gp1g:
Hi! This is the ping planned last month, to know if you'll have more time soon :)

adbenitez:
we are busy with other deadlines we have as part of funding projects, this we didn't receive any funding/support to take care of nor for reviewing etc. but also wrt future maintenance, so this have potential to cause a maintenance work for us while consuming time while we need to care for projects and deadlines that actually bring money/support to the table

p1gp1g:
Sorry if it sounded like a try to rush you, it was simply a gentle ping as suggested in April.

I'll (unfortunately?) have a lot of time in a few weeks, I'd love to offer you to review my own PR, but that wouldn't be very efficient :smiling_face_with_tear:

Anyway, no rush, let's continue when you've time

p1gp1g:
To observers interested in the feature: I've published a temporary fork with this PR: https://codeberg.org/s1m/delta-chaton

It will be automatically updated, as there should not be much conflicts between releases. I expect to discontinue this fork once (if) the support is merged

p1gp1g:
The fork will allow us to get some feedback in the same time

fredix:
Hello all,
IMHO, unifiedpush is a must have when you want to be outside Google .. Unifiedpush allow the user to choice is own notification server which is good for the privacy and good for the battery ..
Morover please allow user to pick non-default distributor when many are available..

adbenitez:
@p1gp1g could you rebase? I feel like this could go in the next release already (and personally I will be an user of this feature since I use a de-googled phone) but don't count on it because I am not the one taking the decisions and some folks are hesitant that merging this will bring more feature requests (someone was asking already for making configurable what notification provider to use, and the PR is not even merged, see comment above:

> Morover please allow user to pick non-default distributor when many are available..

p1gp1g:
I've rebased the PR!

> someone was asking already for making configurable what notification provider to use, and the PR is not even merged, see comment above

That's something I planned to do in a 2nd PR, someone told me to split the PR to not include a single new parameter in the first one.

This new option will be displayed only if the user has 2 or more distributors available

p1gp1g:
Note that I can include it right now if you want, should be very fast to implement

wchen342:
> @p1gp1g could you rebase? I feel like this could go in the next release already (and personally I will be an user of this feature since I use a de-googled phone) but don't count on it because I am not the one taking the decisions and some folks are hesitant that merging this will bring more feature requests (someone was asking already for making configurable what notification provider to use, and the PR is not even merged, see comment above:
> 
> > Morover please allow user to pick non-default distributor when many are available..

I mean, if @p1gp1g can be responsible for maintaining this part of the code base I think it is fine, just like how I maintain the call part. But I think others are not so enthusiastic about this idea so I am not sure this can be agreed up on.

p1gp1g:
I've pushed a few commits to avoid silent fails, and an option to select a non-default distributor

p1gp1g:
Regarding the maintenance, here are the number of commits (except version bump in gradle) in Molly, if we remove the ones related to MollySocket (a workaround required because Signal doesn't support web push (yet hopefully)). And even, the command bellow still counts commits related to MollySocket:

```
~/g/m/a/s/m/j/i/molly (molly-8.18.1)> git log --format='%ad _%s' --date=format:'%Y-%m' unifiedpush | grep -vi "Merge tag\|upstream\|mollysocket\|QR" | sed 's/ _.*//' | sort | uniq -c
      1 2023-03
      4 2024-11
      1 2024-12
      2 2026-02
```

And I'm of course around if needed, if there is something to fix

src/main/java/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
adbenitez: still not solved?

src/main/java/org/thoughtcrime/securesms/notifications/UnifiedPushNotifications.java
```diff
@@ -0,0 +1,115 @@
...
+        context, 0, intent, PendingIntent.FLAG_UPDATE_CURRENT | IntentUtils.FLAG_MUTABLE());
+
+    showNotification(context, R.string.notification_unifiedpush_reconfigure, R.string.notification_title_push_disabled, pi);
+  }
+
+  public static void showRegistrationFailed(Context context, FailedReason reason) {
+    @StringRes int resId = 0;
+    switch (reason) {
+      case INTERNAL_ERROR:
+        // No break;
```
adbenitez: better to remove this comment superfluous comment, consecutive case statements without body is obvious and more readable that all matches have the same code block

adbenitez:
@p1gp1g it is a bit weird that after selecting a provider in the new dialog you added suddenly the subtitle of "Using UnifiedPush" and the option to select one just disappears from notifications settings, user have to go back and enter the section again and then it appears again

this feels very weird and doesn't give the feedback to the user that the action succeeded but rather looks like unifiedpush was disabled

also after switching back and forth between 2 providers, now the provider says that there are 3 apps regestered and it is 3 entries for the same DC app, is this normal??? sounds like a bug

adbenitez:
also some people might install the app from the direct apk download, in that case the support for google play services will be in, but they might be using something like GrapheneOS,

if there is a way to detect that the system doesn't have google play services / push notifications and then if UnifiedPush provider is available use that, would be nice

**But this better comes as a next PR, this PR is already too big**

src/main/java/org/thoughtcrime/securesms/notifications/UnifiedPushNotifications.java
p1gp1g: I did it because some linters yell if you don't use a `break;` and don't comment as intented, but I can remove it

src/main/java/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
p1gp1g: I'm pushing a commit with the different lints, I forgot to do it

p1gp1g:
> @p1gp1g it is a bit weird that after selecting a provider in the new dialog you added suddenly the subtitle of "Using UnifiedPush" and the option to select one just disappears from notifications settings, user have to go back and enter the section again and then it appears again

There is likely something to fix. Do you mind sharing a screen record to illustrate it? The option "Receive notifications with", used to select the provider is removed when we select "Force background Connection". Do you think that'd be better to disable it instead of hiding it ?

> also after switching back and forth between 2 providers, now the provider says that there are 3 apps regestered and it is 3 entries for the same DC app, is this normal??? sounds like a bug

That's definitely a bug, and it needs to be patched on the library. I'm going to fix it.

EDIT: Here is [the patch](https://codeberg.org/UnifiedPush/android-connector/commit/aaa36005a73d6f1a0c24c56f26d89790210e3ffc). I'm publishing 3.3.4-rc1 with this fix. I'll publish a non-rc release in a few days, or once the PR is ready to be merged, in case there is something else to fix

> if there is a way to detect that the system doesn't have google play services / push notifications and then if UnifiedPush provider is available use that, would be nice

That's possible to detect if the system has the play services, and offer UnifiedPush if it doesn't - I can implement in a second PR

adbenitez:
@p1gp1g thanks for fixing, it seems you forgot to run the code formatter

p1gp1g:
I've just comitted a "spotlessApply" :+1:

adbenitez:
@p1gp1g there is a new release, the changelog entry should be moved to the "unreleased" section, maybe it is better to leave the changelog entry out until the end when we are about to merge it to avoid having to change every time, but I hope unifiedpush gets merged for next release

p1gp1g:
Done !

src/main/java/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
```diff
@@ -112,14 +114,15 @@ public static void dozeReminderTapped(Context context) {
     }
   }
 
-  private static boolean isPushAvailableAndSufficient() {
+  private static boolean isPushAvailableAndSufficient(Context context) {
     return ApplicationContext.getDcAccounts().isAllChatmail()
-        && FcmReceiveService.getToken() != null;
+        && (FcmReceiveService.getToken() != null
+            || UnifiedPushUtils.hasPushDistributor(context, true));
```
link2xt: Since there are no named arguments, i'd create a local variable `bool confirmed = true`. I also don't know if passing `true` is correct here and the name `confirmed` is not clear to me yet.

src/main/java/org/thoughtcrime/securesms/notifications/UnifiedPushUtils.java
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  /**
+   * Init UnifiedPush if FCM isn't enabled and UnifiedPush isn't disabled
+   *
+   * @param activity Activity
+   * @param initCallback Callback with [InitStatus]
+   */
+  public static void mayInitUnifiedPush(Activity activity, InitCallback initCallback) {
+    Log.d(TAG, "mayInitUnifiedPush");
```
link2xt: This log should be moved below early returns, otherwise users who use FCM will have `mayInitUnifiedPush` entries which is just noise.

build.gradle
```diff
@@ -250,6 +254,8 @@ dependencies {
     }
     gplayImplementation 'com.google.android.gms:play-services-location:21.3.0'
 
+    implementation('org.unifiedpush.android:connector:3.3.4-rc1')
```
link2xt, line 267: For reviewers: rendered documentation is at https://unifiedpush.org/kdoc/connector/ (found via the readme at https://codeberg.org/UnifiedPush/android-connector/). Don't know if Java/Maven has something similar to https://docs.rs in Rust for rendered library documentation, probably everyone uses IDEs.

build.gradle
```diff
@@ -10,6 +10,9 @@ repositories {
     maven {
         url "https://www.jitpack.io"
         name 'JitPack Github wrapper'
+        content {
+            includeGroupByRegex("com\\.github\\..*")
```
link2xt, line 14: What does this change do, is it related?

adbenitez:
there are merge conflicts, also the changelog changed again, better do what I recommended and postpone the changelog change

src/main/res/values/strings.xml
```diff
@@ -1259,6 +1259,18 @@
...
     <string name="perm_enable_bg_already_done">You already allowed Delta Chat to receive messages in the background.\n\nIf messages still do not arrive in background, please also check your system settings.</string>
+    <string name="dialog_introduce_unifiedpush_selection">You are about to select your UnifiedPush service.</string>
+    <string name="notification_title_push_disabled">Push disabled</string>
+    <string name="notification_unifiedpush_reconfigure">The application has been unregistered for notifications. Open the app to
+        reconfigure.</string>
+    <string name="notification_unifiedpush_internal_error">The UnifiedPush service indicates that it cannot process registrations at this time.\n
+        You may need to take some action there, such as logging in again.</string>
+    <string name="notification_unifiedpush_network_error">The UnifiedPush service indicates that it cannot process registrations at this time, because of lack of connection.\n
+        Please retry later.</string>
+    <string name="notification_unifiedpush_timeout_error">The UnifiedPush service is not responding.\n
```
link2xt, line 1281: Now that arcanechat has released 2.55 with this PR, there are users reporting that they got this notification.

Is it sent simply when the user has F-Droid version, no "force background connection" setting enabled, and no UnifiedPush distributor?

Similarly for the other "notification_unifiedpush_network_error", regardless of when this notification is displayed, what the user is expected to do with the "Please try again later." notfication?

These errors should probably be displayed somewhere in the settings or connectivity view, but not pop up as notifications that users cannot do anything about. E.g. for FCM notification users never get notifications like "cannot obtain the token, please try again later", if there is nothing that the user can do, maybe the error should only be logged but not shown anywhere, this is also one string less to translate.

If it is something expected, e.g. UnifiedPush distributor might simply be not installed, then it can be displayed in the connectivity view or user can be prompted to enable permanent notification with an alternative suggested to setup UnifiedPush distributor. If it is some unexpected error, then there are debugging tools in the distributor UI, the error may be displayed in the logs and connectivity view, but other than asking developers to fix the bug and then upgrade there is nothing the user can do.

link2xt (commented):
I tried to start reading it and this is why there are some comments, but better if android developers review the code.

Mostly wanted to leave the high level comment about error notifications with "try again later" wording.

src/main/res/values/strings.xml
adbenitez: what some users also reported is that if they have the "force background connection" that was enabled automatically before this PR, then the UnifiedPush doesn't work even if they have a UP provider

link2xt: > if they have the "force background connection" that was enabled automatically before this PR, then the UnifiedPush doesn't work even if they have a UP provider

I thought this is on purpose, to not enable UP for those who don't need it and have a setup working already even though it is with the app running in background all the time. So we don't even have the code trying to discover the distributor running when someone has e.g. Conversations set up.

adbenitez: well if they have UP provider app, is because they are using unified push and it makes sense to use that, the reality is that they hear the new about UP support, install UP app and then it just doesn't work and gives some error or so

they didn't enabled the force-background-connection, it was enabled automatically by us and it is no way obvious to the user that it is blocking UP distributor app integration from working, what they actually did manually

also that permanent notification drains battery faster, if the user installed UP app it makes sense to use that in any case

I think we actually should never auto-enable that permanent notification, it also makes the app hard to re-start if it gets in some frozen state etc

src/main/java/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
adbenitez: this is not much of a problem if you are using an IDE like Android Studio, there you see the argument name in-place with the value

build.gradle
p1gp1g: It is a security practice, to avoid someone using jitpack to poison a non-jitpack library

src/main/res/values/strings.xml
p1gp1g: > Similarly for the other "notification_unifiedpush_network_error", regardless of when this notification is displayed, what the user is expected to do with the "Please try again later." notfication?

It is shown to the user if they try to register for push notification when they don't have any network (no wifi/no data), but only if it happens the first time of the registration, else we continue to use the previous endpoint. Maybe we can reword a bit, and say that the registration will happen when they will have network again? Or we can simply ignore that error, and it will be fixed by itself

p1gp1g: > they didn't enabled the force-background-connection, it was enabled automatically by us and it is no way obvious to the user that it is blocking UP distributor app integration from working, what they actually did manually

Oh. I actually didn't think about users who update from a version that had auto-enabled the setting. What do you think about adding a line to the force-background-connection setting, if a UnifiedPush distrib is available: *Disable to use UnifiedPush* ?

src/main/java/org/thoughtcrime/securesms/notifications/UnifiedPushUtils.java
```diff
@@ -0,0 +1,249 @@
...
+
+import android.app.Activity;
+import android.content.Context;
+import android.content.Intent;
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageManager;
+import android.os.Build;
+import android.util.Log;
+import androidx.appcompat.app.AlertDialog;
+import javax.annotation.Nullable;
```
adbenitez: this is wrong `androidx.annotation.Nullable` should be used instead

src/main/java/org/thoughtcrime/securesms/service/UnifiedPushService.java
```diff
@@ -0,0 +1,140 @@
...
+import static org.unifiedpush.android.connector.ConstantsKt.INSTANCE_DEFAULT;
+
+import android.app.ForegroundServiceStartNotAllowedException;
+import android.content.Context;
+import android.os.Build;
+import android.util.Log;
+import androidx.annotation.NonNull;
+import androidx.annotation.RequiresApi;
+import androidx.work.WorkManager;
+import javax.annotation.Nullable;
```
adbenitez: same here `androidx.annotation` should be used

src/main/res/xml/preferences_notifications.xml
```diff
@@ -1,12 +1,18 @@
 <?xml version="1.0" encoding="utf-8"?>
 
-<PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android">
+<PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android"
```
adbenitez:
```suggestion
<PreferenceScreen
    xmlns:android="http://schemas.android.com/apk/res/android"
```

src/main/java/org/thoughtcrime/securesms/notifications/UnifiedPushUtils.java
adbenitez:
```suggestion
import androidx.annotation..Nullable;
import androidx.appcompat.app.AlertDialog;
```

src/main/java/org/thoughtcrime/securesms/service/UnifiedPushService.java
adbenitez:
```suggestion
import androidx.annotation.NonNull;
import androidx.annotation.Nullable;
import androidx.annotation.RequiresApi;
import androidx.work.WorkManager;
```

src/main/res/values/strings.xml
adbenitez: > What do you think about adding a line to the force-background-connection setting, if a UnifiedPush distrib is available: Disable to use UnifiedPush ?

not good, there is no reason to not use UnifiedPush even if that permanent notification is enabled (user can always not grant access from the provider app to Delta Chat etc.), better just always use UnifiedPush if provider is available and check if the permanent notification setting was manually set, if the preference value is undefined/not set meaning that the user is just using defaults, then just use unified push instead of the permanent notification

src/main/java/org/thoughtcrime/securesms/notifications/UnifiedPushUtils.java
```diff
@@ -0,0 +1,249 @@
+package org.thoughtcrime.securesms.notifications;
+
+import android.app.Activity;
+import android.content.Context;
+import android.content.Intent;
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageManager;
+import android.os.Build;
+import android.util.Log;
+import androidx.annotation..Nullable;
```
adbenitez:
```suggestion
import androidx.annotation.Nullable;
```

src/main/java/org/thoughtcrime/securesms/notifications/UnifiedPushUtils.java
```diff
@@ -0,0 +1,249 @@
...
+   *
+   * @param activity Activity
+   * @param initCallback Callback with [InitStatus]
+   */
+  public static void mayInitUnifiedPush(Activity activity, InitCallback initCallback) {
+    Log.d(TAG, "mayInitUnifiedPush");
+    if (Prefs.isFcmPushEnabled(activity)) {
+      initCallback.onInit(InitStatus.HasPush);
+      return;
+    }
```
adbenitez:
```suggestion
    if (Prefs.isFcmPushEnabled(activity)) {
      initCallback.onInit(InitStatus.HasPush);
      return;
    }
    Log.d(TAG, "mayInitUnifiedPush");
```

src/main/java/org/thoughtcrime/securesms/notifications/UnifiedPushUtils.java
```diff
@@ -0,0 +1,249 @@
...
+    Log.d(TAG, "countAvailableDistributor");
+    return UnifiedPush.getDistributors(context).size();
+  }
+
+  public static @Nullable String getDistributorName(Context context) {
+    String distributor = UnifiedPush.getSavedDistributor(context);
+    if (distributor == null) return null;
+    try {
+      ApplicationInfo ai;
+      if (Build.VERSION.SDK_INT >= 33) {
```
adbenitez:
```suggestion
      if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
```

src/main/java/org/thoughtcrime/securesms/service/UnifiedPushService.java
```diff
@@ -0,0 +1,140 @@
...
+    prefixedToken = token;
+    getDcAccounts().setPushDeviceToken(token);
+    KeepAliveService.maybeStopSelf(this);
+    WorkManager.getInstance(this).cancelUniqueWork(FetchWorker.periodicWorkTag);
+  }
+
+  @Override
+  public void onMessage(@NonNull PushMessage _pushMessage, @NonNull String _s) {
+    Log.d(TAG, "New push message received");
+    if (Build.VERSION.SDK_INT < 31) {
```
adbenitez:
```suggestion
    if (Build.VERSION.SDK_INT < Build.VERSION_CODES.S) {
```

src/main/res/values/strings.xml
adbenitez: ok, reading more carefully actually it is already the case that if the setting was not touched manually and if unifiedpush is available then the permanent notification is not started, so maybe it is ok as things are since only very few people go there and change that setting manually and the few that do are advanced users anyways, the ones that install unifiedpush providers are also quite advanced users so they will figure it out, they will likely want to disable permanent notification anyways after installing unifiedpush provider

what should not happen is error toasts that the user can't do anything about as @link2xt said above

adbenitez:
@p1gp1g  to summarize the status of this PR: overall looks good to me and looking forward to merge it, the only pending problem at the moment is https://github.com/deltachat/deltachat-android/pull/4251/changes#r3927378188

src/main/res/values/strings.xml
p1gp1g: *notification_unifiedpush_internal_error*, *notification_unifiedpush_network_error* and *notification_unifiedpush_timeout_error* may be shown only during the first registration to a distributor.

Does your comment apply to both 3 or only the 2 lasts ? As the first one explicitly say that something must be done on the distributor side?

Would changing "Please retry later." to "Please retry when your network access is restored." work for the network related issue ?

Else, should we "silently fail", and wait for the next re-registration to fix the issue by itself in case of network issue ? Maybe timeout too ?

And finally, do you have other ideas ?

p1gp1g:
I've rebased and squashed your last commits (to help maintaining the temporary fork)

adbenitez:
> I've rebased and squashed your last commits (to help maintaining the temporary fork)

maybe you should squash everything into a single commit because every time you rebase with that hell lot of 34 commits it is hiding the comments made previously I am struggling to find my own last comment

src/main/res/values/strings.xml
adbenitez: yes there should not be "please retry later" or anything on network issues, there should not be toast out of the blue when the user are in random parts of the app and they don't even know where the error comes from, any necessary error should likely be prefixed by "UnifiedPush: ..." but as I see it none of them are necessary the registration status is visible in the distributor app and in Delta Chat settings ("Using UnifiedPush") right?

p1gp1g: If one of these errors occur, UnifiedPush is disabled at this moment: https://github.com/deltachat/deltachat-android/pull/4251/changes#diff-62761075dc20c6a30fa31e07daddfa4bef934acb8e55a0da61e4593a3fee9a7cR208

What we can do is:
1. Do not disable UnifiedPush when the error is caused by a missing network access, to let the app re-register when network is back, and remove the "please retry later."

```diff
-     <string name="notification_unifiedpush_network_error">The UnifiedPush service indicates that it cannot process registrations at this time, because of lack of connection.\n
-        Please retry later.</string>
+     <string name="notification_unifiedpush_network_error">The UnifiedPush service indicates that it cannot process registrations at this time, because of lack of connection.</string>
```

2. Reword the message shown on timeout. This should be handled as the other internal error by the user

```diff
     <string name="notification_unifiedpush_timeout_error">The UnifiedPush service is not responding.\n
-         Please try again later.</string>
+         You may need to take some action there, such as logging in again.</string>
```

3. If you prefer avoiding the toasts, we can add the error content to the broadcast receiver that updates the UI based on error to contain a message, that is shown on the NotificationsPreferenceFragment

adbenitez: what I don't understand is: registration is automatic, no manual action of the user, so how would the user "try later" etc. this should just go away and just retry to register later

for timeout also just log it so it is in the logs, there is not much that the user can do with this, just show if UnifiedPush is successfully working/registered as it is done via the "Using UnifiedPush" subtitle 

there is the situation when the user can select a distributor manually when there are several, there it would make sense to report an error after **manual** selection

p1gp1g: OK, I see! I had the workflow in mind where a user enables UnifiedPush - which would be the equivalent to disable "Force Background Connection" here.

So when the user disabled "Force Background Connection", we should try to use the UnifiedPush distributor, and if it fails we simply log that an error occurred, and "Using UnifiedPush" should be shown only if we have successfully received an endpoint.

I'm writing a patch

p1gp1g: Hm, it actually make sense to have those notification in one case: when the user tries to change the distributor, when they have many. Don't you think ?

Fuseteam: they did mention that earlier:

> there is the situation when the user can select a distributor manually when there are several, there it would make sense to report an error after **manual** selection

adbenitez: yep, that is what I said at the end of my last message:

> there is the situation when the user can select a distributor manually when there are several, there it would make sense to report an error after **manual** selection

p1gp1g: https://github.com/deltachat/deltachat-android/pull/4251/commits/d37f4002ac4ddc2d1b0cd3a8a2bd43a9afe128e4 What do you think ?

adbenitez:
@p1gp1g you need to run ./gradlew spotlessApply before submitting your code, CI is failing, thanks in advance

Issue #4158, opened 2026-01-12
The difference between 'deleting' and 'leaving' a chat is confusing


Hocuri:
<details><summary>We got a report that the difference between "deleting" and "leaving" a channel is confusing.</summary>
<p>

> I have encountered two main problems [with channels] so far:
> - [...] 
> - There is a confusing difference between 'deleting' and 'leaving' a chat.
> 
> [...]
> so I have to tell people that hey guyz if you want to leave a channel please hit leave first

</p>
</details> 

Solution: As long as you are a member, only have a "leave" button in channels and groups, no "delete" button. Modify the confirmation dialog:

> Leave group "groupname"?
>
> - Leave
> - Leave and delete for me
> - Cancel

Note that this needs to be changed both in the chatlist (when long-tapping on a chat) and inside the chat.

<details><summary>Click here to see reasoning</summary>
<p>

- As opposed to TG, you can't see the full chat history upon joining a chat. This means that it's nice to be able to leave a group and keep the chat history; when you rejoin, you will lose the messages that were sent while you were gone, but you still have the old messages.
- Therefore, we should copy WA, not TG.
- A "leave" message doesn't disturb other members in any way anymore; it doesn't pop up the chat, and it doesn't recreate the chat if it was deleted.
- One disadvantage of leaving rather than deleting is that a potential spammer will know that you're using this account. However, you can't really defend against a spammer who has your contact, other than switching to a new profile, anyways. So, I think that this is not a concern.

</p>
</details> 

**Open question:** Exact wording in the confirmation dialog - Should the first action be "Leave" or "Leave group"? Should we call it "exit" rather than "leave"? What should the ordering of the options be?

adbenitez:
I can confirm I have seen people deleting chats expecting that makes them leave the group and get annoyed when it reappears, there is a similar problem with bots, people delete the 1:1 chat and expect the bot to be stopped, but that needs to be fixed in core, sending a "delete this 1:1 chat" event to bots contacts so the user's chat/data is not kept and bot can also do other db cleanups

adbenitez:
> Exact wording in the confirmation dialog - Should the first action be "Leave" or "Leave group"? Should we call it "exit" rather than "leave"?

[...]

PR #4262, opened 2026-03-10
leave and delete

r10s:
this PR forces leaving a channel/group before deletion, whenever possible.

- this removes confusion, as usually, if the user wants to have a chat deleted, they do not expect them to pop up over and over again.

- it simplifies UI - there is no either a "Leave" or a "Delete for Me" - but not both at the same time. when tapping "Leave", you can either only "Leave" or "Leave & Delete for Me"

- it fixes the bug of missing avatars after deletion (group/channel avatars are usually not resent when you get a message after leaving)

- cleaning up is still possible using "Clear Chat". also leaving and keeping history is still possible.

- unencrypted groups are left out for leaving

- for multi-select, if a single group/channel is selected, the alert dialog shows "Leave & Delete for Me" - otherwise "Delete"

roughly, this is what whatsapp and probably other messengers are doing.

in the past, we were hesistent to "leave" a group, as this messes up group member list, with group consistency, this seems to be not longer a major concern.

closes https://github.com/deltachat/deltachat-android/issues/4158

once merged and strings are pushed, we need to file corresponding issues for desktop/iOS

### DIFF
--- a/src/main/java/com/b44t/messenger/DcChat.java
+++ b/src/main/java/com/b44t/messenger/DcChat.java
@@ -62,7 +62,7 @@ public class DcChat {
       if (isInBroadcast()) {
         final int[] members = dcContext.getChatContacts(getId());
         return Util.contains(members, DcContact.DC_CONTACT_ID_SELF);
-      } else if (isMultiUser() && isEncrypted() && canSend() && !isOutBroadcast() && !isMailingList()) {
+      } else if (isMultiUser() && isEncrypted() && canSend() && !isOutBroadcast()) {
         return true;
       }
       return false;


src/main/java/com/b44t/messenger/DcChat.java
```diff
@@ -56,6 +58,16 @@ public boolean isMultiUser() {
       return type != DC_CHAT_TYPE_SINGLE;
     }
 
+    public boolean shallLeaveBeforeDelete(DcContext dcContext) {
+      if (isInBroadcast()) {
+        final int[] members = dcContext.getChatContacts(getId());
+        return Util.contains(members, DcContact.DC_CONTACT_ID_SELF);
+      } else if (isMultiUser() && isEncrypted() && canSend() && !isOutBroadcast() && !isMailingList()) {
```
adbenitez: `isEncrypted()` already implies `!isMailingList()`, but it doesn't hurt much to also leave the check if you think there will be encrypted mailing lists at some point in the future :D

Hocuri: Minor note: Instead of `isMultiUser() && isEncrypted() && canSend() && !isOutBroadcast() && !isMailingList()`, we could probably just check `isGroup() && isEncrypted() && canSend()`? (and create a function `isGroup()` for this)

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -528,15 +528,12 @@ public boolean onPrepareOptionsMenu(Menu menu) {
...
+    if (dcChat.shallLeaveBeforeDelete(DcHelper.getContext(context))) {
+      if (dcChat.isInBroadcast()) {
         menu.findItem(R.id.menu_leave).setTitle(R.string.menu_leave_channel).setVisible(true);
-      } else if (dcChat.isEncrypted()
-          && dcChat.canSend()
-          && !dcChat.isOutBroadcast()
-          && !dcChat.isMailingList()) {
-        menu.findItem(R.id.menu_leave).setVisible(true);
       }
+      menu.findItem(R.id.menu_leave).setVisible(true);
```
Hocuri: Minor note: The menu_leave item is set to visible twice here

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -528,15 +528,12 @@ public boolean onPrepareOptionsMenu(Menu menu) {
       menu.findItem(R.id.menu_ephemeral_messages).setVisible(false);
     }
 
-    if (isMultiUser()) {
-      if (dcChat.isInBroadcast() && !dcChat.isContactRequest()) {
+    if (dcChat.shallLeaveBeforeDelete(DcHelper.getContext(context))) {
+      if (dcChat.isInBroadcast()) {
         menu.findItem(R.id.menu_leave).setTitle(R.string.menu_leave_channel).setVisible(true);
```
adbenitez:
```suggestion
        menu.findItem(R.id.menu_leave).setTitle(R.string.menu_leave_channel);
```

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
r10s: good catch! i applied the corresponding  code change from @adbenitez below

src/main/java/com/b44t/messenger/DcChat.java
r10s: no, i do not expect encrypted mailing lists anytime soon at scale :)

i added a commit with your suggestion

r10s: yes, maybe. however, `isMultiUser` is used pervasive also elsewhere, so needs to be understood anyways, so i leave that part as is

PR #4265, opened 2026-03-10
setup spotless for code formatting

adbenitez:
currently code has different formatting per file, sometimes even in the same file there are different uses of 4 or 2 spaces, this PR solves all that and removes the burden of manually caring for code format from the developers

also the code formatter automatically sorts and removes unused imports

- [x] configure spotless gradle plugin in `build.gradle`
- [x] add CI workflow to check for code format
- [x] update `CONTRIBUTING.md` to mention code formatting
- [x] apply code format gradually to the code in different PRs before merging this, to avoid a huge PR adding a lot of changes


spotless/eclipse-wtp-xml.prefs
```diff
@@ -0,0 +1,6 @@
+eclipse.preferences.version=1
+lineWidth=100
+splitMultiAttrs=true
+indentMultipleAttributes=true
+indentationChar=space
+indentationSize=4
```
wchen342, line 6: Maybe `spaceBeforeEmptyCloseTag=true`?

build.gradle
```diff
@@ -151,6 +152,30 @@ android {
 
 }
 
+spotless {
+    java {
+        target 'src/*/java/**/*.java'
+        targetExclude 'src/main/java/chat/delta/**' // ignore auto-generated code
```
wchen342, line 158: Also `build` paths, in case we want to run it locally.

build.gradle
```diff
@@ -151,6 +152,30 @@ android {
...
+        googleJavaFormat()  // Google style = 2-space indent, matches Android Studio defaults
+        removeUnusedImports()
+        trimTrailingWhitespace()
+        endWithNewline()
+    }
+    format 'xml', {
+        target 'src/*/res/**/*.xml', 'src/*/AndroidManifest.xml'
```
wchen342, line 165: Also need to exclude `build`.

build.gradle
adbenitez: build path is not included, only stuff inside src/

spotless/eclipse-wtp-xml.prefs
adbenitez: that seems to be the default

build.gradle
```diff
@@ -151,6 +152,30 @@ android {
...
+        eclipseWtp('xml').configFile('spotless/eclipse-wtp-xml.prefs')
+        trimTrailingWhitespace()
+        endWithNewline()
+    }
+    groovy {
```
wchen342: [Shouldn't this be `groovyGradle`?](https://github.com/diffplug/spotless/blob/main/plugin-gradle/README.md#groovy)

build.gradle
wchen342: Ah ok you are using `src`.

build.gradle
adbenitez: that requires an additional plugin and it seems it will also touch java files and then extra config is needed to exclude java, so I just used this other one that seems simpler

spotless/eclipse-wtp-xml.prefs
wchen342: I can't seem to find the docs on this though. It is a minor thing so I guess it is fine.

build.gradle
wchen342: What do you mean? The example in the doc has `excludeJava()` at the end for `groovy`, and it says the `groovy` one is the one that includes java?

adbenitez: oops! in fact now re-reading the docs I completely mixed that up :rofl: didn't saw the groovyGradle section below at all!

adbenitez: ok, fixed, thanks a lot for spotting!

spotless/eclipse-wtp-xml.prefs
adbenitez: anyways it doesn't hurt to be explicit here so I added it, thanks

wchen342:
We shall also update `.editorconfig` to match the spotless config.

PR #4272, opened 2026-03-11
cache compiled core

adbenitez:
currently every time a PR is open or if any small change/commit is added afterwards to a PR, every time we have to wait for core to compile which takes several minutes

this PR caches the compiled core so it is only build once as long as the core submodule or any of the NDK compilation files don't change

this reduces the apk preview generation time significantly (from 13-14min to 3-4min)

(NOTE: first the spotless PR needs to be merged)

#skip-changelog


wchen342 (approved):
This is nice to have!

PR #4275, opened 2026-03-13
avoid race in processComposeControls during "reply privately"

### LINKED ISSUE #3508: sometimes "reply privately" doesn't quote the message
when you select a message in a group and click the "reply privately" option, sometimes it happens that the 1:1 chat is opened but the message you wanted to reply privately is not shown in the quote in the draft area.

I just tried with one message and it happened, but then going again to the chat and trying again, now it works, so it is not so easy to reproduce

adbenitez:
`processComposeControls()` spawns a background thread that uses the class field `chatId`, but by the time `setDraft(chatId, null)` is called to clear the draft, the `chatId` might be already the new selected chat for "reply privately"

close #3508

### DIFF
--- a/.github/workflows/preview-apk.yml
+++ b/.github/workflows/preview-apk.yml
@@ -14,14 +14,12 @@ jobs:
       - uses: actions/checkout@v5
         with:
           submodules: recursive
-      - uses: Swatinem/rust-cache@v2
-        with:
-          working-directory: jni/deltachat-core-rust
+
       - uses: actions/setup-java@v5
         with:
           java-version: 17
           distribution: 'temurin'
-      - uses: android-actions/setup-android@v3
+
       - uses: actions/cache@v4
         with:
           path: |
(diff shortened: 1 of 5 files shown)


src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1266,48 +1267,36 @@ else if (!body.isEmpty()){
...
-    if (refreshFragment) {
-      fragment.reload(recipient, chatId);
-      try {
-        int accId = rpc.getSelectedAccountId();
-        DcHelper.getNotificationCenter(this).updateVisibleChat(accId, chatId);
-      } catch (RpcException e) {
-        Log.e(TAG, "rpc.getSelectedAccountId() failed", e);
-      }
-    }
-
```
adbenitez, line 1310: it is not clear to me why this was needed at all, sendComplete() is only called on sending messages, I can't think on any case where pressing the send button could trigger a change of chat (maybe if you click a notification super-fast after clicking the send button?? in any case you don't want in that thread of the "send message" action then revert the chatId/selected chat to the chat where the message was sent

wchen342: I guess it is used to refresh that chat after send? But if everything works without it then I guess it is unnecessary.

PR #4276, opened 2026-03-13
Add proper container to AAC streams used in voice messages

### LINKED ISSUE #4274: Problem with playing voice messages in android 7
some android 7 user reported to me and sent videos showing that there the progress bar doesn't move when you play a voice message sometimes, and trying to move the bar while it is playing reverts the position back to where it was

https://github.com/user-attachments/assets/b7b8b924-4f6c-44a1-945c-aef6db5827fc

https://github.com/user-attachments/assets/7f46cd63-91c5-4c04-86b6-d6f3c034299d

wchen342:
Shall fix #4274. The problem is current AAC implementation doesn't have proper seekable metadata thus old devices cannot seek them correctly.

closes #4274

#skip-changelog

### DIFF
--- a/.github/workflows/code-format.yml
+++ b/.github/workflows/code-format.yml
@@ -1,0 +1,35 @@
+name: Code Format
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+jobs:
+  spotless:
+    name: Check code format (Spotless)
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v5
+      - uses: actions/setup-java@v5
+        with:
+          java-version: 17
+          distribution: temurin
+      - uses: actions/cache@v4
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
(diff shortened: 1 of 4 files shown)


adbenitez:
there seems to be unrelated indentation changes, this file is already formatted with spotless

adbenitez:
just in case we need to test that this doesn't break something for iOS receivers, @r10s maybe you want to have a look at it, otherwise I will test on iOS test device on monday

src/main/java/org/thoughtcrime/securesms/audio/AudioCodec.java
```diff
@@ -43,49 +44,150 @@ public AudioCodec(Context context) throws IOException {
...
+                  boolean running = isRunning();
+
+                  handleCodecInput(audioRecord, audioRecordData, mediaCodec, running);
+                  int codecOutputBufferIndex = mediaCodec.dequeueOutputBuffer(bufferInfo, 2000);
+
+                  while (codecOutputBufferIndex != MediaCodec.INFO_TRY_AGAIN_LATER) {
+                    if (codecOutputBufferIndex == MediaCodec.INFO_OUTPUT_FORMAT_CHANGED) {
+                      // Get the format to add track to muxer
+                      if (muxerStarted) {
+                        throw new RuntimeException("Output format changed after muxer started");
```
adbenitez: will this cause a crash? better to just log an error and abort

wchen342: They are caught by outer `try`, but they are unlikely to happen anyway so I changed them to logging.

PR #4289, opened 2026-03-18
hide relays from contacts (unpublish relays)

### LINKED ISSUE #4260: UI changes needed for multi-relay for core 2.45+
in this issue, i am here trying to sum up some of the UI changed that are needed to finalise multi-relay. to get a shared view.

i am creating the issue here, as strings are driven by android and it makes sense to start on android therefore. but i am not sure, when there is a good start, or which api are already there.

partly, these things are tracked in core issues, however, i am not aware of a high-level overview. @link2xt @Hocuri  please correct, if i am wrong :) [^1]

1. ~~add  **Allow Unencrypted Messaging** to classic relay adding/editing
   this replaces "Show All Email"~~ EDIT: as it is unclear when exactly that is removed from core, this is not actionable right now

2. ~~add **IMAP-Folder**  to classic relay adding/editing
  this replaces: **Only fetch from Delta Chat folder**
  this depends on https://github.com/chatmail/core/pull/7780~~ EDIT: as it is unclear when exactly that is removed from core, this is not actionable right now

3. add  **Unpublish** to the relay list, to let a relay fade out (if you want to receive but not give it to others)
  this depends on https://github.com/chatmail/core/pull/7994
  See the discussion [further down](https://github.com/deltachat/deltachat-android/issues/4260#issuecomment-4065759654)
   ~~this replaces "Default" relays (previously, there was only one relay published, non-default were implicitly unpublished)~~ EDIT: "Default" will stay, see below

4. ~~remove **Move automatically to DeltaChat folder**  
  this depends on https://github.com/chatmail/core/pull/7780
  alternatives were shown to the user already in 2.43~~ EDIT: as it is unclear when exactly that is removed from core, this is not actionable right now

5. ~~remove **Delete Messages from Server**
  chatmail handles that automatically,
  this option is needed only for non-chatmail, and there, ppl usually have more space, or there are other options.
  question is, how can core make sure to _not_ delete messages, as this is risky if ppl are using Delta Chat as shared account, even if we tell them not to do~~ EDIT: as it is unclear when exactly that is removed from core, this is not actionable right now

by these changes, one should also be able to drop `is_chatmail` completely from UI.

[^1]: this issue is not about the pro or cons about the options, this was discussed already in the core team (i personally also was not fully part of that). there are as many opinions as devs and users :)

r10s:
this PR adds the "unpublishing" functionality to the UI, however, instead of saying "unpublish" and "default", already saying in the name what it does:

- Used for sending
- Hidden from contacts

the PR requires current main core checked out, or just relased 2.46 - the core change is not part of this PR, as usual.

https://github.com/user-attachments/assets/cb9a45d3-d8c9-4322-9439-fb0b90a47120

  

- adding hint "messages are received on all relays" to the list can be done in another PR. not sure what is an easy approach for that, can also go to another PR. but the string should in any case be reviewed already, on iOS and desktop that is easier

closes #4260

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -760,7 +760,7 @@
     <string name="remove_transport">Remove Relay</string>
     <string name="edit_transport">Edit Relay</string>
     <string name="used_for_sending">Used for sending</string>
-    <string name="hide_from_contacts">Hide From Contacts</string>
+    <string name="hide_from_contacts">Hide from Contacts</string>
     <string name="hidden_from_contacts">Hidden from contacts</string>
     <!-- Hint for the list of relays -->
     <string name="transport_list_hint">Messages are received on all relays.</string>


src/main/java/org/thoughtcrime/securesms/relay/RelayListActivity.java
```diff
@@ -237,21 +239,27 @@ private void onRelayEdit(EnteredLoginParam relay) {
...
         .setPositiveButton(
-            R.string.ok,
-            (dialog, which) -> {
+            R.string.hide_transport_only,
+            (d, which) -> {
+              // TODO: set state to jsonrpc
+            })
+        .setNegativeButton(
+            R.string.remove_transport_now,
+            (d, which) -> {
```
Hocuri: What to do if you want to publish an unpublished relay again?

We could just tell users "this is not possible directly, but you can select the relay as primary, and then select a different primary relay again, because setting as primary will automatically set it as published".

But then the "Hide from contacts" button shouldn't be there if the relay is already unpublished.

r10s: yip, it is easy enough to get that one can use it for sending again, by just tapping around a bit. but i think, this is not a usual case, i mean, one was close to deleting the whole thing and then in the last moment decided to hide it only.

the natural way is to delete it then - and yes, we should not offer hide in that case, i will do a commit.

r10s: hm, otoh, not offering "hide" if already hidden would require a new wording for the text in the alert. not sure if it is worth the effort.

maybe only rename the button to "Keep hidden" in that case? this would also be better for ppl tapping around and would like to see the text again. or if you try to delete only seconds after hiding and saying "great, the warning is gone, it is fine to delete now"

r10s: looks like that now:



"keep hidden" is roughly the same as "cancel", so "cancel" could be removed. otoh, not having "cancel" makes ppl think and confuse more, so maybe fine and good enough

r10s: looks like that now:



- the user gets the same text when they want to verify what they've done
- clear option to leave things as is in case only a moment after hiding, "delete" is tapped again, where the user may think "great, the warning is gone" 
- "hide it first" still matched (the user already did that, it is a general explanation)
- no need for a second text in an already complex area

link2xt: I have clicked around after compiling and before reading the code or comments here and noticed not fitting "hide it first" while testing if it will ask me to confirm deleting the hidden relay. It does not really fit and I rather expected some very short confirmation like "do you really want to remove the relay?" here, but it's fine.

r10s: k, this is roughly also what i understand what @Hocuri meant, so there already two who would expect a simple deletion if the relay is hidden already.

and yes, it seems good as well, if ppl did not read the hidden explanation before, they will not do that the second time. also, nothing too bad happens when removing a relay, you can heal accidental removal.

i do a commit and go for a simple deletion question

src/main/res/values/strings.xml
```diff
@@ -759,13 +759,19 @@
...
     <string name="confirm_add_transport">Add this relay?</string>
     <string name="invalid_transport_qr">The scanned QR code does not contain a valid relay.</string>
     <!-- placeholder will be replaced by a relay server name -->
-    <string name="confirm_remove_transport">Remove relay \"%1$s\"?\n\nYour contacts can only reach you if you\'ve
-        previously contacted them via another relay.\n\nIf in doubt, remove the relay later.</string>
-
+    <string name="confirm_remove_transport_x">Remove relay \"%1$s\"?</string>
+    <!-- placeholder will be replaced by a relay server name -->
+    <string name="confirm_remove_or_hide_transport_x">Remove relay \"%1$s\"?\n\nContacts who only know this relay may not reach
+        you until you message them again.\n\nIf in doubt, hide it first.</string>
```
Hocuri: Thinking of it, "If in doubt, hide it first" sounds a bit like it helps to click "hide from contacts" and then immediately remove it. This impression is exacerbated by the fact that if the relay is hidden already, the confirmation message is way shorter (and does not warn that users may not reach yo). So, the user may think "Great, now that I have hidden the contact before removing it, removing is not dangerous anymore!"


```suggestion
    <string name="confirm_remove_or_hide_transport_x">Remove relay \"%1$s\"?\n\nContacts who only know this relay may not reach
        you until you message them again.\n\nIf in doubt, hide it rather than removing it.</string>
```

Then I think the same confirmation message would also fit even if the relay is hidden already, in order to prevent accidental removing.

link2xt:
One problem is that "Remove relay now" is very close to "hide from contacts" and "close" so it is easy to fat-finger it when you want to just hide: 
image
(not really visible on the screenshot because it is relatively high DPI phone screen, but there is actually not much space between items)

When using it I always had to be careful to press "hide from contacts" and not accidentally clicking "remove relay now". If it cannot be moved to the left like in normal confirmation dialogs, then maybe there should be another confirmation dialog when user clicks "remove relay now" for unhidden relay. Or remove this option, then user will have to hide and then remove hidden relay again, it is probably also fine.

Accidentally irreversibly removing my classic email relay while trying to hide it would be bad.

r10s:
- i changed the button order so hide/delete are not together usually; this matches also more the gist of positive/negative/neutral.
in general, i do not think a second confirmation is needed  - one already selected "delete" before, not "hide".

- also, i realised that "Hidden from contacts" is the default for additional relays of existing multi-relay setups, so one gets into that state without reading the deletion instructions first. an explicit  "Show to contacts" may be helpful in that case - but then, it is advanced, and it is only for ppl doing experiements previously, so maybe it is fine as is.

- another annoyance is, that it is so easy to change "Use for sending" by tapping the row. it comes from the "proxy" activity i think - so one read the alert, hide, understood - and by one tap, everything is different.  the you have to do deletion again, risking again to type smth wrong.

but we can iterate over the latter in a subsequent PR, so we move forward here

src/main/res/values/strings.xml
r10s: this wording seems a good improvement, and indeed fits better if relay is hidden already.

i also had the concern that ppl remove it too fast - "great, there is no warning any longer".

so, let's go with the refined wording and the same dialogs again.

this is also easier to do on other OS and we can move forward.

we can iterate as needed

Hocuri:
This is how the final user-flow looks:

https://github.com/user-attachments/assets/8d8f5df8-d4e1-47c0-952b-8d1e29b8adc1

PR #4290, opened 2026-03-18
replace some hardcoded strings

### LINKED ISSUE #4283: calls: hard-coded strings need to be localized
follow-up issue after https://github.com/deltachat/deltachat-android/pull/4241 gets merged:

there are several strings hard-coded that need to be localized:

- `src/main/res/layout/item_audio_device.xml`
- `src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java`
- `src/main/java/org/thoughtcrime/securesms/calls/CallActivity.java`
- more?

adbenitez:
close #4283 

#skip-changelog

### DIFF
--- a/.github/workflows/cache-core.yml
+++ b/.github/workflows/cache-core.yml
@@ -1,0 +1,50 @@
+name: Core Cache
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'jni/deltachat-core-rust'
+      - 'jni/Android.mk'
+      - 'jni/Application.mk'
+      - 'jni/dc_wrapper.c'
+      - 'scripts/ndk-make.sh'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          submodules: recursive
+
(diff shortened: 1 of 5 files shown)


src/main/res/values/strings.xml
```diff
@@ -441,6 +442,13 @@
     <string name="audio_call">Audio call</string>
     <!-- Used in call bubbles, summaries, notifications -->
     <string name="video_call">Video call</string>
+    <string name="call_initializing">Initializing…</string>
+    <string name="call_incoming">Incoming call</string>
+    <string name="call_ringing">Ringing…</string>
+    <string name="call_reconnecting">Reconnecting…</string>
+    <string name="call_ended">Call ended</string>
+    <string name="call_failed">Call failed</string>
+    <string name="call_no_audio_output_available">No audio devices available</string>
```
r10s: the key does not match the value. if it is about "output" we should say that, so "No audio output available" or "No speaker available".

however, this seems to be a rare error, the user usually does not see or can run into. so we should leave that hardcoded as comparable errors. most other errors are english as well - and esp. for rare errors, this has the advantage that it is easier to search for

PR #4296, opened 2026-03-19
revert line breaks in src/main/res/values/strings.xml

adbenitez:


r10s:
this PR reverts line breaks inserted by auto-format that (1) led to accidental removal of translations (as source key is the english value, thousands of translations were removed by the change of the key) and (2) adds clutter to translators as the line breaks are visible to them, but not meaningful in any way.

moreover, the PR remove strings.xml form auto-format.

little drawback is that the translations will now have linebreak, but this is a minor, as (hopefully!) not meaningful for android. linebreaks in xml depend on context ...

so, i did a `tx push -s -t`

another drawback is that this removes all translations since 2026-03-14 (5 days ago) - but these are only 10 strings for a handful of often updated languages - far better than 80+ strings originally removed for all languages

r10s:
k, this restored most translations. for few languages, it did not work for whatever reason, i restored these strings manually.

so things should be mostly fine again :)

PR #4302, opened 2026-03-20
[do not merge] fix translations linebreak

r10s:
this PR is only for documentation about how to fix the translation lineends.

`convert.py` was created by AI, reviewed, and the result will be checked.

then, we push to transifex and close this PR


r10s:
result is in #4303

PR #4315, opened 2026-03-21
feat: mark messages as "unread"

r10s:
this PR adds an "mark unread" option at the place where otherwise "mark read" is shown.

this is kind of standard of standard on most messengers and one of the most often requested missing features in Delta Chat.

ftr, there is no urgency in merging, we should delay it until 2.47+ is out.

counterpart of https://github.com/deltachat/deltachat-ios/pull/3055

api & discussion: https://github.com/chatmail/core/pull/7990


adbenitez:
with the update to core 2.48.0 at https://github.com/deltachat/deltachat-android/pull/4330/changes#diff-f38d4b28a3999f5176257c614d1c1b50e53cc8a3bc26d56971c0b4118a198160 there is now a new RPC API `markfreshChat` would be nicer if that is used instead of adding more C api stuff

r10s:
yip, feel free to push to this PR once the jsonrpc is updated, or do a subsequent PR.

still, i agree we should wait with merging after 2.48+ is out

adbenitez:
@r10s this kind of deteriorates the following use case:

in the past it was possible to use "select all" + "mark as read" to remove a lot of scattered unread chats, now with this PR you will always get "mark as unread" instead, because there is always some chat that is not possible to mark as unread, like device messages, saved messages and gray-envelop/unencrypted chats, so even if you select "mark all as unread" to then try again and then "mark all as read", the option remains "mark all as unread"

I think it should be the other way around because what is more common to do in-mass is to "mark all as read" on only certain specific chats as unread, so if at least one chat is fresh/unread it should offer "mark all as read" and only offer "mark all as unread" if all selected are unread chats

r10s:
yip, we can change the logic, it is a bit random what i did when there are some unread and some read chats. now we have a good reason to decide what to do :)

r10s:
i added a commit

PR #4319, opened 2026-03-23
only pass URL domain to IDN.toASCII()

### LINKED ISSUE #4318: crash when clicking link in webview
the punycode URL conversion is trying to convert the whole URL instead of just the domain, which causes `IDN.toASCII()` to crash:

```
                                                                                                    java.lang.IllegalArgumentException: Invalid input to toASCII: https://github.com/example/example/issues/2476?email_source=notifications&email_token=LONG-TOKEN#issuecomment-4112428204
                                                                                                    	at java.net.IDN.toASCII(IDN.java:115)
                                                                                                    	at java.net.IDN.toASCII(IDN.java:138)
                                                                                                    	at I7.E1.N(SourceFile:50)
                                                                                                    	at I7.C1.shouldOverrideUrlLoading(SourceFile:7)
                                                                                                    	at android.webkit.WebViewClient.shouldOverrideUrlLoading(WebViewClient.java:92)
                                                                                                    	at I7.C1.shouldOverrideUrlLoading(Unknown Source:5)
                                                                                                    	at org.chromium.android_webview.AwContentsClientBridge.shouldOverrideUrlLoading(chromium-TrichromeWebView6432.apk-stable-768015333:117)
```

adbenitez:
Avoid crashes due to IllegalArgumentException launched by IDN.toASCII() if URL is too long

close #4318 

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/WebViewActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/WebViewActivity.java
@@ -22,6 +22,8 @@ import androidx.webkit.WebSettingsCompat;
 import androidx.webkit.WebViewFeature;
 import chat.delta.rpc.types.SecurejoinSource;
 import java.net.IDN;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import org.thoughtcrime.securesms.qr.QrCodeHandler;
 import org.thoughtcrime.securesms.util.DynamicTheme;
 import org.thoughtcrime.securesms.util.IntentUtils;
@@ -31,6 +33,9 @@ import org.thoughtcrime.securesms.util.ViewUtil;
 public class WebViewActivity extends PassphraseRequiredActionBarActivity
     implements SearchView.OnQueryTextListener, WebView.FindListener {
   private static final String TAG = WebViewActivity.class.getSimpleName();
+  // Regex to extract the host from a URL for IDN conversion.
+  private static final Pattern URL_PATTERN =
+      Pattern.compile("^((?:[a-zA-Z0-9]+://)?)([^/?#]+)(.*)$");
 
   protected WebView webView;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/thoughtcrime/securesms/WebViewActivity.java
```diff
@@ -320,9 +325,17 @@ protected boolean openOnlineUrl(String url) {
     }
 
     if (shouldAskToOpenLink()) {
+      String displayUrl;
+      try {
+        Matcher m = URL_PATTERN.matcher(url);
+        displayUrl = m.find() ? m.group(1) + IDN.toASCII(m.group(2)) + m.group(3) : url;
```
wchen342, line 331: This is problematic if there is a port number, or the url has IPv6 address. I checked and we open user urls, so it is better using `java.net.URI` or `java.net.URL` to avoid the fragility.

adbenitez: main task of this is just to avoid showing punycode domains to people in the dialog previous to opening, I intentionally avoided  using URI because they are too strict, parsing URL with them might fail and then the option is fail to open the link or fallback to displaying original url which potentially has punycode

wchen342: What do you mean too strict? Can you give me some examples?

wchen342: Also it doesn't solve the problem if there is `:` or brackets in the url, which will throw exception with this. Or do you mean the urls passed in cannot have those characters?

adbenitez: URI wants a valid URI and anything that is not valid will throw exception, I don't have an example at hand but in general, we don't need to parse the full URL the only thing that needs to be converted is the domain

I don't understand what exception you mean, why would an exception be thrown if there is : or port?

wchen342: There were some confusion on documentations, it seems to be fine if group 2 captures additional parts of the url.

PR #4322, opened 2026-03-26
update store descriptions

r10s:
this PR updates the "origin" of the app descriptions to the one used roughly used for macOS/iOS after recent discussions.

it makes sense for f-fdroid and gplay at some point as well

### DIFF
--- a/fastlane/metadata/android/en-US/full_description.txt
+++ b/fastlane/metadata/android/en-US/full_description.txt
@@ -2,7 +2,9 @@ Delta Chat is a reliable decentralized instant messenger that is easy and fun to
 
 Features at a glance:
 
-• Anonymous. Instant onboarding without a phone number, e-mail or other private data.
+• Private. Instant on-boarding without a phone number or other private data.
+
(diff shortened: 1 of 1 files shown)


fastlane/metadata/android/en-US/full_description.txt
```diff
@@ -2,7 +2,9 @@ Delta Chat is a reliable decentralized instant messenger that is easy and fun to
 
 Features at a glance:
 
-• Anonymous. Instant onboarding without a phone number, e-mail or other private data.
+• Private. Instant on-boarding without a phone number or other private data.
+
+• Friends & Family: Only chat with people you know. Unknown users cannot message or follow you without mutual consent.
```
adbenitez, line 7: the "without mutual consent" makes it sound like it is still possible after approving some contact request or so would be better to say:

"Strangers can not discover you or your family members,  a spam-free experience!"

or something like that

r10s: maybe on next iteration, i just wanted to sync here what was working elsewhere. for the concrete descriptions, some adaptions are okay/required anyway. this one only goes to f-droid 1:1

PR #4323, opened 2026-03-27
Fix file sharing to certain apps

### LINKED ISSUE #3597: Crashes when sharing from Delta Chat to some apps
@Simon-Laux's file manager (Material Files, from F-Droid) throws this error: 

System.err: java8.nio.file.FileSystemException: content://com.b44t.messenger.attachments/f6ad6d62.log/nixos-infect.log: ContentResolver.query() with content://com.b44t.messenger.attachments/f6ad6d62.log/nixos-infect.log returned null

The fix is probably to return something sensible from `query()` in `AttachmentsContentProvider.java`:

```java
    @Override
    public Cursor query(Uri arg0, String[] arg1, String arg2, String[] arg3,
                        String arg4) {
        return null;
    }
```

Signal has it implemented, so, we can use their code as instructions for how implement it, but we can't copy the code due to licensing problems.

wchen342:
fixes #3597

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Delta Chat Android Changelog
 
+## Unreleased
+
+* Fix file sharing to certain apps (e.g. Material Files, etc.)
+
 ## v2.48.0
 2026-03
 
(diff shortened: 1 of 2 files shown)


inexcode:
Thanks for the fix!

PR #4325, opened 2026-03-28
relay update warning

### LINKED ISSUE #4324: Add warning in the relay dialog that all devices need to be updated
2.43.0 has a bug which we fixed by now: If a message with an attachment (i.e. pre- and post-message) arrives twice, then it will be trashed.
2.47.0 [triggers this bug](https://github.com/chatmail/core/issues/8049) if there are multiple relays, because the message arrives via both relays.

The good thing is that when updating to 2.47.0, all relays get marked as unpublished by default.

I think we should add a warning when opening the relay screen that says e.g.: "Warning: If you change anything here, please make sure that all your devices use at least version 2.47.0. Otherwise, the older devices may miss messages."

cc @adbenitez @r10s

r10s:
this PR adds the hint `transport_list_hint` below the list of relays. this was anyways missing, and is what desktop/iOS are doing. the hint answers the question "what are the other relays doing if only one is use for sending?".

moreover, the PR adds a warning to the hint, saying that one need to update all devices.

advantage in contrast to an - unexpected - alert on opening is that that will be tapped away without reading, after all experience. the hint in the dialog, with the warning symbol seems to be outstanding enough, and will be visible to the user a longer time.

it is fine that the warning is shown when the user is not using multiple devices - we're in the advanced settings here, if they do not have an idea, it is good to leave things as is.

the style is sufficient and good enough - there is few text on the screen anyways and we use that at other places as well, so a red/bold/whatnot would not add much. if would also not match the "warning and maybe" level

another advantage is that desktop/iOS do not to do anything than pulling the strings. and we can remove the warning without changes on all UI later as well.

closes #4324 

Screenshot 2026-03-28 at 12 25 14

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/relay/RelayListActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/relay/RelayListActivity.java
@@ -7,7 +7,6 @@ import android.view.ContextMenu;
 import android.view.MenuItem;
 import android.view.View;
 import android.widget.ScrollView;
-
 import androidx.activity.result.ActivityResultLauncher;
 import androidx.activity.result.contract.ActivityResultContracts;
 import androidx.annotation.NonNull;


src/main/java/org/thoughtcrime/securesms/relay/RelayListActivity.java
```diff
@@ -99,7 +100,8 @@ protected void onCreate(Bundle savedInstanceState) {
     PulsingFloatingActionButton fabAdd = findViewById(R.id.fab_add_relay);
 
     // add padding to avoid content hidden behind system bars
-    ViewUtil.applyWindowInsets(recyclerView);
+    ScrollView scrollView = findViewById(R.id.relay_scroll_view);
```
adbenitez, line 103: this likely breaks edge to edge, probably it is better if the warning comes first then the recycle view without any scrollview (scroll of the scrollview + scroll of recycle view might be problematic?) 

have not tested yet to confirm this doesn't cause problems, OTOH we recently limited the list to 5 relays in core so it is unlikely the list will even need scroll

r10s: please do whatever needed to fix that, i have no experience there.

but yes, the list is always small for now, so that seems fine

Hocuri:
I do wonder if users will notice this text, because this small font is usually used for text that's not necessary to read.

So, what about making the text a bit bigger, and black rather than grey?

r10s:
see note above for the style. there is not much on the screen anyways. it is good enough :)

PR #4332, opened 2026-03-30
remove deprecated stock strings 99 and 100

adbenitez:
recent cores since pre-messages complain that 99 and 100 are no longer valid stock strings

probably the entry at strings.xml could also be removed? this will raise awareness for all other clients they should remove that stock strings, or do you want to remove them later in some "check for unused strings housekeeping" step @r10s ? 

#skip-changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Delta Chat Android Changelog
 
+## Unreleased
+
+* Fix file sharing to certain apps (e.g. Material Files, etc.)
+
 ## v2.48.0
 2026-03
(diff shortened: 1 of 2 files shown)


r10s:
let's go for removing real strings later, but you can mark them as "deprecated" already in this pr

r10s:
i just checked: all other OS, including ubuntutouch, do not use the string any longer and removed the stock string some time ago already.

so, android is the last one - and the strings can be removed directly

adbenitez:
ok, I removed them then, thanks for checking

PR #4335, opened 2026-03-31
get calls out of experimental

adbenitez:
- [x] remove "debug calls"
- [x] bring back the nice styled incoming call notifications
- [x] add entry to changelog

we also need to update the google play description to mention the calls support

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Delta Chat Android Changelog
 
+## Unreleased
+
+* Better incoming call system integration
+* Calls are not experimental anymore and don't need to be manually enabled
+
 ## v2.49.0
 2026-04
 


src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java
```diff
@@ -1320,10 +1320,8 @@ private void showIncomingCallNotification(String callerName, Icon callerIcon) {
     }
 
     // Check full screen intent permission on Android 14+
-    boolean canUseFullScreen = false;
     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
-      canUseFullScreen = canUseFullScreenIntent();
-      if (!canUseFullScreen) {
+      if (!canUseFullScreenIntent()) {
```
wchen342, line 1362: The variable is technically captured locally for thread safety, but in this case edge cases are very unlikely so this simplification is fine. (This is just a note, no changes needed)

wchen342:
Oh also this may be worth mentioning in Changelog.

adbenitez:
> Oh also this may be worth mentioning in Changelog.

thanks for pointing out!

PR #4351, opened 2026-04-08
avoid IllegalArgumentException when taking pictures and video

### LINKED ISSUE #4350: The camera doesn't work in chats
On my device, the camera in the chat refuses to launch (tapping the icon doesn't do anything).
However, scanning the QR code does launch the camera.

- Android version: android 9
- Device: honor 9
- Delta Chat version: 2.48.0
- Expected behavior:  When you click on the camera icon in the chat, the camera should launch.
- Actual behavior: When I click the camera icon in the chat, nothing happens, and the camera doesn't launch.
    I go into the chat
    I click on the camera icon
    Nothing happens (the camera doesn't launch)
- Screenshots:
- Logs:

[deltachat-log-20260407-184823.txt](https://github.com/user-attachments/files/26531889/deltachat-log-20260407-184823.txt)

adbenitez:
in devices with SD card users get:
`IllegalArgumentException: Failed to find configured root that contains /storage/7EBC-C205/Android/data/chat.delta/cache/xxxx.jpg`

adding `/storage/` root-path to `file_provider_paths.xml` solves the problem as already reported by the user at #4350  testing ArcaneChat with this fix already integrated 

close #4350

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Fix file sharing to certain apps (e.g. Material Files, etc.)
 * Remove proxy toggle from profile editing to avoid confusion
+* Fix taking pictures and videos in devices with SD cards
 
 ## v2.48.0
 2026-03


wchen342:
The `<root-path>` hack is a significant security risk, according to [Google](https://developer.android.com/privacy-and-security/risks/file-providers#do_not_use_the_root-path_path_element_in_the_configuration) and security companies like [Guardsquare](https://www.guardsquare.com/blog/android-fileprovider-directories-security-risks) (they created ProGuard).

The attack is not theoretical, there are real exploits on widely used apps, [TikTok](https://blog.oversecured.com/Oversecured-detects-dangerous-vulnerabilities-in-the-TikTok-Android-app/), Samsung Messages/[CVE-2021-25426](https://nvd.nist.gov/vuln/detail/cve-2021-25426) (High), and also [this](https://blog.oversecured.com/Android-security-checklist-theft-of-arbitrary-files/).

wchen342:
Well, if the scope is limited to `/storage/` then it is safer because then the attacker will not be able to access `/data/`, but then they can still read all user data even if they are not supposed to. This is less harmful then the most dangerous exploits, but still pretty bad in term of privacy.

wchen342:
It seems a solution is using `getExternalCacheDirs()` to get external storages (because UUID of SD card is not static at app build time), and then extends `FileProvider` to customize `getUriForFile()`.

adbenitez:
@wchen342 feel free to create an alternative PR if you have a better solution, we can then close this one, I was not fully understanding the implications of this stuff that is why I only added it to ArcaneChat initially

wchen342:
I will work and test if I can make an alternative solution work.

adbenitez:
superseded by https://github.com/deltachat/deltachat-android/pull/4357

PR #4356, opened 2026-04-09
Prevent app launch from creating multiple WelcomeActivity on top of each other

### LINKED ISSUE #4339: Resource Leak
- Android version: 11.0
- Device: Google Pixel
- Delta Chat version: 2.47.0
- Expected behavior: no consistent increase of any resource
- Actual behavior: consistent increase of Views, ViewRootImpl, AppContexts, Activities, Local Binders, Proxy Binders, java heap, native heap, rss
- Steps to reproduce the problem: repeatedly run ⟨ Return to home screen, Resume AUT ⟩ which focuses on onStart()/onRestart() and onStop() callback methods of WelcomeActivity for about 30~50 times
- Screenshots:

Image

- Logs:

Image

wchen342:
See #4339 for details.

Fixes #4339

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
@@ -30,6 +30,7 @@ public abstract class PassphraseRequiredActionBarActivity extends BaseActionBarA
 
     if (!DcHelper.isConfigured(getApplicationContext())) {
       Intent intent = new Intent(this, WelcomeActivity.class);
+      intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP | Intent.FLAG_ACTIVITY_SINGLE_TOP);
       startActivity(intent);
       super.onCreate(savedInstanceState);
       finish();


src/main/java/org/thoughtcrime/securesms/PassphraseRequiredActionBarActivity.java
```diff
@@ -30,6 +30,7 @@ protected final void onCreate(Bundle savedInstanceState) {
 
     if (!DcHelper.isConfigured(getApplicationContext())) {
       Intent intent = new Intent(this, WelcomeActivity.class);
+      intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP | Intent.FLAG_ACTIVITY_SINGLE_TOP);
```
adbenitez, line 33: mmmh... I wonder if this will have unwanted side-effects, did you test this thoroughly like opening dcaccount invite links etc.? (we really need end-to-end tests)

wchen342: It shouldn't change the behavior and should actually be better because with the old behavior you get multiple `WelcomActivity` in the stack, some has the QR in intent and some doesn't. 

I tested `dcaccount` and `dclogin`, works fine for me.

r10s:
let's not introduce a behaviour change to fix a theoretical issue (the memory of the activities is garbage collected once the welcome activity is left finally).

in the past we had hard to get issues with these kind of flags, let's not call for troubles

PR #4361, opened 2026-04-15
Add foreground service for location streaming

### LINKED ISSUE #3563: Location streaming does not work

wchen342:
Remove the legacy `ACCESS_BACKGROUND_LOCATION` permission and add a foreground service when streaming location.

Fixes #3563


src/main/java/org/thoughtcrime/securesms/geolocation/ActiveLocationChats.java
adbenitez, line 1: at some point it would be better if core would provide a getActiveLocationChats() or so instead of having to save chat ids on the UI (ex. chats could be deleted etc

src/main/res/values/strings.xml
```diff
@@ -1039,6 +1039,13 @@
     <string name="new_messages_body">You have new messages</string>
     <string name="n_messages_in_m_chats">%1$d messages in %2$d chats</string>
 
+    <!-- location streaming -->
+    <string name="location_streaming_channel_name">Location Streaming</string>
+    <string name="location_streaming_channel_desc">Channel for On-demand Location Streaming</string>
+    <string name="location_streaming_notification_title">On-demand Location Streaming</string>
```
adbenitez: maybe actually this could be removed and just use the "Location Streaming" from above as title, this one is already too long and just adding "on-demand" that is not telling much

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Better incoming call system integration
 * Calls are not experimental anymore and don't need to be manually enabled
+* Allow sharing location while the app is in background
```
adbenitez: maybe say "Display a notification when sharing location to improve reliability and allow to easily stop it from the notification" because sharing location while the app was in background was already what was there in the past just not reliable 

```suggestion
* Display a permanent notification when doing location streaming and get rid of dangerous "Access Location in Background" permission
```

adbenitez (approved):

adbenitez:
this is really nice improvement! and fixes our oldest issue! :tada: thanks a lot for fixing!

src/main/java/org/thoughtcrime/securesms/geolocation/ActiveLocationChats.java
wchen342: Yeah I agree. I actually asked Hocuri but I just put this in as a workaround for now so we don't need to wait for core.

wchen342: Btw the chat deleted is not a problem because the call has second parameter as 0, so it only disables streaming not starting.

EDIT: I checked core, a non-existing id just got silently skipped, so no crashes.

PR #4363, opened 2026-04-17
allow webxdc to auto-play audio without requiring user gesture/touch

adbenitez:
with this apps like @Amzd is "Radio Station" can play music on opening


wchen342 (approved):
LGTM!

Amzd (approved):

PR #4370, opened 2026-04-20
Remove usages of `class.getSimpleName()`

wchen342:
Currently all TAGs in classes are generated at runtime using `class.getSimpleName()`. This causes problems on release builds because the class names are obfuscated by ProGuard, thus resulting in log lines like these:

```
03-11 22:08:19.022 15827 15827 🔵 j       : startSession: sessionId=[0]
03-11 22:08:19.064 15827 15827 🔵 f       : omitting devices =[]
03-11 22:08:19.066 15827 15827 🔵 p       : getAvailableStartingCallEndpoints: awaitClose
03-11 22:08:19.066 15827 15827 🔵 j       : endSession: sessionId=[0]
03-11 22:08:19.090 15827 15827 🔵 f       : omitting devices =[(type=[18], name=[Pixel 5]),(type=[24], name=[Pixel 5])
```

where j, f, p shall be actual class names.

Retrace does not handle this automatically.

#skip-changelog

### DIFF
--- a/src/main/java/com/b44t/messenger/DcMsg.java
+++ b/src/main/java/com/b44t/messenger/DcMsg.java
@@ -59,7 +59,6 @@ public class DcMsg {
   public static final int DC_VIDEOCHATTYPE_UNKNOWN = 0;
   public static final int DC_VIDEOCHATTYPE_BASICWEBRTC = 1;
 
-  private static final String TAG = "DcMsg";
 
   public DcMsg(DcContext context, int viewtype) {
     msgCPtr = context.createMsgCPtr(viewtype);
--- a/src/main/java/com/b44t/messenger/DcMsg.java
+++ b/src/main/java/com/b44t/messenger/DcMsg.java
@@ -59,7 +59,6 @@ public class DcMsg {
   public static final int DC_VIDEOCHATTYPE_UNKNOWN = 0;
   public static final int DC_VIDEOCHATTYPE_BASICWEBRTC = 1;
 
-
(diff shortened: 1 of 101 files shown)


src/main/java/com/b44t/messenger/DcMsg.java
```diff
@@ -59,7 +59,7 @@ public class DcMsg {
   public static final int DC_VIDEOCHATTYPE_UNKNOWN = 0;
   public static final int DC_VIDEOCHATTYPE_BASICWEBRTC = 1;
 
-  private static final String TAG = DcMsg.class.getSimpleName();
+  private static final String TAG = "DcMsg";
```
adbenitez: this one is unused and can just be removed, we don't use android stuff (Log class) in the core package

```suggestion
```

src/main/java/org/thoughtcrime/securesms/contacts/ContactAccessor.java
```diff
@@ -37,7 +36,7 @@
  * @author Moxie Marlinspike
  */
 public class ContactAccessor {
-  private static final String TAG = ContactSelectionListFragment.class.getSimpleName();
+  private static final String TAG = "ContactSelectionListFragment";
```
adbenitez: this looks like it was copy+paste bug from the past, the class is   ContactAccessor

wchen342: Fixed.

adbenitez (approved):

PR #4374, opened 2026-04-23
allow to share location for 24 hours, remove 5 minutes

adbenitez:
now with the permanent notification is harder to forget you are streaming location so add 24 hours since some people have requested this (ex. people doing long bike trips)
 
5 minutes option is too short and we might not even get gps location in that time, so better remove it, people can always manually disable at any time, and more easily from the notification now

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ShareLocationDialog.java
+++ b/src/main/java/org/thoughtcrime/securesms/ShareLocationDialog.java
@@ -18,19 +18,19 @@ public class ShareLocationDialog {
           switch (which) {
             default:
             case 0:
-              shareLocationUnit = 5 * 60;
-              break;
-            case 1:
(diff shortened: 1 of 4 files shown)


src/main/java/org/thoughtcrime/securesms/ShareLocationDialog.java
```diff
@@ -18,19 +18,19 @@ public static void show(
           switch (which) {
             default:
```
wchen342, line 19: Just wondering, why is `default` at top?

wchen342 (approved):
Makes sense to me, thanks!

r10s:
i am fine with the change, @nicodh does that make sense to you as well?

nicodh:
Yes, makes sense.

PR #4376, opened 2026-04-24
Allow voice messages to be played continuously in a chat

### LINKED ISSUE #4365: playing a voice message doesn't play next one
when playing a voice message, usually the expectation is that when that finishes the next one in the chat is played (from self or from others) this is useful in chats with people that send several voice messages one after the other or to re-play a voice conversation in a chat while we are doing other stuff now that they can play in background

we have APIs to list all the audios in the current chat (ex. used in the media gallery of the chat), when the current msgId finishes it should just jump to the next audio msgId in the list

this is already implemented in Desktop client

wchen342:
Closes  #4365


adbenitez:
this only keeps playing the next if the initial played message was an incoming message, if the starting played message is outgoing/from-self then it doesn't continue playing

this makes it impossible for example to have a list of songs in saved messages and listen to them or to start listening to a voice message conversation from one of your own messages

wchen342:
Do we want to include non-voice messages in the playlist?

adbenitez:
yep, audio messages should also be included, just as on Telegram, and as we list audio+voice in the Audio tab in the chat gallery

wchen342:
> this only keeps playing the next if the initial played message was an incoming message, if the starting played message is outgoing/from-self then it doesn't continue playing

I cannot reproduce, is your outgoing message a non-voice message? If that is the case then it is because of the filtering, I'll change it.

adbenitez:
> I cannot reproduce, is your outgoing message a non-voice message? If that is the case then it is because of the filtering, I'll change it.

super weird, now that you said this I try again and in fact even if it is self message the first one it keeps playing, I think I tested with voice messages at the time, but maybe I am misremembering, so let's forget about it for now

adbenitez:
when user plays an audio from the "all media" gallery it should play the list in that gallery, not only the messages in the chat where the audio belongs to (EDIT: actually in "All Media" it just plays a single audio, not even others from the same chat)

src/main/java/org/thoughtcrime/securesms/components/audioplay/ChatVoiceQueueProvider.java
```diff
@@ -0,0 +1,54 @@
...
+    List<QueueItem> queue = new ArrayList<>();
+    for (int i = startIndex; i < msgIds.length; i++) {
+      DcMsg msg = dcContext.getMsg(msgIds[i]);
+      if (msg.getType() == DcMsg.DC_MSG_VOICE || msg.getType() == DcMsg.DC_MSG_AUDIO) {
+        Uri uri = new AudioSlide(context, msg).getUri();
+        if (uri != null) {
+          queue.add(new QueueItem(msgIds[i], uri));
+        }
+      }
+    }
```
adbenitez: this is not efficient, it is loading all chat messages and then checking if they are voice/audio, there are more efficient ways to get directly voice/audio messages, just as the AllMediaActivity do to show all audios in the audio tab, check `BucketedThreadMediaLoader` and its usage of `DcContext.getChatMedia`

wchen342:
In media tag it needs a separate logic I think, the two lists are different.

Edit: turns out the native method is more straightforward than I thought.

src/main/java/org/thoughtcrime/securesms/components/audioplay/ChatVoiceQueueProvider.java
wchen342: I updated using the native method.

src/main/java/org/thoughtcrime/securesms/components/audioplay/AudioPlaybackViewModel.java
```diff
@@ -78,32 +78,35 @@ public void loadAudioAndPlay(int msgId, Uri audioUri) {
...
-      if (!queue.isEmpty() && queue.get(0).getMsgId() == startMsgId) {
-        for (AudioPlaybackQueueProvider.QueueItem item : queue) {
-          items.add(toMediaItem(item.getMsgId(), item.getUri()));
-        }
-        return items;
-      }
+      items = queueProvider.buildAudioQueue();
+      startIndex = indexOfMediaId(items, msgId);
+    } else {
+      items = Collections.emptyList();
```
adbenitez: this is unnecessary because right below you are overriding items if `startIndex < 0`

I recommend to initialize `startIndex = -1` and remove this whole else branch 🤔

src/main/java/org/thoughtcrime/securesms/components/audioplay/ChatAudioQueueProvider.java
```diff
@@ -0,0 +1,39 @@
...
+    int[] msgIds = dcContext.getChatMedia(chatId, DcMsg.DC_MSG_AUDIO, DcMsg.DC_MSG_VOICE, 0);
+
+    List<MediaItem> items = new ArrayList<>(msgIds.length);
+    for (int msgId : msgIds) {
+      DcMsg msg = dcContext.getMsg(msgId);
+      Uri uri = new AudioSlide(context, msg).getUri();
+      if (uri != null) {
+        items.add(new MediaItem.Builder().setMediaId(String.valueOf(msgId)).setUri(uri).build());
+      }
+    }
```
adbenitez, line 38: this doesn't look good, it is loading the whole list of media messages from the database, ideally this would be done lazily, just create the `MediaItem` with the message ID

~~then in `AudioPlaybackService` override `onAddMediaItems` from parent class `MediaSessionService` and do the loading of DcMsg/URI there~~ (EDIT: actually looking more carefully this solves nothing because the returned ListenableFuture still requires to provide all the resolved items at once :/ there should be another solution)

UPDATE: in `AudioPlaybackService` you are creating the `ExoPlayer` instance, you could use `setMediaSourceFactory` there to then set a `MediaSource.Factory` that will take care of loading the full MediaItem with URI in `createMediaSource`

wchen342: Even if you override the `onAddMediaItems` it is not truly lazy because the whole list is scanned at once, and MediaItem requires Uri, so when they are added they will need Uris anyway. The benefit will be the reading is moved to a background service which is good, but it will still need to read everything at once.

wchen342: Also isn't AudioSlide just reading some metadata? That shall be quick.

adbenitez: see my updated comment, there is a way to load the message lazily, about speed, this is not good, because you are loading the whole DcMsg object from database for a potentially huge list (ex. in "all media") and passing them to AudioSlide that just takes the file path property then

wchen342: Yeah the Factory override is the more "correct" solution but that is more complex. I was hoping to put a simpler solution first but I can go with that.

src/main/java/org/thoughtcrime/securesms/components/audioplay/AudioPlaybackViewModel.java
wchen342: Fixed.

src/main/java/org/thoughtcrime/securesms/components/audioplay/ChatAudioQueueProvider.java
wchen342: Ok I have made the file resolve on load with `DataSource`.

src/main/java/org/thoughtcrime/securesms/service/AudioPlaybackService.java
```diff
@@ -37,10 +47,43 @@ public void onCreate() {
...
+                try {
+                  msgId = Integer.parseInt(host);
+                } catch (NumberFormatException ignored) {
+                }
+              }
+              if (msgId <= 0) {
+                throw new IOException("Invalid dcmsg uri: " + uri);
+              }
+
+              DcMsg msg = DcHelper.getContext(this).getMsg(msgId);
```
adbenitez: what happens if you start playing an audio in one account and switch to another? the URI should be `dcmsg://accid.msgid` to avoid using "current selected account" and load the wrong message

it can happen you start listening to audios of one chat but in the meantime switch to another profile

wchen342: Yeah that can be a problem, but honestly shouldn't everything stop when switching account?

adbenitez: it is already the case that you can switch accounts and keep playing audio, and this is the expected behavior, same way you can switch to other apps you could also click in a notification of a chat in another profile and suddenly the audio would stop for no reason if we change the current behavior to stop on account switching

wchen342: Ok I made it include account id too, now it shall work when switching accounts.

adbenitez:
tested it and now it is all working flawlessly, good work!!! this is a great improvement!!!

adbenitez (approved):

PR #4384, opened 2026-04-28
add highlight to input panel buttons for keyboard navigation

### LINKED ISSUE #4345: Attachment menu has no selection highlight when navigating by keyboard
- Device: Sonim XP5s
- Delta Chat version: 2.48.0
- Expected behavior: When attaching a file, there should be a highlight over menu items as one tabs/arrow keys through them.
- Actual behavior: There is none. Moving the selection still works, but there is no ui feedback and it is confusing.

adbenitez:
close #4345


r10s (approved):
lgtm, thanks a lot for taking care!

PR #4386, opened 2026-04-28
add mapping.txt to symbols zip

### LINKED ISSUE #4342: add mapping file to symbols zip
with every release we upload debugging symbols to https://download.delta.chat/android/symbols/, but now we use code minifier which make reading crash trace reports very hard, this crash traces can easily be deobfuscated using retrace:

`retrace ./path/to/mapping.txt <crash.txt > decoded-crash.txt`

the mapping file is at `./build/outputs/mapping/gplayRelease/mapping.txt` and should be included in the zip generated by `./scripts/upload-release.sh`

adbenitez:
close #4342


wchen342:
Do we upload to play console manually? I think it can also use the mapping to decode stack automatically.

adbenitez:
yep upload there is manually

wchen342 (approved):

r10s (approved):

PR #4388, opened 2026-04-28
fix migration while upgrading from 1.10.1 and similar old versions

### LINKED ISSUE #4328: Upgrading from 1.10.1 to 2.47 does not work
I installed https://download.delta.chat/android/deltachat-gplay-release-1.10.1.apk, created a profile, then upgraded to 2.47.0 from https://download.delta.chat/android/deltachat-gplay-release-2.47.0.apk
Application does not start, it gets ANR on the splash screen.

adbenitez:
the problem was that `AccountManager.migrateToDcAccounts()` was calling `ApplicationContext.getDcAccounts()` internally which would then block on `ensureInitialized()` since `migrateToDcAccounts` is called during initialization process while `isInitialized == false`

this patch just passes `DcAccounts` object directly to `migrateToDcAccounts`

close #4328

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Allow to share location for 24 hours
 * Allow mini-apps to play audio without user interaction
 * Mark chats as unread (long tap a chat and select the corresponding option from the three-dot-menu)
+* Fix process of upgrading from a very old version of the app
 
 ## v2.49.0
 2026-04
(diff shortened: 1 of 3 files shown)


link2xt (approved):
Cannot test because need an old version and fixed new version with the new signature, but code looks good.

adbenitez:
(I tested it with 1.10.1 upgraded to 2.49.0 confirmed it gets stuck every time, then installed this fix and confirmed it works then)

PR #4390, opened 2026-04-28
Setup zizmor workflow to check GitHub Actions workflows

link2xt:


link2xt:
Related:
- https://github.com/deltachat/deltachat-desktop/pull/6307
- https://github.com/deltachat/deltachat-ios/pull/3092
- https://github.com/chatmail/relay/pull/938

adbenitez (approved):
makes sense, I always felt bad about using actions just pointing at a tag, didn't realize you can as well point to direct commits

link2xt:
> makes sense, I always felt bad about using actions just pointing at a tag, didn't realize you can as well point to direct commits

It is still not good because hash pin can point to fork and transitive dependencies are not pinned. Github [plans to introduce proper locking](https://github.blog/news-insights/product-news/whats-coming-to-our-github-actions-2026-security-roadmap/), but as long as we don't have it this is the best workaround. Dependabot supports it, and will even update the comment when upgrading github actions.

PR #4392, opened 2026-04-29
change sticker sort order

### LINKED ISSUE #4391: Newly added stickers appear at the bottom of the list instead of the top
I noticed that when a new sticker is added, it is being appended to the bottom of the sticker list.
The expected behavior is that newly added stickers should appear at the top of the list, above the existing stickers.

adbenitez:
close #4391

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * Allow mini-apps to play audio without user interaction
 * Mark chats as unread (long tap a chat and select the corresponding option from the three-dot-menu)
 * Fix process of upgrading from a very old version of the app
+* Show more recent added stickers at the top of the sticker picker
 
 ## v2.49.0
 2026-04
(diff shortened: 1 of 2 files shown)


wchen342:
Do the file names always increase when saving stickers?

Edit: I did some tests on Desktop and that doesn't seem to be the case, so only reverse the filename sorting won't solve the problem I think?

HeyErfan:
Thanks

adbenitez:
> Do the file names always increase when saving stickers?
> 
> Edit: I did some tests on Desktop and that doesn't seem to be the case, so only reverse the filename sorting won't solve the problem I think?

in android we already use the date timestamp to sort them by added date, desktop is a different story, this PR just reverts the order to have newer at the top as requested by the user

wchen342 (approved):
In that case it makes sense then.

r10s (approved):

r10s:
ftr, this also matches better other sortings, where we as well have "recent is shown directly".

and yes, desktop should adapt to that order as well

PR #4394, opened 2026-04-30
open map if user clicks the "Location streaming enabled by Foo/you" message

adbenitez:
currently clicking such info messages does nothing, which is particularly bad if the receiver side doesn't have location streaming enabled (because there is no way to access the map then)


wchen342 (approved):
Works for me, nice improvement!

PR #4395, opened 2026-05-01
allow to open message links via actions in TalkBack menu

### LINKED ISSUE #4200: [Accessibility] Chat links are not available for Talkback
- Android version: 16
- Talkback version: 16.2
- Delta Chat version: 2.35.0

  1. Open any chat.
  2. Locate or send a message containing one or more URLs.
  3. Move the TalkBack focus to the message element.
  4. Attempt to double-tap to activate the link.
  5. Open TalkBack menu (L-swipe or multi-finger tap) and check for Actions.

- Actual behavior:
  The message element does not respond to the activation gesture (double-tap), and no Custom Actions are available for the URLs within the message.

https://github.com/user-attachments/assets/0261425f-36cd-46cc-bc9d-9e8882191745

- Expected behavior:
  When a message element is focused and activated via a double-tap, the link within it should be triggered.

  However, since a message may contain multiple links, providing distinct child elements for each link is the preferred approach to ensure clarity and precise selection.

  Alternatively, instead of child elements, the message object should implement Custom Actions. These actions allow TalkBack to perform specific functions on the current object and are accessible via the TalkBack menu or the Reading controls menu.

adbenitez:
image

close #4200

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationItem.java
@@ -24,6 +24,7 @@ import android.graphics.PorterDuff;
 import android.graphics.Rect;
 import android.os.Build;
 import android.text.SpannableString;
+import android.text.Spanned;
 import android.text.TextUtils;
 import android.util.AttributeSet;
 import android.util.Log;
@@ -36,6 +37,7 @@ import androidx.annotation.DimenRes;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.appcompat.app.AlertDialog;
+import androidx.core.view.ViewCompat;
 import chat.delta.rpc.RpcException;
 import chat.delta.rpc.types.CallInfo;
 import chat.delta.rpc.types.CallState;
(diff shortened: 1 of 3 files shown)


src/main/java/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -434,6 +447,29 @@ private void setBodyText(DcMsg messageRecord) {
       }
       bodyText.setText(spannable);
       bodyText.setVisibility(View.VISIBLE);
+
+      // Register a TalkBack "Actions" entry for each link in the message
+      if (spannable instanceof Spanned) {
```
wchen342: Isn't this always `true`?

adbenitez: ah! true! looking more carefully `SpannableString` implements ` Spannable` and I didn't realize `Spannable` actually implements `Spanned` I was thinking it was the other way around and some Spannable were Spanned and not that all Spannable are Spanned

wchen342 (approved):
Checked and works for me!

PR #4400, opened 2026-05-04
Use new RPC for updating and stopping location streaming

### LINKED ISSUE #4379: location streaming bugs
currently only chat id is tracked in ActiveLocationChats, so when calling "stop sharing location" from notification it can happen that the currently selected account is not the original account of the chat

also if user tries to share location across 2 different profiles, only the current active profile will get the location (DcContext.setLocation)

we need better core APIs:

- Rpc.setLocation(float, float) - without any account ID parameter that provides location to all profiles
- Rpc.stopLocationStreaming() - that stops streaming location on all chats across accounts that are currently streaming location

in the meantime we might need to patch this on the java side and do things properly if the core API doesn't come before we want to make a new release:

- keep track of (accountId, chatId) instead of just chatId
- iterate over all accounts and call setLocation()

wchen342:
Fixes #4379

#skip-changelog


wchen342:
I'll open a new issue for `stopSendingLocations` to wait for core tag

adbenitez:
this intermediate step is not needed, the new APIs are already implemented in core as said in the linked issue and will be available for the next release

wchen342:
I guess we are not getting it into 2.50 anyway, so let me just wait the new tag and then do it again.

adbenitez:
the new api was introduced at

https://github.com/deltachat/deltachat-android/pull/4438/changes#diff-f38d4b28a3999f5176257c614d1c1b50e53cc8a3bc26d56971c0b4118a198160R1375-R1397

wchen342:
I have updated the PR.

src/main/java/org/thoughtcrime/securesms/geolocation/LocationStreamingService.java
```diff
@@ -37,37 +39,44 @@ public class LocationStreamingService extends Service {
...
 
+  private static boolean isAnySharingActive(Context context) {
+    DcAccounts accounts = DcHelper.getAccounts(context);
+    for (int accountId : accounts.getAll()) {
+      if (accounts.getAccount(accountId).isSendingLocationsToChat(0)) {
+        return true;
+      }
+    }
+    return false;
+  }
```
adbenitez, line 96: `Rpc.isSendingLocations()` should be used instead

src/main/java/org/thoughtcrime/securesms/geolocation/LocationStreamingService.java
```diff
@@ -37,37 +39,44 @@ public class LocationStreamingService extends Service {
 
   /** Register a chat for location updates, then ensure the service is running. */
   public static void startSharing(Context context, int chatId, int durationSeconds) {
-    ActiveLocationChats.add(context, chatId);
     DcHelper.getContext(context).sendLocationsToChat(chatId, durationSeconds);
```
adbenitez: better switch to the Rpc API, the C API will likely get deprecated eventually

wchen342:
I have updated the methods to use RPC calls.

src/main/java/org/thoughtcrime/securesms/geolocation/LocationStreamingService.java
```diff
@@ -36,38 +39,64 @@ public class LocationStreamingService extends Service {
...
 
   public static boolean isRunning() {
     return running;
   }
 
+  private static boolean isAnySharingActive(Context context) {
+    try {
+      Rpc rpc = DcHelper.getRpc(context);
+      DcAccounts accounts = DcHelper.getAccounts(context);
+      for (int accountId : accounts.getAll()) {
```
adbenitez: here you could also just use `Rpc.getAllAccountIds()`

```suggestion
      for (Integer accountId : rpc.getAllAccountIds()) {
```

adbenitez (approved):

src/main/java/org/thoughtcrime/securesms/geolocation/LocationStreamingService.java
wchen342: So are we just moving to rpc for everything? I am actually quite confused because we have two apis now.

adbenitez: not for everything, but for new things where performance is not a problem, yes
and in this case there was new location streaming api in the RPC needed to be used here so it didn't make sense to mix both APIs at the same time

wchen342: Ok I have updated (it can be `int accountId`, doesn't need to be `Integer`).

adbenitez (approved):

adbenitez:
I think this actually deserves a changelog entry, because it is fixing a bug in location streaming: "fix location streaming for multi-profile setups" or something like that

(sorry for not bringing it up earlier)

wchen342:
> I think this actually deserves a changelog entry, because it is fixing a bug in location streaming: "fix location streaming for multi-profile setups" or something like that

Added the entry. Good thing you are fast enough to stop the auto-merge.

adbenitez (approved):
thanks a lot!

PR #4402, opened 2026-05-04
Hide webxdc attach button for channels

### LINKED ISSUE #4398: do not show app picker for channels
most app do not work today in channels

adding a .xdc via file should continue to work, to allow experiments. 

the corresponding tab in the gallery should stay as is

wchen342:
Closes #4398

I didn't re-arrange the buttons because I want to keep the changes minimal

Screenshot_20260504_165929

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
@@ -1116,6 +1116,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     if (chatId == DcChat.DC_CHAT_NO_CHAT)
       throw new IllegalStateException("can't display a conversation for no chat.");
     dcChat = DcHelper.getContext(context).getChat(chatId);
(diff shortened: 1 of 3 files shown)


src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1116,6 +1116,7 @@ private void initializeResources() {
     if (chatId == DcChat.DC_CHAT_NO_CHAT)
       throw new IllegalStateException("can't display a conversation for no chat.");
     dcChat = DcHelper.getContext(context).getChat(chatId);
+    attachmentTypeSelector = null;
```
adbenitez, line 1119: can you explain why this is needed?

wchen342: Because the selector is cached, so if you open a different conversation it will reuse the previous one which may be different right?

adbenitez (approved):

PR #4403, opened 2026-05-04
Get channels out of experimental

### LINKED ISSUE #4399: get "channels" out of experimental
"channels" should be available by default, they're no longer experimental and do not need a dedicated option any longer

(same for "calls", but that is already merged)

wchen342:
Closes #4399

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Better incoming call system integration
 * Calls are not experimental anymore and don't need to be manually enabled
+* Channels are no longer experimental and are enabled by default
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * Better incoming call system integration
 * Calls are not experimental anymore and don't need to be manually enabled
+* Channels are no longer experimental and are enabled by default
```
adbenitez:
```suggestion
* Channels are no longer experimental and are available by default
```

adbenitez (approved):

PR #4404, opened 2026-05-04
don't set draft if user can't send

### LINKED ISSUE #4396: Tapping /text creates drafts in read-only chats
Delta-Chat-Version: Delta Chat for Android (from F-Droid) v2.49.0

# Observed behaviour

1. The context is a read-only mailing list.
2. I received a message that contained the string `/dev/null` -- which was highlighted as if it was a link.
3. Out of curiosity, I tapped the link and nothing seemed to happen.
4. As I return to the chat list (i.e. the initial view), I notice that the chat in question now has `Draft: /dev/null` as the message preview. Since this is a read-only mailing list, I have no way to delete this draft.

# Expected behaviour

I believe that read-only chats shouldn't ever have a draft state. There are a number of ways that could be used to achieve this:

1. Do not highlight strings like `/dev/null` in read-only contexts
2. Do highlight strings like `/dev/null` but do a NO-OP if a user taps the string in read-only contexts
3. Do highlight strings like `/dev/null` but display an error message if a user taps the string in read-only contexts

I personally wouldn't mind option 2 at all. I.e. I tapped the string out of curiosity -- nothing happened -- I didn't mind; but later on I noticed the draft state, which I do mind. Option 3 introduces more verbosity / friction. Since Delta Chat is very cautious with error messages in general, it might be a good idea *not* to introduce this kind of friction. Option 1 might be harder to implement than option 2. If this is the case, pursuing option 1 might not be worth the additional effort.

# Further considerations

1. Due to the bug described above, some existing read-only chats might already have a draft state. It would be nice if this draft state would be cleaned up automatically after some future upgrade.
2. Fixing this may also imply doing some changes to core.

adbenitez:
close #4396

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Fix process of upgrading from a very old version of the app
 * Show more recent added stickers at the top of the sticker picker
 * Allow to open map if user clicks "Location streaming enabled" system message
+* Fix: do not accidentally set draft in chats that don't allow sending messages
 
 ## v2.49.0
 2026-04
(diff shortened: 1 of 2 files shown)


wchen342 (approved):
Tested and works!

PR #4407, opened 2026-05-05
add "mark all as read" to account menu

### LINKED ISSUE #4406: add "mark all as read" from profile switcher
when long tapping a profile in the profile switcher, we currently have the options "mute", "label", "move to top" and delete".

what is missing is an option to mark all chats as read - similar to a chat, to get rid of the unread counter.

desktop has a corresponding option, and there was even a dedicated api added for that at https://github.com/chatmail/core/pull/7709 . string is  `mark_all_as_read`.

for the position: _if_ ppl use that option, this option is likely the one used most, so it should be the first option (all other options are not used on a daily base, more settings)

cmp https://github.com/deltachat/deltachat-ios/issues/3098

adbenitez:
close #4406

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 * Allow to share location for 24 hours
 * Allow mini-apps to play audio without user interaction
 * Mark chats as unread (long tap a chat and select the corresponding option from the three-dot-menu)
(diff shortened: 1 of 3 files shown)


r10s (approved):
nice!

PR #4408, opened 2026-05-05
add a setting to enable/disable calls notification

### LINKED ISSUE #4152: Add UI option for "who can call" me when calls are enabled
Was merged in https://github.com/chatmail/core/pull/7682, will be in core 2.37.0

Once this is decided where the option should go (probably "Notifications" section), open issues for iOS, Desktop and https://codeberg.org/lk108/deltatouch

adbenitez:
close #4152

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Show more recent added stickers at the top of the sticker picker
 * Allow to open links in messages via actions in TalkBack menu
 * Allow to open map if user clicks "Location streaming enabled" system message
+* Allow to disable incoming calls notifications
 * Fix: do not accidentally set draft in chats that don't allow sending messages
 
 ## v2.49.0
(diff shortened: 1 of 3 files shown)


adbenitez:
this is following the recommendations, but I think it is really bad, it will much better if the setting is called "Who can call me?"

wchen342:
Yeah I think the setting looks a bit weird to me. First it is under Notification which is probably fine, but then just call it "Calls" seems confusing. Maybe add some explanation texts to it?

r10s (changes requested):
as just discussed one-to-one, we want to have a simple switch "Calls: Show call screen for accepted contacts", just on/off (for "contacts"/"no one". "everybody" is unused for now, the idea is to disable ringing, not to ring more). if "off", depending on other settings, the normal notification may still be shown, same as for every other boring message, you can also still pick up the call. in prior discussions we came to the point that  "who can call me?" may be regarded as misleading therefore. the option is really about "ringing" only resp. the "system call screen".

see [current desktop](https://github.com/deltachat/deltachat-desktop/pull/6079) and previous discussions there, for iOS i just created a commit as well, see https://github.com/deltachat/deltachat-ios/pull/3095 for a screenshot

anyways, for now, we want to move forward and not open new discussions. it is good enough for now

adbenitez:
I see, I got confused by reading a subsequent desktop PR that had "nobody", "contacts" "everyone" in it but it seems that was just internal stuff, just having a toggle simplifies things a lot and was what I wanted to do initially :(

wchen342:
I just realized this eliminates events from core so we cannot accept calls if this is turned off. Let me make a fix for that and don't merge this yet.

r10s (approved):
nice!

@wchen342 and yes - this setting prevents core from doing the initial INCOMING_CALL events, thanks for looking into that (seems to be needed anyways, this may happen already today for contact requests)

PR #4411, opened 2026-05-06
feat: simplify location streaming wording

r10s:
as an outcome of some recent discussions, this PR renames "On-demand Location Streaming" to only "Location Streaming". the "on demand" was added that time to not confuse ppl - but that was done without any data, and now, ppl _are_ actually confused by that :) also, it makes the things a heavy wording.

moreover, this PR removes some double titles - @wchen342 in general, we do not double strings unless we have hints that it really needed; this is esp. true for experimental features, where also english is fine. we accept sometimes weird translation keys as a drawback.

for the channel description: for more important descriptions, we do not have translations - most ppl never see them, they are hard to translate without context - also, the current one does not make sense to give to translators: title "Channel: Location Streaming" with description "Channel for Location Streaming" - then we can also just strike it (but not sure if a channel can have _no_ description)

both are to remove load from translators - and also from us, as we do not have to deal with questions.

#skip-changelog

once merged, we need to call `./scripts/tx-update-changed-sources.sh`

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/geolocation/LocationStreamingService.java
+++ b/src/main/java/org/thoughtcrime/securesms/geolocation/LocationStreamingService.java
@@ -188,9 +188,9 @@ public class LocationStreamingService extends Service {
       NotificationChannel channel =
           new NotificationChannel(
               CHANNEL_ID,
-              getString(R.string.location_streaming_notification_title),
+              getString(R.string.pref_on_demand_location_streaming),
               NotificationManager.IMPORTANCE_LOW);
-      channel.setDescription(getString(R.string.location_streaming_channel_desc));
+      channel.setDescription("Channel for Location Streaming");
       channel.setShowBadge(false);
       NotificationManager nm = getSystemService(NotificationManager.class);
       if (nm != null) nm.createNotificationChannel(channel);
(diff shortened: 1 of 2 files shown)


src/main/java/org/thoughtcrime/securesms/geolocation/LocationStreamingService.java
```diff
@@ -188,9 +188,9 @@ private void ensureNotificationChannel() {
       NotificationChannel channel =
           new NotificationChannel(
               CHANNEL_ID,
-              getString(R.string.location_streaming_notification_title),
+              getString(R.string.pref_on_demand_location_streaming),
               NotificationManager.IMPORTANCE_LOW);
-      channel.setDescription(getString(R.string.location_streaming_channel_desc));
+      channel.setDescription("Channel for Location Streaming");
```
adbenitez: I don't see the need to do this, while we sometimes take the shortcut of using English, this is really unprofessional and might be really confusing and estrange for some users, we shouldn't assume everyone knows English, and this is notifications settings

by doing this we are not saving anything, this is just another string to translate for translators and if not translated then you get the same English behavior, so we shouldn't be shy on translating things and treating translations as a scarce resource

wchen342: In fact descriptions are optional. If we don't really want description we can just remove this line.

r10s: then let's remove it for now. it does not explain anything saying "channel: location streaming -- channel for location streaming"

if we regard the channel description as important, we should iterate in general. but let's not start with that on an experimental feature, which is subject to change, with an unhelpful wording

adbenitez: perfect, in that case it is better to remove than to have some English translation that doesn't say much

adbenitez (approved):

src/main/java/org/thoughtcrime/securesms/geolocation/LocationStreamingService.java
r10s: k, i removed the description

PR #4415, opened 2026-05-07
Make it possible to answer incoming calls from messages

wchen342:
Prerequisite of #4408.

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationItem.java
@@ -1091,7 +1091,18 @@ public class ConversationItem extends BaseConversationItem {
           if (!messageRecord.isOutgoing() && callInfo.state instanceof CallState.Alerting) {
             int callId = messageRecord.getId();
             CallCoordinator coordinator = CallCoordinator.getInstance(context);
-            coordinator.showIncomingCallScreen(callId);
+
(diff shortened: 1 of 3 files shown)


src/main/java/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -1091,7 +1091,18 @@ public void onClick(final View v, final CallInfo callInfo) {
...
+            if (coordinator.hasActiveCall()) {
+              coordinator.showIncomingCallScreen(callId);
+            } else {
+              if (callInfo.sdpOffer == null) {
+                Toast.makeText(context, R.string.error, Toast.LENGTH_SHORT).show();
+                return;
+              }
+              int accId = dcContext.getAccountId();
+              coordinator.handleIncomingCallFromConversation(
+                  accId, callId, callInfo.sdpOffer, Boolean.TRUE.equals(callInfo.hasVideo));
```
adbenitez: mmh I guess `callInfo.hasVideo.booleanValue()` will be clearer if the intention is to convert from Boolean to boolean, otherwise it is harder to read because the intention is more indirect

wchen342: Yeah don't need that. Was just copying around and didn't realize it.

adbenitez (approved):
tested it with my other PR and it works like charm! thanks a lot!

PR #4416, opened 2026-05-07
unify call buttons

r10s:
this PR is a first step in creating a consistent call experience across platforms:

- use similar colors as on desktop/iOS
- make strings translatable, add missing strings
- for accessibility: set all button labels
- for accessibility: the avatars cannot be tapped and have no action otherwise. the video is read by TalkBack as such, so let's see if that works and wait for feedback


adbenitez (approved):

src/main/res/values/strings.xml
```diff
@@ -206,7 +206,12 @@
     <string name="camera">Camera</string>
     <!-- As in "start a video recording" or "take a photo"; eg. the description of the "shutter button" in cameras -->
     <string name="capture">Capture</string>
+    <!-- Switch audio output eg. from earphone to front speaker -->
+    <string name="switch_speaker">Switch Speaker</string>
```
wchen342, line 210: This is not entirely accurate, it can also switch to bluetooth earphones or even cast devices. But I guess for now it is fine for most common usages.

r10s: i am fine to change that - which wording are  whatsapp/telegram using?

wchen342: Telegram just says "Audio" under the button. Whatsapp doesn't seem to have this directly, only through the phone settings.

r10s: k, then let's stay with that, it is even easier.

i created a PR at https://github.com/deltachat/deltachat-android/pull/4428

PR #4421, opened 2026-05-08
move requestPinShortcut to background

### LINKED ISSUE #4418: Potential main-thread blocking issue in Delta Chat pinned-shortcut flow
Hi Delta Chat Team,

I’m a PhD student researching Android performance issues. My research group recently ran a static analysis scan for thread-affinity and main-thread blocking bugs, and our prototype flagged a potential issue in Delta Chat Android.

## Checked source-level targets

We found two app-owned source callsites in the latest Delta Chat Android codebase:

```text
src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
src/main/java/org/thoughtcrime/securesms/BaseConversationListFragment.java
```

Both callsites invoke:

```java
ShortcutManagerCompat.requestPinShortcut(...)
```

The APK-level trace observed by our analyzer is:

The latest source contains corresponding app-owned pinned-shortcut request callsites through the AndroidX compatibility wrapper.

## What I found

In `WebxdcActivity.java`, the source contains a Webxdc/add-to-homescreen shortcut flow similar to:

```java
Toast.makeText(context, R.string.one_moment, Toast.LENGTH_SHORT).show();

if (!ShortcutManagerCompat.requestPinShortcut(context, shortcutInfoCompat, null)) {
  Toast.makeText(
      context,
      "ErrAddToHomescreen: requestPinShortcut() failed",
      Toast.LENGTH_LONG
  ).show();
}
```

In `BaseConversationListFragment.java`, the source contains another shortcut creation flow similar to:

```java
() -> {
  if (!ShortcutManagerCompat.requestPinShortcut(activity, shortcutInfoCompat, null)) {
    Toast.makeText(
        activity,
        "ErrAddToHomescreen: requestPinShortcut() failed",
        Toast.LENGTH_LONG
    ).show();
  }
}
```

These calls appear to be in UI-facing Activity/Fragment flows and are surrounded by UI feedback code such as `Toast.makeText(...).show()`. We did not see an obvious worker-thread offload around the `requestPinShortcut(...)` calls in these snippets.

On Android API 26+, `ShortcutManagerCompat.requestPinShortcut(...)` goes through the platform pinned-shortcut API using a platform `ShortcutInfo`. The platform `ShortcutManager#requestPinShortcut(ShortcutInfo, IntentSender)` documentation says the method may take several seconds to complete and should only be called from a worker thread.

## Verified source-level traces

```text
Trace 1: WebxdcActivity.java
  -> Webxdc/add-to-homescreen shortcut flow
  -> Toast.makeText(...).show()
  -> ShortcutManagerCompat.requestPinShortcut(context, shortcutInfoCompat, null)
  -> AndroidX pinned-shortcut compatibility wrapper
  -> on newer Android versions, platform ShortcutManager pinned-shortcut API
  -> platform API is documented as potentially slow and worker-thread-only
  -> likely reached from an Activity/UI path
  -> possible UI freeze / delayed response / ANR risk
```

```text
Trace 2: BaseConversationListFragment.java
  -> conversation/home-screen shortcut creation flow
  -> lambda using Activity context
  -> ShortcutManagerCompat.requestPinShortcut(activity, shortcutInfoCompat, null)
  -> AndroidX pinned-shortcut compatibility wrapper
  -> on newer Android versions, platform ShortcutManager pinned-shortcut API
  -> platform API is documented as potentially slow and worker-thread-only
  -> likely reached from a Fragment/UI path
  -> possible UI freeze / delayed response / ANR risk
```

## Why this matters

Requesting a pinned shortcut can involve launcher/system behavior. Even if it usually returns quickly, Android documents the underlying platform API as potentially slow. Running this request on the UI thread can make the shortcut creation flow feel frozen or delayed on some launchers/devices.

## Possible fix

The app can keep the Activity/Fragment flow in the foreground, but move the shortcut request itself to a worker thread and post only UI feedback back to the main thread.

For example:

```java
ExecutorService executor = Executors.newSingleThreadExecutor();

executor.execute(() -> {
  boolean success = ShortcutManagerCompat.requestPinShortcut(
      activity,
      shortcutInfoCompat,
      null
  );

  activity.runOnUiThread(() -> {
    if (!success) {
      Toast.makeText(
          activity,
          "ErrAddToHomescreen: requestPinShortcut() failed",
          Toast.LENGTH_LONG
      ).show();
    }
  });
});
```

One nuance: the platform API requires the caller to have a foreground Activity or foreground service. That does not necessarily mean the API call itself must run on the UI thread; it means the app must still be in a foreground state when the request is made.

## References

* AndroidX wrapper API:
  https://developer.android.com/reference/androidx/core/content/pm/ShortcutManagerCompat#requestPinShortcut(android.content.Context,androidx.core.content.pm.ShortcutInfoCompat,android.content.IntentSender)

* Android platform API:
  https://developer.android.com/reference/android/content/pm/ShortcutManager#requestPinShortcut(android.content.pm.ShortcutInfo,%20android.content.IntentSender)

Relevant Android platform documentation wording:

> This method may take several seconds to complete, so it should only be called from a worker thread.

adbenitez:
close #4418 

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -545,11 +545,18 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
               .build();
 
       Toast.makeText(context, R.string.one_moment, Toast.LENGTH_SHORT).show();
-      if (!ShortcutManagerCompat.requestPinShortcut(context, shortcutInfoCompat, null)) {
-        Toast.makeText(
-                context, "ErrAddToHomescreen: requestPinShortcut() failed", Toast.LENGTH_LONG)
-            .show();
-      }
+      Util.runOnAnyBackgroundThread(
+          () -> {
+            if (!ShortcutManagerCompat.requestPinShortcut(context, shortcutInfoCompat, null)) {
+              Util.runOnMain(
+                  () ->
+                      Toast.makeText(
+                              context,
+                              "ErrAddToHomescreen: requestPinShortcut() failed",
+                              Toast.LENGTH_LONG)
+                          .show());
+            }
+          });
     } catch (Exception e) {
       Toast.makeText(context, "ErrAddToHomescreen: " + e, Toast.LENGTH_LONG).show();
     }
--- a/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -547,7 +547,14 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
       Toast.makeText(context, R.string.one_moment, Toast.LENGTH_SHORT).show();
       Util.runOnAnyBackgroundThread(
           () -> {
-            if (!ShortcutManagerCompat.requestPinShortcut(context, shortcutInfoCompat, null)) {
+            boolean success;
+            try {
+              sucess = ShortcutManagerCompat.requestPinShortcut(context, shortcutInfoCompat, null);
+            } catch (Exception e) {
+              Log.e(TAG, "ErrAddToHomescreen: requestPinShortcut() failed", e);
+              success = false;
+            }
+            if (!success) {
               Util.runOnMain(
                   () ->
                       Toast.makeText(
--- a/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -549,7 +549,7 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
           () -> {
             boolean success;
             try {
-              sucess = ShortcutManagerCompat.requestPinShortcut(context, shortcutInfoCompat, null);
+              success = ShortcutManagerCompat.requestPinShortcut(context, shortcutInfoCompat, null);
             } catch (Exception e) {
               Log.e(TAG, "ErrAddToHomescreen: requestPinShortcut() failed", e);
               success = false;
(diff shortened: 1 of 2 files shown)


src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -545,11 +545,18 @@ public static void addToHomeScreen(Activity activity, int msgId) {
...
 
       Toast.makeText(context, R.string.one_moment, Toast.LENGTH_SHORT).show();
-      if (!ShortcutManagerCompat.requestPinShortcut(context, shortcutInfoCompat, null)) {
-        Toast.makeText(
-                context, "ErrAddToHomescreen: requestPinShortcut() failed", Toast.LENGTH_LONG)
-            .show();
-      }
+      Util.runOnAnyBackgroundThread(
+          () -> {
+            if (!ShortcutManagerCompat.requestPinShortcut(context, shortcutInfoCompat, null)) {
```
wchen342: This is a pre-existing issue, but `requestPinShortcut` can actually throw so we need to wrap it in `try...catch` too.

adbenitez: I was wondering what the outer try-catch was for because after temporarily removing it there was no compilation error not sure if we can move it inside or leave the outer try-catch and add a new one inside

wchen342: I don't know the original intention, but I guess it may be because of `requestPinShortcut`, or all the blob operations. Either way, now the request is in lambda on a separate thread it needs its own catch because the try block on main will not catch on a separate thread.

adbenitez: ok I wrapped this and the other call with a try-catch

wchen342 (approved):
LGTM

PR #4425, opened 2026-05-11
Call: Add notice when offline

### LINKED ISSUE #4423: User can call while offline without any hint
it is confusing that the app allows you to start a call while offline, would be nice to show a dialog to the users notifying they seem to be offline (no connectivity) not sure if we can detect if the phone is connected to some wifi/mobile-data otherwise if it is not reliable the dialog should just show a warning and allow to proceed anyways

wchen342:
Fixes #4423

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/calls/CallUtil.java
+++ b/src/main/java/org/thoughtcrime/securesms/calls/CallUtil.java
@@ -57,15 +57,22 @@ public class CallUtil {
       return;
     }
 
+    Runnable proceedWithCall =
+        () -> {
+          int accId = DcHelper.getContext(context).getAccountId();
+          coordinator.initiateOutgoingCall(accId, chatId, startsWithVideo);
+        };
+
     if (!isNetworkAvailable(context)) {
       new AlertDialog.Builder(context)
           .setMessage(context.getString(R.string.call_requires_connection))
-          .setPositiveButton(android.R.string.ok, null)
+          .setPositiveButton(R.string.perm_continue, (dialog, which) -> proceedWithCall.run())
+          .setNegativeButton(android.R.string.cancel, null)
           .show();
+      return;
     }
 
-    int accId = DcHelper.getContext(context).getAccountId();
-    coordinator.initiateOutgoingCall(accId, chatId, startsWithVideo);
+    proceedWithCall.run();
   }
 
   @Nullable
--- a/src/main/java/org/thoughtcrime/securesms/calls/CallUtil.java
+++ b/src/main/java/org/thoughtcrime/securesms/calls/CallUtil.java
@@ -10,8 +10,10 @@ import android.util.Log;
 import android.widget.Toast;
 import androidx.annotation.Nullable;
 import androidx.annotation.RequiresApi;
+import androidx.appcompat.app.AlertDialog;
 import androidx.core.telecom.CallEndpointCompat;
 import com.b44t.messenger.DcChat;
+import com.b44t.messenger.DcContext;
 import com.bumptech.glide.load.engine.DiskCacheStrategy;
 import org.thoughtcrime.securesms.R;
 import org.thoughtcrime.securesms.connect.DcHelper;
@@ -52,7 +54,17 @@ public class CallUtil {
       return;
     }
 
-    int accId = DcHelper.getContext(context).getAccountId();
+    DcContext dcContext = DcHelper.getContext(context);
+
+    if (dcContext.getConnectivity() < DcContext.DC_CONNECTIVITY_WORKING) {
+      new AlertDialog.Builder(context)
+          .setMessage(context.getString(R.string.call_requires_connection))
+          .setPositiveButton(android.R.string.ok, null)
+          .show();
+      return;
+    }
+
+    int accId = dcContext.getAccountId();
     coordinator.initiateOutgoingCall(accId, chatId, startsWithVideo);
   }
 
--- a/src/main/java/org/thoughtcrime/securesms/calls/CallUtil.java
+++ b/src/main/java/org/thoughtcrime/securesms/calls/CallUtil.java
@@ -5,6 +5,9 @@ import android.content.Intent;
 import android.graphics.Bitmap;
 import android.graphics.drawable.Drawable;
 import android.graphics.drawable.Icon;
+import android.net.ConnectivityManager;
+import android.net.Network;
+import android.net.NetworkCapabilities;
 import android.os.Build;
 import android.util.Log;
 import android.widget.Toast;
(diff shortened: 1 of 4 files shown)


src/main/java/org/thoughtcrime/securesms/calls/CallUtil.java
```diff
@@ -52,7 +54,17 @@ private static void startCall(Context context, int chatId, boolean startsWithVid
       return;
     }
 
-    int accId = DcHelper.getContext(context).getAccountId();
+    DcContext dcContext = DcHelper.getContext(context);
+
+    if (dcContext.getConnectivity() < DcContext.DC_CONNECTIVITY_WORKING) {
```
adbenitez: I think at the moment if any of the relays is not connected the connectivity will be "not connected" then, this is rather a bug in core but until that is fixed we can not merge this since it will make calling impossible

wchen342: Ah ok, that is unfortunate. I can change it to just use the Android OS's detection instead for now.

I suppose [this](https://github.com/chatmail/core/issues/8141) is the core issue?

wchen342: I have updated it to use Android's own detection.

src/main/res/values/strings.xml
```diff
@@ -440,6 +440,7 @@
     <string name="call_answered_elsewhere">Call answered on another device</string>
     <string name="call_requires_camera_permission">Camera permission is required for video calls</string>
     <string name="call_requires_mic_permission">Microphone permission is required for calls</string>
+    <string name="call_requires_connection">Cannot start call. Please check your internet connection and try again.</string>
```
r10s: slightly more actionable, and less vague advice:

```suggestion
    <string name="call_requires_connection">Cannot start call. Make sure your device has an internet connection and try again.</string>
```

r10s (commented):
i am unsure about that.

connectivity is flaky, at other places, eg. in core, we on purpose do not check a "connected" state as that tends to be invalid at the moment it is requested. detection of "device is offline" failed in the past. instead, we just try to connect.

checking telegram and signal in flight mode, this is what they're doing as well. whatsapp otoh has a similar question, and checking that before placing a call is okay, and another things as "sending a message". 

so, not against merging, just want to mention previous thoughts in that area.

cc @link2xt

src/main/java/org/thoughtcrime/securesms/calls/CallUtil.java
```diff
@@ -138,4 +150,17 @@ public static int getIconResByCallEndpoint(CallEndpointCompat endpoint) {
...
+    ConnectivityManager manager =
+        (ConnectivityManager) context.getSystemService(Context.CONNECTIVITY_SERVICE);
+    if (manager == null) return true;
+
+    Network network = manager.getActiveNetwork();
+    if (network == null) return false;
+
+    NetworkCapabilities caps = manager.getNetworkCapabilities(network);
+    return caps != null && caps.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET);
+  }
```
adbenitez, line 179: will this return `true` if user has hotspot setup?? had the experience in the past when trying to use things offline ex in a mesh network where one device would act as hotspot that some apps would not allow to use some feature that would absolute work

for example if user sets up a hotspot and uses yggmail to talk to another user connected to their hotspot and then they can't do calls because the client will think there is no wifi connected but in fact user is the hotspot so there is some network

otherwise if it is not possible to determine, it is better if the dialog instead of being a blocker dialog it is just a warning and has a button "call" besides "cancel" in so the user can continue calling anyways if they really want to try to, this is a rather corner case tho so maybe it is ok like that for now

wchen342: What do you mean hotspot? If you are connected to a hotspot then you are connected to internet right? Or if upstream is not connected then you will not be able to reach a server anyway, so either way it is just like connecting to another wifi.

adbenitez: not connected to a hotspot but when **you** are the hotspot, other people are the one to connect to your hotspot/network so your wifi connection is off

image


also if you are connected to some wifi intranet like in a University in Cuba or street mesh, I am afraid NET_CAPABILITY_INTERNET might be false while they are actually using a local email server in LAN

wchen342: For hotspot, if you are still connected to mobile network it will just go through mobile, in that case it is fine.

In the case of a limited network, `NET_CAPABILITY_INTERNET` only detects if the phone is connected to some kind of network using i.e. wifi or mobile. It does not try to connect to google server, so it shall be fine even in LAN. The one you are thinking is probably `NET_CAPABILITY_VALIDATED`. See https://developer.android.com/reference/android/net/NetworkCapabilities.

wchen342:
@r10s I think in this case it is fine, because in the current implementation we are not actually detecting connections to specific servers. The only thing get checked is if the phone has a wifi or mobile connection up, which is very basic and does not require reaching a specific server.

src/main/java/org/thoughtcrime/securesms/calls/CallUtil.java
adbenitez: I confirmed that when I connect to another phone's (off-grid) hotspot it allows me to call, but when I am the hotspot (my phone has a hotspot setup and people are connected to me but I am not connected to any wifi because I am the network/mesh center) then I can not start a call, I think you could then check if `caps.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET) || dcContext.getConnectivity() >= DcContext.DC_CONNECTIVITY_WORKING` this check would be safer because you rather have internet connectivity or core is connected successful to a local/LAN server like p2p yggmail server in mesh and there is no need to restrict calling since core seems to be connected

I think this is better since it is just about some warning that catches most cases and not blocking things completely for the user and not being able to call when it would actually be possible

wchen342: Yeah that can work too. Also I'm going to downgrade it to a warning and does not stop the call from going forward just in case.

wchen342: Updated.

adbenitez (approved):

adbenitez:
it would be nicer if the call is only started after user clicks "OK" (which should probably something like "continue" or "call") in the dialog to give chance to click cancel and enable internet

with the current more flexible check you probably can abort calling as well that will be fine

wchen342:
@adbenitez I have changed it to use a dialog with `Continue` for continue the call regardless and `Cancel` to cancel the call.

PR #4426, opened 2026-05-11
migrate to ViewPager2

### LINKED ISSUE #4422: Page navigation in RTL languages
- Android version: 15
- Delta Chat version: 2.46.0
- Steps to reproduce the problem: Change DeltaChat language to Persian, click on the QR icon at the top bar and try to navigate between pages
- Actual behavior: Page changes in the wrong way

adbenitez:
currently we use the old unmaintained `androidx.viewpager.widget.ViewPager`, this is not recommended anymore and the recommended maintained alternative is `androidx.viewpager2.widget.ViewPager2` which is its successor, solving many of the unsolved problems with `ViewPager` like lack of RTL support as reported at #4422 

the Android docs have a migration guide at https://developer.android.com/develop/ui/views/animations/vp2-migration#java

- [x] migrate `MediaPreviewActivity` to ViewPager2
- [x] migrate `QrActivity`  to ViewPager2
- [x] migrate `AllMediaActivity`  to ViewPager2
- [x] update changelog

close #4422


src/main/java/org/thoughtcrime/securesms/qr/QrActivity.java
```diff
@@ -228,47 +224,24 @@ public void onClick(View v) {
...
-          break;
-
-        default:
-          fragment = new QrScanFragment();
-          break;
-      }
-
-      return fragment;
+    public Fragment createFragment(int position) {
+      return position == TAB_SHOW ? qrShowFragment : new QrScanFragment();
```
wchen342: According to the migration guide, `createFragment` needs to supplies a new fragment instance each time the function is called.

adbenitez: I had noticed but was not sure if it was really important, fixed it now to always create a new fragment

wchen342 (approved):
LGTM!

PR #4428, opened 2026-05-12
prefer 'audio' over 'switch speaker', 

r10s:
'audio' is more correct, others are using that as well, and the translations are there already

see discussion at https://github.com/deltachat/deltachat-android/pull/4416#pullrequestreview-4252074882

to be clear: the string is used as invisible label only

#skip-changelog


wchen342 (approved):

adbenitez (approved):
small note: I think "audio output" would be more clear than just "audio" and btw telegram was using "speaker" for it (but maybe they are smart and adapt the name when you actually have headsets so I don't know)

r10s:
telegram iOS uses "speaker" as well, but i think, it's fine

PR #4430, opened 2026-05-13
rencode several videos at the same time when sharing to a single chat

### LINKED ISSUE #3928: videos not always recoded

adbenitez:
I gave up for now in trying to also encode them when sharing to several chats at the same time after trying several things, because the current sharing logic is too complicated heavily relays on finishing the `ConversationListRelayingActivity` which makes it impossible to display the `wait a moment...` dialog and also the way we send to multiple chats needs to change first to instead of iterating over every chat and then send all messages into it, to: iterate over every message and send to all chats, otherwise the same video is re-encoded over and over again one time per chat it is sent so the waiting time multiplies per chat

close #3928


CHANGELOG.md
```diff
@@ -20,7 +20,11 @@
 * Allow to disable incoming calls notifications
 * Fix: do not accidentally set draft in chats that don't allow sending messages
 * Fix swipe navigation between tabs in RTL languages
+<<<<<<< adb/issues-3928
+* Compress videos when sharing multiple files into a chat
```
wchen342: Conflict not resolved?

src/main/java/org/thoughtcrime/securesms/util/SendRelayedMessageUtil.java
```diff
@@ -99,28 +106,65 @@ public static void sendMultipleMsgs(
...
       }
       for (Uri uri : uris) {
-        dcContext.sendMsg(chatId, createMessage(context, uri, null));
+        DcMsg msg = createMessage(context, uri, null);
+        if (!isFinishing && isVideoUri(context, uri)) {
+          if (!VideoRecoder.prepareVideo(context, chatId, msg)) {
+            Log.w(TAG, "prepareVideo failed for " + uri);
+          }
+        }
+        dcContext.sendMsg(chatId, msg);
```
wchen342, line 140: Is it intentional to send the message even the video fails?

src/main/java/org/thoughtcrime/securesms/util/SendRelayedMessageUtil.java
```diff
@@ -99,28 +106,65 @@ public static void sendMultipleMsgs(
     ArrayList<Uri> uris = sharedUris;
     String text = sharedText;
 
+    AtomicReference<ProgressDialog> progressDialogRef = new AtomicReference<>(null);
+    boolean hasVideos = containsVideoType(context, uris);
+    boolean isFinishing = context instanceof Activity && ((Activity) context).isFinishing();
+    if (hasVideos && !isFinishing) {
```
wchen342, line 112: The finishing check needs to be on main, otherwise it may be outdated when `ProgressDialog.show` runs.

src/main/java/org/thoughtcrime/securesms/util/SendRelayedMessageUtil.java
```diff
@@ -99,28 +106,65 @@ public static void sendMultipleMsgs(
...
+        if (!isFinishing && isVideoUri(context, uri)) {
+          if (!VideoRecoder.prepareVideo(context, chatId, msg)) {
+            Log.w(TAG, "prepareVideo failed for " + uri);
+          }
+        }
+        dcContext.sendMsg(chatId, msg);
       }
     }
+
+    ProgressDialog dialog = progressDialogRef.get();
```
wchen342, line 144: This shall be on main too I think.

CHANGELOG.md
adbenitez: damn! I resolved it via the GitHub UI (just clicking buttons without actually editing the file manually), it was having glitches yesterday and somehow did that :/

src/main/java/org/thoughtcrime/securesms/util/SendRelayedMessageUtil.java
adbenitez: yes, if re-encoding/compressing the video fails we just do the same as before and try to send the video without compression, otherwise they will not be able to send them at all

adbenitez:
superseded by https://github.com/deltachat/deltachat-android/pull/4573

PR #4431, opened 2026-05-13
add "Enforce e2ee" switch

adbenitez:
- [x] add new switch/toggle to "Enforce Encryption" in relay settings
- [x] add new "Enforce Encryption" string
- [x] use new core API to actually set the setting

this requires a core with https://github.com/chatmail/core/pull/8220


r10s:
rebased PR and fixed assignment to the "more options" group, also moving it down there

PR #4432, opened 2026-05-14
Parse invitation link from search

### LINKED ISSUE #4340: allow to use invite link by pasting in search
this is already supported by desktop, if you paste an invite link in the search it will offer you to chat with the contact in case it is a 1:1 invite or join the group/channel in case is a group/channel invite

this is something users might try to do naturally to use an invite link since the "paste from clipboard" option in the QR scanner is harder to discover

wchen342:
Fixes #4340

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
@@ -245,8 +245,9 @@ public class ConversationListItem extends RelativeLayout
       avatar.setAvatar(glideRequests, recipient, false);
       avatar.setSeenRecently(contact.wasSeenRecently());
     } else {
-      GeneratedContactPhoto photo = new GeneratedContactPhoto(inviteData.getDisplayTitle());
-      avatar.setImageDrawable(photo.asDrawable(getContext(), 0xFF808080, true));
+      avatar.setImageDrawable(
+          new GeneratedContactPhoto("+")
+              .asDrawable(getContext(), ThemeUtil.getDummyContactColor(getContext())));
       avatar.setSeenRecently(false);
     }
   }
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
@@ -44,9 +44,11 @@ import org.thoughtcrime.securesms.components.AvatarView;
 import org.thoughtcrime.securesms.components.DeliveryStatusView;
 import org.thoughtcrime.securesms.components.FromTextView;
 import org.thoughtcrime.securesms.connect.DcHelper;
+import org.thoughtcrime.securesms.contacts.avatars.GeneratedContactPhoto;
 import org.thoughtcrime.securesms.database.model.ThreadRecord;
 import org.thoughtcrime.securesms.mms.GlideRequests;
 import org.thoughtcrime.securesms.recipients.Recipient;
+import org.thoughtcrime.securesms.search.QrInviteData;
 import org.thoughtcrime.securesms.util.DateUtils;
 import org.thoughtcrime.securesms.util.ThemeUtil;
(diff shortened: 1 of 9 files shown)


src/main/java/org/thoughtcrime/securesms/search/QrInviteData.java
```diff
@@ -0,0 +1,91 @@
...
+    int contactId = 0;
+
+    switch (state) {
+      case DcContext.DC_QR_ASK_VERIFYCONTACT:
+      case DcContext.DC_QR_FPR_OK:
+      case DcContext.DC_QR_ADDR:
+        contactId = qrParsed.getId();
+        DcContact contact = dcContext.getContact(contactId);
+        title = contact.getDisplayName();
+        subtitle = contact.getAddr();
```
adbenitez: we don't show addr for contacts, subtitle could be `R.strings.start_chat`

src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
```diff
@@ -216,6 +218,39 @@ public void bind(
...
+    if (inviteData.getContactId() > 0) {
+      DcContext dcContext = DcHelper.getContext(getContext());
+      DcContact contact = dcContext.getContact(inviteData.getContactId());
+      Recipient recipient = new Recipient(getContext(), contact);
+      avatar.setAvatar(glideRequests, recipient, false);
+      avatar.setSeenRecently(contact.wasSeenRecently());
+    } else {
+      GeneratedContactPhoto photo = new GeneratedContactPhoto(inviteData.getDisplayTitle());
+      avatar.setImageDrawable(photo.asDrawable(getContext(), 0xFF808080, true));
+      avatar.setSeenRecently(false);
```
adbenitez: instead of having such gray letter avatar it will be better if it was like the (+) gray icons in contact list or perhaps with the QR icon like the "New Contact" avatar in the contact list

wchen342: This is only for groups and channels, for normal accounts it will go the first branch.

src/main/java/org/thoughtcrime/securesms/search/QrInviteData.java
wchen342: Changed.

src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
adbenitez: yep I know, and for contacts it will have the contact color already so that is ok, but for groups and channels it looks weird that the letter avatar is first gray and will immediately get color after clicking the link, so better not pretend that is the channel avatar and show something that better indicate it is about joining it

wchen342: Oh ok, let me just show a plus for them then.

wchen342: Updated.

src/main/res/values/strings.xml
```diff
@@ -1053,6 +1053,7 @@
     <string name="set_name_and_avatar_explain">Set a name that your contacts will recognize. You can also set a profile image.</string>
     <string name="please_enter_name">Please enter a name.</string>
     <string name="qraccount_qr_code_cannot_be_used">The scanned QR code cannot be used to set up a new profile.</string>
+    <string name="qr_invitation_header">Invitation Link</string>
```
adbenitez: maybe instead of adding a new string just use the existing "Link" string that is more generic? because a proxy is not really an invitation link etc

adbenitez (approved):

src/main/res/values/strings.xml
wchen342: True, that is simpler.

adbenitez (approved):

PR #4434, opened 2026-05-15
Improve workflows

wchen342:
Fixes https://github.com/deltachat/deltachat-android/security/code-scanning/1. Also add additional pinned commits.


wchen342:
It is showing two comments because the old workflow is still being triggered from `workflow_run` that runs on `main`, not PR branch. It will resolve itself when this is merged.

.github/workflows/preview-apk.yml
```diff
@@ -70,7 +74,22 @@ jobs:
...
+          ARTIFACT_URL: ${{ steps.upload.outputs.artifact-url }}
+        with:
+          script: |
+            const url = process.env.ARTIFACT_URL;
+            await github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              body: `**To test the changes in this pull request, install this apk:**\n\n[📦 app-preview.apk](${url})`,
+            });
```
adbenitez, line 95: super cool to get rid of 3rd party actions doing things like this!

adbenitez (approved):

PR #4443, opened 2026-05-27
Fix debouncing in chatlist and search

link2xt:
Reset inLoadChatlist only right before exit
in loadChatlistAsync().
Otherwise if loadChatlistAsync() is called
while the loop is running loadChatlist() or Util.sleep(100),
loadChatlistAsync() will see inLoadChatlist=false,
set needsAnotherLoad=true and start a second loop.
This way it is possible to spawn any number of background loops
running loadChatlist() simultaneously
so no debouncing actually happened.

Debouncing in SearchViewModel.updateQuery()
is fixed similarly by copying the code structure from loadChatlistAsync().
Previously it did not even have the lock.

Seems to be introduced in https://github.com/deltachat/deltachat-android/pull/2001

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListFragment.java
@@ -240,9 +240,9 @@ public class ConversationListFragment extends BaseConversationListFragment
 
   public void loadChatlistAsync() {
     synchronized (loadChatlistLock) {
-      needsAnotherLoad = true;
       if (inLoadChatlist) {
         Log.i(TAG, "chatlist loading debounced");
+        needsAnotherLoad = true;
         return;
       }
       inLoadChatlist = true;
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListFragment.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListFragment.java
@@ -251,6 +251,9 @@ public class ConversationListFragment extends BaseConversationListFragment
     Util.runOnAnyBackgroundThread(
         () -> {
           while (true) {
+            Log.i(TAG, "executing debounced chatlist loading");
+            loadChatlist();
+
             synchronized (loadChatlistLock) {
               if (!needsAnotherLoad) {
                 inLoadChatlist = false;
@@ -259,8 +262,6 @@ public class ConversationListFragment extends BaseConversationListFragment
               needsAnotherLoad = false;
             }
 
-            Log.i(TAG, "executing debounced chatlist loading");
-            loadChatlist();
             Util.sleep(100);
           }
         });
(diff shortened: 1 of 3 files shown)


src/main/java/org/thoughtcrime/securesms/ConversationListFragment.java
```diff
@@ -259,8 +262,6 @@ public void loadChatlistAsync() {
               needsAnotherLoad = false;
             }
 
-            Log.i(TAG, "executing debounced chatlist loading");
-            loadChatlist();
             Util.sleep(100);
```
adbenitez, line 265: this is wrong, now if loadChatlistAsync() is called again while previous call is currently busy in loadChatlist();  it will set  needsAnotherLoad = true; and return thinking the current running thread will take care about it, but then the thread finishes loadChatlist(); and proceeds to set needsAnotherLoad = false;
I guess to fix it you need to set needsAnotherLoad = false; before calling loadChatlist(); and remove the setting to false from the lines after loadChatlist();

edit: ah! nevermind the while(true) will run again, now every call to `loadChatlistAsync` will always run `loadChatlist();` twice tho

link2xt: > edit: ah! nevermind the while(true) will run again, now every call to loadChatlistAsync will always run loadChatlist(); twice tho

It either detects that `needsAnotherLoad` has been set, sleeps and runs `loadChatlist` another time, or, if nobody called `loadChatlistAsync` while the loop was running, just exits. I think it only runs twice if someone actually requested another loading, in which case second load happens 100 ms later.

wchen342:
Wait, isn't the original code correct? The changes will make `loadChatlist()` run twice because it is moved out of `synchronized`?

Edit: I meant above, not out.

link2xt:
> Wait, isn't the original code correct? The changes will make `loadChatlist()` run twice because it is moved out of `synchronized`?

`loadChatlist` is not in `synchronized` already. The change is about whether we set the flags before or after `loadChatlist`.

wchen342:
But we shall avoiding doing that I think? Because `loadChatlist` seems slow to me.

src/main/java/org/thoughtcrime/securesms/ConversationListFragment.java
adbenitez: `needsAnotherLoad` is set to true even the first time the function is called so the loop will run another time even with a single call to `loadChatlistAsync`, maybe it should only be set to true if `inLoadChatlist` is true

link2xt: I have moved `needsAnotherLoad` into `if (inLoadChatlist)`.

Still not great that this will not be triggered if the loop has just exited because it has no work to do, debouncing is not really doing anything then. But at least it works sometimes if another request arrives when chatlist is actually loading.

Maybe we should just drop the sleep statement as well. Nobody noticed so far that it was not working and we even ran concurrent chatlist loading.

adbenitez (approved):
at least this seems to be an improvement, if it can be further improved it can come in another PR

wchen342 (approved):
The current version looks correct to me. I think we shall keep the sleep and use this version.

PR #4449, opened 2026-05-28
fix sharing contact across profiles

### LINKED ISSUE #4445: share contact's vcard + switch profile doesn't work
when yo go to a contact's profile there is the option to "share" this is to share the contact's vcard, if you click there you get to the sharing screen, similar to when sharing other attachments, **but:**

- there is no option to save the vcard to local storage
- if you select "switch profile" then the vcard is no longer available when sharing to a chat (or a vcard of a wrong contact is shared instead)

adbenitez:
close #4445


wchen342 (approved):
LGTM

PR #4451, opened 2026-05-29
Reset scroll position after switching account

### LINKED ISSUE #4446: wrong chatlist possition when switching accounts
when you switch accounts, if you had the current chatlist scrolled somewhere down in the chat list, and then switch to a new profile, instead of the chatlist appearing at the top of the list for the new selected profile, the current chatlist scroll position is kept, which is a bit weird and unexpected

wchen342:
Fixes #4446


adbenitez (approved):
lgtm! thanks

PR #4460, opened 2026-06-02
Save audio draft on interrupt

### LINKED ISSUE #4437: Voice message is lost when receiving a phone call during recording
When recording a voice message in Delta Chat, there is a bug where the recording gets lost if a regular phone call interrupts the app.

Steps to reproduce
Open Delta Chat.
Start recording a voice message.
Receive an incoming phone call while recording.
Answer or dismiss the call.
Return to Delta Chat.
Expected behavior

The voice message recording should be automatically saved (or at least recoverable) after the interruption, or the app should prompt the user to continue/discard it.

Actual behavior

The recorded voice message is lost and not saved.

Environment
Delta Chat version: 2.49.0
OS: Android 16

This can be frustrating, especially for longer voice messages, because the recording disappears completely after a phone call interruption.

wchen342:
- Also fixes pre-existing bugs while attaching audio files as draft.
- When rejecting a call from notification, it no longer goes through activity (no black screen for a split second)
- Removed the deprecated `AsyncTask` in [AttachmentManager.java‎](https://github.com/deltachat/deltachat-android/pull/4460/changes#diff-610e8defba98a716becd78e3122eba4380fb7144eb464bc501204f372ce15459)
- Audio attachments are now saved properly like other file attachments when in a draft

Fixes #4437

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## Unreleased
 
 * Fix: avoid crashes in Media preview sometimes
+* Fix: Incorrect total time when attaching audio files as draft
+* Fix: Audio files in draft showing total time from wrong file
+* Voice recording will be automatically saved as draft when interrupted 
 
 ## 2.51.0
 2026-05
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -494,6 +494,11 @@
             android:foregroundServiceType="camera|microphone|phoneCall" />
 
         <receiver
+            android:name=".calls.CallActionReceiver"
+            android:enabled="true"
+            android:exported="false" />
+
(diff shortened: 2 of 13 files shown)


adbenitez:
now when I select a song as draft, play it, then get out of the chat, the first time it stops, but then I enter the chat again and play it and go out again then continues to play despite when entering the chat the play button is in "play" state

wchen342:
> now when I select a song as draft, play it, then get out of the chat, the first time it stops, but then I enter the chat again and play it and go out again then continues to play despite when entering the chat the play button is in "play" state

The audio attachment wasn't set correctly in draft before. I made a fix for that.

adbenitez (approved):
lgtm!

PR #4461, opened 2026-06-03
feat: use message style notifications

d2weber:
With MessagingStyle, Notifications can be expanded to show more text of the message than just a single line.

This feature has been requested in https://support.delta.chat/t/swipe-down-on-notification-to-expand-it/2742

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
+++ b/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
@@ -810,11 +810,12 @@ public class NotificationCenter {
     }
   }
 
-  public IconCompat getAvatarIcon(DcContact contact) {
-    return IconCompat.createWithBitmap(getAvatar(new Recipient(context, contact)));
+  public @Nullable IconCompat getAvatarIcon(DcContact contact) {
+    Bitmap avatar = getAvatar(new Recipient(context, contact));
+    return avatar != null ? IconCompat.createWithBitmap(avatar) : null;
   }
 
-  public Bitmap getAvatar(Recipient recipient) {
+  public @Nullable Bitmap getAvatar(Recipient recipient) {
     try {
       Drawable drawable;
       ContactPhoto contactPhoto = recipient.getContactPhoto(context);
--- a/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
+++ b/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
@@ -63,8 +63,18 @@ public class NotificationCenter {
   private volatile long lastAudibleNotification = 0;
   private static final long MIN_AUDIBLE_PERIOD_MILLIS = TimeUnit.SECONDS.toMillis(2);
 
-  // Map<accountId, Map<chatId, lines>, contains the last lines of each chat for each account
-  private final HashMap<Integer, HashMap<Integer, LinkedHashMap<Integer, String>>> inboxes =
+  private static class NotifData {
+    final Person sender;
+    final String text;
+
+    NotifData(Person sender, String text) {
+      this.sender = sender;
+      this.text = text;
+    }
+  }
+
+  // notification history of each chat for each account
+  private final HashMap<Integer, HashMap<Integer, LinkedHashMap<Integer, NotifData>>> inboxes =
       new HashMap<>();
 
   public NotificationCenter(Context context) {
@@ -412,27 +422,31 @@ public class NotificationCenter {
           DcMsg dcMsg = dcContext.getMsg(msgId);
           NotificationPrivacyPreference privacy = Prefs.getNotificationPrivacy(context);
 
+          DcContact sender = dcContext.getContact(dcMsg.getFromId());
+          String senderName = dcMsg.getSenderName(sender);
           String shortLine =
               privacy.isDisplayMessage()
                   ? dcMsg.getSummarytext(2000)
                   : context.getString(R.string.notify_new_message);
           String tickerLine = shortLine;
-          if (!dcChat.isMultiUser() && privacy.isDisplayContact()) {
-            tickerLine =
-                dcMsg.getSenderName(dcContext.getContact(dcMsg.getFromId())) + ": " + tickerLine;
 
-            if (dcMsg.getOverrideSenderName() != null) {
-              // There is an "overridden" display name on the message, so, we need to prepend the
-              // display name to the message,
-              // i.e. set the shortLine to be the same as the tickerLine.
-              shortLine = tickerLine;
-            }
+          NotifData notifData =
+              new NotifData(
+                  new Person.Builder()
+                      .setName(senderName)
+                      .setIcon(getAvatarIcon(sender))
+                      .setBot(sender.isBot())
+                      .build(),
+                  shortLine);
+
+          if (!dcChat.isMultiUser() && privacy.isDisplayContact()) {
+            tickerLine = senderName + ": " + tickerLine;
           }
 
           DcMsg quotedMsg = dcMsg.getQuotedMsg();
           boolean isMention = dcChat.isMultiUser() && quotedMsg != null && quotedMsg.isOutgoing();
 
-          maybeAddNotification(accountId, dcChat, msgId, shortLine, tickerLine, true, isMention);
+          maybeAddNotification(accountId, dcChat, msgId, notifData, tickerLine, true, isMention);
         });
   }
 
@@ -449,6 +463,7 @@ public class NotificationCenter {
           }
 
           DcContact sender = dcContext.getContact(contactId);
+          String senderName = dcMsg.getSenderName(sender);
           String shortLine =
               context.getString(
                   R.string.reaction_by_other,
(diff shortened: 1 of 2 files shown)


src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -797,8 +810,11 @@ private void rebuildNotification(
     }
   }
 
-  public Bitmap getAvatar(DcChat dcChat) {
-    Recipient recipient = new Recipient(context, dcChat);
+  public IconCompat getAvatarIcon(DcContact contact) {
+    return IconCompat.createWithBitmap(getAvatar(new Recipient(context, contact)));
```
wchen342: Need a null check for `getAvatar`.

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -694,14 +692,29 @@ private void buildAndShowChatNotification(
...
+          style.setConversationTitle(dcChat.getName());
+          for (Map.Entry<Integer, String> msgEntry : messagesForInbox.entrySet()) {
+            DcMsg msg = dcContext.getMsg(msgEntry.getKey());
+            DcContact senderContact = dcContext.getContact(msg.getFromId());
+            Person sender =
+                new Person.Builder()
+                    .setName(msg.getSenderName(senderContact))
+                    .setIcon(getAvatarIcon(senderContact))
+                    .build();
+            style.addMessage(msgEntry.getValue(), msg.getSortTimestamp(), sender);
```
wchen342: `getSortTimestamp` returns timestamp in seconds, but `addMessage` expect timestamps in ms, so you need to multiple it by 1000.

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -694,14 +692,29 @@ private void buildAndShowChatNotification(
...
+              new Person.Builder()
+                  .setName(selfContact.getDisplayName())
+                  .setIcon(getAvatarIcon(selfContact))
+                  .build();
+          NotificationCompat.MessagingStyle style = new NotificationCompat.MessagingStyle(self);
+          style.setGroupConversation(dcChat.isMultiUser());
+          style.setConversationTitle(dcChat.getName());
+          for (Map.Entry<Integer, String> msgEntry : messagesForInbox.entrySet()) {
+            DcMsg msg = dcContext.getMsg(msgEntry.getKey());
+            DcContact senderContact = dcContext.getContact(msg.getFromId());
```
wchen342: I think this does not work for reactions or maybe webxdc, but I am not entirely sure. @adbenitez

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
d2weber: Good catch, thank you! `setIcon` accepts null, so I just adjusted `getAvatarIcon` to handle null.

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
d2weber: Just checked and it is indeed broken currently, the sender is wrong for reactions and webxdc updates.

This is what it looks like when Bob reacts to a message. I'm not sure what we want to show here. Currently the translation string includes the name of the sender of the reaction, if we also set the `sender` in `addMessage`  (hence there would be "Bob" instead of "Me") the sender name would be duplicated. Maybe that's fine for now. 

image_2026-06-03_20-40-37 (Edited)


I started to draft a fix for these issues, I think I will put a class `NotifData` instead of just `String` in the `messagesForInbox`.

~I'm not sure though what to put in that class besides the message text: either~

~* the senders contact id: this would allow for easier avatar retrieval but I think overrideSenderName could not be supported. The reason is that we cannot pass a meaningful contact id if the overrideSenderName is set. If we try to get the overrideSenderName inside the inner `buildAndShowChatNotification` method, we don't know anymore if its a reaction/webxdc in which case we should never use the messages overrideSenderName.~
~* the senders name and avatar icon.~

d2weber: I just pushed my attempt to fix reactions and webxdc updates. Webxdc updates just don't set a sender name, they should just appear without a name, and with the group image as avatar.

Feedback welcome!

wchen342: Can you check when the Notification->Show setting is changed to "Name only", does it still show sender name in a group chat? I tested and it works correctly in a 1 to 1 chat, but in a group chat it no longer shows the sender name.

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -694,14 +719,24 @@ private void buildAndShowChatNotification(
...
+          Person self =
+              new Person.Builder()
+                  .setName(selfContact.getDisplayName())
+                  .setIcon(getAvatarIcon(selfContact))
+                  .build();
+          NotificationCompat.MessagingStyle style = new NotificationCompat.MessagingStyle(self);
+          style.setGroupConversation(dcChat.isMultiUser());
+          style.setConversationTitle(dcChat.getName());
+          for (Map.Entry<Integer, NotifData> msgEntry : messagesForInbox.entrySet()) {
+            long timestamp_ms = dcContext.getMsg(msgEntry.getKey()).getSortTimestamp() * 1000;
```
wchen342, line 790: Can we cache it in `NotifData` instead of read messages in a loop?

d2weber: Hm reading this now.. Should we add the timestamp to NotifData? Not sure what effects that would have.

wchen342: On a second thought this is probably fine because reading db isn't that slow. My original thought is to add a timestamp field to `NotifData` and include that in `notifyMessage`, but I think this is fine.

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
d2weber: Your are right, I just pushed d757ef59c which re-adds the removed code which was used in that case.

wchen342 (approved):
Need a rebase, otherwise lgtm

d2weber:
CI fails because this PR comes from my fork I guess..

wchen342:
Ah interesting, I thought removing the restriction on running was enough. I'm making another change.

wchen342:
It seems the original zizmor security alarm was a blank one and the original pattern is actually acceptable. In any case we shall merge this for now. The workflow problem is rarely encountered here anyway and we can adjust that later.

@adbenitez I actually can't merge. Do you have override?

adbenitez:
> @adbenitez I actually can't merge. Do you have override?

this should not be merged yet, I want to test and review carefully

adbenitez:
this is looking very promising! but it has some bugs and behavior changes:

when someone else send an info message inside a webxdc in a group it is showing sender "Me:" prefix when it was not me the sender of the status update, below you can see how it looks in ArcaneChat (old behavior before this change):

image

the sender is duplicated in the collapsed chat view for group chats:
image

there is no longer images for chats in the collapsed view (once you expand you start seeing avatars, not sure if this is expected behavior, needs to be tested in Signal, WhatsApp, Telegram):
image


when sending webxdc notification in 1:1 chat, instead of the entry appearing as yet another message from the person, it duplicates the contact (you see two avatars of the same person as if it was two messages from different person) besides showing it as any other text from that sender, since a notification from webxdc is a bit special, a solution here could be to display the webxdc app icon as the avatar, that way it would be very much what you already have but no duplicate avatar for the same contact but an indication that that message comes from the webxdc:

image

d2weber:
> there is no longer images for chats in the collapsed view (once you expand you start seeing avatars, not sure if this is expected behavior, needs to be tested in Signal, WhatsApp, Telegram):

I checked with signal, and indeed they have nice big icons also in collapsed view. Turns out to make this work `builder.setShortcutInfo(..)` (or similar) is needed, see f0760d669. To create the shortcut I took inspiration from https://github.com/deltachat/deltachat-android/blob/4bdfe8e390edbd01190ba439f0674cd7d92a66a0/src/main/java/org/thoughtcrime/securesms/connect/DirectShareUtil.java#L124-L142

It seems to work but it feels weird to reuse `ShareActivity`.

d2weber:
> when someone else send an info message inside a webxdc in a group it is showing sender "Me:" prefix when it was not me the sender of the status update, below you can see how it looks in ArcaneChat (old behavior before this change):

Now I apply the messaging style also when not `isDisplayMessage()`, that way we don't have to prefix senders in the notification anymore. In the case where not `isDisplayMessage()` the inbox hash map is cleared so that only one "New Messages" is shown.

> when sending webxdc notification in 1:1 chat, instead of the entry appearing as yet another message from the person, it duplicates the contact 

I decided to handle webxdcs in 1:1 chats separately so they use the same sender name/avatar of the 1:1 chat instead of the webxdc name/avatar. That way the 1:1 chat stays clean with only a single sender name/avatar.

adbenitez (approved):
wow! awesome! great work! now it is much nicer! thanks a lot 🙏

PR #4462, opened 2026-06-05
Fix call not ending correctly under certain conditions

### LINKED ISSUE #4454: Device still on call mode after hang up
- Android version:
- Device: Samsung A15 5G 
- Delta Chat version: 2.50.0
- Expected behavior: Phone volumes back at normal after hang up.
- Actual behavior: if both sides hit the hang up button, one or both phones will stay on calling mode volume, killing alarms and notifications. The phone only actually hang up after killing ArcaneChat.
The call appears on the chat as ended, so if you guys ensure a hang up after the timer is stopped or something, pehaps it will avoid it to happening.

wchen342:
Fixes #4454

#skip-changelog


adbenitez (approved):
great thanks for fixing!

PR #4471, opened 2026-06-09
docs: fix links in BUILDING.md

r10s:
the redirect from `master` to `main` does not or no longer work or deeplinks.

came over that when checking which NDK version is actually recommended (it is NDK 27 :)

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -47,7 +47,7 @@ Nix development environment contains Rust with cross-compilation toolchains and 
 
 To [build an APK](https://developer.android.com/studio/build/building-cmdline) run the following 2 steps.
 Note that the first step may take some time to build for all architectures. You can optionally read 
-[the first comment block in the `ndk-make.sh` script](https://github.com/deltachat/deltachat-android/blob/main/scripts/ndk-make.sh)
+[the first comment block in the `ndk-make.sh` script](./scripts/ndk-make.sh)
 for pointers on how to build for a specific architecture.
 ```
 $ scripts/ndk-make.sh
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -135,8 +135,8 @@ See https://wiki.archlinux.org/index.php/Podman#Rootless_Podman for more informa
 
 To setup build environment manually:
 - _Either_, in Android Studio, go to "Tools / SDK Manager / SDK Tools", enable "Show Package Details",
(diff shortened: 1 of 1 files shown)


BUILDING.md
```diff
@@ -47,7 +47,7 @@ Nix development environment contains Rust with cross-compilation toolchains and
 
 To [build an APK](https://developer.android.com/studio/build/building-cmdline) run the following 2 steps.
 Note that the first step may take some time to build for all architectures. You can optionally read 
-[the first comment block in the `ndk-make.sh` script](https://github.com/deltachat/deltachat-android/blob/master/scripts/ndk-make.sh) 
+[the first comment block in the `ndk-make.sh` script](https://github.com/deltachat/deltachat-android/blob/main/scripts/ndk-make.sh)
```
adbenitez: better use relative links to avoid future problems etc, also more comprehensible when reading the file offline / in the local filesystem:

```suggestion
[the first comment block in the `ndk-make.sh` script](./scripts/ndk-make.sh)
```

BUILDING.md
```diff
@@ -135,8 +135,8 @@ See https://wiki.archlinux.org/index.php/Podman#Rootless_Podman for more informa
 
 To setup build environment manually:
 - _Either_, in Android Studio, go to "Tools / SDK Manager / SDK Tools", enable "Show Package Details",
-  select "CMake" and the desired NDK (install the same NDK version as the [Dockerfile](https://github.com/deltachat/deltachat-android/blob/master/Dockerfile)), hit "Apply".
-- _Or_ read [Dockerfile](https://github.com/deltachat/deltachat-android/blob/master/Dockerfile) and mimic what it does.
+  select "CMake" and the desired NDK (install the same NDK version as the [Dockerfile](https://github.com/deltachat/deltachat-android/blob/main/Dockerfile)), hit "Apply".
+- _Or_ read [Dockerfile](https://github.com/deltachat/deltachat-android/blob/main/Dockerfile) and mimic what it does.
```
adbenitez: same here:

```suggestion
  select "CMake" and the desired NDK (install the same NDK version as the [Dockerfile](./Dockerfile)), hit "Apply".
- _Or_ read [Dockerfile](./Dockerfile) and mimic what it does.
```

adbenitez (approved):

PR #4473, opened 2026-06-09
update NDK version in CI workflows

adbenitez:
wanted to update to 30, but it seems that is not "stable release" yet, latest stable release at https://developer.android.com/ndk/downloads/revision_history is r29


wchen342 (approved):
lgtm, thanks!

PR #4476, opened 2026-06-10
Call maybeNetwork() on network unblock

### LINKED ISSUE #4459: maybe_network is called when network is blocked
> it's indeed a bug that we call maybe_network when the network is blocked.

_Originally posted by @link2xt in https://github.com/chatmail/notifiers/issues/85#issuecomment-4595432054_

---

I checked [uses of maybe network](https://github.com/search?q=repo%3Adeltachat%2Fdeltachat-android+%28maybe_network+OR+maybenetwork%29&type=code) and it seems only the json rpc api is used for that, not cffi. The [json rpc docs for `maybeNetwork`](https://github.com/chatmail/core/blob/50e83f2072e4e0a5db1d1f52909b14b27d47067d/deltachat-jsonrpc/src/api.rs#L2071-L2072) say:
```
    /// Indicate that the network likely has come back.
    /// or just that the network conditions might have changed
```
Should I open a PR on core to remove the second line and add a separate method for `maybeNetworkLost` which can then be used in android?

wchen342:
Fixes #4459

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
+++ b/src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
@@ -275,7 +275,9 @@ public class ApplicationContext extends MultiDexApplication {
               Log.i(
                   "DeltaChat",
                   "++++++++++++++++++ NetworkCallback.onAvailable() #" + debugOnAvailableCount++);
-              getDcAccounts().maybeNetwork();
+              if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) {
+                getDcAccounts().maybeNetwork();
+              }
             }
 
             @Override
@@ -283,8 +285,13 @@ public class ApplicationContext extends MultiDexApplication {
                 @NonNull android.net.Network network, boolean blocked) {
               Log.i(
                   "DeltaChat",
-                  "++++++++++++++++++ NetworkCallback.onBlockedStatusChanged() #"
+                  "++++++++++++++++++ NetworkCallback.onBlockedStatusChanged("
+                      + blocked
+                      + ") #"
                       + debugOnBlockedStatusChangedCount++);
+              if (!blocked) {
+                getDcAccounts().maybeNetwork();
+              }
             }
 
             @Override
--- a/src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
+++ b/src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
@@ -275,6 +275,7 @@ public class ApplicationContext extends MultiDexApplication {
               Log.i(
                   "DeltaChat",
                   "++++++++++++++++++ NetworkCallback.onAvailable() #" + debugOnAvailableCount++);
+              // onBlockedStatusChanged is only available on API 29+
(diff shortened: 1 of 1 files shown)


wchen342:
The reason is currently, when the device is in Doze, the system blocks network and IMAP receive ECONNABORTED. Core then tries  reconnection cycle and hits rate limite and waits long time before retrying.

When an FCM notification arrives and the system temporarily unblocks network, `onBlockedStatusChanged` is called but the core is not notified, which stucks waiting, wasting the limited 20s window. This is what is causing missing notifications.

wchen342:
Also the high-priority FCM messages shall be independent of this fix.

adbenitez (approved):

link2xt:
This does not fix the problem that maybeNetwork is called when the network is known to be blocked.

The issue is that the network becomes available, maybeNetwork is called and connection fails, then the network becomes unblocked slightly later. Calling maybeNetwork then will not really work because previous connection already failed and there will be a backoff before trying again (with "IMAP got rate limited, waiting for ..." log) if it fails too often.

wchen342:
> This does not fix the problem that maybeNetwork is called when the network is known to be blocked.

That problem doesn't actually exists on Android (well technically there is one call but it is only below API 24, and our lowest API is 23 so almost no one will hit it).

> The issue is that the network becomes available, maybeNetwork is called and connection fails

I don't get this part, what do you mean network becomes available and connection fails? If the server is being blocked by, say ISP then it is a different problem.

link2xt:
What I mean is that when onAvailable is called, it is possible that the network is blocked, but we still call maybeNetwork and then connection will fail.

The change is an improvement because we at least don't call maybeNetwork when onBlocked clearly says the network has just become blocked, but this is maybe not enough to avoid calling maybeNetwork in all cases where the network is known to be blocked.

wchen342:
@link2xt Ah I finally understand what you meant. See my updated code, I added a version check to the `onAvailable` call. It cannot be completely removed because `onBlockedStatusChanged` doesn't fire under API 24-28 and there is no reliable ways to tell the blocking state (unless we ping something manually) to detect if network is blocked.

src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
```diff
@@ -275,16 +275,23 @@ public void onAvailable(@NonNull android.net.Network network) {
               Log.i(
                   "DeltaChat",
                   "++++++++++++++++++ NetworkCallback.onAvailable() #" + debugOnAvailableCount++);
-              getDcAccounts().maybeNetwork();
+              if (Build.VERSION.SDK_INT < Build.VERSION_CODES.Q) {
```
link2xt, line 279: `Q` is because `onBlockedStatusChanged` does not exist before Q? A comment explaining this would be useful.

link2xt (approved):

src/main/java/org/thoughtcrime/securesms/ApplicationContext.java
wchen342: Comment added.

PR #4482, opened 2026-06-11
don't create mailto unencrypted chat if force-encryption is enabled

### LINKED ISSUE #4481: check if offering to start a chat on `mailto:` link makes sense
came over that at https://github.com/deltachat/deltachat-desktop/issues/5737 - we should also on android check if force_encryption is set - and if so, show the standard "get in contact" dialog.

changes since the original issue: even if some bots have exceptions, since 2.51, the client discards unencrypted messages by default.

adbenitez:
close #4481 

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/qr/QrCodeHandler.java
+++ b/src/main/java/org/thoughtcrime/securesms/qr/QrCodeHandler.java
@@ -314,6 +314,12 @@ public class QrCodeHandler {
 
   private void showFingerprintOrQrSuccess(
       AlertDialog.Builder builder, DcLot qrParsed, String name) {
+    if (qrParsed.getState() == DcContext.DC_QR_ADDR
+        && dcContext.getConfigInt(DcHelper.CONFIG_FORCE_ENCRYPTION) == 1) {
+      DcHelper.prepareInvalidUnencryptedDialog(activity, builder);
+      return;
+    }
+
     @StringRes
     int resId =
         qrParsed.getState() == DcContext.DC_QR_ADDR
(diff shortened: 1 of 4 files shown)


src/main/java/org/thoughtcrime/securesms/qr/QrCodeHandler.java
```diff
@@ -314,6 +314,12 @@ private void showDoneToast() {
 
   private void showFingerprintOrQrSuccess(
       AlertDialog.Builder builder, DcLot qrParsed, String name) {
+    if (qrParsed.getState() == DcContext.DC_QR_ADDR
+        && dcContext.getConfigInt(DcHelper.CONFIG_FORCE_ENCRYPTION) == 1) {
+      DcHelper.prepareInvalidUnencryptedDialog(activity, builder);
```
wchen342, line 319: I wonder if it is possible when `contact != null` but `getChatIdByContactId` returns 0?

wchen342 (approved):
Lgtm, the other comment is just a question and non-critical.

src/main/java/org/thoughtcrime/securesms/qr/QrCodeHandler.java
adbenitez: yep, it is possible that the contact exists but no 1:1/dm chat exists with that contact, in that case getChatIdByContactId will be zero, but how is that relevant here in this place where you are commenting about it? 🤔

wchen342: So when a chat doesn't exist we do not need to check the `force_encryption` option? Because it will go to the next branch which does not have the check.

adbenitez: what next branch?? in this whole `showFingerprintOrQrSuccess` function `getChatIdByContactId` is not used at all, please provide more context

what I can tell is: in any case, if there is a chat or not it doesn't matter because all what the existing code does is "if there is not 1:1 chat yet, then create it" and what we want to do is prevent any chat creation when clicking a mailto link,

what could happen is the contrary: if the unencrypted chat already exists, but we still want to show the error because sending in that chat will not work anyways

PR #4485, opened 2026-06-12
Add notification for missed calls

### LINKED ISSUE #4468: missed calls don't trigger notification
when the app is online, a new call comes, it rings, but if you were away from the phone, and the person cancels the call after long waiting, when you come back to the phone there is no trace of the missed call

we can add a "missed call" notification after removing the "ringing incoming call" notification checking if the status of the call is missed/canceled

there is https://github.com/chatmail/core/pull/7955 that would make this simpler but for now it is also possible to implement without it

this is only for calls that are received while the app is online and a ringing notification was displayed, if the app was offline and you then connect, already missed notifications are already received/notified as missing calls

wchen342:
Fixes #4468

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java
+++ b/src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java
@@ -73,7 +73,16 @@ public class CallCoordinator implements DcEventCenter.DcEventDelegate {
   private static final String CHANNEL_ID_ONGOING = "voip_ongoing_calls";
   private static final String CHANNEL_ID_MISSED = "voip_missed_calls";
   private static final int NOTIFICATION_ID_CALL = 1001;
-  private static final int NOTIFICATION_ID_MISSED_CALL = 1002;
+  static final int NOTIFICATION_ID_MISSED_CALL = 1002;
+
+  private static final int PI_ANSWER = 0;
+  private static final int PI_DECLINE = 1;
+  private static final int PI_FULLSCREEN = 2;
+  private static final int PI_HANGUP = 3;
+  private static final int PI_ONGOING_CONTENT = 4;
+  private static final int PI_MISSED_CONTENT = 5;
+  private static final int PI_MISSED_CALLBACK = 6;
+  private static final int PI_MISSED_MESSAGE = 7;
 
   private static final String CALL_IDENTIFIER_SCHEME = "deltachat:";
 
@@ -883,7 +892,7 @@ public class CallCoordinator implements DcEventCenter.DcEventDelegate {
               // This event is problematic because it can trigger in both directions,
               // in addition to multiple other scenarios which cannot easily be distinguished
               // May cause problems in edge cases
-              onCallEnded(accId, callId);
+              onCallEnded(accId, callId, startsWithVideo);
               break;
           }
         });
@@ -1034,7 +1043,7 @@ public class CallCoordinator implements DcEventCenter.DcEventDelegate {
     showOrUpdateOngoingNotification(appContext.getString(R.string.call_with, calleeName));
   }
 
-  private synchronized void onCallEnded(int accId, int callId) {
+  private synchronized void onCallEnded(int accId, int callId, boolean startsWithVideo) {
     Log.d(TAG, "onCallEnded: accId=" + accId + ", callId=" + callId);
 
     if (!hasActiveCall()) {
@@ -1069,7 +1078,7 @@ public class CallCoordinator implements DcEventCenter.DcEventDelegate {
     }
 
     if (isIncomingCall && !wasAnsweredLocally) {
-      showMissedCallNotification(activeAccId, activeChatId);
+      showMissedCallNotification(activeAccId, activeChatId, startsWithVideo);
     }
 
     // Clear active states
@@ -1434,7 +1443,7 @@ public class CallCoordinator implements DcEventCenter.DcEventDelegate {
     PendingIntent answerPendingIntent =
         PendingIntent.getActivity(
             this.appContext,
-            0,
+            PI_ANSWER,
             answerIntent,
             PendingIntent.FLAG_UPDATE_CURRENT | PendingIntent.FLAG_IMMUTABLE);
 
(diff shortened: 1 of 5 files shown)


src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java
```diff
@@ -1490,6 +1507,52 @@ private void showIncomingCallNotification(String callerName, Icon callerIcon) {
...
+        PendingIntent.getActivity(
+            appContext,
+            NOTIFICATION_ID_MISSED_CALL,
+            intent,
+            PendingIntent.FLAG_UPDATE_CURRENT | PendingIntent.FLAG_IMMUTABLE);
+
+    Notification.Builder builder =
+        new Notification.Builder(appContext, CHANNEL_ID_MISSED)
+            .setSmallIcon(R.drawable.icon_notification)
+            .setContentTitle("Missed call")
```
adbenitez: hardcoded string should use the existing missed call string

adbenitez:
the notification is a bit plain/not-so-useful, it doesn't have button to "reply" to send some text back or maybe a "call back" button, but this is just my guess what such notifications usually have, needs to be tried what Telegram/WhatsApp do here, already a good step to have the notification tho

wchen342:
Signal doesn't have a separate "Missed Call" type notification. They actually use the normal message notification with an entry of the "missed call" event, so it will look the same as other messages. Shall we do the same approach?

Also @r10s

adbenitez:
> Signal doesn't have a separate "Missed Call" type notification. They actually use the normal message notification with an entry of the "missed call" event, so it will look the same as other messages. Shall we do the same approach?

I think what makes more sense is to look into Telegram/WhatsApp, long term, Signal is not great at UI/UX, but if re-using existing notification is easy so at least the mark-read & reply button are there, that will be good enough for now

src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java
wchen342: Fixed.

wchen342:
I looked up Whatsapp notification and it seems they just have a notification with title as the caller, "Missed Call" as text and two action buttons. I have made the notification the same way.

adbenitez:
if you click the "Message" option, the notification is not removed

wchen342:
> if you click the "Message" option, the notification is not removed

Ok that is fixed now.

adbenitez:
something is weird about the way the buttons work, selecting either of both wouldn't focus the activity that is launched but stay on the notifications view, when the notification is clicked it opens the activity properly :thinking:

wchen342:
It is by design, the action buttons on notifications are meant to keep notifications in foreground, see https://developer.android.com/develop/ui/compose/notifications/create-notification#Actions.
We can technically dismiss the notification manually but that will go against the design intention of the buttons.

adbenitez:
the notification is dismissed, and an outgoing call notification is created, the problem here is the way the starting of the call notification is triggered

not so long ago we had the issue that "answer" / "decline" actions in a call notification was in fact opening activity in full screen even for decline briefly, so we just need to do whatever we where doing there

what contradicts the docs you sent is the `But these action buttons must not duplicate the action performed when the user taps the notification.`  because we add a "Message" action that is basically the same as clicking the notification, I think it still makes sense to have the explicit action tho

wchen342:
What do you mean? The actions doesn't go through activities so that problem doesn't apply here. I think we are not talking about the same thing? Maybe give me a recording to show what you were talking about initially?

adbenitez:
1. expand the notifications view of the phone completely so it covers all the screen
2. click the "call back" action in the missed call notification
3. the view remains the same, the list of notifications, the missed call notification disappear and now there is an outgoing call notification, but the call activity is covered still by the notifications list overlay, the activity is opened in the background

wchen342:
Yeah that is the intended behavior. The message button is probably redundant, so we can remove that one. But the action button normally do not launch activity on its own.
But I think there can be an argument about call shall be foreground by default, which is technically not the intended usage of the action buttons. Then again Whatsapp does that anyway so I guess we can just launch an activity from that.

wchen342:
Actually it is very difficult to launch activity from action buttons on Android 15+, it has multiple BAL restrictions:

```
Activity start ONLY allowed by BAL_ALLOW_NON_APP_VISIBLE_WINDOW uid has non-app visible window {2000=1, 2013=1, 2024=2, 2040=1}: [callingPackage: chat.delta.beta; callingPackageTargetSdk: 36; callingUid: 10226; callingPid: -1; appSwitchState: 2; callingUidHasVisibleActivity: false; callingUidHasVisibleNotPinnedActivity: false; callingUidHasNonAppVisibleWindow: false; callingUidProcState: LAST_ACTIVITY; isCallingUidPersistentSystemProcess: false; allowBalExemptionForSystemProcess: true; intent: Intent { act=chat.delta.beta.CALL_BACK flg=0x14000000 xflg=0x4 cmp=chat.delta.beta/org.thoughtcrime.securesms.calls.CallActivity (has extras) }; callerApp: null; balAllowedByPiCreator: BSP.ALLOW_BAL; resultIfPiCreatorAllowsBal: BAL_BLOCK; callerStartMode: MODE_BACKGROUND_ACTIVITY_START_ALLOWED; hasRealCaller: true; isCallForResult: false; isPendingIntent: true; autoOptInReason: null; realCallingPackage: com.android.systemui; realCallingPackageTargetSdk: 36; realCallingUid: 10185; realCallingPid: 1077; realCallingUidHasVisibleActivity: false; realCallingUidHasVisibleNotPinnedActivity: false; realCallingUidHasNonAppVisibleWindow: true; realCallingUidProcState: PERSISTENT_UI; isRealCallingUidPersistentSystemProcess: true; originatingPendingIntent: PendingIntentRecord{beee31a chat.delta.beta startActivity (allowlist: 2505e83:+30s0ms/0/NOTIFICATION_SERVICE/NotificationManagerService)}; realCallerApp: ProcessRecord{e0e8dce 1077:com.android.systemui/u0a185}; realInVisibleTask: false; balAllowedByPiSender: BSP.ALLOW_BAL; resultIfPiSenderAllowsBal: BAL_ALLOW_NON_APP_VISIBLE_WINDOW; realCallerStartMode: MODE_BACKGROUND_ACTIVITY_START_ALWAYS; balDontBringExistingBackgroundTaskStackToFg: true]
```

Note I have already used `MODE_BACKGROUND_ACTIVITY_START_ALLOWED` to opt-in, but it is still blocked.

There *may* be a workaround, but at this point I think it is not worth the trouble. The call will start in background with the ongoing call notification, but without the screen opened immediately.

adbenitez:
ok, then let's leave it like that for now, but I doubt that is how it works in WhatsApp etc

adbenitez (approved):
thanks a lot for the PR and for trying to solve the situation with the buttons!

PR #4488, opened 2026-06-17
fix: edit avatar btn: Group Name and Image -> Edit

WofWca:
Since recently there is also "description"
in the "Edit Group" / "Edit Channel" dialog,
so the string is pretty much unapplicable anymore.

There is also an MR to remove its usage from Desktop:
https://github.com/deltachat/deltachat-desktop/pull/6484.
On iOS looks like it's already unused.

TODO:
- [x] Do the regular string removal procedure, unless it's enough to just remove it from the .xml files here.
- [ ] Assess the overall idea of this MR.

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -288,7 +288,7 @@
     <string name="menu_send">Send</string>
     <string name="menu_toggle_keyboard">Toggle Emoji Keyboard</string>
     <string name="menu_edit_group">Edit Group</string>
-    <!-- deprecated: "Edit Group" dialog usually also contains the "Group Description" field -->
+    <!-- deprecated: "Edit Group" dialog usually also contains the "Group Description" field, use global_menu_edit_desktop instead -->
     <string name="menu_group_name_and_image">Group Name and Image</string>
     <string name="menu_show_global_map">All Locations</string>
     <string name="menu_archive_chat">Archive Chat</string>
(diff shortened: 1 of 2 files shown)


adbenitez:
I don't like that the string says _desktop in general we should avoid such string keys, the string comment could say if it is desktop-only at that time, then now it would be a matter of updating the comment without triggering translations

there is also menu_edit_group that seems to be unused now

cc @r10s

adbenitez:
@WofWca I recommend you update this PR to only remove the string usage from the code and leave the string there for it to be later deprecated/removed that would make the PR easier to review (and the string might also be in use by other clients like ios or Ubuntu Touch etc), that or do 2 commits one with the code and another with the removal from xml

r10s:
it is an old string, the _desktop suffix is not used since years. we have lots of inconsistencies in the keys over the years, it is like it is. changing keys is not worth the effort and discussions. at least not now.

and yes, the string to be removed should just me marked as "deprecated" - there are lots of them, just do the same. the "deprecated" comment should go to english source only - all other language files are generated anyways and will get it on next translation pull


from time to time we check if a string is really unused, and then finally remove in a batch

WofWca:
Right, I forgot about the "deprecated" thing.

adbenitez (approved):

r10s (approved):

src/main/res/values/strings.xml
```diff
@@ -288,6 +288,7 @@
     <string name="menu_send">Send</string>
     <string name="menu_toggle_keyboard">Toggle Emoji Keyboard</string>
     <string name="menu_edit_group">Edit Group</string>
+    <!-- deprecated: "Edit Group" dialog usually also contains the "Group Description" field -->
```
r10s:
```suggestion
    <!-- deprecated: "Edit Group" dialog usually also contains the "Group Description" field, use global_menu_edit_desktop instead -->
```

r10s (approved):

PR #4491, opened 2026-06-22
allow to select multiple files

adbenitez:
some users have asked me for this here and there, and also happens quite often to myself when I want to send images as files to avoid compression, have to do it one-by-one 

also made the "allow multi-selection" flag more explicit than implicit and removed the unnecessary overridden variants

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
@@ -441,32 +441,19 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
           else mediaType = MediaType.IMAGE;
           setMedia(singleUri, mediaType);
         } else {
-          final ClipData multipleUris = data.getClipData();
-          if (multipleUris != null) {
-            final int uriCount = multipleUris.getItemCount();
-            if (uriCount > 0) {
-              ArrayList<Uri> uriList = new ArrayList<>(uriCount);
-              for (int i = 0; i < uriCount; i++) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -441,32 +441,19 @@ public void onActivityResult(final int reqCode, int resultCode, Intent data) {
...
+          sendMultipleMsgs(data);
         }
         break;
 
       case PICK_DOCUMENT:
-        final String docMimeType = MediaUtil.getMimeType(this, data.getData());
-        final MediaType docMediaType =
-            MediaUtil.isAudioType(docMimeType) ? MediaType.AUDIO : MediaType.DOCUMENT;
-        setMedia(data.getData(), docMediaType);
+        if (data.getData() != null) { // single Uri
```
wchen342, line 449: We shall check `clipData.getItemCount() > 1`, otherwise if the user select only one file then they will for example lose the opportunity to add message, etc.

adbenitez: mmmh, is it really needed? I tested selecting a single file (by long-press and entering selection mode) and it attaches it just fine, same for images

wchen342: Interesting, so if I pick from the default file picker then it does work as expect. I guess that shall be good enough.

wchen342 (approved):
LGTM, thanks!

PR #4493, opened 2026-06-22
properly hide all input&drafting elements on in-chat search

### LINKED ISSUE #4492: draft attachment not hidden when search is started
when you start a search in a chat, the input bar is hidden, but if there is an attachment draft, it will remain visible covering the view

adbenitez:
close #4492

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
@@ -1142,17 +1142,17 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
 
   private void setInputPanelVisibility(boolean isInitialization) {
     int inputPanelVisibility;
-    boolean isAttamchentHidden;
+    boolean isAttachmentHidden;
     if (dcChat.canSend()) {
       inputPanelVisibility = View.VISIBLE;
-      isAttamchentHidden = false;
+      isAttachmentHidden = false;
       // FIXME: disabled for now to avoid problems with chat scrolling and keyboard covering input
       // bar
       // ViewUtil.forceApplyWindowInsets(findViewById(R.id.root_layout), true, false, true, true);
       // fragment.handleRemoveBottomInsets();
     } else {
       inputPanelVisibility = View.GONE;
-      isAttamchentHidden = true;
+      isAttachmentHidden = true;
       hideSoftKeyboard();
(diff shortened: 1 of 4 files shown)


adbenitez:
while reviewing use "hide whitespaces" to make it easier to review changes in conversation_activity.xml

r10s (approved):
lgtm, thanks for fixing!

wchen342:
There is a minor UI problem: the voice recording button and the "slide up" thing is now clipped by the bounding box of the input panel. Especially the sliding is invisible.

adbenitez:
@wchen342 good that you noticed that! I reverted back to the "dumb"/simpler logic of just keeping track of view visibility and individually hiding them, in the process fixed an existing bug in that logic: if while user is searching the chat status changes (ex. the user gets removed from group) then the `beforeSearch*Visibility` variable had an outdated visibility and would then make the `inputPanel` visible again when it shouldn't, this is still not 100% bulletproof and I would have liked to avoid that situation altogether by having some way to hide visibility of all elements at once but I don't know if there is such a simple solution without breaking the layout

wchen342:
Yeah I tested and if you are searching and for example group name changes and then the input panel will re-appear. Maybe add

```
if (!searchActive) {
        inputPanel.setVisibility(View.VISIBLE);
        ...
```

which shall be able to force the input not show up when search is on?

adbenitez:
ah yes, I had seen that in the past, had forgot about it!

adbenitez:
I pushed a better fix

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1136,21 +1136,23 @@ private void initializeResources() {
...
 
-    setComposePanelVisibility(true);
+    setInputPanelVisibility(true);
     initializeContactRequest();
   }
 
-  private void setComposePanelVisibility(boolean isInitialization) {
+  private void setInputPanelVisibility(boolean isInitialization) {
+    int inputPanelVisibility;
+    boolean isAttamchentHidden;
```
wchen342: Typo?

wchen342 (approved):
Everything looks good now!

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
adbenitez: looks more like dyslexia 😵‍💫 LOL

PR #4495, opened 2026-06-23
properly handle messages with ~overrideSenderName

### LINKED ISSUE #4494: notifications are not respecting ~overridden sender name
when a message has overridden sender name, it should appear in the notification, not the name of the original contact, this happens with bots that use the overridden sender name property when federating messages from other platforms, or in channels when they use the channel name as display name

this bug was introduced in #4461

cc @d2weber

adbenitez:
close #4494 

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
+++ b/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
@@ -429,11 +429,11 @@ public class NotificationCenter {
 
           DcContact sender = dcContext.getContact(dcMsg.getFromId());
           String senderName = dcMsg.getSenderName(sender);
-          String contactId = String.valueOf(sender.getId());
+          String personId = accountId + "-" + dcMsg.getFromId();
           if (dcMsg.getOverrideSenderName() != null) {
             // we need to treat the contact as a separate Person with different ID
             // otherwise the name will be overwritten by future notifications
-            contactId += "-" + senderName;
+            personId += "-" + senderName;
           }
           String text =
               privacy.isDisplayMessage()
@@ -450,7 +450,7 @@ public class NotificationCenter {
                       .setName(senderName)
                       .setIcon(getAvatarIcon(sender))
                       .setBot(sender.isBot())
-                      .setKey(accountId + "-" + contactId)
+                      .setKey(personId)
                       .build(),
                   text);
 
@@ -478,12 +478,11 @@ public class NotificationCenter {
             // just do nothing.
           }
 
-          DcContact sender = dcContext.getContact(contactId);
-          String senderName = dcMsg.getSenderName(sender);
+          DcContact contact = dcContext.getContact(contactId);
           String text =
               context.getString(
                   R.string.reaction_by_other,
-                  sender.getDisplayName(),
+                  contact.getDisplayName(),
                   reaction,
                   dcMsg.getSummarytext(2000));
           DcChat dcChat = dcContext.getChat(dcMsg.getChatId());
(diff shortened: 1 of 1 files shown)


src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -444,7 +450,7 @@ public void notifyMessage(int accountId, int chatId, int msgId) {
                       .setName(senderName)
                       .setIcon(getAvatarIcon(sender))
                       .setBot(sender.isBot())
-                      .setKey(accountId + "-" + sender.getId())
+                      .setKey(accountId + "-" + contactId)
```
d2weber: Could we maybe just add `senderName` to the key? That would avoid the separte `getOverrideSenderName() != null` branch and `senderName` should already provide the required uniqueness because it respects overrideSenderName.

We might want to use the same fix in these places:
https://github.com/deltachat/deltachat-android/blob/dfd323a89fb627631ff15014c69139755393fd23/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java#L491


https://github.com/deltachat/deltachat-android/blob/dfd323a89fb627631ff15014c69139755393fd23/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java#L447

https://github.com/deltachat/deltachat-android/blob/dfd323a89fb627631ff15014c69139755393fd23/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java#L545

adbenitez: > could we maybe just add senderName to the key?

no because we don't want 2 contact with the same name to be treated as the same, and most of the time there is no overrideSenderName so using the contact ID is fine

> We might want to use the same fix in these places: ...

thanks for pointing that out! I realized there is a bug in the `notifyReaction` if it ever happens the message has overriden-sender-name then the reaction will show that name instead of the name of the person that is reacting to the message, this in practice never happens atm in existing clients, but it is unnecessary and confusing nonetheless that in notifyReaction we use `dcMsg.getSenderName(sender)` it should use always the contact's name, not the msg's author since that is just the target message from self that is being reacted upon

will fix that in the process

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -472,23 +478,22 @@ public void notifyReaction(int accountId, int contactId, int msgId, String react
             // just do nothing.
           }
 
-          DcContact sender = dcContext.getContact(contactId);
-          String senderName = dcMsg.getSenderName(sender);
```
adbenitez, line 476: for `notifyReaction` we always want to notify with the name of the person that is reacting, not the sender of the message that is being reacted upon (that will always be ourselves) probably dcMsg.getSenderName() function should check that dcMsg.getFromId() matches sender.getId() and throw exception to avoid such errors of passing a wrong contact, or maybe even remove the function altogether, it hurts more than what it helps

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -539,10 +544,10 @@ public void notifyWebxdc(int accountId, int contactId, int msgId, String text) {
             notifData =
                 new NotifData(
                     new Person.Builder()
-                        .setName(dcMsg.getSenderName(sender))
```
adbenitez, line 542: for `notifyWebxdc` we also don't want to use `dcMsg.getSenderName` because dcMsg might be the parent Webxdc instance message, the only situation when dcMsg is a message that could contain override-sender-name is when it is an info-message, but even then: the current webxdc API doesn't support sending an status update with overriden-sender-name so it is impossible the message would have one, bots better relay on setting the name as prefix to the notification text instead in that case and that will work even in case the notification has not corresponding in-chat info-message

wchen342:
The two comments from adb needs to be addressed, otherwise I don't see any other issues.

adbenitez:
> The two comments from adb needs to be addressed, otherwise I don't see any other issues.

my comments are just explaining why I did remove the `dcMsg.getSenderName()` usage that is meant for getting overridden-sender-name

wchen342:
Oh sorry I thought [d2weber](https://github.com/d2weber) made the PR, got confused because they made the original change to the notification.

wchen342 (approved):

PR #4496, opened 2026-06-25
Fix video capture to query camera formats and enforce aspect ratio

### LINKED ISSUE #4489: WebRTC adaptive video resolution
Right now video stream sent through WebRTC is hardcoded to 1280x720, which may not have the same aspect ratio as the preview. The unintended part of camera capture sent to remote may be a privacy concern in some cases.

Ref: https://github.com/deltachat/calls-webapp/pull/12, https://support.delta.chat/t/camera-coverage-differs-between-local-preview-and-what-is-sent-to-the-peer/5452

wchen342:
Fixes #4489

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Allow to select multiple files for sending
 * Add notifications for missed calls
+* Video call preview now accurately shows what is sent to remote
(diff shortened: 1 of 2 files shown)


adbenitez (approved):
tested it and it seems to work fine

PR #4497, opened 2026-06-26
when a webxdc is deleted close WebxdcActivity

adbenitez:
also if a chat is deleted, close webxdc, and chat instead of showing weird "ghost chat"

this can happen in situations like: 

* bot+webxdc, the bot deletes old webxdc instance and sends new one but user still has old one open and might try to use or not notice there is a new instance in the chat meanwhile, DC desktop already handles this properly
* you delete a chat in desktop while the chat is open in android, the chat remains open in android as a "ghost chat"

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
@@ -278,6 +278,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
     eventCenter.removeObservers(this);
 
     eventCenter.addMultiAccountObserver(DcContext.DC_EVENT_CHAT_MODIFIED, this);
+    eventCenter.addMultiAccountObserver(DcContext.DC_EVENT_CHAT_DELETED, this);
     eventCenter.addMultiAccountObserver(DcContext.DC_EVENT_CHAT_EPHEMERAL_TIMER_MODIFIED, this);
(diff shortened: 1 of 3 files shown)


src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1869,6 +1870,8 @@ public void handleEvent(@NonNull DcEvent event) {
       DcContact contact = recipient.getDcContact();
       titleView.setSeenRecently(
           contact != null ? dcContext.getContact(contact.getId()).wasSeenRecently() : false);
+    } else if (eventId == DcContext.DC_EVENT_CHAT_DELETED && event.getData1Int() == chatId) {
```
wchen342, line 1873: Shall we also clear shortcuts here?

adbenitez: doesn't make much sense here since this is only when the chat is open, if done probably better in a more global place and not depending on the chat being open

wchen342:
There is an edge case where you have a launcher shortcut created without this PR, and then if webxdc is opened from the shortcut when a chat is deleted then it will not close.

adbenitez:
> There is an edge case where you have a launcher shortcut created without this PR, and then if webxdc is opened from the shortcut when a chat is deleted then it will not close.

I added a commit to take care of being backward compatible with existing shortcuts that don't set  `EXTRA_CHAT_ID`

wchen342 (approved):
LGTM!

PR #4499, opened 2026-06-26
Update Dockerfile for add openjdk-17-jdk

klim-iv:
This patchset fix compile error inside docker:
* What went wrong: Execution failed for task ':compileFossReleaseJavaWithJavac'.
> Error while evaluating property 'javaCompiler' of task ':compileFossReleaseJavaWithJavac'.
   > Failed to calculate the value of task ':compileFossReleaseJavaWithJavac' property 'javaCompiler'.
         > Toolchain installation '/usr/lib/jvm/java-17-openjdk-amd64' does not provide the required capabilities: [JAVA_COMPILER]


klim-iv:
@adbenitez < could  you review this patchset, please

Dockerfile
```diff
@@ -8,8 +8,11 @@ wget \
 curl \
 unzip \
 openjdk-17-jre \
+openjdk-17-jdk \
+git tig gawk \
```
link2xt: We likely don't want vim, tig, gawk and other tools that are not related to building the app. Otherwise everyone will add their preferred text editors here.

This dockerfile is only for building, not a development environment.

link2xt:
For reference, buildings steps are at https://github.com/deltachat/deltachat-android/blob/main/BUILDING.md#build-using-dockerfile

Dockerfile
klim-iv: ok,  fixed

adbenitez (approved):

PR #4504, opened 2026-06-28
cancel search mode when back is pressed

### LINKED ISSUE #4501: pressing back while in in-chat search ends activity
instead, the expected behavior is that the search would be canceled

adbenitez:
close #4501


wchen342 (approved):
LGTM!

PR #4508, opened 2026-07-01
change default visibility of scroll_to_bottom_button to invisible

### LINKED ISSUE #4502: scroll-to-bottom button not appearing when "show in chat" is used
when inside a webxdc if you use the "show in chat" option you will go up in chat history to the webxdc instance, however the "scroll-to-bottom" floating `V` button will not appear in this situation, this might also happen with the "show in chat" option of media gallery etc

adbenitez:
close #4502

it seems the problem was because a race condition the animation being triggered when the view had zero size or something like that, I was a bit throwing spaghetti into the wall here but setting the visibility to invisible so it is size is calculated sooner seems to fix the problem

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * Fix: properly hide draft attachment during in-chat search
 * Fix: close mini-apps and chats if they are deleted
 * Fix: cancel in-chat search when back is pressed, instead of directly returning to chatlist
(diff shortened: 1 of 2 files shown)


wchen342 (approved):
So the problem seems to be the dimension of the button is 0 because of `gone` and animation plays with the 0 dimension which just ends in 0 too. Changing it to `invisible` is right.

PR #4509, opened 2026-07-02
Prevent scroll listener double initialization

### LINKED ISSUE #4507: addOnScrollListener added twice due ConversationFragment.onNewIntent()
while debugging a bit https://github.com/deltachat/deltachat-android/issues/4502 I discovered that two `OnScrollListener` objects are being registered because `initializeResources()` is called on `onViewCreated` but also on `ConversationFragment.onNewIntent()` so when `onNewIntent()` is called a new listener is registered causing the events to be processed double

this is not the root cause of #4502 even if you open the activity in a way that onNewIntent is avoided, the problem still happens, hence why I opened a separated issue, this just makes things harder to debug

wchen342:
Fixes #4507

#skip-changelog


adbenitez (approved):
lgtm, thanks for fixing!

PR #4510, opened 2026-07-03
remove unused legacy sms-related code from Signal

adbenitez:
this was from when Signal supported SMS, it was never used by us

#skip-changelog


src/main/res/values/attrs.xml
```diff
@@ -59,12 +59,6 @@
...
         name="conversation_background"
         format="reference|color" />
-    <attr
-        name="conversation_transport_sms_indicator"
-        format="reference" />
-    <attr
-        name="conversation_transport_push_indicator"
-        format="reference" />
     <attr
         name="conversation_transport_popup_background"
```
wchen342: This is also no longer used I think?

wchen342:
Also `AttachButtonLongClickListener` in `ConversationActivity` does nothing now.

adbenitez:
I pushed a commit removing the other stuff you pointed out, thanks!

wchen342 (approved):
LGTM!

PR #4515, opened 2026-07-06
Use notification channel settings for ringtone

### LINKED ISSUE #4513: Configuring incoming call notification channel sound is ignored
- Android version: can reproduce on both Android 16 and Android 17, running GrapheneOS
- Device: Pixel 9A
- Delta Chat version: Arcanechat 2.53
- Expected behavior: Ability to reconfigure sound for incoming call notifications
- Actual behavior: When configuring incoming call notification sound it's silently ignored or better to say not handled
    * Go to the system Settings app
    * Find notifications, then App notifications, then find Deltachat or Arcanechat in the list.
    * Under notification categories press others, then find Incoming calls and press that too.
    * Notice there is a way to change the sound for this notification channel.
    * Assign a different sount than the default, close the Settings app and wait for incoming call to ring.
    * When incoming call is ringing after you have changed the Incoming calls notification sound as described, the default device ring tone is played.

wchen342:
So that incoming call sound loop correct and can be customized by users.

For new users this works automatically. For existing installations, there may be a left over channel in their settings, but if one didn't change anything there then no actions are needed.

Fixes #4513

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java
+++ b/src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java
@@ -15,6 +15,9 @@ import android.content.Intent;
 import android.content.ServiceConnection;
 import android.content.pm.PackageManager;
 import android.graphics.drawable.Icon;
+import android.media.AudioAttributes;
+import android.media.AudioManager;
+import android.media.RingtoneManager;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Handler;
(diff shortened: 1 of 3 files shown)


src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java
```diff
@@ -170,11 +174,22 @@ public static synchronized CallCoordinator getInstance(Context context) {
   }
 
   private void createNotificationChannels() {
+    notificationManager.deleteNotificationChannel(CHANNEL_ID_INCOMING_OLD);
```
adbenitez: CHANNEL_ID_INCOMING_OLD is only used here, it is ok to just put the value here directly and keep the variable CHANNEL_ID_INCOMING with the new value

wchen342: True, the change is small enough so the variable is probably unnecessary. Updated.

adbenitez (approved):
tested and seems to work fine, good work!

with the previous behavior the notification sound options were even the normal notifications options, now it properly let you select ringtones instead

PR #4519, opened 2026-07-07
Linkifier: shorten links

### LINKED ISSUE #4518: URLs in message bubbles are too long
currently when user shares links in chats, the links can be very long and ugly, this is specially true for our own invite links, we could shorten the links similar to what Mastodon and other sites do:

Image

adbenitez:
close #4518 

## before
image

## after 
image

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/util/Linkifier.java
+++ b/src/main/java/org/thoughtcrime/securesms/util/Linkifier.java
@@ -1,6 +1,6 @@
 package org.thoughtcrime.securesms.util;
 
-import android.text.SpannableString;
+import android.text.SpannableStringBuilder;
 import android.text.Spanned;
 import android.text.style.URLSpan;
 import android.text.util.Linkify;
@@ -8,6 +8,9 @@ import java.util.regex.Pattern;
 
 /* Utility for text linkify-ing */
 public class Linkifier {
+  private static final int MAX_DISPLAY_LINK_LENGTH = 32;
+  private static final String ELLIPSIS = "…";
+
   private static final Pattern CMD_PATTERN =
       Pattern.compile("(?<=^|\\s)/[a-zA-Z][a-zA-Z@\\d_/.-]{0,254}");
   private static final Pattern PROXY_PATTERN =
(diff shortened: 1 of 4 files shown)


src/main/java/org/thoughtcrime/securesms/util/Linkifier.java
```diff
@@ -33,41 +36,90 @@ private static boolean internalPhoneLinkifierNeeded() {
...
     if (Linkify.addLinks(messageBody, PROXY_PATTERN, null, null, null)) {
-      replaceURLSpan(
-          messageBody); // replace URLSpan so that it is not removed on the next addLinks() call
+      // replace URLSpan so that it is not removed on the next addLinks() call
+      replaceURLSpan(messageBody, true, clickable);
     }
 
     int flags;
-    if (internalPhoneLinkifierNeeded()) {
+    if (clickable && internalPhoneLinkifierNeeded()) {
```
wchen342, line 114: Is there a reason why `internalPhoneLinkifierNeeded` is no longer used when `clickable` is false?

adbenitez: because that is only about linkifying phone numbers, if clickable is false we don't need to linkify so it is pointless to enter that branch and linkify phone numbers just to remove the links again

wchen342 (approved):
Tested and works fine, thanks!

src/main/java/org/thoughtcrime/securesms/util/Linkifier.java
```diff
@@ -76,7 +128,7 @@ public static SpannableString linkify(SpannableString messageBody) {
 
     // linkyfiy urls etc., this removes all existing URLSpan
     if (Linkify.addLinks(messageBody, flags)) {
-      replaceURLSpan(messageBody);
+      replaceURLSpan(messageBody, true, clickable);
```
adbenitez, line 131: here we call `Linkify.addLinks(messageBody, flags)` and `replaceURLSpan` even if clickable is false just so it shorten links, otherwise in multi-select mode suddenly the messages would have the original links and become much bigger which feels glitchy/hacky

wchen342: No that doesn't happen to me? I enter selection mode and the links are still shortened.

wchen342: Oh nvm you are just making a comment.

adbenitez: yes, because what I just said, if you would put `clickable &&` before `Linkify.addLinks` in that `if`, then you would get the original links without shortening, what I meant with my comment is just to explain why even if clickable is false (multi-select mode) we still need to call Linkify.addLinks there, just to then shorten the links even if there is no need for it to be clickable

wchen342: Yes I know. It is just my mind default to a comment under PR as some sort of issue, that's why I keep getting confused when you make explanatory comments like this😂

PR #4506, opened 2026-06-29
webxdc: hide "add to home screen" and "show in chat" when instance is just a draft

adbenitez:
close #4503 

#skip-changelog

PR #4521, opened 2026-07-07
webxdc drafts: show state in title

r10s:
show draft state already in title.
we got reports of ppl not aware they have to send an app for collaboration. to mitigate that, we set the subtitle to "tap 'send' to share". showing 'draft' when enlarged makes additionally clear that this is not sent out yet.
moreover, it removes clutter - the title and the chat is not really needed as the context is very much clear (more than eg. opening an app by icon from the media view)

counterpart of https://github.com/deltachat/deltachat-ios/pull/3178 cmp https://github.com/deltachat/deltachat-android/pull/4506

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Add notifications for missed calls
 * Video call preview now accurately shows what is sent to remote
 * Enhance display of messages with long links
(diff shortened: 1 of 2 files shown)


adbenitez (approved):
makes sense, thanks!

PR #4524, opened 2026-07-07
increase link size to 80

adbenitez:
#skip-changelog


wchen342 (approved):

r10s (approved):

r10s:
successor of #4519

we discussed that we want to change MAX_DISPLAY_LINK_LENGTH from 32 to 80 characters and give that a try

most "nice URLs" and short URLs should not be truncated
for invite links, 80 is even better as the group name is shown shomehow
in general, mastodon, which is not only for "friends & family" also shortens links. telegram does not shorten links, however, allows to display smth else than the link - if scamming is a concern, that is much worse

r10s:
what a speed this evening 🐇

PR #4528, opened 2026-07-08
Refresh menu after conversation states are changed

### LINKED ISSUE #4526: Call icon doesn't occur immediately after adding new contact
Once the three-dot menu is clicked or backpress->open again the call icon will occur correctly.

Reported in https://github.com/deltachat/deltachat-android/issues/4517#issuecomment-4912292744

wchen342:
Fixes #4526

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
@@ -1856,6 +1856,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
       initializeSecurity(isSecureText, isDefaultSms);
       setInputPanelVisibility(false);
       initializeContactRequest();
+      invalidateOptionsMenu();
     } else if ((eventId == DcContext.DC_EVENT_INCOMING_MSG
             || eventId == DcContext.DC_EVENT_MSG_READ)
         && event.getData1Int() == chatId) {


src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1856,6 +1856,7 @@ public void handleEvent(@NonNull DcEvent event) {
       initializeSecurity(isSecureText, isDefaultSms);
       setInputPanelVisibility(false);
       initializeContactRequest();
+      invalidateOptionsMenu();
```
adbenitez, line 1859: I am wondering if it makes sense to move the `eventId == DcContext.DC_EVENT_CONTACTS_CHANGED` at the top to a dedicated branch because contact changes can't never change the status of the input bar, or invalidate menu, because it will be a bit annoying if the menu gets constantly invalidated (and hence closed) every time some contact in a big group changes

I don't remember now if chat-modified is triggered when contact changes name/avatar, probably no so for the title we might still need to watch that event, otherwise it could be removed completely 🤔

wchen342: `DC_EVENT_CONTACTS _CHANGED`  can probably be changed? The reason why the icon doesn't show is because when a new contact is added `canSend` is false initially. I can move the menu refresh to only for `canSend` though.

adbenitez (approved):
it is probably ok to merge this as-is for now, we can try to optimize stuff in another PR in the future

PR #4533, opened 2026-07-13
don't send pasted images as sticker

### LINKED ISSUE #4532: Pasted images are sent as sticker
we sent any image from keyboard as sticker, but images can also be pasted from clipboard, in that case they shouldn't be sent as sticker

adbenitez:
close #4532 

this is also to what Signal is doing to detect stickers: https://github.com/signalapp/Signal-Android/blob/c47a74ba0961903eeaeda6d22347a990056c3e83/app/src/main/java/org/thoughtcrime/securesms/keyboard/KeyboardUtil.kt#L31C1-L34C4

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
@@ -1601,10 +1601,14 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
 
   // media selected by the system keyboard
   @Override
-  public void onMediaSelected(@NonNull Uri uri, String contentType) {
+  public void onMediaSelected(@NonNull Uri uri, String contentType, boolean isSticker) {
     if (isEditing) return;
     if (MediaUtil.isImageType(contentType)) {
-      sendSticker(uri, contentType);
+      if (isSticker) {
+        sendSticker(uri, contentType);
+      } else {
+        setMedia(uri, MediaType.IMAGE);
+      }
     } else if (MediaUtil.isVideoType(contentType)) {
       setMedia(uri, MediaType.VIDEO);
     } else if (MediaUtil.isAudioType(contentType)) {
--- a/src/main/java/org/thoughtcrime/securesms/components/ComposeText.java
+++ b/src/main/java/org/thoughtcrime/securesms/components/ComposeText.java
@@ -4,6 +4,7 @@ import android.content.ClipData;
 import android.content.ClipboardManager;
 import android.content.Context;
 import android.content.res.Configuration;
+import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
 import android.text.Spannable;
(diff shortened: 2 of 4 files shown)


adbenitez:
also while with Gboard and such keyboards pasting images worked, while testing in GrapheneOS with basic open source keyboards, pasting image via keyboard button or via "paste" in context menu (long-press in the input field) was not pasting the image at all, `onCommitContent` is not called at all for that case

I pushed a commit to also fix that problem

src/main/java/org/thoughtcrime/securesms/components/ComposeText.java
```diff
@@ -198,13 +228,20 @@ public boolean onCommitContent(
...
         return true;
       }
 
       return false;
     }
+
+    private boolean isSticker(Uri uri) {
+      String uriString = uri.toString();
+      return uriString.contains("sticker")
+          || uriString.contains("com.touchtype.swiftkey.fileprovider/share_images");
```
wchen342, line 244: Have you checked if the keyboard send you any additional metadata like MIME? I am not sure this is the best approach, but maybe it is good enough for now.

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
```diff
@@ -1601,10 +1601,14 @@ public void refreshStickerPicker() {
...
   @Override
-  public void onMediaSelected(@NonNull Uri uri, String contentType) {
+  public void onMediaSelected(@NonNull Uri uri, String contentType, boolean isSticker) {
     if (isEditing) return;
     if (MediaUtil.isImageType(contentType)) {
-      sendSticker(uri, contentType);
+      if (isSticker) {
+        sendSticker(uri, contentType);
+      } else {
+        setMedia(uri, MediaType.IMAGE);
```
wchen342, line 1610: Do we want to handle GIF differently?

adbenitez: wtym? picking GIF is handled the same way as pasting image, and that is ok, some stickers are actually GIF they will be sent fine as sticker

wchen342: Because there is a `MediaType.GIF` which is separate from `IMAGE`?

src/main/java/org/thoughtcrime/securesms/components/ComposeText.java
adbenitez: didn't check, but it seems there is no standard way, but if Signal is doing the same it is probably good enough

src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
adbenitez: core will convert GIF images to GIF type, anyways GIF type has not much meaning the way we are using it, it is just the same as IMAGE, gif files are just images, the only reason to have a special GIF type is if we would convert gif to mp4 to compress them as Telegram and other apps do, and then you need such special GIF type but at the moment we don't do

wchen342: Hmm, so that type is actually redundant? Well at least it doesn't matter here then.

wchen342 (approved):
Tested and works

PR #4535, opened 2026-07-13
Change missed call notification implementation

### LINKED ISSUE #4520: replace "message" with "reply" in missed call notification
for missed call notification we have a "message" action that just opens the chat, duplicating the action of clicking the notification but with worse UX, opening it in background

it will be more useful to have the same "reply" action we already have in normal message notifications, so you can quickly reply from notification with something like "I am busy" etc

I just realized we also already have "missed call" notifications that come from core as normal messages when the device was offline, and they already have the "reply" as they are normal messages from the point of view of the android client, that would also make then the style more similar for the more special "missed call" notifications we add on the android side after a missed ringing call

### LINKED ISSUE #4534: Missing call notifications overwrite each other for different accounts
As title says, currently two missing call notifications will overwrite one another which is a bug.

wchen342:
- Change style to use default message notification
- Fix an issue with multiple accounts
- Fix an issue where wrong notification is shown going online when there was a missed call during offline

Fixes #4534
Fixes #4520

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
+++ b/src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
@@ -76,10 +76,12 @@ public class NotificationCenter {
   private static class NotifData {
     final Person sender;
     final String text;
+    final boolean isCall;
 
-    NotifData(Person sender, String text) {
+    NotifData(Person sender, String text, boolean isCall) {
       this.sender = sender;
       this.text = text;
+      this.isCall = isCall;
     }
   }
 
@@ -470,7 +472,8 @@ public class NotificationCenter {
                       .setBot(sender.isBot())
                       .setKey(personId)
                       .build(),
-                  text);
+                  text,
+                  dcMsg.getType() == DcMsg.DC_MSG_CALL);
 
           String tickerLine = text;
           if (privacy.isDisplayContact()) {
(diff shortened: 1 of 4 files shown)


adbenitez:
now the notification has 3 buttons: `[reply] [mark as read] [call back]` it feels a bit like too much to me, @r10s what do you think?

what do WhatsApp and telegram display in that case?

wchen342:
Whatsapp doesn't have the `[mark as read]` one. It makes sense though, a call doesn't need to be marked although technically it is a message for us.

src/main/java/org/thoughtcrime/securesms/notifications/NotificationCenter.java
```diff
@@ -716,6 +734,13 @@ private void buildAndShowChatNotification(
         }
       }
 
+      boolean isCall = false;
+      try {
+        isCall = dcContext.getMsg(msgId).getType() == DcMsg.DC_MSG_CALL;
+      } catch (Exception e) {
+        Log.w(TAG, e);
+      }
```
adbenitez, line 749: loading the message from DB again here for every notification sounds overkill, would be better to pass a isCall parameter in the code that is creating the missed call notification

wchen342: Ok I've updated it to be passed in.

adbenitez (approved):
nice! thanks a lot 👍

PR #4538, opened 2026-07-15
Improve image preview

### LINKED ISSUE #4512: Image viewer never enters fullscreen on Android 16
Device: Google Pixel 8a Android: 16 Delta Chat: 2.53.0 (Google Play)

When opening any image, the top toolbar is always visible. Tapping the image, zooming, panning, or rotating the device does not hide the toolbar. Images can never be viewed in true fullscreen mode.
Expected behavior: The toolbar should auto-hide after a tap or while viewing the image, allowing fullscreen viewing.

wchen342:
- The top bar is now hidden on tap
- Top bar is semi transparent so image can extend behind it
- Increase max zoom allowed

Fixes #4512

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/components/MediaView.java
+++ b/src/main/java/org/thoughtcrime/securesms/components/MediaView.java
@@ -86,4 +86,8 @@ public class MediaView extends FrameLayout {
       this.videoView.get().cleanup();
     }
   }
+
+  public void setOnMediaTapListener(@Nullable Runnable listener) {
+    imageView.setOnMediaTapListener(listener);
+  }
 }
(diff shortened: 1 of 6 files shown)


src/main/java/org/thoughtcrime/securesms/components/MediaView.java
```diff
@@ -86,4 +86,8 @@ public void cleanup() {
       this.videoView.get().cleanup();
     }
   }
+
+  public void setOnMediaTapListener(@Nullable Runnable listener) {
+    imageView.setOnMediaTapListener(listener);
+  }
```
adbenitez: this says "on media tap" but seems to be more accurately "onImageTap" because it doesn't apply to videos

adbenitez (approved):
works like a charm, really nice improvement!

Issue #4540, opened 2026-07-16
Core 2.54+ breaking changes


wchen342:
There are a few changes needed to UI:
1. Currently only `listTransportsEx` is actually used and `listTransports` is only used for logging. We need to switch to use `listTransports` and make sure unpublished relays are hidden.
2. The current "Remove or hide" dialog shall be simplified, we no longer need the "Hide" button and keep the "Delete" button, but change the "Delete" button to call `setTransportUnpublished` instead of `remove_transport`.
3. As per https://github.com/deltachat/deltachat-android/pull/4480, we can keep the `is_chatmail` for now so that doesn't need to change right away.

r10s:
desktop counterpart: https://github.com/deltachat/deltachat-desktop/pull/6556

PR #4542, opened 2026-07-20
update string for removing relays

r10s:
targets https://github.com/deltachat/deltachat-android/issues/4540 , the PR can be merged without a new core, however

change for help at https://github.com/deltachat/deltachat-pages/pull/1383

#skip-changelog

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -721,6 +721,8 @@
     <string name="confirm_add_transport">Add this relay?</string>
     <string name="invalid_transport_qr">The scanned QR code does not contain a valid relay.</string>
     <!-- placeholder will be replaced by a relay server name -->
+    <string name="confirm_remove_relay_x">Remove relay \"%1$s\"?</string>
+    <!-- deprecated -->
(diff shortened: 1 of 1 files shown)


wchen342 (approved):

adbenitez (approved):

WofWca (approved):

WofWca:
Note that with this the UIs will not in any way (besides the logs) expose information about this "hidden" thing going on behind the scenes.
Maybe this dialog message would be the right place to provide users with that info. "Deleted relays will linger around for a while before getting finally deleted" or something along the lines (based on my understanding of https://github.com/chatmail/core/pull/8428).

r10s:
see comment at https://github.com/deltachat/deltachat-pages/pull/1383 , primary UI might not be the correct place, many ppl do not have an idea about what that means, might overestimate things etc. esp. as we want to have that automated, we like to have as less as possible here. maybe it is a thing for the upcoming security whitepaper

r10s:
we finally decided to add a little hint at https://github.com/deltachat/deltachat-android/pull/4569

PR #4548, opened 2026-07-20
remove "hide relay"

### LINKED ISSUE #4540: Core 2.54+ breaking changes

adbenitez:
close #4540

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/relay/RelayListActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/relay/RelayListActivity.java
@@ -17,7 +17,7 @@ import androidx.recyclerview.widget.LinearLayoutManager;
 import androidx.recyclerview.widget.RecyclerView;
 import chat.delta.rpc.Rpc;
 import chat.delta.rpc.RpcException;
-import chat.delta.rpc.types.TransportListEntry;
+import chat.delta.rpc.types.EnteredLoginParam;
 import com.b44t.messenger.DcContext;
 import com.b44t.messenger.DcEvent;
 import com.google.zxing.integration.android.IntentIntegrator;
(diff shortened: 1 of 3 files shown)


wchen342 (approved):
`hide_from_contacts`, `hidden_from_contacts`, `confirm_remove_or_hide_transport_x` can be removed?

Otherwise LGTM

src/main/java/org/thoughtcrime/securesms/relay/RelayListActivity.java
```diff
@@ -234,38 +234,27 @@ public boolean onContextItemSelected(@NonNull MenuItem item) {
...
         new AlertDialog.Builder(this)
-            .setTitle(R.string.remove_transport)
-            .setMessage(getString(R.string.confirm_remove_or_hide_transport_x, relay.param.addr))
+            .setMessage(getString(R.string.confirm_remove_relay_x, relay.addr))
             .setPositiveButton(
                 R.string.remove_transport,
                 (d, which) -> {
                   try {
-                    rpc.deleteTransport(accId, relay.param.addr);
+                    rpc.deleteTransport(accId, relay.addr);
```
r10s:
```suggestion
                    rpc.setTransportUnpublished(accId, relay.addr, true);
```

(did not test :)

r10s (changes requested):

r10s:
> hide_from_contacts, hidden_from_contacts, confirm_remove_or_hide_transport_x can be removed?

these strings should be marked `<!-- deprecated -->` as usual, we'll remove them alltogether these days then, when most platforms have updated (reminder: strings are used by iOS, desktop, and maybe even ubuntu touch as well)

r10s (approved):

PR #4549, opened 2026-07-20
remove DcAccounts.isAllChatmail()

adbenitez:
this is tackles the last open point of #4540

#skip-changelog

### DIFF
--- a/src/main/java/com/b44t/messenger/DcAccounts.java
+++ b/src/main/java/com/b44t/messenger/DcAccounts.java
@@ -70,14 +70,4 @@ public class DcAccounts {
   private native long getAccountCPtr(int accountId);
 
   private native long getSelectedAccountCPtr();
-
-  public boolean isAllChatmail() {
-    for (int accountId : getAll()) {
-      DcContext dcContext = getAccount(accountId);
-      if (dcContext.getConfigInt("is_chatmail") == 0) {
-        return false;
-      }
-    }
-    return true;
-  }
 }
--- a/src/main/java/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
+++ b/src/main/java/org/thoughtcrime/securesms/components/reminder/DozeReminder.java
@@ -15,7 +15,6 @@ import androidx.core.content.ContextCompat;
 import com.b44t.messenger.DcContact;
 import com.b44t.messenger.DcContext;
 import com.b44t.messenger.DcMsg;
-import org.thoughtcrime.securesms.ApplicationContext;
 import org.thoughtcrime.securesms.R;
 import org.thoughtcrime.securesms.connect.DcHelper;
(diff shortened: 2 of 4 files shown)


wchen342 (approved):

r10s:
let's not rush with merging this PR - it potentially worsen situation on many phones that do not need background permission or have unreliable background task, which are a lot. 

obvious issue is that more battery drain will happen, ppl get questions that should not be needed, any a permanent notification they have to deal with. but apart from that, also the notification situation may get worse in the long run:

1. forcing background will make issues we have with PUSH less visible - esp. as devs tend to have less problems with that. we were noticing that in the past as well
2. for many, [forcing background is not working permanently](https://dontkillmyapp.com), so it is really needed to have reliable PUSH, which is more complex because of 1.

in general, Delta Chat _should_ work on phones that support PUSH _without_ requiring background [^1]

for removal of chatmail, we have other options as well:

- only ask to run in background on fdroid flavour (probably already the case)
- we want to get rid of "is_chatmail" on the network level - however, we can set a flag when the profile was created manually. we anyway may want to do that because the "Edit Relay" is weird for chatmail - that could be a good enough hint for force background as well. if we do not want a flag, we can ask on manual creation in general

to be clear, i am not against this PR in general, but there is more to consider. if Signal / Whatsapp / Telegram are also asking for "force background" unconditionally even tho PUSH is avaiable, that would be good to know, and would be some plus points for this PR :)

[^1]:  also, in general, it is not unhealthy if the app is terminated from time to time, memory leaks may pile up etc. - not saying that it is the case, just a general, honestly minor comment, that is also not part of above considerations

adbenitez:
to clarify "force background connection" is a different, thing, we are not having this enabled, on the contrary, we are not enabling it anymore automatically with this, so in that regard several of the points you said about "it is good if the app can be killed" actually apply IF we merge this PR, since currently we are enabling that "force background connection" on noticing there is no push notifications available, and maybe is not needed at all and it is good enough with the "do not optimize" option

which is what this PR is doing: we ask for the "disable battery optimization" option that people that had problems with battery drain in the past is what they use, instead of the "force background connection" option

r10s:
after some one-to-one discussions, i am more in favour of this PR, esp. to move forward.

in practise, the question was done often anyways (and always in the past), @adbenitez also got a phone which reliably does not get a PUSH token, so we may get a better grip on that. and, even with PUSH, disabling battery optimisation may help eg. if you are in a slow network and the 15 seconds window (it is shorter than apple's!) may really be too small

r10s (approved):
actually, this PR does not always ask for "disable battery optimisation", but does _not_ ask when a push notification token is received. so it is actually implementing the alternative approach above.

after more discussion, it is even a bit unclear what exactly "isable battery optimisation" is doing on modern androids, and how much it help if it is enabled unconditionally.

so, after another discussion with @adbenitez, let's merge the PR as is. we can always iterate :)

PR #4553, opened 2026-07-21
chore: scripts/upload*.sh: update upload user

j4n:
The renewed delta.chat webserver now has per-service upload users; update script accordingly.

### DIFF
--- a/scripts/upload-release.sh
+++ b/scripts/upload-release.sh
@@ -21,14 +21,8 @@ APK="deltachat-gplay-release-$VERSION.apk"
 ls -l $APK
 read -p "upload this apk and belonging symbols to download.delta.chat/android? ENTER to continue, CTRL-C to abort."
 
-# you need the private SSH key of the jekyll user; you can find it in this file:
-# https://github.com/hpk42/otf-deltachat/blob/master/secrets/delta.chat
-# It is protected with [git-crypt](https://www.agwa.name/projects/git-crypt/) -
-# after installing it, you can decrypt it with `git crypt unlock`.
-# If your key isn't added to the secrets, and you know some of the team in person,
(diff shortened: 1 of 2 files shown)


scripts/upload-release.sh
```diff
@@ -21,22 +21,22 @@ APK="deltachat-gplay-release-$VERSION.apk"
 ls -l $APK
 read -p "upload this apk and belonging symbols to download.delta.chat/android? ENTER to continue, CTRL-C to abort."
 
-# you need the private SSH key of the jekyll user; you can find it in this file:
+# you need the private SSH key of the android user; you can find it in this file:
```
missytake:
```suggestion
# you need the private SSH key of the www-android user; you can find it in this file:
```

scripts/upload-release.sh
```diff
@@ -21,22 +21,22 @@ APK="deltachat-gplay-release-$VERSION.apk"
...
 # https://github.com/hpk42/otf-deltachat/blob/master/secrets/delta.chat
 # It is protected with [git-crypt](https://www.agwa.name/projects/git-crypt/) -
 # after installing it, you can decrypt it with `git crypt unlock`.
 # If your key isn't added to the secrets, and you know some of the team in person,
 # you can ask on irc #deltachat for access.
 # Add the key to your `~/.ssh/config` for the host, or to your ssh-agent, so rsync is able to use it)
```
missytake: Let's just say that you need to give your public SSH key to sysadmin@testrun.org if you want to be able to upload there, and remove all the other comments. If the apk is always uploaded by devs, and not by github actions, let's not store the private key in a secrets management tool at all, and remove the comments here.

missytake (approved):

adbenitez (approved):

PR #4560, opened 2026-07-22
allow to react in channels

adbenitez:
this PR makes it possible for channels subscribers to react to posts ONLY with the default reactions provided by the client

this change requires a core counterpart that actually allows sending the reaction message, latest core 2.56.0 will reject the attempt to send with an error 

- [x] allow subscribers to react to messages
- [x] only allow default reactions in channels (in the future this should be customizable)
- [x] allow to mute/unmute an outgoing channel
- [x] don't notify reactions in outgoing-channels if channel is muted (do not consider them as mentions)
- [x] improve reactions details to be more like WhatsApp's and to support channels reactions
- [x] wait for new core with https://github.com/chatmail/core/pull/8450 merged in
- [x] update changelog

## Screenshots

Channel reactions details:
image

1:1 and groups reactions details:
image


adbenitez:
the android part is ready for review, can't be merged until there is a new core with https://github.com/chatmail/core/pull/8450 merged in

wchen342:
Generally looks good, a minor nit: `getReactionView` can be removed.

Need to wait for core to actually test it end-to-end.

PR #4561, opened 2026-07-23
properly update message views counter for channel owner

### LINKED ISSUE #4546: use new MsgReadCountChanged event
with the core 2.54.0 there is a new MsgReadCountChanged event that can be used to properly refresh post view count for the channel creator's posts

adbenitez:
in the past the message would be updated only with the first read receipt, but further receipts would not update the message counter, admin needed to leave and enter the chat again for it to refresh

close #4546


src/main/java/org/thoughtcrime/securesms/ConversationFragment.java
```diff
@@ -1159,6 +1170,7 @@ public void handleEvent(@NonNull DcEvent event) {
       case DcContext.DC_EVENT_MSG_DELIVERED:
       case DcContext.DC_EVENT_MSG_FAILED:
       case DcContext.DC_EVENT_MSG_READ:
+      case DcContext.DC_EVENT_MSG_READ_COUNT_CHANGED:
         if (event.getData1Int() == chatId) {
           reloadList();
```
adbenitez, line 1175: btw, I think at some point would be nice to improve performance here, it is always just reloading the whole list, when an event is received that correspond to a single message changing, there is no need to reload the whole message list but instead the adapter could be notified with the ID of the specific item that changed instead of calling `notifyDataSetChanged()`

wchen342: Yeah it is a good idea, but probably separately and we need to make sure the reload is not affect by cache in this case.

wchen342 (approved):
LGTM!

PR #4565, opened 2026-07-25
Make call better support multi-profiles

### LINKED ISSUE #4562: Call multi-profile conflict
Currently the call part doesn't handle multiple profiles correctly in multiple places and is causing problems when adding new functionalities. Need some refactors to make it better support the multi-profile nature of the app better.

wchen342:
Closes #4562

#skip-changelog


src/main/java/org/thoughtcrime/securesms/calls/CallCoordinator.java
```diff
@@ -926,25 +928,66 @@ private synchronized void onIncomingCall(
...
+    addCallToTelecom(session, callAttributes);
 
     // Show CallStyle notification
     showIncomingCallNotification(callerName, callerIcon);
 
     startAndBindService();
   }
 
+  private synchronized void handleSecondaryIncomingCall(int accId, int callId, String offerSdp) {
+    if (isConfidentlyIndependent(offerSdp)) {
```
adbenitez: `isConfidentlyIndependent` looks unnecessary to me, the `callId` is always unique per call so if two calls have different ID there is no need to compare the offer SDP 🤔

wchen342: That is for each account right? Are they unique between profiles?

adbenitez: you have the accId + callId pair

adbenitez: or is this about identifying if an incoming call is from ourselves but from another profile in the same app?
maybe some doc comment for the function would help

wchen342: Yeah it's for detecting loopback calls. Even on the same phone if you have two profiles, you will get two sets of different `(accId, callId)` on the two profiles if they call each other.

adbenitez: I think that is really not something we should care to support much, so rather cancel all incoming calls or just keep current behavior and don't cancel any, just ignore them

there are better ways to detect if call is from self: 
* if contact fingerprint matches one of our accounts
* if the sdp offer in the incoming call message is the exact same string as the one in our current outgoing call message

parsing sdp offer sounds dangerous and unnecessary, but as said better just keep it simple for now and don't care if it is from self or not and just ignore the incoming calls

wchen342: It is possible but then it will not be extendable to the case to support the case when e.g. P1 is calling and a genuine new call to P2 comes in. But if we really want a simple solution then we can ignore that use case for now.

wchen342: I removed the sdp detection part. It shall be simpler now.

adbenitez (approved):
lgtm, also tested it and seems to work as expected, thanks!

PR #4573, opened 2026-07-30
Fix videos not being recoded when sending multiple/from file picker

### LINKED ISSUE #3928: videos not always recoded

### LINKED ISSUE #4541: Videos are lost during recoding or compression
- Android version:
- Device:
- Delta Chat version: 2.53.0
- Expected behavior:
When sending a video it should either be compressed and sent or if it's too big it should warn the sender
- Actual behavior:
On the receiver side I very often get 140 bytes videos from android users
Not 100% sure but probably send a large video
- Screenshots:
- Logs:

wchen342:
Fixes #3928
Fixes #4541

There is one minor inconvenience that if there are multiple videos to be encoded, each one will get a dialog. But it is not very common and increase complexity of the patch so I left it as-is.

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/util/SendRelayedMessageUtil.java
+++ b/src/main/java/org/thoughtcrime/securesms/util/SendRelayedMessageUtil.java
@@ -27,6 +27,7 @@ import org.thoughtcrime.securesms.ConversationListRelayingActivity;
 import org.thoughtcrime.securesms.connect.DcHelper;
 import org.thoughtcrime.securesms.mms.PartAuthority;
 import org.thoughtcrime.securesms.providers.PersistentBlobProvider;
+import org.thoughtcrime.securesms.video.recode.VideoRecoder;
 
 public class SendRelayedMessageUtil {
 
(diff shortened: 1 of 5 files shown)


adbenitez:
by "each one will get a dialog" you mean that after one is finished its dialog closes and a new one appears for the next one? that would be a bit hacky but I am ok with it, what I would not like is if ALL dialogs get opened at the same time making the background darker and darker etc

btw this supersedes https://github.com/deltachat/deltachat-android/pull/4430 thanks for taking over that issue

wchen342:
I should be more precise, the dialog I was talking about doesn't normally occur. They are only shown when re-encode failed for some reason, and only then multiple dialogs will be shown if multiple files failed to re-encode. So it is a very rare edge-case and I think we don't need to worry about too much.

The pre-sending confirmation is always only once.

adbenitez (approved):

src/main/java/org/thoughtcrime/securesms/util/SendRelayedMessageUtil.java
```diff
@@ -95,30 +96,26 @@ private static void handleForwarding(Context context, int chatId, int[] forwarde
...
-  public static boolean containsVideoType(Context context, ArrayList<Uri> uris) {
-    for (final Uri uri : uris) {
-      final String mimeType = MediaUtil.getMimeType(context, uri);
-      if (MediaUtil.isVideoType(mimeType)) {
-        return true;
-      }
+  private static void sendMsgRecodingVideo(Context context, int chatId, DcMsg msg) {
+    if (msg.getType() == DcMsg.DC_MSG_VIDEO && !VideoRecoder.prepareVideo(context, chatId, msg)) {
+      return;
     }
```
adbenitez, line 117: mmmh, with this the video is not sent, maybe it makes more sense to just send as file then as in the past? there is not much the user can do about the error other than manually selecting the video via "attach file", so showing an error is not super useful

wchen342: Yeah that is true, I was originally thinking it may hit the server size limit so sending it kind of meaningless, but I think maybe some relays support larger files so it doesn't hurt to try.

wchen342: I made it so it will send the video even the recode failed.

PR #4576, opened 2026-07-31
make big badge counters more human readable and avoid badge overflow

### LINKED ISSUE #4564: badge counter is unbounded
the unread messages badge counter seems to be unbounded and it looks overflown with big numbers:

Image

Image

in apps like Telegram they would switch to `1k`, `2k`, etc. for big numbers so they take less space, that would be nicer but a simpler solution would be to just display `99+` which gives less information but probably good enough, since most of the time one doesn't have so many unread messages and being so precise about the big number also doesn't help much

adbenitez:
if number is bigger than 999 there is no much space left in the badge-counter bubbles, it also doesn't make much sense to be so precise at that point so changing shape of badge counter to accommodate hard to read big numbers is not helpful, instead this PR does similar to Telegram and start counting in k, ex `1k, 2k, 3k, etc.` this gives a bit of more information than just 99+ which would only provide accurate counting on `< 99`

## Chat list
image

## Title bar badge

image

## Account switcher

image

----

close #4564

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/util/Util.java
+++ b/src/main/java/org/thoughtcrime/securesms/util/Util.java
@@ -110,6 +110,13 @@ public class Util {
     }
   }
 
+  public static String humanReadableCount(int count) {
+    if (count >= 1000) {
+      return (count / 1000) + "k";
+    }
+    return String.valueOf(count);
+  }
+
   public static List<Integer> toList(int[] array) {
     ArrayList<Integer> list = new ArrayList<>(array.length);
     for (int i : array) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Fix: Mic will stay active during a call when screen is off on Android 14+
 * Fix: allow to pick image without requiring storage permission
+* Fix: avoid overflow with big numbers in badge counters
 
 ## v2.57.0
 2026-07
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListActivity.java
@@ -421,6 +421,7 @@ public class ConversationListActivity extends PassphraseRequiredActionBarActivit
     if (unreadCount == 0) {
       unreadIndicator.setVisibility(View.GONE);
     } else {
+      String badgeText = Util.humanReadableCount(unreadCount);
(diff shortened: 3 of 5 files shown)


src/main/java/org/thoughtcrime/securesms/util/Util.java
```diff
@@ -110,6 +110,13 @@ public static void redButton(AlertDialog dialog, int whichButton) {
     }
   }
 
+  public static String humanReadableCount(int count) {
+    if (count >= 1000) {
+      return (count / 1000) + "k";
```
wchen342, line 115: In this case won't 1999 show as 1k? That is kind of confusing. Maybe make it show `1.9k` instead?

src/main/java/org/thoughtcrime/securesms/util/Util.java
```diff
@@ -110,6 +110,13 @@ public static void redButton(AlertDialog dialog, int whichButton) {
...
+    }
+    return String.valueOf(count);
```
wchen342, line 117: Also shall we localize this?

src/main/java/org/thoughtcrime/securesms/util/Util.java
adbenitez: `1.9k` is 4 characters just as `1999`, idea is to keep it shorter, also we don't need so much precision when numbers go so high, the approach in this PR is a bit more useful than the `99+` approach but also not intending to be precise, just to give a rough idea, anyways the situation when the number will grow so big is rare, and at that point it doesn't matter much if it was 1k or 1999 unread messages

src/main/java/org/thoughtcrime/securesms/util/Util.java
adbenitez: not sure if it makes sense, we can keep it like that and change it if some of our translators give feedback, or check what Telegram is doing

adbenitez: also because if the localized way of shorten the number is too verbose that would defeat the point

src/main/java/org/thoughtcrime/securesms/util/Util.java
wchen342: Ah ok I see, that nakes sense.

wchen342 (approved):

PR #4579, opened 2026-07-31
allow to pick image even without storage permission

### LINKED ISSUE #4563: Unnecessary storage permission
seems denying storage access does not only make the camera roll unavailable, but blocks picking an image from the gallery at all.

that uses to work as described at https://support.delta.chat/t/unnecessary-storage-permission/5434/3, but maybe it has changed for some android

adbenitez:
the minimum android version we support is android 5 and there already is possible to pick files via the system file picker without storage permission

we ask for storage permission for the camera roll, but if the permission is denied we should still proceed to the picker

in more recent android versions we can also use a new image&video-focused picker it seems, but that can be done at later point, this already solves the problem at hand

close #4563

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Fix: Mic will stay active during a call when screen is off on Android 14+
+* Fix: allow to pick image without requiring storage permission
 
 ## v2.57.0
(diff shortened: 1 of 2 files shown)


wchen342 (approved):
Checked and lgtm!

PR #4581, opened 2026-08-02
i try to resolve audio conflict issue #4578

### LINKED ISSUE #4578: Conflict between audio playback and recording
- Android version: Android 16
- Device: Motorola Edge 70
- Delta Chat version:2.57.0
- Expected behavior: When I press record for a new audio, but I'm listening to another one, that audio I'm listening to has to stop
- Actual behavior:When I press to record a new audio track while I'm listening to another audio, that audio doesn't stop, making itself heard in the audio I'm recording.
- Steps to reproduce the problem:turn on an audio and then start recording

Description of the BUG:when I listen to an audio, I can simultaneously record, causing the audio to overlap

GHManu:
Closes #4578
Hi! I submitted a small fix for issue #4578 created by me. 
The problem is audio overlapping, that is, when I listen to an audio I can also record, but it's not good, when I listen to an audio and press to record, it has to stop the audio I'm listening to, so I tried to look for where the audio was managed (both listening to audio and the start of a recording) and where the startRecording method was called, then I checked that if the PlaybackStatus of PlaybackState is = PLAYING, and before starting startRecording() of AudioRecorder, you have to set the PlaybackStatus = PAUSED, and then I called the pause method to stop the current audio;

I'm very sorry but this is one of my first times trying to modify a large project and I'm having trouble managing the whole environment and i wasn't able to set up the local environment to test it, so I would appreciate it if someone could double-check/test it. In the meantime I'm trying to set up the whole environment. Thanks!

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
@@ -99,6 +99,7 @@ import org.thoughtcrime.securesms.components.InputPanel;
 import org.thoughtcrime.securesms.components.KeyboardAwareLinearLayout.OnKeyboardShownListener;
 import org.thoughtcrime.securesms.components.ScaleStableImageView;
 import org.thoughtcrime.securesms.components.SendButton;
+import org.thoughtcrime.securesms.components.audioplay.AudioPlaybackState;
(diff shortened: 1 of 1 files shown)


adbenitez:
closed in favor of #4585  thanks @GHManu

PR #4582, opened 2026-08-03
Add simple update flow

### LINKED ISSUE #4575: self-update functionality
this issue is about letting ppl update the app in case they are not using gplay, fdroid or other stores.

- create a mockup-function `get_latest_versions()` returning a JSON as follows:   
    ```json
   {
        "deltachat": 
        {
          "gplay": {
                "version_integer": 754,
                "version_string": "2.57.0",
                "release_timestamp": 1785499168,
                "download_url": "https://github.com/deltachat/deltachat-android/releases/download/v2.57.0/deltachat-gplay-release-2.57.0.apk"
            } 
        }
    }
    ```  
    `get_latest_versions()` is to be implemented in a way, that it is cheap to call it. no network or so should be involved.
later on, this will become a core function that probably get the information from the relays. later we also might want to add other flavours and systems, that eg. inform about updates in case auto-update for some store is disabled.
`release_timestamp` is seconds since 00:00 hours, Jan 1, 1970 UTC.

- check `get_latest_versions()` regularly, eg. where other device messages are scheduled already

- if `version_integer` is larger than current version, and release_timestamp is one week in the past, add a device message:  
    > 🎉 New version available. Tap this this message to update and enjoy the new features
    
    the first words appear in chat summary, so should already let ppl get a quick idea (we were thinking about a banner as well, but that might be clicked away even more eagerly, os let's stay with a device message for the first implementation, we can always iterate :)

- if message is tapped, show a download progress dialog as follows:
    ```
    Downloading 2.57.0 ...
    [std progress bar of android]
                         [cancel]
    ```

- once download finishes, make sure, that the app is really from us, so that a relay cannot provide a malfeasant apk. eg. check that downloaded-app-id matches self-app-ip - for same app-id, signatures do not match, later install will fail

- then, the app update shall start. this is probably done by calling `DcHelper.openForViewOrShare()` on the downloaded file (similar to when an apk is tapped in a regular file-message with an apk)

gist of the UI actions is to provide as few options and taps as possible and to guide the user. this may or may not require additional hints wrt "allow installation from unknown sources" or others. the fact that an APK is installed via Firefox does not mean that delta is allowed to install APK without changing that option

ftr, @adbenitez and me crafted this issue together

wchen342:
A very simple self-update flow, starting from device message:

Screenshot_20260803_200452

Then a simple progess dialog:

Screenshot_20260803_200517

And going into install dialog:

Screenshot_20260803_200644

Permission check happens before downloading to avoid wasting a download, other problems like app id mismatch are shown as toasts.

resolve #4575

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -74,6 +74,7 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_PHONE_CALL" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_LOCATION" />
     <uses-permission android:name="android.permission.MANAGE_OWN_CALLS" />
+    <uses-permission android:name="android.permission.DOWNLOAD_WITHOUT_NOTIFICATION" />
 
     <!-- force compiling libs on older sdk than supported; runtime checks are required -->
     <uses-sdk tools:overrideLibrary="androidx.core.telecom,im.conversations.webrtc" />
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -74,7 +74,6 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_PHONE_CALL" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_LOCATION" />
     <uses-permission android:name="android.permission.MANAGE_OWN_CALLS" />
-    <uses-permission android:name="android.permission.DOWNLOAD_WITHOUT_NOTIFICATION" />
 
     <!-- force compiling libs on older sdk than supported; runtime checks are required -->
     <uses-sdk tools:overrideLibrary="androidx.core.telecom,im.conversations.webrtc" />
--- a/src/main/java/org/thoughtcrime/securesms/updater/AppUpdate.java
+++ b/src/main/java/org/thoughtcrime/securesms/updater/AppUpdate.java
@@ -1,0 +1,117 @@
+package org.thoughtcrime.securesms.updater;
+
+import android.content.Context;
+import androidx.fragment.app.FragmentActivity;
+import com.b44t.messenger.DcContact;
+import com.b44t.messenger.DcContext;
+import com.b44t.messenger.DcMsg;
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.thoughtcrime.securesms.BuildConfig;
+import org.thoughtcrime.securesms.R;
+import org.thoughtcrime.securesms.connect.DcHelper;
+
+public class AppUpdate {
+  private static final String TAG = "AppUpdate";
+
+  private static final String APP_KEY = "deltachat";
+
+  private static final long ONE_WEEK_SECONDS = 7L * 24 * 60 * 60;
+
+  private static final String MOCK_LATEST_VERSIONS_JSON =
+      "{"
+          + "  \"deltachat\": {"
+          + "    \"gplay\": {"
+          + "      \"version_integer\": 755,"
+          + "      \"version_string\": \"2.57.0\","
+          + "      \"release_timestamp\": 1782499168,"
+          + "      \"download_url\": \"https://github.com/deltachat/deltachat-android/releases/download/v2.57.0/deltachat-gplay-release-2.57.0.apk\""
+          + "    }"
+          + "  }"
+          + "}";
+
+  public static class LatestVersion {
+    public final int versionInteger;
+    public final String versionString;
+    public final long releaseTimestamp;
+    public final String downloadUrl;
+
+    LatestVersion(
+        int versionInteger, String versionString, long releaseTimestamp, String downloadUrl) {
+      this.versionInteger = versionInteger;
+      this.versionString = versionString;
+      this.releaseTimestamp = releaseTimestamp;
+      this.downloadUrl = downloadUrl;
+    }
+  }
+
+  /** Mock function. */
+  public static JSONObject getLatestVersions() {
+    try {
+      return new JSONObject(MOCK_LATEST_VERSIONS_JSON);
+    } catch (JSONException e) {
+      return null;
+    }
+  }
+
+  public static LatestVersion getLatestForThisBuild() {
+    JSONObject all = getLatestVersions();
+    if (all == null) return null;
+    JSONObject app = all.optJSONObject(APP_KEY);
+    if (app == null) return null;
+    JSONObject flavor = app.optJSONObject(BuildConfig.FLAVOR);
+    if (flavor == null) return null;
+    try {
+      return new LatestVersion(
+          flavor.getInt("version_integer"),
+          flavor.getString("version_string"),
+          flavor.getLong("release_timestamp"),
+          flavor.getString("download_url"));
+    } catch (JSONException e) {
+      return null;
+    }
+  }
+
+  public static boolean isUpdateAvailable() {
+    LatestVersion latest = getLatestForThisBuild();
+    return latest != null && latest.versionInteger > BuildConfig.VERSION_CODE;
+  }
+
+  public static String buildUpdateText(Context context) {
+    return context.getString(R.string.update_available_msg);
+  }
+
+  public static boolean isEligible(Context context) {
+    LatestVersion latest = getLatestForThisBuild();
+    if (latest == null) return false;
+    if (latest.versionInteger <= BuildConfig.VERSION_CODE) return false;
+    long nowSeconds = System.currentTimeMillis() / 1000L;
+    return latest.releaseTimestamp <= nowSeconds - ONE_WEEK_SECONDS;
+  }
+
+  public static void addUpdateDeviceMsg(Context context) {
+    LatestVersion latest = getLatestForThisBuild();
+    if (latest == null) return;
+    DcContext dcContext = DcHelper.getContext(context);
+    DcMsg msg = new DcMsg(dcContext, DcMsg.DC_MSG_TEXT);
+    msg.setText(buildUpdateText(context));
+    dcContext.addDeviceMsg("update_" + latest.versionInteger, msg);
+  }
+
+  public static boolean isUpdateDeviceMsg(Context context, DcMsg msg) {
+    return msg != null
+        && msg.getFromId() == DcContact.DC_CONTACT_ID_DEVICE
+        && isUpdateAvailable()
+        && buildUpdateText(context).equals(msg.getText());
+  }
+
+  public static void updateDeviceMsgTapped(FragmentActivity activity) {
+    LatestVersion latest = getLatestForThisBuild();
+    if (latest == null || activity == null) return;
+    if (!DcHelper.canInstallApks(activity)) {
+      DcHelper.requestInstallApksPermission(activity);
+      return;
+    }
+    AppUpdateDialogFragment.show(activity, latest);
+  }
+}
--- a/src/main/java/org/thoughtcrime/securesms/updater/AppUpdate.java
+++ b/src/main/java/org/thoughtcrime/securesms/updater/AppUpdate.java
@@ -26,7 +26,7 @@ public class AppUpdate {
   private static final String SOURCE_ID = "gplay";
   private static final String PLAY_STORE_PACKAGE = "com.android.vending";
 
-  private static final long CHECK_INTERVAL_MS = 24 * 60 * 60 * 1000L;
+  private static final long CHECK_INTERVAL_MS = 12 * 60 * 60 * 1000L;
 
   private static final String SEPARATOR = "\n";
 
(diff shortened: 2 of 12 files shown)


r10s:
some quick feedback:

- we want to have a button in the update message bubble, similar to webxdc or full-message (we can reuse that probably). button should read "Update Now"

- before we open the system settings about "install apk from unknown sources", we want to have a dialog, explaining what is going on: "To install the update, please first allow Delta Chat to install apps in the system settings." buttons: "Open Settings" (string exist as `open_settings`) and "Cancel".

- if possible: when "Open Settings" is hit, set a bool to remember that state and open system settings. when the activity resumes, the bool is set, and the user granted the permission, continue update

- if smth goes wrong, better show the error in a dialog. these toast are not good for that, no one can read that fast :)

wchen342:
I have added the button on message:

Screenshot_20260806_184441

And the dialog before opening the settings:

Screenshot_20260806_184514

The installation will resume when the user comes back.

The errors are shown in the same dialog now.

adbenitez:
after talking with @r10s wrt the core side of this etc. we also came over some change of requirements:

* do not wait for update to be older than 1 week to add the device message, we can better control this server-side
* after update is finished, remove the device message, otherwise it will stay "update available" even after user updated, which is confusing. Since the app is restarted after upgrade, I guess one idea to know if it was upgraded is to save the `versionInteger` + device-msg ID in some preferences right before giving the apk to the system installer, then on app start check that preference and if set, and current app's version number `>=` the set version number then remove the corresponding device message using the existing APIs to remove messages given the msg ID and clear the stored preferences

about that last point: I also realized that we are adding that device message for every selected profile, better do it only in one profile, and in the rest set it to null as we do already in `ConversationListActivity.addDeviceMessages()`, maybe move `AppUpdate.addUpdateDeviceMsg(context);`  there instead of doing it in `ConversationListFragment`

r10s:
then we probably can also drop `release_timestamp`? that would make providing/checking the json by hand much easier, and one think less to care/think about :)

still, it makes sense to do the update check only once per day or every some hours, at a somehow random time

wchen342:
> then we probably can also drop release_timestamp? that would make providing/checking the json by hand much easier, and one think less to care/think about :)

It may be useful later so I say we just keep it, it basically costs nothing.

src/main/AndroidManifest.xml
```diff
@@ -74,6 +74,7 @@
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_PHONE_CALL" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_LOCATION" />
     <uses-permission android:name="android.permission.MANAGE_OWN_CALLS" />
+    <uses-permission android:name="android.permission.DOWNLOAD_WITHOUT_NOTIFICATION" />
```
r10s: why is the permission needed? if we do not have that, a notification is visible, which seems fine. https://developer.android.com/reference/kotlin/android/app/DownloadManager.Request

wchen342: It is to avoid confusion and edge cases, because user will be able to, for example, stop the download through download manager, etc. which makes things more complicated to handle. Now it is just handled through the dialog completely.

adbenitez: I think it makes sense to show the notification tho, having an extra permission is something we should really avoid unless strictly necessary, and I guess if user cancel via notification we also get something via the call we are using already, which would be the same for network errors etc :thinking:

adbenitez:
> > we are adding that device message for every selected profile, better do it only in one profile, and in the rest set it to null as we do already in `ConversationListActivity.addDeviceMessages()`
> 
> It may be useful later so I say we just keep it, it basically costs nothing.

it is not about cost, it is about annoyance, if you have, for example, 5 profiles, you will have to deal 5 times with a device message chat with the same unread message, also for removing it later is much easier because you have to only remove it in 1 place

wchen342:
Strange, github quoted the wrong message. I was replying to the timestamp thing @r10s mentioned.

adbenitez:
> Strange, github quoted the wrong message. I was replying to the timestamp thing r10s mentioned.

ahh! wrt that the thing is that core will just not provide the field

the whole format will change so better wait in that regard to just use the core api, not much to do here than removing the whole mock data with timestamp and stuff

wchen342:
I updated the update flow to show only in the active profile. Also the message shall remove itself after a new version is installed now. The one-week check is also removed.

src/main/AndroidManifest.xml
wchen342: Yeah, actually I fixed some bugs with the download progress bar and now the cancel is handled already, so I reverted this part so there is no additional permissions needed now.

r10s:
i did the core part at https://github.com/chatmail/core/pull/8557 - but please wait for replacing MOCKUP by real calls until things are settled there; there are some open ends to be discussed

src/main/java/org/thoughtcrime/securesms/updater/AppUpdateDialogFragment.java
```diff
@@ -0,0 +1,369 @@
...
+
+  private void onDownloadSucceeded(String localPath) {
+    stopPolling();
+
+    Context context = getContext();
+    Activity activity = getActivity();
+    if (context == null || activity == null) return;
+
+    PackageInfo info = context.getPackageManager().getPackageArchiveInfo(localPath, 0);
+    if (info == null || !context.getPackageName().equals(info.packageName)) {
```
link2xt, line 337: For other reviewers: this is the place where we check that downloaded APK is an update and not something entirely unrelated.

r10s:
> Also the message shall remove itself after a new version is installed now. The one-week check is also removed.

as we have the hook for that now, let's also add a message "Successfully updated to v1.2.3" - the update needs to be triggered manually by the user, not giving feedback or only implicit feedback (app restarted, message deleted) is not great.

this is also what eg. systems are doing on manual updates, also other apps, eg. here is firefox:

Screenshot 2026-08-13 at 11 56 16

the link to the changelog is not needed tho, if we have smth to say, we anyway add a device message.

to be clear: that message shall _only_ be added if the user started the update with the new manual flow. for completely automated updates from stores, it is not needed - and confusing, as the user usually does not even know

adbenitez:
@r10s this is really not needed, the system already gives feedback that the app got updated after the update finishes, see attached screenshots of an apk update process, showing a message afterwards in-app only makes sense if you are going to announce what is new / changelog, otherwise you are just stating the same again:


image
image

wchen342:
@r10s @adbenitez I kind of agree with adb here. Although I think we can re-purpose the device message we have now to show new changes. Currently that message is static and doesn't do much after a user have installed Delta.

r10s:
k, if that is the normal flow, i am fine for now with not saying smth in device chat after installation (only removing existing "New version available" message)

r10s:
ftr, the [core PR](https://github.com/chatmail/core/pull/8557) and [relay PR](https://github.com/chatmail/relay/pull/1038) are merged, and this PR should use the new API get_app_version("deltachat", "gplay"), returning an json as 
```
{
    "versionInteger": 754,
    "versionString": "2.57.0",
    "downloadUrl": "https://github.com/deltachat/deltachat-android/releases/download/v2.57.0/deltachat-gplay-release-2.57.0.apk"
}
```

for testing: this is already deployed on nine.testrun.org, so faking a lower versionInteger should start the update flow (that should not be part of the PR, just mentioning that so one can test :)

adbenitez:
this branch can update to main that contains the new core with the new API needed

wchen342:
I have tested the flow right before the actual installation, because the signature won't match. I am not sure if there is a good way to test with actual signed apk other than build locally with the signing key.

src/main/java/org/thoughtcrime/securesms/updater/AppUpdate.java
```diff
@@ -0,0 +1,208 @@
...
+      Log.w(TAG, "could not delete " + apk);
+    }
+  }
+
+  public static boolean isUpdateDeviceMsg(Context context, DcChat dcChat, DcMsg msg) {
+    if (msg == null) return false;
+    StoredUpdate stored = StoredUpdate.load(context);
+    return stored != null
+        && stored.msgId != 0
+        && stored.version > BuildConfig.VERSION_CODE
```
adbenitez: this line is not necessary, an "app update" message is so even if the version code is smaller than current version, which in practice should never happen because such message should be deleted on app start after an update

src/main/res/values/strings.xml
```diff
@@ -1271,4 +1273,9 @@
 
     <!-- device messages for updates -->
     <string name="update_2_0">What\'s new?\n\n💯 End-to-end encryption is reliable and forever now. Padlocks 🔒 are gone!\n\n✉️ Classic email without end-to-end encryption is marked with a letter symbol\n\n😻 New enhanced profile screen for all your contacts\n\n🔲 New button for quick access to apps used in a chat\n\n❤️ Please donate to help us remain independent and continue to bring improvements: %1$s</string>
+    <string name="update_available_msg">🎉 New version available. Tap this message to update and enjoy the new features</string>
```
adbenitez: I think it would be useful if the version number is here so people can already recognize what they are about to download, otherwise the message will look exactly the same every time and users might even think it is that we are poking them / re-notifying the same app update over and over without realizing it is actually yet another new update, by showing the number here it also reflects more the urgency to update if the number is way bigger than the version they are currently using

they will see the number upon clicking "update now" but at that point the apk is already start to download

I suggest using:
```
🎉 New version %1$s available. Tap this message to update and enjoy the new features
```

cc @r10s

r10s: makes sense!

src/main/java/org/thoughtcrime/securesms/updater/AppUpdate.java
```diff
@@ -0,0 +1,208 @@
...
+import org.thoughtcrime.securesms.util.Prefs;
+
+public class AppUpdate {
+  private static final String TAG = "AppUpdate";
+
+  private static final String CLIENT_ID = "deltachat";
+  private static final String SOURCE_ID = "gplay";
+  private static final String PLAY_STORE_PACKAGE = "com.android.vending";
+
+  private static final long CHECK_INTERVAL_MS = 24 * 60 * 60 * 1000L;
```
adbenitez: this could reduced to 12 hours instead of 24, the call to the core API is really cheap, it just returns a cached value, doesn't really do any networking etc.

```suggestion
  private static final long CHECK_INTERVAL_MS = 12 * 60 * 60 * 1000L;
```

r10s: ftr, 12 hours is also the (current) time where core aims to update the versions, so that matches as well. so, when there are no errors, a new update is announced within a day, often faster, that seems fine

src/main/java/org/thoughtcrime/securesms/updater/AppUpdate.java
```diff
@@ -0,0 +1,208 @@
...
+
+      if (TextUtils.isEmpty(source.versionString) || TextUtils.isEmpty(source.downloadUrl)) {
+        Log.w(TAG, "incomplete version info");
+        return;
+      }
+
+      DcContext dcContext = DcHelper.getContext(context);
+      DcMsg msg = new DcMsg(dcContext, DcMsg.DC_MSG_TEXT);
+      msg.setText(context.getString(R.string.update_available_msg));
+      int msgId = dcContext.addDeviceMsg(null, msg);
```
adbenitez: this looks rather dangerous, we don't do this anywhere else in the app, we always use a label, in the past this PR was doing:

```
dcContext.addDeviceMsg("update_" + latest.versionInteger, msg);
```

this would avoid adding the device message multiple times in some bug/rare scenario, and would make it possible to do better cleanup in the future if core exposes some api to get device messages' label

wchen342: This is to avoid cases when the message is deleted for some reason but then we want to show the update from the same version again, but with label we need a new label every time because you cannot use the same label twice.

I'll change it to use a suffix instead.

src/main/java/org/thoughtcrime/securesms/updater/AppUpdate.java
wchen342: Yeah this makes sense.

src/main/java/org/thoughtcrime/securesms/updater/AppUpdate.java
wchen342: The duplicate shouldn't be possible because it is also tracked by Prefs (that is why the label can be changed to null). You can argue that you trust the core's tracking more, but that is not a very strong argument IMO.

src/main/res/values/strings.xml
wchen342: Added.

src/main/java/org/thoughtcrime/securesms/updater/AppUpdate.java
adbenitez: in which situation you would delete the message and want the same update message to appear again? :thinking:

adbenitez (approved):
code lgtm, also tested the whole update process (needed to modify/hack around to force the update of a local debug build from custom mocked URL, etc) we will be able to test it further after merging

thanks a lot for pushing this forward!

src/main/java/org/thoughtcrime/securesms/updater/AppUpdate.java
wchen342: For example the server posted for example 760, then 761, then somehow pulled that and posted 758.
Another example is when a backup is restored with a label already used.

Generally very rare, but not impossible I think.

PR #4584, opened 2026-08-05
properly restore search mode

### LINKED ISSUE #4580: receiving message while in chat search mode cancels search
while in search mode inside a chat, if you receive a message or the chat status is refreshed somehow by some incoming update/message, the search is canceled, but what is worse: the input bar stays hidden since it is hidden in search mode since https://github.com/deltachat/deltachat-android/pull/4493

adbenitez:
close #4580


wchen342 (approved):
Works for me!

PR #4585, opened 2026-08-05
Add audio focus handling to recording

### LINKED ISSUE #4578: Conflict between audio playback and recording

wchen342:
Fixes #4578

This changes a few things:
1. When recording starts, it will stop audio playback (requests audio focus)
2. When recording stops, playback will resume, as long as the playing app respects audio focus
3. During recording, play/pause/seek on audios are disabled in a chat
4. If some other app requests audio focus during recording, the recording will stop and save to draft

Supersede #4581

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Fix: avoid overflow with big numbers in badge counters
 * Fix: make video players truly fullscreen
 * Fix: avoid ending search mode when a new message arrive to the chat
+* Fix: Recording now request and respect audio focus
 
 ## v2.57.0
 2026-07
--- a/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationActivity.java
@@ -388,7 +388,7 @@ public class ConversationActivity extends PassphraseRequiredActionBarActivity
   protected void onPause() {
     super.onPause();
 
-    if (inputPanel.isRecording() && inputPanel.getRecordingDuration() > 1000) {
+    if (shouldSaveRecording()) {
       saveRecording();
     } else {
       processComposeControls(ACTION_SAVE_DRAFT);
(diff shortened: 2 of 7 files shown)


adbenitez:
> If some other app requests audio focus during recording, the recording will stop and save to draft

this parts sounds dangerous/unwanted, does this mean that if user is recording a voice message and a notification sound plays the recording will stop?? that sound inconvenient, in that case I would just keep recording as usual until the user manually stops recording, because I fail to see a situation where this is what the user want

ex. if we allow to keep recording then the user could resume some music etc. if they really want the sound to keep going while recording

wchen342:
No it doesn't stop on notification sound, that is definitely not the case. The recording only stops when some apps request `AUDIOFOCUS_GAIN` or `AUDIOFOCUS_GAIN_TRANSIENT`. Some cases probably include, for example, user opens a YouTube player playing video or uses some other app to call, or opens some other app for voice recognition.

I am not sure it is a good idea to keep recording in those cases, because if some app grabs audio focus we are supposed to stop, and it doesn't really make sense to continue in most cases.

adbenitez (approved):
code lgtm, tested it and it works nicely! thanks

btw, would be nice if we also do (in another PR ofc) something similar when playing voice messages and return the focus to the music player etc. after voice messages finish playing, right now they remain paused and user have to manually play again

wchen342:
Yeah I will check that in a separate issue.

PR #4591, opened 2026-08-08
Improve screen reader accessibility of chat list items (fixes #4590)

### LINKED ISSUE #4590: A11Y: When navigating in the list of chats the fact conversation is pinned, muted or features location streaming is not reported to screen reader users
- Android version: tested with android 17
- Device: Pixel 9A
- Delta Chat version: 2.57
- Expected behavior: It should be possible to find out if conversation is pinned, muted or has location streaming enabled when browsing the list of chats with a screen reader.
- Actual behavior: These details are not yet available in the list of chats
    * Open deltachat with talkback running
    * Tap and hold some chats, then mute or pin them from the navigation menu
    * Now touch the same chats and pay attention to the fact pinned, muted and location streaming status is not reported.

pvagner:
ConversationListItem: Improve content descriptions so muted, pinned and location streaming is reported for screen reader users (Fixes #4590)


adbenitez:
closed in favor of #4596 
thanks @pvagner

PR #4596, opened 2026-08-11
Add ConversationListItem state announcement to Talkback

### LINKED ISSUE #4590: A11Y: When navigating in the list of chats the fact conversation is pinned, muted or features location streaming is not reported to screen reader users

wchen342:
Fixes #4590

Supersede #4591

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Add speed button for voice message playback
+* Improve a11y on Conversation List
 * Fix: support scanning a channel-invite while creating a new profile
 * Fix: random long delay when reopening a WebXDC app
 
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
@@ -157,7 +157,7 @@ public class ConversationListItem extends RelativeLayout
     DcContact contact = recipient.getDcContact();
     avatar.setSeenRecently(contact != null && contact.wasSeenRecently());
 
-    setFromViewState(isSelected(), thread.isMuted());
+    setMutedState(thread.isMuted());
   }
 
   public void bind(
@@ -168,7 +168,7 @@ public class ConversationListItem extends RelativeLayout
     Recipient recipient = new Recipient(getContext(), contact);
 
     fromView.setText(getHighlightedSpan(contact.getDisplayName(), highlightSubstring));
-    setFromViewState(false, false);
+    setMutedState(false);
     subjectView.setVisibility(GONE);
     dateView.setText("");
     setContentState(false, false);
@@ -192,7 +192,7 @@ public class ConversationListItem extends RelativeLayout
     Recipient recipient = new Recipient(getContext(), sender);
 
     fromView.setText(recipient, true);
-    setFromViewState(false, false);
+    setMutedState(false);
     subjectView.setVisibility(VISIBLE);
     subjectView.setText(getHighlightedSpan(messageResult.getSummarytext(512), highlightSubstring));
 
@@ -218,7 +218,7 @@ public class ConversationListItem extends RelativeLayout
     this.selectedThreads = Collections.emptySet();
 
     fromView.setText(inviteData.getDisplayTitle());
-    setFromViewState(false, false);
+    setMutedState(false);
     subjectView.setVisibility(VISIBLE);
     subjectView.setText(inviteData.getDisplaySubtitle());
     subjectView.setTypeface(LIGHT_TYPEFACE);
@@ -300,6 +300,7 @@ public class ConversationListItem extends RelativeLayout
 
     if (!showUnreadBadge) {
       unreadIndicator.setVisibility(View.GONE);
+      unreadIndicator.setContentDescription(null);
     } else {
       boolean isMuted = thread.isMuted() || chatId == DcChat.DC_CHAT_ID_ARCHIVED_LINK;
       final int color =
@@ -321,6 +322,9 @@ public class ConversationListItem extends RelativeLayout
               .endConfig()
               .buildRound(badgeText, color));
       unreadIndicator.setVisibility(View.VISIBLE);
+      unreadIndicator.setContentDescription(
+          getResources()
+              .getQuantityString(R.plurals.chat_n_unread_messages, unreadCount, unreadCount));
     }
   }
 
@@ -334,19 +338,10 @@ public class ConversationListItem extends RelativeLayout
     }
   }
 
-  private void setFromViewState(boolean selected, boolean muted) {
+  private void setMutedState(boolean muted) {
     fromView.setCompoundDrawablesWithIntrinsicBounds(
         muted ? R.drawable.ic_volume_off_grey600_18dp : 0, 0, 0, 0);
-
-    if (!selected && !muted) {
-      fromView.setContentDescription(null);
-      return;
-    }
-    CharSequence description = fromView.getText();
-    if (muted) {
-      description = withStatus(description, R.string.muted);
-    }
-    fromView.setContentDescription(description);
+    fromView.setContentDescription(muted ? withStatus(fromView.getText(), R.string.muted) : null);
   }
 
   private void setContentState(boolean pinned, boolean sendingLocations) {
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
@@ -32,6 +32,7 @@ import android.widget.RelativeLayout;
 import android.widget.TextView;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.StringRes;
 import com.amulyakhare.textdrawable.TextDrawable;
 import com.b44t.messenger.DcChat;
 import com.b44t.messenger.DcContact;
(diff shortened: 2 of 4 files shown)


src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
```diff
@@ -334,6 +345,32 @@ private void setBgColor(ThreadRecord thread) {
...
+        states.add(getContext().getString(R.string.muted));
+      }
+      if (thread.isSendingLocations()) {
+        states.add(getContext().getString(R.string.pref_on_demand_location_streaming));
+      }
+    }
+
+    if (isSelected()) {
+      states.add(getContext().getString(R.string.selected));
+    }
```
adbenitez: not sure this is needed, otherwise @pvagner would have added it 🤔  but I suspect this should be already handled by the system/list and its state "selected" or? sounds odd to me that such common state of lists would be up to the app to report when list elements are selected, but I don't know for sure, maybe @pvagner can confirm if this is really necessary

pvagner: Yes, selecting by long tapping on a list item has already been handled fine up to this point. It sets is_selected property on the corresponding AccessibilityNodeInfo object.

setStateDescription is not yet supported by all 3rd party screen readers on android so being somewhat selfish I went by tweaking contentDescription in my original PR this one supersedes.

Still I understand this is the correct thing to do so I'll just test this PR with talkback and I'll report my findings soon.

Huge thanks for working on this.

wchen342: @adbenitez Yeah makes sense. I added it just in case but seems unnecessary.

@pvagner  I can make a different implementation using `contentDescription` which is a bit more complicated. I didn't know `setStateDescription` is not supported by third party screen reader. I'll change the implementation so it can work with `contentDescription`.

pvagner: I have already contacted two screen reader developers, so [Corvus](https://corvuskit.com/en) and [Jieshuo](https://github.com/nirenr/jieshuo) will get it supported soonish.

I have seen you have used state description for the call screen as well.


The best thing we can do is to implement it with content description for now and change it to state description in a few months when those screen readers are ready.

However if you find it too complicated or consider it a lot of work let's go with state description and we'll wait for the screen readers to catch up.

pvagner:
@wchen342 Please, Is this handling items inside both lists, items in the list of conversations on the main screen as well as list of messages inside a chat?
I am not sure I am miss reading but so far I can only see this PR handling list of conversations.

wchen342:
@pvagner This is only for list of chats. I will work on a different PR for messages later, since that has some different problems and need to be fixed separately.

pvagner:
@wchen342 Oh thank you very much.

Please also pay attention to the importantForAccessibility by explicitly hiding elements that are presented as a part of content description and / or state description. It has two purposes.

The first one is to avoid presenting it twice, the second one is optimizing the list navigation in a way so it takes as few gestures to navigate from the selected item to the next / previous one as possible.

Also I am not asking you to do it right now but let's consider an alternative way of activating the reply action for each message by adding it as an accessibility action similar how @adbenitez added actions that enable activating links for the future. I know it's inside the three dots menu when the item is selected and when swiping to the side already but reply is something that should really be easy to discover and quick to activate.

pvagner:
Thanks @wchen342 for the updated PR.

I have tested both the implementations with multiple screen readers:

* Talkback as shipped by google
* Talkback foss as shipped by Graphene OS
* Corvus (used extensivelly in central europe)
* Jieshuo
* Talk Forward (which has started as a talkback fork)

This updated contentDescription based implementation is working great with all the screen readers I have tested.

The previous implementation with state description is working with all the screen readers I have tested except of Corvus.

As Corvus developers have already expressed their desire to get it fixed on their end feel free to decide what's the best for the deltachat project. They haven't promised an ETA but they are updating their app approx three times a year.

Thanks and greetings

wchen342:
@pvagner Thank you for the testing! I think I will keep this implementation since it works fine. I will take a look at the navigation and check if Talkback includes too many things.

wchen342:
I also added announcement for unread counts.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Delta Chat Android Changelog
 
+## Unreleased
+
+* Improve a11y on Conversation List
+
```
adbenitez, line 9: this needs to be rebased on main

src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
```diff
@@ -334,6 +334,47 @@ private void setBgColor(ThreadRecord thread) {
     }
   }
 
+  private void setFromViewState(boolean selected, boolean muted) {
+    fromView.setCompoundDrawablesWithIntrinsicBounds(
+        muted ? R.drawable.ic_volume_off_grey600_18dp : 0, 0, 0, 0);
+
+    if (!selected && !muted) {
```
adbenitez: I am struggling a bit to understand why selected plays a role here, it is only used in this condition, could you explain a bit?

wchen342: It was a residual from the previous version, I got a bit confused there. Fixed now.

CHANGELOG.md
wchen342: Rebased.

pvagner:
@wchen342 Oh excuse me please there is something I have missed. Now the word read is always included in the content description even when there are unread messages. So it says something like"read, 1 unread message". My guess is that there is "read" contentDescription duplicated within the layout somewhere.
I've just tested todays apk explicitly marking conversation as unread.

Could you please try to investigate?

pvagner:
@wchen342 
Okay, I think I have partially tracked it down, now we just need to come up with a solution improving it:

* conversation_list_item_view layout has two image views: *delivery_indicator* and *unread_indicator*.
* unread_indicator shows how many messages are unread.
* One of the states for the delivery_indicator is seen that might be translated as read for some languages.


My thinking is that if there is something to report for the unread_indicator, and the state of delivery_indicator is seen we may silence delivery_indicator. However I am not sure I am interpreting it correctly.

Thanks and greetings

wchen342:
@pvagner You are right. I made it announce "Delivery status: Read" instead of just "read" in the list view. Can you test if that sounds better now?

pvagner:
@wchen342 Most likelly My explanation is not clear enough.

Try to follow my steps like this please:

* Imagine conversation between chat partners called *Phone* and *Laptop*. Obviously *Phone* is running this test build.
* *Phone* sends a message with text `Hello`. The content description says: `Laptop , Me: Hello , 3 min , Delivery status: Delivered ` - It's all expected so far.
* The *Laptop* reads the message and the content description on the *Phone* changes to: `Laptop , Me: Hello , 4 min , Delivery status: Read ` - Still all expected.
* Then *Laptop* replies with word `Hi` and the content description on the *Phone* changes to: `Laptop , Hi , Now , 1 unread message` - I think you don't have to introduce a new translation string here, there already is a string saying "%d new message".
* Now *Phone* opens the conversation and presses the back button to display the list of conversations again. The content description updates to: `Laptop , Hi , 4 min ` - Still expected.
* Now *Phone* long presses this conversation, then presses *Mark as unread* in the three dots menu. The content description updates to: `Laptop , Hi , 10 min , 1 unread message` - kind of expected plus the translation string warning above.
* Now multiple things happen one after the other:
    - *Phone* replies with a message `How are you`
    - *Laptop* reads the message
    - *Phone* long presses the conversation, then presses *Mark as unread* in the three dots menu
* After all of this, the content description on the *Phone* changes to: `Laptop , Me , How are you , Delivery status: Read , 1 unread message ` - This is the edge case where both *Delivery status: Read* and *1 unread message* are included. My question is if this is expected and if tthe delivery indicator should remain visible in such a scenario. Conversation being read with 1 unread message sounds confusing although technically it's correct that the other party has read the chat and I have marked it as unread.

Note, Long form of `Delivery status: <insert state here>` translation string is a different issue and I suggest to handle it later when improving list of messages inside a conversation as it's also being used there and it's also affecting deltachat-desktop.

Please, Let me know if it's still difficult to understand what I am asking.

wchen342:
@pvagner Thank you for the very detailed explanation. I will try to reproduce with your steps. I thought it was a general problem that the delivery status and unread counter are shown together in the list, but from your description it seems a very specific case that it happens, so I need to investigate more about that. I suspect what you just described is actually a bug that is not limited to Talkback, but probably a general bug in the list view that also affects normal visual presentation.
The change I did was assuming delivery status and unread counter occurs at the same time and the only change I did was adding the prefix to make it clear the first announcement is for delivery status, but I see that is a separate problem now.

adbenitez:
@wchen342 in some rare cases it can happen that both  are displayed IIRC, like when you still have unread messages in that chat (still badge counter for the chat) but the last message was from you, maybe this can be triggered by replying from notification, or otherwise in some rare syncing races 

maybe we should just fix it in the UI and don't show the marker if the chat has an unread badge

wchen342:
@adbenitez Yeah I think there is a bug somewhere for the status and I will make a separate PR to fix that if I can find it, before this PR.

adbenitez (approved):
changelog conflict needs to be resolved, otherwise looks good to me

pvagner:
@wchen342 Thanks again, all is fine after todays update except of that thing with both indicators showing in rare cases you plan to look at as a different issue.

wchen342:
@pvagner The bug with both delivery status and unread count are shown together is fixed now, and the newest test apk shall include the fix. It shall also be fixed for Talkback too.

pvagner:
@wchen342 Huge thanks, It's working great for me now.

PR #4593, opened 2026-08-08
ConversationItemFooter: improve screen reader accessibility

pvagner:
(Fixes #4592)

* Since there are no actionable controls inside ConversationItemFooter, make all of it not important for accessibility
* Expand its getDescription() method so it describes all the missing indicators that were inaccessible up to this point making it similar as other children inside ConversationItem.
* Shorten all the accessibility specific strings describing delivery status of a chat messages (This contributes to https://github.com/deltachat/deltachat-desktop/issues/6605 ).

PR #4597, opened 2026-08-12
a11y: Renovate chat/messages screen

### LINKED ISSUE #4592: A11Y: When browsing list of messages in a chat with the screen reader running some message indicators are not read, others are too verbose or inconsistent
- Android version: Test with Android 17 Graphene OS
- Device: Pixel 9A
- Delta Chat version: 2.57
- Expected behavior: I'd like more consistent presentation of individual messages in a chat.
- Actual behavior: When reading individual messages in the chat with a screen reader running there are some usability and consistency issues:
    * The following message indicators are not read with talkback running: Edited, Saved, Location
    * Delivery status is not always read even if its read, it's too verbose.
    * With some 3rd party screen reader apps such as [Corvus](https://corvuskit.com/en) the message item and message footer are presented separated as two navigation elements.
    * Read counter for outgoing channel messages is presented with no context
    * Open deltachat with a screen reader running
    * Open any chat
    * Touch individual messages or use screen reader gestures to navigate over the list.

Related deltachat-desktop issue is https://github.com/deltachat/deltachat-desktop/issues/6605

wchen342:
Main changes:
- Make each row a screen reader node, mark item roots screenReaderFocusable, suppress decorative descendants, and label media wrappers
- Build descriptions from DcMsg/DcChat instead of aggregation of child view text
- Announce the Edited, Saved, Location and not-encrypted indicators, channel read count, quoted messages, reactions, and info message timestamps
- Add labeled actions: Play, Open image, Open file, Start app, Download, Show full message, Show error, Learn more
- Fix minor recycling and formatting issues

Fixes #4592

Supersede #4593

### DIFF
--- a/src/main/res/values/strings.xml
+++ b/src/main/res/values/strings.xml
@@ -1198,15 +1198,39 @@
     <string name="insert_newline">Insert Newline</string>
 
     <!-- accessibility, the general idea is to use the normal strings for accessibility hints wherever possible -->
-    <string name="a11y_delivery_status_error">Delivery status: Error</string>
-    <string name="a11y_delivery_status_sending">Delivery status: Sending</string>
+    <!-- deprecated -->
+    <string name="a11y_delivery_status_error">Error</string>
+    <string name="a11y_delivery_status_sending">Sending</string>
+    <!-- deprecated -->
     <string name="a11y_delivery_status_draft">Delivery status: Draft</string>
-    <string name="a11y_delivery_status_delivered">Delivery status: Delivered</string>
-    <string name="a11y_delivery_status_read">Delivery status: Read</string>
+    <string name="a11y_delivery_status_delivered">Delivered</string>
+    <string name="a11y_delivery_status_read">Read</string>
+    <!-- deprecated -->
     <string name="a11y_delivery_status_invalid">Invalid delivery status</string>
+    <!-- deprecated -->
     <string name="a11y_message_context_menu_btn_label">Message actions</string>
+    <!-- deprecated -->
     <string name="a11y_background_preview_label">Wallpaper preview</string>
+    <!-- deprecated -->
     <string name="a11y_disappearing_messages_activated">Disappearing messages activated</string>
+    <string name="a11y_msg_not_encrypted">Not end-to-end encrypted</string>
+
+    <plurals name="a11y_msg_read_by">
+        <item quantity="one">Read by %d</item>
+        <item quantity="other">Read by %d</item>
+    </plurals>
+    <string name="a11y_msg_reply_to">Replying to %1$s: %2$s</string>
+    <string name="a11y_msg_reply_to_no_author">Replying to: %1$s</string>
+
+    <string name="a11y_action_open_image">Open image</string>
+    <string name="a11y_action_open_video">Open video</string>
+    <string name="a11y_action_open_file">Open file</string>
+    <string name="a11y_action_open_contact">Open contact</string>
+    <string name="a11y_action_show_error">Show error</string>
+
+    <string name="a11y_reactions">Reactions: %1$s</string>
+    <string name="a11y_reactions_more">%1$d more</string>
+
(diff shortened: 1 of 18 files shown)


wchen342:
@pvagner Can you also test the apk in this PR? I have made some major changes to how everything on messages are announced, and I hope it work better than before, but it will be great if you can give some feedbacks so I can further tune it.

pvagner:
@wchen342 Huge thanks you are trying to improve all of this. Excuse me for delay in testing.

There are things I like and other things I wish to improve even further.

Now each message in the list is taking two accessibility nodes to explore and navigate through.

First the delivery state indicator is presented with the date, then the message content.
It's nice the message content is presented with no content description as there is no longer a need for adding accessibility actions for activating links inside the message content. However the fact there are two gestures required for navigating from one message to the other is something I'd like to try improving.

About the individual accessibility actions per message content such as open document, play video and similar, I understand it's a nice implementation but I find these somewhat redundant as there is still the default action saying activate which does the same thing on a simple click.

You are calling setScreenReaderFocusable() and setting state description on the message container and then keeping the other view for the content.

My assumption is that we don't need to handle screen reader focus seperate from the keyboard focus and we should do everything at the single node when looking at the accessibility tree.

I don't really know my-self how to make it the best, however I am thinking we should preserve the click behavior so the default action is performed by clicking. That's usefull for both screen reader users and other users that are not using a screen reader.

The individual descriptions for all the content types such as file, video, voice message, voice call are okay. The fact each message is reported as multiple objects affects usability.

pvagner:
@wchen342 I have only tested this with talkback so far. I will be testing and working with this more. I plan to take a look into the strings you have added if I can identify all of them when using with a screen reader.

Apparently in my previous comment I was wrong, the issue we are visiting two accessible object when navigating come from the fact date time and content are two seperate views.

Greetings

Peter

wchen342:
@pvagner Thank you for your comprehensive testings! It is very helpful and catches a few things I don't realize.

I think I can further combine the two nodes into a single one, but I need to figure out how the tree works when the framework builds it.

About the actions, yes I think they are redundant if they are already announced by clicking. I will check again and probably remove those.

I am not very sure about the screen reader focus versus keyboard focus thing, I will need to do more research into that, but I will try to improve on that.

src/main/res/values/strings.xml
```diff
@@ -1198,15 +1198,39 @@
...
+    <string name="a11y_msg_reply_to_no_author">Replying to: %1$s</string>
+
+    <string name="a11y_action_open_image">Open image</string>
+    <string name="a11y_action_open_video">Open video</string>
+    <string name="a11y_action_open_file">Open file</string>
+    <string name="a11y_action_open_contact">Open contact</string>
+    <string name="a11y_action_show_error">Show error</string>
+
+    <string name="a11y_reactions">Reactions: %1$s</string>
+    <string name="a11y_reactions_more">%1$d more</string>
```
r10s, line 1232: in general, we try to do support a11y greatly, at the same time to be as close to plain, visible UI. this means, that we use the same strings, the same controls etc.

so, the strings should become just `open_image`, `show_error` etc, so we do not need to duplicate them when needing elsewhere [^1]. it is known that other projects do it differently, but that is how we handle our strings - we duplicate only when they become ambiguous

[^1]: the existing a11y strings are partly "wrong" as well, but we do not change keys

pvagner:
Thanks @wchen342. Feel free to ask if you need my opinion on related stuff.

I am thinking one possible approach might be to hide everything except message body from screen readers by using `importantForAccessibility=no` recursivelly, then add everything including author, time and indicators when visible into state description for the message body view. Downside about this approach is that it will be difficult to copy or spell someone's name. Positive thing about not using content description will be that screen readers will be able to retrieve link spans dirrectly from the message body view thus those actions can also be removed.

Other approach might be just add everythink into the content description for the parent view and hide everything inside. When doing this, we'll also need accessibility actions for the links.

I can't make up my mind which of these approaches might be better or even if there might be a different approach as the balance between these two.

PR #4598, opened 2026-08-13
Block WebRTC per renderer instead of per webxdc app

### LINKED ISSUE #4436: Mini-apps loading times vary widely
- Android version: 16
- Device: XQ-ES54
- Delta Chat version: 2.49.0
- Expected behavior: All small apps load equally fast
- Actual behavior: Sometimes XDC apps load SLOW (10+ seconds) on Anroid while the same app loads fast on Linux.
- The problem occurs sometimes when I load apps (seems independent of specific app). E.g.:
   - If I load the app "chess", it loads quickly; If I load the app "Event Countdown" it delays 10s before starting (while starting immediately on linux)
  - If I load "chess" it is fast, If I close it and re-open, it is slow.
  I am unable to deterministically replicate. It usually happens every 2-3 times I open an app.
- Screenshots:
- Logs: sent per chat to ADB

wchen342:
Chromium's RTCPeerConnection limit is enforced per renderer process, so we allocate the whole budget in a hidden WebView, blockinng WebRTC in every WebView of the app.

WebRTC is now also blocked in the HTML mail viewer, help, connectivity and app store views, but this is not an issue since none of them needs WebRTC.

Closes #4436

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/WebRtcHolder.java
+++ b/src/main/java/org/thoughtcrime/securesms/WebRtcHolder.java
@@ -1,0 +1,206 @@
+package org.thoughtcrime.securesms;
+
+import android.annotation.SuppressLint;
+import android.content.Context;
+import android.os.Build;
+import android.util.Log;
+import android.webkit.JavascriptInterface;
+import android.webkit.RenderProcessGoneDetail;
+import android.webkit.WebResourceRequest;
+import android.webkit.WebResourceResponse;
+import android.webkit.WebSettings;
+import android.webkit.WebView;
+import android.webkit.WebViewClient;
+import androidx.annotation.RequiresApi;
+import java.io.ByteArrayInputStream;
+import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
+import org.thoughtcrime.securesms.util.Util;
+
+/**
+ * Holds the whole per-renderer RTCPeerConnection budget in a hidden WebView, so that no webxdc can
+ * use WebRTC.
+ *
+ * <p>WebView gives an app process a single renderer process, and Chromium's limit of 500
+ * RTCPeerConnections is enforced per renderer. Allocating all of them here blocks WebRTC in every
+ * WebView of this app.
+ */
+public final class WebRtcHolder {
+
+  private static final String TAG = "WebRtcHolder";
+
+  // RTCPeerConnection is only available in secure contexts
+  private static final String HOLDER_URL = "https://webrtc-holder.localhost/holder.html";
+
+  private static final long WAITER_TIMEOUT_MS = 15000;
+
+  public enum State {
+    IDLE,
+    FILLING,
+    CONFIRMED,
+    EMPTY,
+    PARTIAL,
+  }
+
+  // This is the mechanism itself. Also, there is no actual leak:
+  // the non-static classes only capture the singleton.
+  @SuppressLint("StaticFieldLeak")
+  private static WebRtcHolder instance;
+
+  private final Context appContext;
+  private final List<Runnable> waiters = new ArrayList<>();
+  private volatile WebView webView;
+  private volatile State state = State.IDLE;
+
+  private WebRtcHolder(Context context) {
+    this.appContext = context.getApplicationContext();
+  }
+
+  public static synchronized WebRtcHolder getInstance(Context context) {
+    if (instance == null) {
+      instance = new WebRtcHolder(context);
+    }
+    return instance;
+  }
+
+  public void start() {
+    Util.assertMainThread();
+    if (state != State.IDLE) {
+      return;
+    }
+    Log.i(TAG, "Starting fill");
+    state = State.FILLING;
+    createWebView();
+    webView.loadUrl(HOLDER_URL);
+  }
+
+  public State getState() {
+    if (!Util.isMainThread()) {
+      Log.w(TAG, "getState() off main thread, result may be stale");
+    }
+    return state;
+  }
+
+  private boolean isSettled() {
+    return state == State.CONFIRMED || state == State.EMPTY || state == State.PARTIAL;
+  }
+
+  public void awaitSettled(final Runnable callback) {
+    if (!Util.isMainThread()) {
+      Util.runOnMain(this::start);
+      return;
+    }
+    if (isSettled()) {
+      callback.run();
+      return;
+    }
+    final boolean[] fired = new boolean[1];
+    final Runnable once =
+        () -> {
+          if (!fired[0]) {
+            fired[0] = true;
+            callback.run();
+          }
+        };
+    waiters.add(once);
+    Util.runOnMainDelayed(
+        () -> {
+          if (!fired[0]) {
+            Log.w(TAG, "timed out waiting for holder");
+          }
+          once.run();
+        },
+        WAITER_TIMEOUT_MS);
+  }
+
+  public void onRendererGone() {
+    if (!Util.isMainThread()) {
+      Util.runOnMain(this::start);
+      return;
+    }
+    Log.w(TAG, "Renderer gone");
+    destroyWebView();
+    state = State.IDLE;
+    start();
+  }
+
+  private void createWebView() {
+    destroyWebView();
+
+    WebView wv = new WebView(appContext);
+    WebSettings settings = wv.getSettings();
+    settings.setJavaScriptEnabled(true);
+    settings.setBlockNetworkLoads(true);
+    settings.setAllowFileAccess(false);
+    settings.setAllowContentAccess(false);
+    settings.setGeolocationEnabled(false);
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+      settings.setSafeBrowsingEnabled(false);
+    }
+    wv.setWebViewClient(new HolderWebViewClient());
+    wv.addJavascriptInterface(new HolderJSApi(), "HolderJSApi");
+    webView = wv;
+  }
+
+  private void destroyWebView() {
+    if (webView != null) {
+      webView.destroy();
+      webView = null;
+    }
+  }
+
+  private void handleFillFinished(int count, String error, boolean stillHolding) {
+    Util.assertMainThread();
+    if (state != State.FILLING) {
+      return;
+    }
+    if (error == null) {
+      state = State.CONFIRMED;
+      Log.i(TAG, "Confirmed, holding " + count + " connections");
+    } else if (stillHolding) {
+      state = State.PARTIAL;
+      Log.w(TAG, "Holding " + count + " connections but not the whole budget: " + error);
+    } else {
+      state = State.EMPTY;
+      Log.w(TAG, "Fill failed after " + count + " connections, released: " + error);
+      destroyWebView();
+    }
+    List<Runnable> pending = new ArrayList<>(waiters);
+    waiters.clear();
+    for (Runnable waiter : pending) {
+      waiter.run();
+    }
+  }
+
+  private WebResourceResponse serve(String url) {
+    if (url != null && url.startsWith(HOLDER_URL)) {
+      InputStream stream = appContext.getResources().openRawResource(R.raw.webrtc_holder);
+      return new WebResourceResponse("text/html", "UTF-8", stream);
+    }
+    Log.w(TAG, "Blocked unexpected request from holder: " + url);
+    return new WebResourceResponse("text/plain", "UTF-8", new ByteArrayInputStream(new byte[0]));
+  }
+
+  private class HolderWebViewClient extends WebViewClient {
+    @Override
+    public WebResourceResponse shouldInterceptRequest(WebView view, WebResourceRequest request) {
+      return serve(request.getUrl().toString());
+    }
+
+    @Override
+    @RequiresApi(Build.VERSION_CODES.O)
+    public boolean onRenderProcessGone(WebView view, RenderProcessGoneDetail detail) {
+      Log.w(TAG, "Holder renderer gone, didCrash=" + detail.didCrash());
+      Util.runOnMainDelayed(WebRtcHolder.this::onRendererGone, 0);
+      return true;
+    }
+  }
+
+  private class HolderJSApi {
+    @JavascriptInterface
+    public void onFillFinished(final int count, final String error, final boolean stillHolding) {
+      Util.runOnMain(() -> handleFillFinished(count, error, stillHolding));
+    }
+  }
+}
--- a/src/main/java/org/thoughtcrime/securesms/WebRtcHolder.java
+++ b/src/main/java/org/thoughtcrime/securesms/WebRtcHolder.java
@@ -40,7 +40,6 @@ public final class WebRtcHolder {
     FILLING,
     CONFIRMED,
     EMPTY,
-    PARTIAL,
   }
 
   // This is the mechanism itself. Also, there is no actual leak:
@@ -83,7 +82,7 @@ public final class WebRtcHolder {
   }
 
   private boolean isSettled() {
-    return state == State.CONFIRMED || state == State.EMPTY || state == State.PARTIAL;
+    return state == State.CONFIRMED || state == State.EMPTY;
   }
 
   public void awaitSettled(final Runnable callback) {
@@ -150,20 +149,16 @@ public final class WebRtcHolder {
     }
   }
 
-  private void handleFillFinished(int count, String error, boolean stillHolding) {
-    Util.assertMainThread();
+  private void handleFillFinished(String result, String reason) {
     if (state != State.FILLING) {
       return;
     }
-    if (error == null) {
+    if ("confirmed".equals(result)) {
       state = State.CONFIRMED;
-      Log.i(TAG, "Confirmed, holding " + count + " connections");
-    } else if (stillHolding) {
-      state = State.PARTIAL;
-      Log.w(TAG, "Holding " + count + " connections but not the whole budget: " + error);
+      Log.i(TAG, "Confirmed, holding the whole budget");
     } else {
       state = State.EMPTY;
-      Log.w(TAG, "Fill failed after " + count + " connections, released: " + error);
+      Log.w(TAG, "Released: " + reason);
       destroyWebView();
     }
     List<Runnable> pending = new ArrayList<>(waiters);
@@ -199,8 +194,8 @@ public final class WebRtcHolder {
 
   private class HolderJSApi {
     @JavascriptInterface
-    public void onFillFinished(final int count, final String error, final boolean stillHolding) {
-      Util.runOnMain(() -> handleFillFinished(count, error, stillHolding));
+    public void onFillFinished(final String result, final String reason) {
+      Util.runOnMain(() -> handleFillFinished(result, reason));
     }
   }
 }
--- a/src/main/java/org/thoughtcrime/securesms/WebRtcHolder.java
+++ b/src/main/java/org/thoughtcrime/securesms/WebRtcHolder.java
@@ -81,13 +81,19 @@ public final class WebRtcHolder {
     return state;
   }
 
-  private boolean isSettled() {
+  /**
+   * Whether the fill has finished. Callers rely on this meaning exactly CONFIRMED or EMPTY.
+   *
+   * <p>Callers treat "settled but not CONFIRMED" as EMPTY, i.e. as permission to fill the budget
+   * themselves. A state where the holder has partial budget must not be settled.
+   */
+  public boolean isSettled() {
     return state == State.CONFIRMED || state == State.EMPTY;
   }
 
   public void awaitSettled(final Runnable callback) {
     if (!Util.isMainThread()) {
-      Util.runOnMain(this::start);
+      Util.runOnMain(() -> awaitSettled(callback));
       return;
     }
     if (isSettled()) {
@@ -115,7 +121,7 @@ public final class WebRtcHolder {
 
   public void onRendererGone() {
     if (!Util.isMainThread()) {
-      Util.runOnMain(this::start);
+      Util.runOnMain(this::onRendererGone);
       return;
     }
     Log.w(TAG, "Renderer gone");
--- a/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -293,14 +293,14 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
             if (isFinishing() || isDestroyed() || webView == null) {
               return;
             }
-            WebRtcHolder.State holderState = holder.getState();
             if (!holder.isSettled()) {
-              Log.e(TAG, "Cannot block WebRTC (" + holderState + "), refusing to load");
+              Log.e(TAG, "Cannot block WebRTC (" + holder.getState() + "), refusing to load");
               finish();
               return;
             }
             webView.loadUrl(
-                buildBootstrapUrl(holderState == WebRtcHolder.State.CONFIRMED, finalEncodedHref));
+                buildBootstrapUrl(
+                    holder.getState() == WebRtcHolder.State.CONFIRMED, finalEncodedHref));
           });
     }
 
--- a/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
+++ b/src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
@@ -294,8 +294,7 @@ public class WebxdcActivity extends WebViewActivity implements DcEventCenter.DcE
               return;
             }
             WebRtcHolder.State holderState = holder.getState();
-            if (holderState != WebRtcHolder.State.CONFIRMED
-                && holderState != WebRtcHolder.State.EMPTY) {
+            if (!holder.isSettled()) {
               Log.e(TAG, "Cannot block WebRTC (" + holderState + "), refusing to load");
               finish();
               return;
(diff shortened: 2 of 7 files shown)


wchen342:
This finally solves the problem when webxdc randomly lags for ~30s when opened more than once.

WofWca (commented):
Wow, this indeed seems to work, nice job!

However, am I getting this right that we create the `RTCPeerConnection` objects on one origin (URL, seems to be `https://webrtc-holder.localhost/` in this case), and then let the WebXDC apps that have different origins rely on that?
So different origins, and even different apps aren't process-isolated?
If that's the case then I'm wondering if it *should* be the case...

Looking at
- https://chromium.googlesource.com/chromium/src/+/main/docs/process_model_and_site_isolation.md
  > No Site Isolation
  > Android WebView does not yet support multiple renderer processes or out-of-process iframes.
- https://www.chromium.org/Home/chromium-security/site-isolation/
  > Site Isolation is not yet supported in Android WebView
- https://chromium.googlesource.com/chromium/src/+/HEAD/android_webview/renderer/README.md
  > Apps can create multiple WebView Profiles, in which case each Profile gets its own renderer process. Please see [WebViewCompat.setProfile](https://developer.android.com/reference/androidx/webkit/WebViewCompat#setProfile%28android.webkit.WebView,java.lang.String%29)

Maybe we'd rather want to `setProfile` to get process isolation?

---
In case process isolation is not possible to achieve, I guess we can take advantage of how things are now, but we also should be prepared for process isolation to become available in WebView, in case that happens before we get a more proper way to disable WebRTC.

WofWca:
> but we also should be prepared for process isolation to become available in WebView

By this I mean that we might need to basically revert this change then.

wchen342:
@WofWca Basically the original implementation depended on the absence of isolation for iframes, the holder additionally depends on it for the top-level document. This is the only additional dependency, everything else is effectively the same.

About the process, currently there is strictly one process per app so there is nothing to isolate anyway, and between apps.the processes are sandboxed. In the original implementation, all webxdc, even opened separately, *still shares the same renderer* because that is how webview works today, so this is no worse than the original.

Also the assuring part is that if isolation is ever implemented (which is unlikely in any near future), this does not fail silently as top will throw ERROR_A, so no worry on that part.

wchen342:
Also profiles are for data-separation, not for site isolation if I'm not mistaken.

WofWca (commented):
> the original implementation depended on the absence of isolation for iframes

Oof, right. That is also going to come around when cross-site isolation starts working, so we're gonna be in trouble anyway...

> so this is no worse than the original

My point is not that this is worse than the original but that we might want to improve security by enabling process-isolation, which would make this MR not work.

> Also profiles are for data-separation, not for site isolation if I'm not mistaken.

Why not both? See the quote above, 
> each Profile gets its own renderer process

src/main/res/raw/webxdc_wrapper.html
```diff
@@ -42,120 +42,148 @@
       style="display: none"
     ></iframe>
     <script>
-      const loadingDiv = document.getElementById("loading");
-      const iframe = document.getElementById("frame");
-
-      let fill500 = async (href) => {
-        const connections = [];
+        const loadingDiv = document.getElementById("loading");
```
WofWca: Nitpick: not sure why an extra indentation level is needed now

src/main/res/raw/webxdc_wrapper.html
```diff
@@ -42,120 +42,148 @@
...
+                console.log("WEBRTC-WG: giving up");
+                return false;
+              }
+              new Array(1024*1024).fill(0);
+              await new Promise((res) => setTimeout(res, 250));
+              console.log("WEBRTC-WG: waiting for gc");
             }
+          }
+          console.log("WEBRTC-WG: done");
+          window.__webrtcWgConnections = connections;
```
WofWca: This allows the guest WebXDC app to close the `RTCPeerConnection` objects with `window.top.__webrtcWgConnections[0].close()`, thus break the isolation, when the fallback is used, or am I missing something?

WofWca (commented):
In case we're not going for trying to add process isolation: the high-level idea, i.e. "exhaust the limit just once for all apps" looks good to me!
One thing I shall remind ourselves about: we must ensure that the `RTCPeerConnection`s don't get garbage-collected while an app is already open.

As to the code: pretty hard to review, lots of refactors mixed in, weird-looking `MessageChannel` usage, seeming WIP leftovers. Not sure if this MR is to be treated as a prototype.

src/main/res/raw/webxdc_wrapper.html
wchen342: Yes, this is a real issue which I accidentally introduced. It is fixed in the new version.

src/main/res/raw/webxdc_wrapper.html
wchen342: Android Studio doesn't have JS/html formatting..Maybe we shall add JS/HTML formatting to Spotless too? @adbenitez

adbenitez:
I guess this should be marked as closing https://github.com/deltachat/deltachat-android/issues/4436

src/main/res/raw/webxdc_wrapper.html
adbenitez: probably makes sense, or generate the file outside in a small web subproject with vite + typescript + prettier in the repo or so , the webxdc.js file is actually a mess and a bit buggy IIRC because it is basically written as a raw file without any kind of linter checking javascript errors etc

wchen342: I mean it's JS right now, linter won't work very well, we need proper TypeScript set up and compile them to JS, which needs some additional set up.

adbenitez: it is possible to do typescript checking in js files (and type hints via comments) but yes, my previous comment was getting bit offtopic, to answer your question more concretely: yes I think it makes sense for now to use spotless for formatting js files if it is easy to solve

the bugs in webxdc.js etc are unrelated and needs to be checked in the future

wchen342:
@WofWca I have cleaned up and simplified the JS/HTML part, the diff for the wrapper shall be much easier to see now.

> That is also going to come around when cross-site isolation starts working, so we're gonna be in trouble anyway

In that case the whole FILL500 trick stops working, so it is kind of moot to argue for it here. We at least will catch it with `ERROR_C` if that really happens.

>  we might want to improve security by enabling process-isolation, which would make this MR not work

If you mean site isolation for each webxdc-app: yes we can use profiles for that, but there are reasons against that:

 - If we open each WebXDC app in a separate profile, it will need to start a separate rendering process+data I/O for every cold start, adding RAM and CPU usage for each app opened (simultaneously) which will be problematic on low-end or old devices
 
  - It will require WebView somewhere around mid-2023 up, again depending on if the device have updated WebView apk installed. Also some 32-bit devices don't support out-of-process renderer, which makes profile unusable.
  
  - We need to figure out how to migrate existing apps because everything is in the default profile now.
  
  - There is a huge problem with lifetime management, because we have app+msg as identifiers for an webxdc, so we need to manager, for example, when a message is deleted, since profiles are persistent, not in RAM.
  
  In any case I think the profile/site isolation needs its own discussion, and is beyond the scope of this pr so let us not getting too deep into it.

src/main/res/raw/webxdc_wrapper.html
```diff
@@ -61,23 +61,46 @@
...
+          loadingProgress.value = 0;
+          while (connections.length < 500) {
+            try {
+              connections.push(new RTCPeerConnection(cert));
+              if (connections.length%50 == 0) {
+                loadingProgress.value = connections.length;
+                await new Promise((res) => setTimeout(res));
+              }
+            } catch (error) {
+              new Array(1024*1024).fill(0);
```
WofWca:
```suggestion
              loadingProgress.value++;
              new Array(1024*1024).fill(0);
```

src/main/res/raw/webxdc_wrapper.html
```diff
@@ -111,15 +140,15 @@
...
-
                   sandboxedIframe.remove();
-                  window.removeEventListener("message", messageListener);
+                  finish(event.data.rtcpcCreationFailed);
                   break;
                 }
               }
             }
             window.addEventListener("message", messageListener);
+            retry = setInterval(askIframeToPerformCheck, 250);
```
WofWca, line 147: What changed in this MR that made the retry necessary? In which circumstances does it happen that the iframe doesn't respond?
Is it simply because now we get to this check faster and sometimes the iframe doesn't manage to load yet?

src/main/res/raw/webrtc_holder.html
```diff
@@ -0,0 +1,114 @@
...
+    <meta charset="UTF-8">
+</head>
+<body>
+<script>
+    // Allocates the whole per-renderer RTCPeerConnection budget so no other document in
+    // this app process can create one. See WebRtcHolder.java for additional info.
+    //
+    // `connections` is a top-level `const` and it must stay reachable for the lifetime
+    // of this document.
+    const connections = [];
```
WofWca: Nitpick: Not sure if it's good to diverge from the original wrapper's approach in this regard.

src/main/res/raw/webrtc_holder.html
```diff
@@ -0,0 +1,114 @@
...
+    (async () => {
+      const EXPECTED_CAPACITY = 500;
+      const HARD_LIMIT = 2000;
+      // 20 * 500ms = 10s, which must stay below WAITER_TIMEOUT_MS in WebRtcHolder.java.
+      const TOP_UP_ATTEMPTS = 20;
+      const TOP_UP_DELAY_MS = 500;
+
+      const release = () => {
+        for (const pc of connections) {
+          try { pc.close(); } catch (e) {}
```
WofWca: Does `close()` ever throw? If it does, does it make sense to proceed with an unreleased `RTCPeerConnection`?

src/main/java/org/thoughtcrime/securesms/WebRtcHolder.java
```diff
@@ -0,0 +1,206 @@
...
+  }
+
+  private void destroyWebView() {
+    if (webView != null) {
+      webView.destroy();
+      webView = null;
+    }
+  }
+
+  private void handleFillFinished(int count, String error, boolean stillHolding) {
```
WofWca: Nitpick:
Maybe it's just me but I find this API unnecessarily complex.
At least the `count`, which is only used for logging, shouldn't be the first argument?
Better yet it can probably be merged into `error` as the only case where there is no error is when it's 500.

And the `stillHolding` parameter is also weird. It's unused when `error == null` and the "PARTIAL" state sort of assumes that `stillHolding` corresponds to the fact that we couldn't claim the budget.

We can simply accept an enum of three (string or number), and a nullable error that's only used for logging.

src/main/res/raw/webrtc_holder.html
```diff
@@ -0,0 +1,114 @@
...
+        }
+        connections.length = 0;
+      };
+
+      try {
+        if (typeof RTCPeerConnection === "undefined") {
+          HolderJSApi.onFillFinished(0, "RTCPeerConnection unavailable", false);
+          return;
+        }
+
```
WofWca: Nitpick:
We're not really expecting this to happen. Even if it does happen, we'll just fall through to `catch` and all that's gonna be different is the precise error wording.

```suggestion
```

src/main/java/org/thoughtcrime/securesms/WebxdcStoreActivity.java
```diff
@@ -102,15 +105,23 @@ public WebResourceResponse shouldInterceptRequest(
...
+          }
         });
 
     WebSettings webSettings = webView.getSettings();
     webSettings.setJavaScriptEnabled(true);
     webSettings.setAllowFileAccess(false);
     webSettings.setAllowContentAccess(false);
     webSettings.setGeolocationEnabled(false);
-    webSettings.setAllowFileAccessFromFileURLs(false);
-    webSettings.setAllowUniversalAccessFromFileURLs(false);
```
WofWca, line 113: Is this relevant to this MR?

src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -629,13 +643,17 @@ public void onActivityResult(int reqCode, int resultCode, final Intent data) {
...
+    final int accountId = dcContext.getAccountId();
+    final int msgId = dcAppMsg.getId();
+    Util.runOnAnyBackgroundThread(
+        () -> {
+          try {
+            rpc.leaveWebxdcRealtime(accountId, msgId);
+          } catch (RpcException e) {
+            e.printStackTrace();
+          }
+        });
```
WofWca, line 655: Is this relevant to this MR?

src/main/res/raw/webrtc_holder.html
```diff
@@ -0,0 +1,114 @@
...
+        // Closing frees most of the memory (~46 KB instead of ~84 KB per connection) but does
+        // not free the slot. The process-wide counter is decremented in the destructor, not in
+        // close(). Note that an unclosed connection is kept alive by the engine, but a closed one
+        // is kept alive only by our reference to it. The `connections` array must not be cleared
+        // or go out of scope.
+        for (const pc of connections) {
+          pc.close();
+        }
```
WofWca: I'm personally not quite confident that a top-level `const` is enough to ensure that it's not gonna get GCd. And the check below won't catch it if the garbage collection doesn't run immeditaly after `close()`.
Firstly maybe it'd be better to `window.__connections = []`, we can afford it in this script since it's not accessible by apps.
Secondly: maybe 2x difference in memory usage is not that bad?

src/main/res/raw/webrtc_holder.html
```diff
@@ -0,0 +1,114 @@
...
+          attempts++;
+          await new Promise((res) => setTimeout(res, TOP_UP_DELAY_MS));
+          await fillUntilRejected();
+        }
+
+        if (connections.length < EXPECTED_CAPACITY) {
+          HolderJSApi.onFillFinished(
+            connections.length, "could not claim the whole budget", true);
+          return;
+        }
```
WofWca: Why do we `release()` before returning all other errors, but not for this one?

src/main/res/raw/webrtc_holder.html
```diff
@@ -0,0 +1,114 @@
...
+          }
+          return false;
+        };
+
+        if (!(await fillUntilRejected())) {
+          const cl = connections.length;
+          release();
+          HolderJSApi.onFillFinished(cl, "no connection limit enforced", false);
+          return;
+        }
```
WofWca: This whole check can be covered by a single `new RTCPeerConnection();` at the end, which we already have (its error message will need to be adjusted though).
Again, I doesn't look like diverging from the original script is paying off.

WofWca: Nitpick:
This whole check can be covered by a single `new RTCPeerConnection();` at the end, which we already have (its error message will need to be adjusted then though). The return value (bool) is also not needed then.

```suggestion
        await fillUntilRejected();
```

Again, it doesn't look like diverging from the original script is paying off.

src/main/res/raw/webrtc_holder.html
```diff
@@ -0,0 +1,114 @@
...
+          certificates: [
+            await RTCPeerConnection.generateCertificate({
+              name: "ECDSA",
+              namedCurve: "P-256",
+            }),
+          ],
+        };
+
+        const fillUntilRejected = async () => {
+          while (connections.length < HARD_LIMIT) {
```
WofWca: Nitpick:
yet another case against diverging from the original script: this script will happily report success if it manages to allocate 667 and no more connections.

WofWca (commented):
I haven't thoroughly reviewed the Java code as I'm unfortunately not super familiar with it and Android.
But I confirm that, again, as long as we don't suddenly GC the objects, the security is not weakened. The WebXDC wrapper script still checks if it can create a new object in either case. And if the holder doesn't report success, then the original script falls back to trying to create 500 objects, and the holder only reports success if it manages to create 500 (or more, see my comment).

src/main/java/org/thoughtcrime/securesms/WebxdcStoreActivity.java
wchen342: These are no longer necessary, we usually don't open a pr for every single small thing here so just fix these along the rest.

src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
wchen342: This is also a bug fix, same as above.

src/main/res/raw/webrtc_holder.html
wchen342: `const` and `window` shall both be effective because we are at top level and the page is not going away. The original reason to use `const` is because variable on `window` can be deleted but yeah it is not accessible so it is fine here.
I am not sure about the memory usage is fine or not, I measured on emulator which means without closing we keep about 90MB permanently with the app and with closing about 40MB. It is moderately large but maybe fine.

WofWca: If you have another script, this const is not going to be available to it. So what's holding a reference to it?

src/main/res/raw/webrtc_holder.html
wchen342: The intention was that holding a partial capacity is better than releasing it because probably it is caused by some other webview not fully released their connections yet in which case we are still blocking, but this is probably too complicated.

src/main/res/raw/webrtc_holder.html
wchen342: In that case it means the capacity is 667 and we are at the capacity, no?

The intention is to give a heads up in case the capacity change, but in any case we can just do the original fill as above and it is simpler.

WofWca: Hypothetically, this can mean that the capacity was raised to 1000, but another 333 connections are held by another script, which is what the surrounding code seems to be trying to account for.

src/main/res/raw/webrtc_holder.html
wchen342: What do you mean? The scripts are not moduled so two scripts will share the scope. If you are talking about webxdc script then they are different documents so they cannot see each other from the beginning.

The document in the webview is never closed, so the document itself is holding the variable. That is the whole point.

WofWca: Hm you're right. Still, assigning to `globalThis` would ensure this.

src/main/res/raw/webxdc_wrapper.html
wchen342: Yes, it's because the iframe may not be loaded, but you are right the retry is not needed either way because the listener is attached already.

src/main/res/raw/webrtc_holder.html
wchen342: We can change it to use `window`, see below.

src/main/res/raw/webrtc_holder.html
wchen342: In any case, using `window` is equivalent in this case because it is not reachable by other webviews, so I will just change it to be on `window`.

src/main/res/raw/webrtc_holder.html
wchen342: It used to throw `InvalidStateError` I think but that seems to be changed recently. I'll remove the catch.

wchen342:
I have simplified the script and removed most of the additional functionalities. It shall be very close to the wrapper script now.

src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
```diff
@@ -272,23 +281,29 @@ public boolean onShowFileChooser(
...
+      // Wait until the state of the RTCPeerConnection budget is settled.
+      final WebRtcHolder holder = WebRtcHolder.getInstance(this);
+      holder.awaitSettled(
+          () -> {
+            if (isFinishing() || isDestroyed() || webView == null) {
+              return;
+            }
+            WebRtcHolder.State holderState = holder.getState();
+            if (holderState != WebRtcHolder.State.CONFIRMED
+                && holderState != WebRtcHolder.State.EMPTY) {
```
WofWca: Nitpick:

```suggestion
            if (!holder.isSettled()) {
```

With the removal of `PARTIAL` this is now equivalent.
Up to you to decide if the surrounding code (specifically the error message) still makes sense.

src/main/java/org/thoughtcrime/securesms/WebRtcHolder.java
```diff
@@ -0,0 +1,201 @@
...
+  // RTCPeerConnection is only available in secure contexts
+  private static final String HOLDER_URL = "https://webrtc-holder.localhost/holder.html";
+
+  private static final long WAITER_TIMEOUT_MS = 15000;
+
+  public enum State {
+    IDLE,
+    FILLING,
+    CONFIRMED,
+    EMPTY,
```
WofWca, line 42: Nitpick: With the removal of `PARTIAL` maybe this needs to be renamed to simply `FAILED` or something.

src/main/res/raw/webrtc_holder.html
```diff
@@ -0,0 +1,73 @@
...
+      try {
+        await fill500();
+
+        // Closing frees most of the memory (~46 KB instead of ~84 KB per connection) but does
+        // not free the slot. The process-wide counter is decremented in the destructor, not in
+        // close(). Note that an unclosed connection is kept alive by the engine, but a closed one
+        // is kept alive only by our reference to it. The `connections` array must not be cleared
+        // or go out of scope.
+        for (const pc of window.__connections) {
+          pc.close();
```
WofWca: BTW if we really decide to go for this, then it's probably better to do it right inside the loop which creates the objects. Less peak memory consumption (probably) and more likely to catch the possible error where closing the connections frees up the pool.

WofWca (commented):
I verified that the new script is basically copy-pasted old one with some plumbing adjusted.

The assignment to `window` is good,
but I am still not super confident about the `pc.close()` thing, I also didn't measure the memory usage.

Besides that, the JS part and the overall approach get an "approved" from me.

If someone is going to review the Java part and you want my advice: again, please ensure that the holder can' suddenly stop existing and drop the `RTCPeerConnection` objects while a WebXDC app is still open (as it might re-acquire them).

I re-tested the last build, tried waiting out for GC, but all works well.

src/main/res/raw/webxdc_wrapper.html
adbenitez: for the record, things like this is what I was talking about with bugs in webxdc.js: https://github.com/deltachat/deltachat-android/pull/4632

adbenitez:
@wchen342 could you update the branch and add a changelog entry? ex. "improve opening speed of in-chat apps"

src/main/java/org/thoughtcrime/securesms/WebRtcHolder.java
```diff
@@ -0,0 +1,201 @@
...
+
+  private boolean isSettled() {
+    return state == State.CONFIRMED || state == State.EMPTY;
+  }
+
+  public void awaitSettled(final Runnable callback) {
+    if (!Util.isMainThread()) {
+      Util.runOnMain(this::start);
+      return;
+    }
```
adbenitez, line 98: callback is never called if `awaitSettled` was not called from the main thread, this makes a bit hard to think about it, if this was intentional better add a doc string describing what is "callback" when does it get called

also passing a callback while calling this function not from main thread would be a programming error then, so you should assert it is null in such case or better decouple this logic

src/main/java/org/thoughtcrime/securesms/WebxdcStoreActivity.java
adbenitez: in general, we don't want unrelated changes in PRs, it doesn't matter if it is a 2 lines change PR

src/main/java/org/thoughtcrime/securesms/WebxdcActivity.java
adbenitez: this time leave it like that, but in the future, please, move such change to another branch/pr that depends in your base PR to keep PRs like this one that are not so small easier to review and more on-point

src/main/java/org/thoughtcrime/securesms/WebRtcHolder.java
wchen342: It is an error when editing, thanks for catching that. The intention is just to re-run on main thread if not already on it.

src/main/res/raw/webrtc_holder.html
wchen342: The closing is moved into the loop now.

adbenitez (approved):
this patch solves a big road block in practical webxdc usage, without this it was very annoying and buggy experience, thanks a lot!!!

one thing I noticed is that now this slows down the opening of the mini-apps store and even the offline help, I guess also the opening of html emails, the first time one WebviewActivity child is opened

I think for the case of the html pages like offline help and xdc store it is better to not trigger the FILL500 and trigger `WebRtcHolder.start()` only on WebxdcActivity (this will still affect the map, so one could claim it is only really necessary to call it if internet access flag is false, but would be good enough even without this optimization since map is experimental feature)

for the case of the mini-apps, the first time you open them there is a complete blank screen for a second, no progress bar like in the past, I guess in most fast devices this is just 1 sec blank screen and it is only the first time, but in older/slower devices it might be more noticeable, an spinning wheel or other type of undefined progress bar could have been shown there but maybe good enough as it is for now, we can always further improve if this becomes a problem people report

wchen342:
> I think for the case of the html pages like offline help and xdc store it is better to not trigger the FILL500 and trigger WebRtcHolder.start() only on WebxdcActivity

Yes that is doable, I can change that later as it are already a lot if changes here.

> an spinning wheel or other type of undefined progress bar could have been shown

That is just some JS changes to add the animation, probably reuse the existing progress bar. We can revisit this later.

PR #4613, opened 2026-08-20
remove unused leftover variable in AppUpdateDialogFragment

adbenitez:
just remove an unused variable forgotten from #4582 

#skip-changelog


wchen342 (approved):
Good catch!

PR #4621, opened 2026-08-24
Add speed button for voice/audio messages

### LINKED ISSUE #4588: allow to change voice message playing speed
iOS and desktop already allow to control playback speed, only Android is missing the feature

there is [some PR](https://github.com/deltachat/deltachat-android/pull/4429) from some external contributor, maybe we can reuse some of the logic there, but UI-wise we want to:
 
- simplify the current player and have a single track time instead of the too long `00:00 / 00:13`
- when the audio is played, add a speed button after the track time with `1x` label, upon click, loop over the possible values (`1x, 1.5x, 2x`)

unlike the current open PR, the button should have a background or border

wchen342:
Partially based on #4429

Closes #4588

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationItem.java
@@ -291,8 +291,8 @@ public class ConversationItem extends BaseConversationItem {
     try (TypedArray attrs = context.obtainStyledAttributes(attributes)) {
       incomingBubbleColor = attrs.getColor(0, Color.WHITE);
       outgoingBubbleColor = attrs.getColor(1, Color.WHITE);
-      incomingTextSecondaryColor = attrs.getColor(2, Color.WHITE);
-      outgoingTextSecondaryColor = attrs.getColor(3, Color.WHITE);
+      incomingTextSecondaryColor = attrs.getColor(2, Color.BLACK);
+      outgoingTextSecondaryColor = attrs.getColor(3, Color.BLACK);
     }
   }
 
--- a/src/main/java/org/thoughtcrime/securesms/ConversationItem.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationItem.java
@@ -131,6 +131,8 @@ public class ConversationItem extends BaseConversationItem {
 
   private int incomingBubbleColor;
   private int outgoingBubbleColor;
+  private int incomingTextSecondaryColor;
+  private int outgoingTextSecondaryColor;
 
   public ConversationItem(Context context) {
     this(context, null);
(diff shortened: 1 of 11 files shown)


adbenitez:
testing it, the button is very hard to tap (and more when using the thumb with single-hand handling) and easy to end up clicking progress bar instead because the border of the progress bar is too close to the button see:
image

wchen342:
I'm not sure what is the best way here. Originally the button is taller, but then it just looks out of place in the bottom line because it is bigger. Maybe make the whole row taller?

wchen342:
And yeah probably should added padding there too, although that will make the whole bubble larger.

wchen342:
I have moved the time and button to the same row as the date, below the play button.
Also made the button bigger so it is easier to click.

image

adbenitez:
@wchen342 nice! but I think it is better to keep the audio filename where it was below the progress bar, also could we keep the footer as it was?

wchen342:
Do you mean make the footer in xml using a similar style to the rest? Because nothing in the footer class changed, the changes are just the same as in thumbnail bubbles, etc.

adbenitez:
I mean that the footer now doesn't look the same as it was in the past (it has orange color) didn't look in the internals if that was the only change you did to the footer

wchen342:
Yeah that is because the added one in xml is not copy pasted exactly. I'll check that.

wchen342:
@adbenitez Ok, the footer shall be fixed now. Filename is moved back to under seekbar.

adbenitez:
much nicer! it is looking almost finished, but there is a small detail left: the time is a bit too on the left compared to the play button:

image

for reference, this is how it looks in signal:
image


btw, code formatter needs to be ran

wchen342:
Hmm I definitely ran code formatter last time, maybe last minute change messed it up again.

Anyway, now the text shall be centered to the button.

wchen342:
Wait, something is wrong with the formatter. I formatted locally but the CI still complains.

EDIT: I checked, the branch is up to date, and both `spotlessCheck` and `spotlessApply` passes for me locally.

src/main/java/org/thoughtcrime/securesms/ConversationItem.java
```diff
@@ -283,10 +285,14 @@ private void initializeAttributes() {
...
           R.attr.conversation_item_incoming_bubble_color,
           R.attr.conversation_item_outgoing_bubble_color,
+          R.attr.conversation_item_incoming_text_secondary_color,
+          R.attr.conversation_item_outgoing_text_secondary_color,
         };
     try (TypedArray attrs = context.obtainStyledAttributes(attributes)) {
       incomingBubbleColor = attrs.getColor(0, Color.WHITE);
       outgoingBubbleColor = attrs.getColor(1, Color.WHITE);
+      incomingTextSecondaryColor = attrs.getColor(2, Color.WHITE);
+      outgoingTextSecondaryColor = attrs.getColor(3, Color.WHITE);
```
adbenitez: this should never happen, but if it would, the background was set to white as fallback, so this ends on white-on-white

```suggestion
      incomingTextSecondaryColor = attrs.getColor(2, Color.BLACK);
      outgoingTextSecondaryColor = attrs.getColor(3, Color.BLACK);
```

adbenitez:
this is how it looks when you set an audio as draft
image

wchen342:
Ok the draft problem is fixed now. I am trying to debug the Spotless problem, haven't figured out what is going on here.

wchen342:
Tested it now and confirmed, it is Java 17 that is causing the problem, change Spotless to use Java 21 passes.

Context: See [this comment](https://github.com/diffplug/spotless/issues/2503#issuecomment-2953146277).

adbenitez (approved):
all seems good now, thanks a lot! this is a nice improvement 🚀

PR #4623, opened 2026-08-25
update dependencies and remove unused ones

### LINKED ISSUE #4622: androidx.multidex dep not needed for minSdk >= 21
from gradle output while building:

```
WARNING: The multidex library is included as a dependency, but it is not needed for apps
with minSdk >= 21. Please remove dependency 'androidx.multidex:multidex:2.0.1' from ':'.
See https://developer.android.com/build/multidex for more information.
```

quote from https://developer.android.com/build/multidex#mdex-on-l

> Android 5.0 (API level 21) and higher uses a runtime called ART that natively supports loading multiple DEX files from APK files. ART performs pre-compilation at app install time, scanning for classesN.dex files and compiling them into a single OAT file for execution by the Android device. Therefore, if your minSdkVersion is 21 or higher, multidex is enabled by default and you don't need the multidex library.

adbenitez:
close #4622 

#skip-changelog

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -200,7 +200,6 @@ dependencies {
     implementation 'androidx.lifecycle:lifecycle-livedata-ktx:2.9.4'
     implementation 'androidx.sharetarget:sharetarget:1.2.0'
     implementation 'androidx.webkit:webkit:1.14.0'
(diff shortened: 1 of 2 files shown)


wchen342 (approved):
LGTM, thanks!

PR #4632, opened 2026-08-28
fix: type check of webxdc realtime data

d2weber:


d2weber:
Related PRs:
https://codeberg.org/lk108/deltatouch/pulls/319
https://github.com/deltachat/deltachat-ios/pull/3274

adbenitez (approved):
thanks! as I had commented at https://github.com/deltachat/deltachat-android/pull/4598#discussion_r3783926032 the current situation with webxdc.js file is very bad, there is no checking/linter etc. it is handled as raw text we really need to improve this in the future

PR #4634, opened 2026-08-31
Remove unused CaptureActivity declaration from AndroidManifest.xml

### LINKED ISSUE #4606: Remove resizability and orientation restrictions in your app to support large screen devices
**Warning From Google Play Console:**
From Android 16, Android will ignore resizability and orientation restrictions for large screen devices, such as foldables and tablets. This may lead to layout and usability issues for your users.

Image

learn more: https://developer.android.com/about/versions/16/behavior-changes-16#large-screens-form-factors

wchen342:
Closes #4606

#skip-changelog

### DIFF
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -413,6 +413,10 @@
             android:configChanges="touchscreen|keyboard|keyboardHidden|orientation|screenLayout|screenSize" />
 
         <activity
+            android:name="com.journeyapps.barcodescanner.CaptureActivity"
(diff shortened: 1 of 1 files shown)


adbenitez (approved):
thanks for figuring this out!

maybe put a comment above saying something like "CaptureActivity sets some deprecated orientation parameters, we always use IntentIntegrator.setCaptureActivity to provide our own custom activity instead of relaying on the default, so it can be removed"

adbenitez (approved):

PR #4639, opened 2026-09-04
Hide delivery status when there are unread messages

### LINKED ISSUE #4638: Conflicting icons when a chat is marked unread after replying
From #4596, here is the real problem. The reproduce steps are [here](https://github.com/deltachat/deltachat-android/pull/4596#issuecomment-5497237914):

The delivery status and unread count are independent of each other, which is normally fine because usually an unread message is the last message. The problem is "Mark as unread" breaks that, because it can mark an older message that is not the newest in the chat. In this case, the conversation ends with both the delivered status and the unread count, because technically both are true, but they are actually describing two different messages.

Now the question is, what do we actually want to do with that scenario? The code is technically not broken and is doing exactly what it is supposed to do, but the UX is the problem. We have a few options:

 1. Hide the delivery status when the unread count is shown
 2. Show both, and make the row consistent (currently the bold text is not applied)
 3. Hide the unread count. But this is problematic because user will see no change after pressing "Mark as unread" in this specific case
 4. Disable "Mark as unread" when the newest message is an outgoing message

I lean towards 1 or 4, but I want to hear your opnions. @r10s @adbenitez

wchen342:
Closes #4638

#skip-changelog

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
+++ b/src/main/java/org/thoughtcrime/securesms/ConversationListItem.java
@@ -268,6 +268,9 @@ public class ConversationListItem extends RelativeLayout
   }
 
   private void setStatusIcons(ThreadRecord thread, int state) {
+    int unreadCount = thread.getUnreadCount();
+    boolean showUnreadBadge = unreadCount > 0 && !thread.isContactRequest();
+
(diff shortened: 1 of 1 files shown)


adbenitez (approved):
lgtm!

PR #4641, opened 2026-09-09
Add local network permission check/request

### LINKED ISSUE #4624: Local network permission required for apps targeting Android 17
Quote from https://developer.android.com/about/versions/17/behavior-changes-17#local-network-protection-permission

Android 17 introduces the [ACCESS_LOCAL_NETWORK](https://developer.android.com/reference/android/Manifest.permission#ACCESS_LOCAL_NETWORK) runtime permission to protect users from unauthorized local network access. Because this falls under the existing NEARBY_DEVICES permission group, users who have already granted other NEARBY_DEVICES permissions aren't prompted again. This new requirement prevents malicious apps from exploiting unrestricted local network access for covert user tracking and fingerprinting. By declaring and requesting this permission, your app can discover and connect to devices on the local area network (LAN), such as smart home devices or casting receivers.

Apps targeting Android 17 (API level 37) or higher now have two paths to maintain communication with LAN devices: Adopt system-mediated, privacy-preserving device pickers to skip the permission prompt, or explicitly request this new permission at runtime to maintain local network communication.

For more information, see the [Local network permission](https://developer.android.com/privacy-and-security/local-network-permission) documentation.

----

I tested changing `targetSdkVersion` to `37` and in fact, adding from second device etc stops working, we need to ask for the new permission, until this problem is solved we are stuck/blocked in SDK 36, eventually targeting 37 will be mandatory

wchen342:
- Request permission for: Profile transfer, Call, WebXDC Realtime Channel, Relay with local addresses
- Raise Target SDK to 37

I only tested the permission request itself. For Call and WebXDC the request is shown only once, since they have backup channels from relay.

Closes #4624

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,7 +35,7 @@ android {
         multiDexEnabled true
 
         minSdkVersion 21
-        targetSdkVersion 36
+        targetSdkVersion 37
 
         vectorDrawables.useSupportLibrary = true
 
--- a/src/main/AndroidManifest.xml
+++ b/src/main/AndroidManifest.xml
@@ -46,6 +46,7 @@
         android:name="android.permission.BLUETOOTH"
         android:maxSdkVersion="30" />
     <uses-permission android:name="android.permission.BLUETOOTH_CONNECT" />
+    <uses-permission android:name="android.permission.ACCESS_LOCAL_NETWORK" />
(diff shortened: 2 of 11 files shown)


adbenitez (approved):
code lgtm and tested it and all works fine, thanks a lot

PR #4642, opened 2026-09-10
fix: apply maximum file size limit to videos as is, without 5/4 factor

link2xt:
UI should not need to do any calculations before comparing the file size to sys.msgsize_max_recommended. Any overhead, e.g. base64 overhead, is taken into account in the core already.

Factor 5/4 seems to come from https://github.com/deltachat/deltachat-android/pull/883, commit https://github.com/deltachat/deltachat-android/commit/a59f6fd69b6f27f5d5d044a881e4767a3c48ff1f

Looking at this code because of https://github.com/chatmail/core/issues/8678

### DIFF
--- a/src/main/java/org/thoughtcrime/securesms/video/recode/VideoRecoder.java
+++ b/src/main/java/org/thoughtcrime/securesms/video/recode/VideoRecoder.java
@@ -590,7 +590,7 @@ public class VideoRecoder {
       VideoEditedInfo vei = getVideoEditInfoFromFile(inPath);
       if (vei == null) {
         Log.w(TAG, String.format("Recoding failed for %s: cannot get info", inPath));
-        if (msg.getFilebytes() > MAX_BYTES + MAX_BYTES / 4) {
+        if (msg.getFilebytes() > MAX_BYTES) {
           alert(context, TOO_BIG_FILE);
           return false;
         }
@@ -673,6 +673,7 @@ public class VideoRecoder {
               vei.originalDurationMs,
               vei.originalAudioBytes);
 
+      // If estimated size is higher than the limit by 25%, don't even try to reencode.
       if (vei.estimatedBytes > MAX_BYTES + MAX_BYTES / 4) {
         alert(context, TOO_BIG_FILE);
         return false;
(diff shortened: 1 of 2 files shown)


src/main/java/org/thoughtcrime/securesms/video/recode/VideoRecoder.java
```diff
@@ -590,7 +590,7 @@ public static boolean prepareVideo(Context context, int chatId, DcMsg msg) {
       VideoEditedInfo vei = getVideoEditInfoFromFile(inPath);
       if (vei == null) {
         Log.w(TAG, String.format("Recoding failed for %s: cannot get info", inPath));
-        if (msg.getFilebytes() > MAX_BYTES + MAX_BYTES / 4) {
```
link2xt, line 593: At least here this factor should not apply, this is a precise file size as we are not going to do any recoding.

src/main/java/org/thoughtcrime/securesms/video/recode/VideoRecoder.java
```diff
@@ -673,7 +673,7 @@ public static boolean prepareVideo(Context context, int chatId, DcMsg msg) {
               vei.originalDurationMs,
               vei.originalAudioBytes);
 
-      if (vei.estimatedBytes > MAX_BYTES + MAX_BYTES / 4) {
```
link2xt, line 676: This I'm not sure about, maybe it is a workaround because the size is "estimated".

adbenitez (approved):

r10s (approved):

r10s (commented):
sorry, a bit too fast with approval.

the check for estimation might still be okay as it is a conservative estimation. this is to not try to dod time-consuming recode if it is already clear that the result is too big.

what seems to be missing, and what might be worth being added additionally, is to check the actually recoded file size

src/main/java/org/thoughtcrime/securesms/video/recode/VideoRecoder.java
r10s: yeah, reading twice, i also think, it makes some sense to keep this in. the estimation is only to prevent trying to recode huge files where it is clear the result will be too big

nicodh:
I just tested with a 20MB video with the provided apk. 
SOmehow the aspect ratio was inverted (was portrait is landscape now)

link2xt:
> what seems to be missing, and what might be worth being added additionally, is to check the actually recoded file size

I added two more checks: after reencoding and after failed reencoding.

adbenitez (approved):
the changelog comment is a bit outdated now that this also "enforce file size limit after encoding video"

wchen342:
Need a rebase.

link2xt:
Rebased and updated the changelog.

PR #4660, opened 2026-09-15
CI android-actions/setup-android temporary fix

wchen342:
See https://github.com/android-actions/setup-android/issues/537

#skip-changelog


adbenitez (approved):
thanks for fixing

PR #4662, opened 2026-09-16
Remove "introduced by" and verification icon from profile

### LINKED ISSUE #4661: Remove verification icon and "Introduced by" line from profile
UI changes for https://github.com/chatmail/core/pull/8691

wchen342:
Closes #4661


src/main/java/com/b44t/messenger/DcContact.java
```diff
@@ -68,12 +68,8 @@ public String toString() {
 
   public native boolean isBlocked();
 
-  public native boolean isVerified();
-
   public native boolean isKeyContact();
 
-  public native int getVerifierId();
-
```
adbenitez, line 76: this needs to also be removed from jni/dc_wrapper.c

wchen342: Oh ok I thought that file is auto generated.

wchen342: Functions are removed.

adbenitez (approved):
this is labeled with "wait-for-core" but actually with the update to jni/dc_wrapper.c now when core gets updated nothing will break and this is already ready and can be merged without blocking on waiting for core, since it is just not using some API, not depending on new API that is not available in current core

wchen342:
> this is labeled with "wait-for-core" but actually with the update to jni/dc_wrapper.c now when core gets updated nothing will break

Yeah that's true. The only dependency is actually the FAQ link (which is in pages not core) but we will wait for a release anyway so I think it is fine.

PR #4670, opened 2026-09-17
Video thumbnail get dimensions through MediaMetadataRetriever

### LINKED ISSUE #4651: attached video is too wide
attaching the following _portrait_ video on android will show it super-wide, as width/height are swapped.

this might be  a device-specific issues, it happens definetely on one of the devices of @adbenitez 

https://github.com/user-attachments/assets/2691adbe-e99c-448f-8571-49082fe763d9

wchen342:
Fixes #4651

#skip-changelog


wchen342:
@adbenitez I cannot reproduce it here, can you test if it solves the problem on your device?

adbenitez:
tested, it seems to be wrong orientation still? it is a portrait video but I just sent it and it is displayed in landscape mode in the chat

image

image

wchen342:
Hmm, I thought it was only the thumbnail that has the dimensions swapped. But the video itself is sent with wrong dimensions too?
The root cause is the same though I think. But can you check if the thumbnail (before sending I think) has the right dimension?

adbenitez:
> Hmm, I thought it was only the thumbnail that has the dimensions swapped. But the video itself is sent with wrong dimensions too?

actually with or without this PR the video is played fine without being wide, what I was saying is in fact just the thumbnail that has dimension swapped when send via gallery option, if sent via "File" button then the dimension seems fine

sadly it seems the test video doesn't help, the original that caused the problem was deleted already because it was a real faces recording and we couldn't share in github

wchen342:
If you happen to get another video that have the same problem then you can send it to me so I can see what is in the video metadata actually. Although the difference between file and gallery is interesting. They shall use the same path which they aren't now I guess.

PR #4673, opened 2026-09-17
process invite links via search in "New Chat" screen

### LINKED ISSUE #4647: allow to paste invite links to contact search
allow to paste invite links to "New Chat / Contact search".

desktop already allows that, android only allows to paste these links to the global search.

cmp. https://github.com/deltachat/deltachat-ios/issues/3285

adbenitez:
this patch makes it possible to paste invite links in the search input bar in the "New Chat" screen -- accessed via menu or the (+) floating button in the chat list

image

close #4647


wchen342:
This doesn't work when you go from the three dot menu -> New Chat. Is that intentional?

adbenitez:
> This doesn't work when you go from the three dot menu -> New Chat. Is that intentional?

@wchen342 good point! I had set the Intent extra flag only for the intent created from the (+) button and forgot to set it also for the menu action, to fix this and avoid similar de-sync in the future I just set the flag in a single place in NewConversationActivity so now there is no need to set it on every place where  NewConversationActivity is being started

wchen342 (approved):
Tested and works. Nice addition!

PR #4675, opened 2026-09-17
remove usage of addr and configured_addr

### LINKED ISSUE #4672: do not use `addr` or `configured_addr` configs
it is mainly used as a fallback if `displayname` is unset.

for that case, add a new string `<string name="unnamed">Unnamed</string>` and use that. "Unnamed" is better than "Unknown" or so as it implies that the user can take action and name their profile (by default, we force a name, so it is a rare case anyways)

once done, file corresponding issues for iOS/desktop/ubuntutouch

adbenitez:
close #4672 

#skip-changelog


src/main/java/org/thoughtcrime/securesms/qr/QrShowActivity.java
```diff
@@ -60,7 +60,7 @@ protected void onCreate(Bundle savedInstanceState) {
                   .CONFIG_DISPLAY_NAME); // we cannot use MrContact.getDisplayName() as this would
```
r10s: "Mr. Contact" is long gone :)

src/main/java/org/thoughtcrime/securesms/preferences/widgets/ProfilePreference.java
```diff
@@ -62,7 +62,7 @@ public void onBindViewHolder(@NonNull PreferenceViewHolder viewHolder) {
   public void refresh() {
     if (profileNameView == null) return;
 
-    final String address = DcHelper.get(getContext(), DcHelper.CONFIG_CONFIGURED_ADDRESS);
+    final String address = "acc-" + DcHelper.getContext(getContext()).getAccountId();
```
wchen342, line 65: Ah ok, this makes sense.

wchen342 (approved):
LGTM, thanks

PR #4678, opened 2026-09-18
reduce usage of email addresses

### LINKED ISSUE #4671: remove email addresses from group creation dialog
we do not want to display any email address, however, when creating a group, the email address is still shown in the memberlist

remove the email address there, and check if it is displayed somewhere else. the only place where it is acceptable is for unencrypted 1:1 chats (however, do not add it there, if it is not there, this issue is only about removing email addresses :)

adbenitez:
also remove unused code that was using email addresses like importing from phone address book and copying email to clipboard in the no longer existing email address field in ProfileActivity

close #4671 

#skip-changelog


wchen342 (approved):
That's quite some dead code, nice

PR #4679, opened 2026-09-18
reduce usage of email address for key-contacts (part 2)

adbenitez:
#skip-changelog


src/main/java/org/thoughtcrime/securesms/ConversationTitleView.java
```diff
@@ -118,7 +118,7 @@ public void setTitle(@NonNull GlideRequests glideRequests, @NonNull DcContact co
     avatar.setSeenRecently(contact.wasSeenRecently());
 
     title.setText(contact.getDisplayName());
-    subtitle.setText(contact.getAddr());
+    if (!contact.isKeyContact()) subtitle.setText(contact.getAddr());
```
r10s: usually, we also indent one-line-if

```suggestion
    if (!contact.isKeyContact()) {
      subtitle.setText(contact.getAddr());
    }
```

src/main/java/org/thoughtcrime/securesms/ProfileAdapter.java
```diff
@@ -183,11 +183,11 @@ public void onBindViewHolder(@NonNull RecyclerView.ViewHolder viewHolder, int i)
       } else {
         dcContact = dcContext.getContact(contactId);
         name = dcContact.getDisplayName();
-        addr = dcContact.getAddr();
+        if (!dcContact.isKeyContact()) subtitle = dcContact.getAddr();
```
r10s: same:

```suggestion
        if (!dcContact.isKeyContact()) {
          subtitle = dcContact.getAddr();
        }
```

r10s (approved):

src/main/java/org/thoughtcrime/securesms/ConversationTitleView.java
adbenitez: makes sense, mainly did it because it was easier when grepping to know which remaining getAddr() was only used for non-key contacts

PR #4680, opened 2026-09-19
tweak transport_list_hint

### LINKED ISSUE #4676: Update "Messages are received on all relays." (`transport_list_hint`)
See https://github.com/deltachat/deltachat-desktop/pull/6747#issuecomment-5728324415.

https://github.com/deltachat/deltachat-android/blob/bc92423e464dcaf80e664a4bdcfc83ec2099acca/src/main/res/layout/activity_relay_list.xml#L31-L38

- [ ] Do the same on other platforms

Related:
- https://github.com/deltachat/deltachat-android/issues/4665

adbenitez:
based on suggestion at https://github.com/deltachat/deltachat-android/issues/4676#issuecomment-5731846492

feel free to suggest changes or tweak directly

close #4676


src/main/res/values/strings.xml
```diff
@@ -737,7 +737,7 @@
     <string name="hidden_from_contacts">Hidden from contacts</string>
     <string name="enforce_e2ee">Enforce Encryption for All Relays</string>
     <!-- Hint for the list of relays -->
-    <string name="transport_list_hint">Messages are received on all relays.</string>
+    <string name="transport_list_hint">Relays are used for sending and receiving messages. Having more than one increases resilience.</string>
```
r10s: i have the feeling that "resilience" is a little too techy, also for translations.

eg. german would be "Resilienz" or "Belastbarkeit", both not really how many ppl would talk.

```suggestion
    <string name="transport_list_hint">Relays are used for sending and receiving messages. Having more than one keeps your connection reliable.</string>
```

r10s (approved):
i left a suggestions that i think is easier to understand, otherwise lgtm, thanks for pushing that forward

src/main/res/values/strings.xml
adbenitez: for me personally resilience feels not related to tech at all, ex. Cuban politicians use it a lot when talking about being a resilient country. Also the first definition given by "The American Heritage® Dictionary of the English Language" is "The ability to recover quickly from illness, change, or misfortune; buoyancy." only the 9th alternative definition is IT related, but I am fine with your suggestion

adbenitez: (btw, "reliable" is not so easy to translate to Spanish, it translates to "confiable" which is more like "trustworthy" than unfailing but I could use equivalent of "stable" in Spanish as translation)

PR #4681, opened 2026-09-19
add context to strings, unify string usage

r10s:
this PR is a follow-up of the appupdate changes:

- add context to the new string placeholders - it is otherwise hard for the translators to translate them, and/or causes more support from us

- we do not make "rare" errors translatable: they are often enough hard to translate, cost resources, human and machine - and when asking for support it is even better if it is english

- took the chance to make some errors more specific


wchen342 (approved):
Ah ok, makes sense. Thank you for correcting them!

PR #4685, opened 2026-09-21
update privacy policy link

adbenitez:
#skip-changelog


src/main/java/org/thoughtcrime/securesms/LocalHelpActivity.java
```diff
@@ -106,7 +106,7 @@ public boolean onOptionsItemSelected(MenuItem item) {
       openOnlineUrl("https://delta.chat");
       return true;
     } else if (itemId == R.id.privacy_policy) {
-      openOnlineUrl("https://delta.chat/gdpr");
+      openOnlineUrl("https://delta.chat/privacy");
```
r10s, line 109: when there is a define, it makes sense to reuse it here as well, smth as 


```
      openOnlineUrl(InstantOnboardingActivity.PRIVACY_URL);
```

r10s (approved):

Issue #4640, opened 2026-09-09
Sending an image while not connected to the internet will block future notifications


Amzd:
- Android version: 
- Device:
- Delta Chat version: 2.59
- Expected behavior:
Notification process should not try to send images
- Actual behavior:
Notification process tries to send images (and probably gets killed for high memory)
- Steps to reproduce the problem:
  - Alice turn on airplane mode
  - Alice sends image to bob
  - Alice closes the app
  - Alice turns off airplane mode
  - Bob sends message to Alice
  - Now bob receives the premessage for the image message, and Alice does not receive any notification
- Screenshots:
- Logs:

Yes it is nice that the background process sends messages that were queued up previously but that should not happen at the cost of not receiving notifications.

adbenitez:
I tried this yesterday with @Amzd and it worked fine for me, when fetching we get a permanent notification so the app is considered to be in foreground, we need more information/logs of the device where this causes problems

adbenitez:
since we lack more info and can't reproduce the problem so easily, and since the theoretical cause of this problem got fixed in core side (background fetch not doing sending if I/O is started etc) and since we also have an open issue to add permanent notification when there are messages pending to send (#4648) which would also avoid the mentioned situation completely,  I think we can close this issue here

PR #4688, opened 2026-09-22
enable fetch worker on all builds

wchen342:
Needs https://github.com/chatmail/core/pull/8592

Partial workaround for #4640 before larger refactor.

#skip-changelog


adbenitez (approved):
tested it and all works as expected when receiving push notifications, thanks

PR #4692, opened 2026-09-22
fix android-actions/setup-android  in cache-core workflow

adbenitez:
#4660 fixed it for apk preview but core cache was not updated

#skip-changelog


wchen342 (approved):
LGTM, thanks

Issue #4687, opened 2026-09-21
adapt to new contact.get_freshness


r10s:
replace contact.was_seen_recently() by contact.get_freshness()

in contact list (new contact, memberlist, maybe more), add a subtitle in case contact.get_freshness() returns OLD:

the subtitle should show the age of the contact using the strings `seen_n_months_ago`, `seen_n_years_ago`, `never_seen`, based on `contact.getLastSeen()` (0=never)

the contact's profile itself can probably stay as is for now, it already shows last seen

counterpart of https://github.com/deltachat/deltachat-desktop/issues/6769

PR #4693, opened 2026-09-22
cleanup ContactSelectionListItem

adbenitez:
base cleanup as preparation for targeting #4687 

I was looking into targeting #4687 and came across this unused label while refactoring `ContactSelectionListItem.set()` method, moved it to a separate PR to avoid cluttering the main PR

also removes `name` from class field since it is needed only locally in the methods it is passed to


r10s (approved):

adbenitez:
superseded by https://github.com/deltachat/deltachat-android/pull/4694

PR #4694, opened 2026-09-22
cleanup ContactSelectionListItem

adbenitez:
base cleanup as preparation for targeting #4687 

I was looking into targeting #4687 and came across this unused label while refactoring `ContactSelectionListItem.set()` method, moved it to a separate PR to avoid cluttering the main PR

also:
- remove `name` from class field since it is needed only locally in the methods it is passed to
- remove unneeded boilerplate around ViewHolder


src/main/java/org/thoughtcrime/securesms/contacts/ContactSelectionListItem.java
```diff
@@ -127,21 +136,21 @@ public void setChecked(boolean selected) {
   }
 
   public void unbind(GlideRequests glideRequests) {
-    if (recipient != null) {
-      recipient.removeListener(this);
+    if (recipientListener != null && recipient != null) {
```
wchen342, line 142: This is a slight correctness problem, I think `setQrInviteData` can set `recipient` without setting `recipientListener`?

r10s (approved):

src/main/java/org/thoughtcrime/securesms/contacts/ContactSelectionListItem.java
adbenitez: yes, setQrInviteData doesn't need to set the listener because there is no contact that can change at this point (theoretically yes, if the invite is from a contact you already have and then contact changes casually at the same time due to incoming message from that person at the same time you have their invite in the search, but it is not worth updating the contact in that case and for the "online" green dot we also don't care to show it in that case either

wchen342: What I was trying to say is we shall do `recipient = null;` without putting it inside `recipientListener != null`.

adbenitez: I pushed a commit resetting both recipient and recipientListener in the binding of new item

wchen342 (approved):
LGTM!

PR #4695, opened 2026-09-22
add a device message for 2.60+

r10s:
this PR adds a new device message dor 2.60+. the strings are split into general/os-specific parts, so that desktop/iOS can add their own ending

the last device message was about 2.0, idea is to mention large, new features since then - user may not have discovered them, also we were hesitant about many things in the past, for $reasons

this is how it looks like:

Screenshot 2026-09-22 at 20 35 22


wchen342 (approved):

wchen342:
Need a code formatting.

src/main/res/values/strings.xml
```diff
@@ -1301,7 +1301,9 @@
...
 
     <!-- device messages for updates -->
-    <string name="update_2_0">What\'s new?\n\n💯 End-to-end encryption is reliable and forever now. Padlocks 🔒 are gone!\n\n✉️ Classic email without end-to-end encryption is marked with a letter symbol\n\n😻 New enhanced profile screen for all your contacts\n\n🔲 New button for quick access to apps used in a chat\n\n❤️ Please donate to help us remain independent and continue to bring improvements: %1$s</string>
+    <string name="update_2_6">Guess what\'s new!\n\n🌐 Your messages now travel smarter: With multiple relays, one going down won\'t stop them\n\n📞 Calls grew up and are no longer experimental\n\n📢 Channels are here to stay - now even with reactions\n\n✍️ Add a chat description, so no one\'s left wondering\n\n👌 Sent images now look sharper\n\n🔵 Mark chats as unread, so you won\'t forget</string>
+    <string name="update_2_6_android_additions">🐇 Speed-up and auto-play voice messages\n\nFinally: We got a basic sticker picker! Please donate to help us remain independent and continue to bring improvements ❤️ https://delta.chat/donate</string>
```
adbenitez: for the voice message point, the ⏩ (fast-forward) emoji is likely more is more on-topic (and usually orange like our player) but the bunny is also ok

r10s: idea of 🐇 was about speed, if also seems more funny, the bunny

adbenitez (approved):

